PR #126, opened 2023-06-12
Postgres Peer Revitalised

Amogh-Bharadwaj:
Adds a lot of functionality to current state of our postgres peer such as CRUD, data type support (ex: timestamp, json etc.).


nexus/server/tests/results/actual/postgres.sql.out
iskakaushik, line 1: don't commit files in the actual directory

nexus/peer-postgres/src/ast.rs
```diff
@@ -1,21 +1,82 @@
...
             }
             ControlFlow::<()>::Continue(())
         });
     }
+
+    #[tracing::instrument(name = "peer_postgres::rewrite_statement", skip(self))]
+    pub fn rewrite_statement(&self, stmt: &mut Statement) -> anyhow::Result<()> {
+        /* could be written less clunkily when https://github.com/rust-lang/rust/issues/53667 is stabilized.
+        Tracking in #225 */
+        let table_name: &mut ObjectName = if let Statement::Insert {
```
iskakaushik, line 28: can we use the visitor pattern to simplify a lot of these for example we seem to be working with TableFactor::Table in multiple branches.

iskakaushik:
we seem to be failing with this error

```[server/tests/server_test.rs:110] query.as_str() = "SELECT COUNT(*) FROM FIRST_POSTGRES.PUBLIC.ILMETZIK;"
thread 'server_test' panicked at 'Failed to query: Error { kind: Db, cause: Some(DbError { severity: "ERROR", parsed_severity: None, code: SqlState(EXX000), message: "Internal error: error getting schema: db error: ERROR: cross-database references are not implemented: \"first_***.public.ilmetzik\"", detail: None, hint: None, position: None, where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: None, line: None, routine: None }) }', server/tests/server_test.rs:119:18```

Could it be that in rewriting we aren't removing the peer name?

iskakaushik (approved):
LGTM

PR #129, opened 2023-06-14
Composite primary keys support in MIRROR for CDC.

saisrirampur:
Also fixed bug where we are assuming primary key as **id** in NormalizeRecords


heavycrystal:
implemented as part of #499

PR #145, opened 2023-06-19
Kafka Peer Creation and Peer Validity

Amogh-Bharadwaj:
- Introduces a peer creation flow for Apache Kafka, with the help of [Rdkafka](https://docs.rs/rdkafka/0.32.2/rdkafka/index.html). Uses SASL SSL authentication.
- Adds configuration validity checking for all peers before creation, with appropriate error messages.
- Warnings in Nexus codebase brought down to 3.


nexus/server/src/main.rs
```diff
@@ -158,6 +170,101 @@ impl NexusBackend {
                     peer,
                     if_not_exists: _,
                 } => {
+                    // Checking if you can connect to the peer with the config
+                    match peer.clone().config {
+                        Some(Config::BigqueryConfig(bq_config)) => {
+                            let project_id = bq_config.clone().project_id;
```
iskakaushik, line 176: move to check to the executor interface

PR #196, opened 2023-07-07
Fixes for PG Numeric and CREATE Peer

Amogh-Bharadwaj:
- Maps `Numeric` in PG to a new `PostgresNumeric` value of Rust type decimal.
- Does not use QueryExecutor's Execute function for BigQuery in CREATE PEER check as it fails.
- Removes peer name from server/main's executor map after check is done. CREATE PEER will now not, under any circumstance, add peer executors to the `NexusBackend` state.
- Removes `expect` in BigQuery stream.rs ; this prevents crashing for invalid timestamps


iskakaushik:
as discussed make this the default numeric and delete the other one.

nexus/peer-bigquery/src/stream.rs
```diff
@@ -127,9 +127,11 @@ impl BqRecordStream {
                     FieldType::Float | FieldType::Float64 => {
                         result_set.get_f64_by_name(field_name)?.map(Value::Double)
                     }
-                    FieldType::Bignumeric | FieldType::Numeric => result_set
-                        .get_string_by_name(field_name)?
-                        .map(Value::Numeric),
+                    FieldType::Bignumeric | FieldType::Numeric => {
+                        let result_string = result_set.get_string_by_name(field_name)?;
+                        result_string.map(|s| Value::Numeric(s.parse().expect("Invalid numeric")))
```
iskakaushik, line 132: rather than expect, propagate error upwards.

nexus/value/src/lib.rs
```diff
@@ -164,7 +165,7 @@ impl Value {
                 } else if let Some(f) = n.as_f64() {
                     Value::Double(f)
                 } else {
-                    Value::Numeric(n.to_string())
+                    Value::Numeric(n.to_string().as_str().parse::<Decimal>().unwrap())
```
iskakaushik, line 168: rather than `unwrap` send the error upwards if possible.

Amogh-Bharadwaj: Would require refactoring return types and values of this function and functions where this is called.. don't see an easy way right now

iskakaushik:
thanks @Amogh-Bharadwaj !

PR #198, opened 2023-07-07
fixing remaining tests as well as expanding QValue type system coverage

heavycrystal:


flow/model/qvalue/qvalue.go
```diff
@@ -59,10 +59,11 @@ func (q *QValue) Equals(other *QValue) bool {
 }
 
 func (q *QValue) GoTimeConvert() (string, error) {
-	if q.Kind == QValueKindTime {
+	if q.Kind == QValueKindTime || q.Kind == QValueKindTimeTZ {
 		return q.Value.(time.Time).Format("15:04:05.999999"), nil
-	} else if q.Kind == QValueKindTimeTZ {
-		return q.Value.(time.Time).Format("15:04:05.999999-0700"), nil
+		// no connector supports time with timezone yet
```
iskakaushik, line 64: is this comment still valid? looks like you added `|| q.Kind == QValueKindTimeTZ ` above

heavycrystal: comment is valid, I'm parsing both in the same manner to remove the timezone component, as BQ errors out if it sees a timezone in the string

iskakaushik (approved):
LGTM

PR #206, opened 2023-07-11
S3 for Streaming Query

### LINKED ISSUE #202: Add support for S3 peer for Query Replication
## Context:

When we replicate a query to snowflake, we already stage to an S3 bucket. We write avro files to that S3 bucket and then we issue a COPY command to copy from the external S3 stage to the snowflake table. We can think of this as a transitional stage.

### code references: 

* `aws.go` (credentials)
* `snowflake/qrep_avro_sync.go`
* `snowflake/qrep.go` is the entry point for query replication.

## Requirement:

Treat S3 as a first class Peer so that we can write Avro to the S3 destination as a final point.

## Approach:

1. In `peers.proto` add definition for an S3 peer. This would be the following
    a. S3 url: `s3://<bucket_name>/<prefix>`
2. create a connector for `S3`, the same way we do it for Snowflake, BQ, etc. All the CDC methods will error saying not supported, only query replication will be supported.
3. When we get a batch of records from the source, we will write them as avro files to the given s3 url. Most of this logic already exists in `snowflake/qrep_avro_sync.go`, refactor to reuse.
4. Make it so that down the line its easy to support `parquet` etc.

Amogh-Bharadwaj:
- Introduces streaming queries on Postgres to S3 buckets in the form of Avro files.
- New S3 connector

Fixes #202

### DIFF
--- a/flow/connectors/s3/qrep.go
+++ b/flow/connectors/s3/qrep.go
@@ -27,6 +27,9 @@ func (c *S3Connector) SyncQRepRecords(
 	partition *protos.QRepPartition,
 	records *model.QRecordBatch,
 ) (int, error) {
+	if len(records.Records) == 0 {
+		return 0, nil
+	}
 	dstTableName := config.DestinationTableIdentifier
 	avroSchema, err := getAvroSchema(dstTableName, records.Schema)
 	if err != nil {


flow/connectors/s3/avro_writer.go
```diff
@@ -0,0 +1,127 @@
+package conns3
```
iskakaushik, line 1: lets share this class between s3 and sf

iskakaushik (commented):
add some test, see eventhub_test file as an example

flow/connectors/s3/s3.go
```diff
@@ -0,0 +1,118 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type S3Connector struct {
+	url    string
+	client s3.S3
+}
+
+func NewS3Connector(ctx context.Context,
+	S3ProtoConfig *protos.S3Config) (*S3Connector, error) {
```
iskakaushik, line 20: variable names should start with lower case letters

flow/connectors/s3/s3.go
```diff
@@ -0,0 +1,118 @@
...
+	log.Infof("Close not needed for S3")
+	return nil
+}
+
+func (c *S3Connector) ConnectionActive() bool {
+	_, err := c.client.ListBuckets(nil)
+	return err == nil
+}
+
+func (c *S3Connector) NeedsSetupMetadataTables() bool {
```
iskakaushik, line 41: for all the non-qrep methods (CDC methods like most of these in this file, can you return error:

`cdc based replication is not currently supported for S3 target`.

flow/e2e/s3_helper.go
```diff
@@ -0,0 +1,96 @@
...
+	bucketName string = "peerdb-test-bucket"
+	prefixName string = "test-s3"
+)
+
+type S3TestHelper struct {
+	client   *s3.S3
+	s3Config *protos.S3Config
+}
+
+func NewS3TestHelper(pgConf *protos.PostgresConfig) (*S3TestHelper, error) {
```
iskakaushik, line 24: we don't need PG conf here

iskakaushik (commented):
left some minor comments, looks great overall

PR #242, opened 2023-07-22
Allow connecting to servers that enforce tls

iskakaushik:
We don't verify the server cert yet, this matches the default behaviour of psql.


iskakaushik:
cc: @saisrirampur

PR #247, opened 2023-07-25
Wires Eventhub Peer

Amogh-Bharadwaj:
- Nexus now takes in the `metadata_db` for Eventhub as a connection string.
- Fixes lints


iskakaushik:
can you revert all the protofile changes? They seem to be only whitespaces

PR #264, opened 2023-08-01
Check for existing table in CDC 

### LINKED ISSUE #256: Table matching needs to be uniform across streaming query and cdc
1. The assumption in qrep is that there will already be a table on the destination side with compatible columns.
2. On CDC side, PeerDB creates the destination table.

Lets move to a uniform assumption that does the following:
1. Checks if the table already exists, if it exists
    a. Check if its compatible with the transferred rows -- fail if it isn't.
    b. If it is compatible use this as the target.
2. If there is no table, create a compatible table on the target and use it.

cc: @saisrirampur

Amogh-Bharadwaj:
This PR introduces check for an existing destination table for CDC for Snowflake, Bigquery and Postgres as sinks. Every column and its data type is matched with that of the source table.

Fixes #256

### DIFF
--- a/flow/connectors/bigquery/bigquery.go
+++ b/flow/connectors/bigquery/bigquery.go
@@ -837,8 +837,16 @@ func (c *BigQueryConnector) SetupNormalizedTable(
 	table := c.client.Dataset(c.datasetID).Table(req.TableIdentifier)
 
 	// check if the table exists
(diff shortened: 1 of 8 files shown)


Amogh-Bharadwaj:
Closing this as we now have https://github.com/PeerDB-io/peerdb/pull/756

PR #266, opened 2023-08-02
making CREATE MIRROR command block until SetupFlow workflow completes

heavycrystal:
This fixes two issues:

1) CREATE MIRROR [for CDC] was completely asynchronous earlier, which ended up masking issues with the SetupFlow step which could fail for multiple reasons.
2) DROP MIRROR would not work for failed mirrors like this, since Temporal was spinning on SetupFlow.


iskakaushik:
obsolete.

PR #384, opened 2023-09-14
Adds actions files for ui

Amogh-Bharadwaj:
- Linter for UI
- UI builder

### DIFF
--- a/.github/workflows/ui-build.yml
+++ b/.github/workflows/ui-build.yml
@@ -11,8 +11,10 @@ on:
 jobs:
   build-test:
     name: Build & Test UI
-    runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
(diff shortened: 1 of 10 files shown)


.github/workflows/ui-build.yml
```diff
@@ -0,0 +1,26 @@
...
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build-test:
+    name: Build & Test UI
+    runs-on: ubuntu-latest
```
iskakaushik, line 14: can you change runs-on to be similar to https://github.com/PeerDB-io/peerdb/pull/383/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR11 for both this and the other file?

For context, we are testing out ubicloud runners for lower price and better perf.

PR #387, opened 2023-09-15
support for mirror if not exists

Amogh-Bharadwaj:


nexus/server/src/main.rs
```diff
@@ -199,79 +199,134 @@ impl NexusBackend {
                         "OK", None,
                     ))])
                 }
-                PeerDDL::CreateMirrorForCDC { flow_job } => {
+                PeerDDL::CreateMirrorForCDC {
```
iskakaushik, line 202: can we refactor handling this branch to its own function? It seems pretty complex now, can you refactor to make it readable.

iskakaushik (changes requested):

iskakaushik (approved):

PR #420, opened 2023-09-25
[DRAFT] CDC mirrors with SCHEMA MAPPING support

heavycrystal:
support only for Postgres and Snowflake for now


protos/flow.proto
```diff
@@ -127,6 +124,9 @@ message SetupReplicationInput {
   bool do_initial_copy = 5;
   string existing_publication_name = 6;
   string existing_replication_slot_name = 7;
+
+  // if length > 0, ignore table name mapping and use schemas as basis of publication creation
```
iskakaushik, line 128: instead of this, can we pass the mapping type here?

protos/flow.proto
```diff
@@ -110,12 +110,9 @@ message TableIdentifier {
   }
 }
 
-message EnsurePullabilityOutput {
-  TableIdentifier table_identifier = 1;
-}
-
 message EnsurePullabilityBatchOutput {
   map<string, TableIdentifier> table_identifier_mapping = 1;
+  MappingType mapping_type = 2;
```
iskakaushik, line 115: Why does the output contain the mapping type?

protos/flow.proto
```diff
@@ -310,3 +310,29 @@ message ReplayTableSchemaDeltaInput {
   FlowConnectionConfigs flow_connection_configs = 1;
   TableSchemaDelta table_schema_delta = 2;
 }
+
+message PopulateTableMappingFromSchemasInput {
```
iskakaushik, line 314: ListTablesInSchemaInput seems like a better name here

protos/flow.proto
```diff
@@ -310,3 +310,29 @@ message ReplayTableSchemaDeltaInput {
...
+  peerdb_peers.Peer peer_connection_config = 1;
+  map<string, string> schema_mapping = 2;
+}
+
+message TablesList {
+  repeated string tables = 1;
+}
+
+message PopulateTableMappingFromSchemasOutput {
```
iskakaushik, line 323: ListTablesInSchemaOutput

protos/flow.proto
```diff
@@ -310,3 +310,29 @@ message ReplayTableSchemaDeltaInput {
...
+  map<string, TablesList> table_mapping = 1;
```
iskakaushik, line 324: schema_to_tables

protos/flow.proto
```diff
@@ -56,6 +58,10 @@ message FlowConnectionConfigs {
   // the below two are for eventhub only
   int64 push_batch_size = 21;
   int64 push_parallelism = 22;
+
+  MappingType mapping_type = 23;
+  // if MappingType is SCHEMA, this contains the schema level mappings.
+  map<string, string> schema_mapping = 24;
```
iskakaushik, line 64: lets also take a boolean here: `allow_table_additions`, I think it will be useful for when people don't want to allow new tables.

flow/connectors/core.go
```diff
@@ -31,14 +31,15 @@ type CDCPullConnector interface {
...
 	// PullRecords pulls records from the source, and returns a RecordBatch.
 	// This method should be idempotent, and should be able to be called multiple times with the same request.
-	PullRecords(req *model.PullRecordsRequest) (*model.RecordsWithTableSchemaDelta, error)
+	PullRecords(req *model.PullRecordsRequest) (*model.RecordsWithDeltaInfo, error)
 
 	// PullFlowCleanup drops both the Postgres publication and replication slot, as a part of DROP MIRROR
 	PullFlowCleanup(jobName string) error
+
+	// GetAllTablesInSchemas... gets all the tables in multiple schemas
+	GetAllTablesInSchemas(schemas []string) (map[string][]string, error)
```
iskakaushik, line 42: `ListTablesInSchemas`

flow/model/model.go
```diff
@@ -282,6 +284,8 @@ type SyncResponse struct {
 	TableSchemaDelta *protos.TableSchemaDelta
 	// to be stored in state for future PullFlows
 	RelationMessageMapping RelationMessageMapping
+	// to be used to create additional tables, for MappingType SCHEMA
+	AdditionalTableInfo *protos.AdditionalTableInfo
```
iskakaushik, line 288: Can we push this info to `TableSchemaDelta`?

flow/model/model.go
```diff
@@ -318,3 +324,22 @@ func (r *RelationRecord) GetItems() *RecordItems {
 }
 
 type RelationMessageMapping map[uint32]*protos.RelationMessage
+
+type AddedTableRecord struct {
```
iskakaushik, line 328: Why do we need these? Can you add a comment?

iskakaushik (changes requested):
Please let me know once these are fixed, I'll take another look.

iskakaushik (commented):
Please let me know once these are fixed, I'll take another look.

protos/flow.proto
heavycrystal: removed

protos/flow.proto
heavycrystal: added

saisrirampur:
+1 this feature would have helped a customer we chatted with today.

PR #437, opened 2023-09-26
added overall throughput metric and fixed records pulled

heavycrystal:


docker-compose.yml
```diff
@@ -222,7 +222,7 @@ services:
       context: .
       dockerfile: stacks/ui.Dockerfile
     ports:
-      - 3000:3000
+      - 3001:3001
```
iskakaushik, line 225: It should be `3001:3000`

flow/connectors/utils/metrics/metrics.go
```diff
@@ -118,3 +119,14 @@ func LogQRepNormalizeMetrics(ctx context.Context, flowJobName string,
 	recordsSyncedPerSecondGauge.Update(float64(normalizedRecordsCount) / duration.Seconds())
 	totalRecordsAtTargetGauge.Update(float64(totalRecordsAtTarget))
 }
+
+func LogCDCOverallMetrics(ctx context.Context, flowJobName string, throughput float64) {
```
iskakaushik, line 123: lets be consistent -- call it Throughput (instead of Overall)

flow/activities/flowable.go
```diff
@@ -238,6 +239,8 @@ func (a *FlowableActivity) StartFlow(ctx context.Context,
 		log.WithFields(log.Fields{
 			"flowName": input.FlowConnectionConfigs.FlowJobName,
 		}).Info("no records to push")
+		metrics.LogSyncMetrics(ctx, input.FlowConnectionConfigs.FlowJobName, 0, 1)
+		metrics.LogNormalizeMetrics(ctx, input.FlowConnectionConfigs.FlowJobName, 0, 1, 0)
```
iskakaushik, line 243: what about Throughput? Should we zero it out as well?

flow/activities/flowable.go
```diff
@@ -351,6 +364,13 @@ func (a *FlowableActivity) StartNormalize(
 		return nil, err
 	}
 
+	throughput, err := a.CatalogMirrorMonitor.GetThroughputForCDCBatch(ctx, input.FlowConnectionConfigs.FlowJobName,
```
iskakaushik, line 367: Why are we doing this twice?

iskakaushik (changes requested):

flow/activities/flowable.go
heavycrystal: the first occurrence is the case where a connector doesn't support normalize, we need to log metrics despite not running normalize

PR #451, opened 2023-09-28
Update base64 0.13 to base64 0.21. Replace use of deprecated chrono method

serprex:


nexus/value/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 use array::ArrayValue;
+use base64::Engine;
```
iskakaushik, line 2: What do you think about [something like](https://docs.rs/base64/latest/base64/prelude/index.html):

```
use base64::prelude::{Engine as _, BASE64_STANDARD}
```

We can then use `BASE64_STANDARD`.

iskakaushik (approved):
Thanks for the contribution. Looks great overall. Left one minor feedback, let me know what you think.

PR #456, opened 2023-09-29
UI for Create Postgres Peer

Amogh-Bharadwaj:
- UI for Create Peer step where user enters configuration details
- Wired validate peer API to the UI
- Wired create peer API to the UI
- Form validation, loading text 
- Written in a way where adding new peers is easy
- Tested the create peer flow (including psql-ing the peer) for postgres

APIs for create peer and validate peer are present in `app/api/peers/validate` and `app/api/peers/create` respectively.

Makes progress towards #431


ui/app/peers/create/configuration/helpers/pg.ts
```diff
@@ -0,0 +1,36 @@
...
+  {
+    label: 'Host',
+    stateHandler: (value: string, setter: PeerSetter) =>
+      setter((curr) => ({ ...curr, host: value })),
+  },
+  {
+    label: 'Port',
+    stateHandler: (value: string, setter: PeerSetter) =>
+      setter((curr) => ({ ...curr, port: parseInt(value, 10) })),
+    type: 'number', // type for textfield
```
Jobandeep2417, line 13: The Port should have a min max range, it should not allow negative values. 

image


The way PgConfig and Config form components are setup there is no way to send this info from the settings props.

ui/app/peers/create/configuration/configForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+interface ConfigProps {
+  settings: Setting[];
+  setter: PeerSetter;
+}
+
+export default function PgConfig(props: ConfigProps) {
+  return (
+    <>
+      {props.settings.map((setting, id) => {
+        return (
```
Jobandeep2417, line 22: I think making the form config-driven is a good idea. But we maybe need to revisit things like how default values can be set. For Example port default value is 5432 in config state variable, but we are not using it anywhere. perhaps a crude way to do it could be as shown below by adding a default value
 
```

export default function PgConfig(props: ConfigProps) {
  return (
    <>
      {props.settings.map((setting, id) => {
        let _defaultValue = setting.type === "number" && setting.label === "Port" ? "5432" : "";
        return (
          <RowWithTextField
            key={id}
            label={<Label as='label'>{setting.label}</Label>}
            action={
              <TextField
                variant='simple'
                type={setting.type}
                defaultValue={_defaultValue}
                onChange={(e) =>
                  setting.stateHandler(e.target.value, props.setter)
                }
              />
            }
          />
        );
      })}
    </>
  );
```

image

ui/app/peers/create/page.tsx
```diff
@@ -27,13 +39,25 @@ export default function CreatePeer() {
...
         />
       </Panel>
       <Panel>
         <ButtonGroup>
-          <Button>Cancel</Button>
-          <Button variant='normalSolid'>Continue</Button>
+          <Button as={Link} href='/peers'>
+            Cancel
+          </Button>
+          <Button
```
Jobandeep2417, line 52: This is confusing we should block at Data source selection only or grey out the unavailable datasource types

image

ui/app/peers/create/page.tsx
```diff
@@ -1,21 +1,33 @@
...
+        <Action
+          icon={<Icon name='help' />}
+          href='https://docs.peerdb.io/sql/commands/create-peer'
+          target='_blank'
+        >
+          Learn about peers
+        </Action>
       </Panel>
       <Panel>
         <Label colorName='lowContrast' variant='subheadline'>
```
Jobandeep2417, line 33: I think both source and Datasource is not required any one is sufficient
image

ui/app/peers/create/configuration/handlers.ts
```diff
@@ -0,0 +1,86 @@
...
+    method: 'POST',
+    body: JSON.stringify({
+      name,
+      type,
+      config,
+    }),
+  }).then((res) => res.text());
+
+  if (statusMessage !== 'created') {
+    setMessage({ ok: false, msg: statusMessage });
```
Jobandeep2417, line 78: Is this supposed to show an error message on UI? Coz even for 500 code I am not seeing any error message?

image

ui/app/peers/create/configuration/schema.ts
```diff
@@ -0,0 +1,20 @@
...
+  database: Joi.string().min(1).max(100).required(),
+  user: Joi.string().alphanum().min(1).max(64).required(),
+  password: Joi.string().min(1).max(100).required(),
+  transactionSnapshot: Joi.string().max(100).allow('').optional(),
+});
+
+export const checkFormFields = (peerType: string, config: PeerConfig) => {
+  switch (peerType) {
+    case 'POSTGRES':
+      return pgSchema.validate(config);
```
Jobandeep2417, line 16: nit: we are showing extra double quotes for error field
image

ui/app/api/peers/validate/route.ts
```diff
@@ -0,0 +1,29 @@
...
+  ValidatePeerResponse,
+  ValidatePeerStatus,
+} from '@/grpc_generated/route';
+import { GetFlowServiceClient } from '@/rpc/rpc';
+import { typeMap } from '../peerTypeMap';
+
+export async function POST(request: Request) {
+  let body = await request.json();
+  let name = body.name;
+  let type = body.type;
```
Jobandeep2417, line 12: Can be refactored as 

```
let { name, type, config } = body;
```

ui/app/peers/create/configuration/handlers.ts
Jobandeep2417: The same issue for validate button

Jobandeep2417:
@iskakaushik Did we consider using material-ui even if we want to use storybook, material-ui can be integrated with storybook (https://storybook.js.org/recipes/@mui/material). The out-of-the-box components provided by the storybook lib look pretty primitive (in styling, shimmering/loading behavior, etc.) . I think we should consider libraries like material-ui/Chara-ui etc from the starting stage of the UI, Please let me know your thoughts on this

ui/app/peers/create/configuration/configForm.tsx
Amogh-Bharadwaj: The default values are not really meant to be used in UI, but rather because protobuf-grpc doesn't seem to allow undefined values so we need dummies.

ui/app/peers/create/configuration/handlers.ts
Amogh-Bharadwaj: Hey Joban. Couldn't reproduce this. Maybe we could investigate separately if this persists

ui/app/peers/create/configuration/helpers/pg.ts
Amogh-Bharadwaj: We could prevent typing of non-numbers but right now I personally think it's fine. Happy to discuss though

ui/app/peers/create/configuration/schema.ts
Amogh-Bharadwaj: I think this is fine as it quotes does make the error more clear in a way

ui/app/peers/create/page.tsx
Amogh-Bharadwaj: Yup. Changed. Thanks

ui/app/peers/create/page.tsx
Amogh-Bharadwaj: Yup. Singled out postgres (for now) as the only option in the data source selection step.

ui/app/peers/create/configuration/configForm.tsx
Amogh-Bharadwaj: Also thanks for the catch: the config value for the state here should be made generic to peer. fixed.

ui/app/api/peers/validate/route.ts
Amogh-Bharadwaj: Fixed

ui/app/peers/create/configuration/schema.ts
Jobandeep2417: For the Name field double quotes were not there, which is why I pointed it out. Looks a bit inconsistent

ui/app/peers/create/page.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </Panel>
+      <Panel>
+        <ButtonGroup>
+          <Button as={Link} href='/peers'>
+            Cancel
+          </Button>
+          <Button
+            disabled={!peerType}
+            onClick={() => {
+              router.push(`/peers/create/configuration?dbtype=${peerType}`);
```
iskakaushik, line 52: should we use `next/link` instead? https://nextjs.org/blog/next-13-2#statically-typed-links

ui/app/api/peers/create/route.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  CreatePeerRequest,
+  CreatePeerResponse,
+  CreatePeerStatus,
+} from '@/grpc_generated/route';
+import { GetFlowServiceClient } from '@/rpc/rpc';
+import { constructPeer } from '../util';
```
iskakaushik, line 7: lets move this method to this class, there isn't much happening in the utils still.

ui/app/api/peers/create/route.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export async function POST(request: Request) {
```
iskakaushik, line 9: can we move this to `peers/route.ts` and service both validate and create based on a param in the json or something else?

ui/app/peers/create/configuration/handlers.ts
```diff
@@ -0,0 +1,83 @@
...
+    setLoading(false);
+    return;
+  } else {
+    setMessage({ ok: true, msg: 'Peer is valid' });
+  }
+  setLoading(false);
+};
+
+// API call to create peer
+export const handleCreate = async (
```
iskakaushik, line 55: lets discuss the need for this

iskakaushik (changes requested):

iskakaushik (approved):

ui/app/peers/create/configuration/schema.ts
Amogh-Bharadwaj: Hey Joban there are no quotes now. Thanks

PR #500, opened 2023-10-10
Add support for more properties to be passed with TableMapping

iskakaushik:


saisrirampur:
You can now specify any (non CTID) column as a partition column. The column needs to be of the type integer and timestamp and need not be monotonically increasing. PeerDB under the covers takes of creating equal sized partitions for streaming to the target. Below is an example of how the MIRROR would be modified:

```
CREATE MIRROR postgres_to_snowflake
FROM postgres_peer TO snowflake_peer
WITH TABLE MAPPING (public.pgbench_branches:public.pgbench_branches,
                    public.pgbench_tellers:public.pgbench_tellers, -- if partition column is not specified in table name mapping, CTID will be used
                    {
                    from: public.pgbench_accounts,
                    to: public.pgbench_accounts,
                    key: aid -- partition column
                    })
WITH (
  do_initial_copy = true,
  snapshot_sync_mode='avro',
  snapshot_num_rows_per_partition = 500000,
  snapshot_max_parallel_workers = 4,
  snapshot_num_tables_in_parallel = 1,
  snapshot_staging_path = ''
);
```

PR #503, opened 2023-10-11
Return only at commit message in Postgres CDC

iskakaushik:


heavycrystal (approved):

heavycrystal:
Also fixes lingering issues with schema changes support:

1) DROP COLUMN is no longer supported, columns dropped will be detected but not propagated to the destination.
2) correctness improvements to ALTER TABLE... ADD COLUMN

PR #506, opened 2023-10-12
Add gRPC method to get the status of a mirror

iskakaushik:


flow/cmd/mirror_status.go
```diff
@@ -0,0 +1,256 @@
...
+	err = h.pool.QueryRow(context.Background(),
+		"SELECT config_proto FROM flows WHERE name = $1", flowJobName).Scan(&configBytes)
+	if err != nil {
+		logrus.Warnf("unable to query qrep config from catalog: %s", err.Error())
+		return nil
+	}
+
+	err = proto.Unmarshal(configBytes, &config)
+	if err != nil {
+		return nil
```
Amogh-Bharadwaj, line 216: Any reason why we don't want to return an error here like in the above function?
```go
if err != nil {
    return nil, fmt.Errorf("unable to unmarshal qrep config: %w", err)
}
```

Amogh-Bharadwaj (approved):
lgtm

flow/cmd/mirror_status.go
iskakaushik: in case where initial load is done for cdc we do not add the entry to flows table

PR #507, opened 2023-10-12
CDC to S3/GCS

Amogh-Bharadwaj:
- Mandatory Metadata DB field added for the S3/GCS peer just like in Eventhub.
```sql
CREATE PEER gcs_peer FROM S3 WITH
(
    url = 's3://<gcs/s3_bucket_name>/<prefix>', -- bucket should exist
    access_key_id = '<hmac_key>', -- or AWS equivalent
    secret_access_key = '<hmac_secret>', -- or AWS equivalent
    region = 'auto', -- change this for S3
    endpoint = 'https://storage.googleapis.com', -- or empty for S3
    metadata_db = 'host=<host> port=<port> user=<user> password=<password> database=<database>'
);
```
- Destination File Structure:
Initial Load: One folder per table
CDC is a separate folder

Tests added for both S3 and GCS versions of the peer (CDC)


flow/connectors/s3/s3.go
```diff
@@ -63,6 +76,246 @@ func (c *S3Connector) Close() error {
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	first := true
+	var firstCP int64 = 0
+	tableNameRowsMapping := make(map[string]uint32)
+
+	for _, record := range req.Records.Records {
```
iskakaushik, line 148: don't we do this elsewhere too? can we refactor to share the code?

iskakaushik (approved):

PR #516, opened 2023-10-15
Geospatial support for Snowflake

Amogh-Bharadwaj:
- Supports GEOGRAPHY and GEOMETRY of PostGIS from PostgreSQL to Snowflake - both CDC and QRep
- Supports Postgres' POINT data type to Snowflake
- Adds unique timestamp in prefix for S3 test

### DIFF
--- a/flow/connectors/utils/postgres.go
+++ b/flow/connectors/utils/postgres.go
@@ -6,7 +6,6 @@ import (
 	"net/url"
 
 	"github.com/PeerDB-io/peer-flow/generated/protos"
-	"github.com/PeerDB-io/peer-flow/model/qvalue"
 	"github.com/jackc/pgx/v5/pgxpool"
 )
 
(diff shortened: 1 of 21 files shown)


flow/model/conversion_avro.go
```diff
@@ -75,6 +76,7 @@ func GetAvroSchemaDefinition(
 	nullableFields := map[string]bool{}
 
 	for _, qField := range qRecordSchema.Fields {
+		log.Infof("qField name: %s, qField type: %s", qField.Name, qField.Type)
```
iskakaushik, line 79: remove log

flow/model/column.go
```diff
@@ -0,0 +1,6 @@
+package model
+
+type ColumnInformation struct {
```
iskakaushik, line 3: add comments on what column map example would look like

flow/connectors/utils/postgres.go
```diff
@@ -20,3 +23,30 @@ func GetPGConnectionString(pgConfig *protos.PostgresConfig) string {
...
+	defer pool.Close()
+
+	err = pool.QueryRow(ctx, "SELECT typname FROM pg_type WHERE oid = $1", dataType).Scan(&typeName)
+	if err != nil {
+		return qvalue.QValueKindString, fmt.Errorf("failed to query pg_type for custom type handling: %w", err)
+	}
+
+	var qValueKind qvalue.QValueKind
+	switch typeName {
+	case "geometry":
```
iskakaushik, line 44: "point"?

iskakaushik (commented):
lets store the typid to typename in the constructor rather than passing around the ConnStr, and pass around the map as needed.

flow/connectors/utils/postgres.go
Amogh-Bharadwaj: Point in PostGIS comes under GEOMETRY. Example : `GEOMETRY('POINT(1 2)')`, like `GEOMETRY('LINESTRING(1 2, 3 4)')`
The point data type I'm adding support for in this PR is Postgres' point type

PR #531, opened 2023-10-18
add scripts for running PeerDB using prebuilt artifacts and development

### LINKED ISSUE #465: [docker] add scripts in repo to deploy from prebuilt artifacts instead of from source

heavycrystal:
closes #465

### DIFF
--- a/run-peerdb.sh
+++ b/run-peerdb.sh
@@ -8,4 +8,4 @@ then
 fi
 
 docker compose pull
-docker compose -f docker-compose.yml up
+docker compose -f docker-compose.yml up --no-attach catalog --no-attach temporal --no-attach temporal-ui --no-attach temporal-admin-tools


dev-peerdb.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+if ! command -v docker &> /dev/null
+then
+    echo "docker could not be found on PATH"
+    exit 1
+fi
+
+docker compose -f docker-compose-dev.yml up --build
```
iskakaushik, line 10: always have empty lines at end of files

iskakaushik:
@heavycrystal  can we leverage https://docs.docker.com/compose/multiple-compose-files/include/ and move temporal and catalog service definition to a separate file.

iskakaushik (approved):

PR #555, opened 2023-10-22
remove code for int,timestamp partitioning using min,max values

heavycrystal:
Only mechanism for partitioning will be `num_rows_per_partition` which is now a mandatory option. Unrecognized options in QRep mirror commands will now cause an error from the SQL interface layer.


iskakaushik:
@heavycrystal  can you let the protos exist as is, but remove the usage of it?

iskakaushik (approved):

Issue #439, opened 2023-09-27
Take in metadata_schema for snowflake peer


Amogh-Bharadwaj:
For the CREATE PEER command for Snowflake, let's take in a parameter called `metadata_schema`. Mirrors will use this value instead of `PUBLIC` schema.

```sql
CREATE PEER sf_test FROM SNOWFLAKE WITH(
  account_id= "...",
  ...
  metadata_schema="test_schema"
)
```

PR #560, opened 2023-10-23
added metadata_schema option for SF and PG peers

heavycrystal:
All metadata tables and Snowflake stages for CDC and QRep mirrors will be created in the provided schema, schema will be auto-created. Default is `_peerdb_internal`

fixes: https://github.com/PeerDB-io/peerdb/issues/439


iskakaushik:
I need to figure out a migration strategy for existing users whose metadata is in public schema for QRep mirrors. Let us hold off on landing this right away.

iskakaushik (approved):

PR #572, opened 2023-10-25
Better peer checks

### LINKED ISSUE #482: Create PEER SF check if database exists

### LINKED ISSUE #484: Refactor Peer Validation To Use API Validate Peer
We are maintaining a ValidatePeer endpoint for our UI in GRPC backend. Let's hit that in the Rust side as well for all peers to check their validity before creating peer.

Amogh-Bharadwaj:
Fixes #482 
Fixes #484


iskakaushik (approved):

flow/connectors/s3/s3.go
```diff
@@ -141,8 +141,8 @@ func ValidCheck(s3Client *s3.S3, bucketURL string, metadataDB *metadataStore.Pos
 }
 
 func (c *S3Connector) ConnectionActive() bool {
-	_, err := c.client.ListBuckets(nil)
-	return err == nil
+	_, listErr := c.client.ListBuckets(nil)
```
heavycrystal, line 144: This doesn't seem like a strong enough check for the connection being active.
Something along the lines of the `ValidCheck` function seems better, the bucket seems to be constant anyway.

nexus/server/src/main.rs
```diff
@@ -209,50 +208,36 @@ impl NexusBackend {
         }
     }
 
-    async fn validate_peer<'a>(&self, peer_type: i32, peer: &Peer) -> anyhow::Result<()> {
-        if peer_type != 6 {
-            let peer_executor = self.get_peer_executor(peer).await.map_err(|err| {
+    async fn validate_peer<'a>(&self, peer: &Peer) -> anyhow::Result<()> {
+        //if flow handler does not exist, skip validation
+        if self.flow_handler.is_none() {
+            return Ok(());
```
heavycrystal, line 214: Should this always happen? A lot of our peer types are not usable from the query layer and having flow not validate them doesn't seem strong enough.

heavycrystal (changes requested):

heavycrystal:
If Snowflake/BigQuery are using an external stage, `ConnectionActive()` should validate that as well.

flow/connectors/s3/s3.go
Amogh-Bharadwaj: The ValidCheck function is performed as part of validation

iskakaushik (approved):

heavycrystal (approved):

PR #578, opened 2023-10-26
store sync state in catalog as well

iskakaushik:
- this lets catalog sync state to be ahead of target sync state and flush the lsn at source


iskakaushik:
cc: @heavycrystal

PR #589, opened 2023-10-29
mixed case support for CDC mirrors in Snowflake and Postgres

heavycrystal:


flow/connectors/postgres/client.go
```diff
@@ -497,6 +497,7 @@ func (c *PostgresConnector) generateFallbackStatements(destinationTableIdentifie
 		}
 	}
 	flattenedCastsSQL := strings.TrimSuffix(strings.Join(flattenedCastsSQLArray, ","), ",")
+	parsedDstTable, _ := utils.ParseSchemaTable(destinationTableIdentifier)
```
serprex, line 500: put this at top before `normalizedTableSchema` would probably run into a nil dereference & handle errors

flow/connectors/postgres/client.go
```diff
@@ -529,6 +530,7 @@ func (c *PostgresConnector) generateMergeStatement(destinationTableIdentifier st
 	for i, columnName := range columnNames {
 		columnNames[i] = fmt.Sprintf("\"%s\"", columnName)
 	}
+	parsedDstTable, _ := utils.ParseSchemaTable(destinationTableIdentifier)
```
serprex, line 533: On one level it feels like this could be made more efficient with a `utils.SanitizeSchemaTable` where it'd avoid allocations in case where `destinationTableIdentifier == parsedDstTable.String()` but in another level it seems like you're already too late here if for stranger names like those containing a period. But maybe character stripping avoids that for now

flow/connectors/snowflake/client.go
```diff
@@ -82,3 +85,18 @@ func (c *SnowflakeConnector) getTableCounts(tables []string) (int64, error) {
 	}
 	return totalRecords, nil
 }
+
+func snowflakeIdentifierNormalize(identifier string) string {
+	// https://www.alberton.info/dbms_identifiers_and_case_sensitivity.html
+	// Snowflake follows the SQL standard, but Postgres does the opposite.
+	// Ergo, we suffer.
+	if strings.ToLower(identifier) == identifier {
```
serprex, line 93: A `utils.IsLower(string)` / `utils.IsUpper(string)` could be implemented using `unicode` package's `IsLetter` / `IsUpper` / `IsLower`

https://pkg.go.dev/unicode

flow/connectors/snowflake/qrep_avro_consolidate_handler.go
```diff
@@ -0,0 +1,259 @@
...
+
+	appendMode := true
+	if config.WriteMode != nil {
+		writeType := config.WriteMode.WriteType
+		if writeType == protos.QRepWriteType_QREP_WRITE_MODE_UPSERT {
+			appendMode = false
+		}
+	}
+
+	switch appendMode {
```
serprex, line 242:
```suggestion
	if appendMode {
```

PR #590, opened 2023-10-29
CDC stream

iskakaushik:
- this is not ready to land
- tests have been deleted, need to re-add them


saisrirampur:
I am very excited about this feature @iskakaushik !! I think it would significantly improve performance both latency/lag and throughput. I did some initial tests and here are the results:

1. **Scale test** - 10-12k TPS - lag looks good - 1-5 seconds
2. Transaction with 1 million rows looks good
3. Transaction with 10 million rows is slowly progressing but isn't full baked. Needs to be tested more thoroughly.

PR #595, opened 2023-10-30
[snowflake] Add basic schema changes support for QRep

iskakaushik:


saisrirampur:
@iskakaushik Works well until i add just 1 column.

But if i add 2 columns before a qrepflow and then add data for both those columns at once, I get this error:

  "message": "failed to sync records: failed to alter destination table: 001003 (42000): SQL compilation error:\nsyntax error line 1 at position 89 unexpected 'COLUMN'.\nsyntax error line 1 at position 103 unexpected 'EXISTS'.",

saisrirampur:
@iskakaushik 

Looks like we might need to handle case better. For example if I add a column with all lower case as this `ALTER TABLE test_xmin_schema_change ADD COLUMN c3 real;`. On snowflake side, you are creating it with lower case "c3" and I was ran in the below issue:
```
peerdb=> SELECT * FROM snowflake_peer.sai_tests.public.test_xmin_schema_change_2 WHERE c3 IS NOT NULL LIMIT 100 ;
ERROR:  Internal error: Unexpected response: 422
{
  "code" : "000904",
  "message" : "SQL compilation error: error line 1 at position 63\ninvalid identifier 'C3'",
  "sqlState" : "42000",
  "statementHandle" : "01b001fc-0001-7679-0000-b17901bd956e"
}
peerdb=> SELECT * FROM snowflake_peer.sai_tests.public.test_xmin_schema_change_2 WHERE "c3" IS NOT NULL LIMIT 100 ;
   ID   | C1 |   C2   | _PEERDB_IS_DELETED | c3
--------+----+--------+--------------------+-----
 332128 |  1 | ernala | false              | 1.4
(1 row)
```
As caps is the default case in Snowflake, we could do something on these lines:
1. If column name is all small case - on snowflake create it with upper case
2. If column name is all upper case - on snowflake create it with upper case
3. If column name is mixed case - create it with mixed case.

iskakaushik:
@saisrirampur -- I fixed the above issues, I am not handling mixed case columns yet, but will get to them as part of follow-up PR.

PR #601, opened 2023-11-01
Column Exclusion

### LINKED ISSUE #428: Add support to exclude columns for tables
Spec, see `exclude_cols`

```
CREATE MIRROR [IF NOT EXISTS] <mirror_name>
FROM <source_peer> TO <target_peer>
WITH TABLE MAPPING
(
  <schema_name>.<source_table_name>:<schema_name>.<target_table_name>,
  <schema_name>.<source_table_name>:<schema_name>.<target_table_name>,
  {
    from: <schema_name>.<source_table_name>,
    to: <schema_name>.<target_table_name>,
    key: <partition_key>,
    exclude_cols: [ <col_1>, <col_2>, <col_3> ]
  }
)
WITH (
  ...
);
```

Expectation:

When a user specifies the columns to exclude, they will be created on the destination but will not be populated / be nulled out.

serprex:
If a mirror is created with exclusions omit those columns from schema of target table

Syntax is in `TABLE MAPPING` of `CREATE MIRROR`, `exclude:[col1,col2]`

Ignores column appearing in schema changes too

Fixes #428


flow/model/model.go
```diff
@@ -12,6 +12,11 @@ import (
 	"github.com/PeerDB-io/peer-flow/model/qvalue"
 )
 
+type NameAndExclude struct {
+	Name string
+	Exclude []string
```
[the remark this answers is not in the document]
serprex: `map[string]struct{}``

flow/connectors/postgres/postgres.go
```diff
@@ -782,9 +782,16 @@ func (c *PostgresConnector) SetupReplication(signal *SlotSignal, req *protos.Set
 		return fmt.Errorf("error checking for replication slot and publication: %w", err)
 	}
 
+	tableNameMapping := make(map[string]model.NameAndExclude)
+	for k, v := range req.TableNameMapping {
```
serprex, line 786: req is `protos.SetupReplicationInput`. Could pull exclusions & then populate a new TableNameMapping field to carry exclusions through. But since it works, not sure this is necessary. Maybe useful to somehow stop source from sending excluded columns altogether?

iskakaushik: this seems good for now :-)

iskakaushik (approved):

iskakaushik (changes requested):
@serprex  can we think about how we can test this change?

serprex:
Edgecase: with schema changes, drop excluded column. Expectation: ignore. Then add excluded column back. Expectation: continue excluding?

iskakaushik:
> Edgecase: with schema changes, drop excluded column. Expectation: ignore. Then add excluded column back. Expectation: continue excluding?

Yes, continue excluding. Also currently mirrors are static, I think eventually we would want to make them dynamic, where a user can exclude more columns, but thats a problem for another day.

iskakaushik (approved):

PR #605, opened 2023-11-02
PAUSE MIRROR support

heavycrystal:
`PAUSE MIRROR` supports both CDC and QRep mirrors. This is implemented by signalling the Temporal workflow. A PauseSignal delivered to the workflow should pause it, while a NoopSignal delivered to the workflow should resume it. Signals that don't make sense for the current workflow state are ignored. Currently supported only via the query layer, not from the UI. The following caveats exist for now:

1) QRep mirrors cannot be paused during partition reading or replicating, only after a run completes successfully and the signals are read. This is usually not an issue but could be problematic for QRep runs with a lot of partitions are running.
2) CDC mirrors similarly will only process signals and pause after a pull + sync + normalize cycle is run. Due to 1), a CDC mirror cannot pause during the initial snapshot phase, which is implemented as a QRep mirror internally. This will be marked as unpausable in the UI.


flow/workflows/cdc_flow.go
```diff
@@ -159,16 +145,43 @@ func GetChildWorkflowID(
...
 // CDCFlowWorkflowResult is the result of the PeerFlowWorkflow.
-type CDCFlowWorkflowResult = CDCFlowState
+type CDCFlowWorkflowResult = CDCFlowWorkflowState
+
+func (w *CDCFlowWorkflowExecution) signalHandler(state *CDCFlowWorkflowState, v shared.CDCFlowSignal) {
+	w.logger.Info("received signal - ", v)
+	if v == shared.ShutdownSignal {
+		w.logger.Info("received shutdown signal")
+		state.ActiveSignal = v
+	} else if v == shared.PauseSignal {
```
serprex, line 155: So PauseSignal triggers flipping pause state? Seems like a source of flakiness compared to having explicit Pause/Unpause signal. But then you'd have to consider what to do when both signals are set. Are signals not able to carry boolean data?

flow/workflows/qrep_flow.go
```diff
@@ -254,6 +279,33 @@ func (q *QRepFlowExecution) waitForNewRows(ctx workflow.Context, lastPartition *
 	return nil
 }
 
+func (q *QRepFlowExecution) signalHandler(v shared.CDCFlowSignal) {
```
serprex, line 282: is there some way to share this shutdown/pause logic?

Also seems like `activeSignal` shouldn't be switched from `ShutdownSignal` to `PauseSignal`. So should be checking activeSignal to prevent that sort of state transition

flow/workflows/qrep_flow.go
```diff
@@ -365,6 +394,30 @@ func QRepFlowWorkflow(
 		"Last Partition", lastPartition,
 		"Number of Partitions Processed", numPartitionsProcessed)
 
+	// here, we handle signals after the end of the flow because a new workflow does not inherit the signals
+	// and the chance of missing a signal is much higher if the check is before the time consuming parts run
+	q.receiveAndHandleSignal(ctx)
+	if q.activeSignal == shared.ShutdownSignal {
```
serprex, line 400: Can move this comparison after the PauseSignal comparison & then remove exit handling from being duplicated after PauseSignal for loop

nexus/server/src/main.rs
```diff
@@ -540,6 +540,64 @@ impl NexusBackend {
...
+                    if self.flow_handler.is_none() {
+                        return Err(PgWireError::ApiError(Box::new(PgError::Internal {
+                            err_msg: "flow service is not configured".to_owned(),
+                        })));
+                    }
+
+                    let catalog = self.catalog.lock().await;
+                    tracing::info!("[PAUSE MIRROR] mirror_name: {}, if_exists: {}", flow_job_name, if_exists);
+                    let workflow_details = catalog
+                    .get_workflow_details_for_flow_job(flow_job_name)
```
serprex, line 556: should be indented

flow/workflows/cdc_flow.go
heavycrystal: I agree that the current logic is a little flaky, but at the same time I can't think of an obvious way it will fail, since we only read a signal at a time and we only maintain one state variable. Temporal Signals are able to carry serializable data, but carrying a boolean value on the Pause signal isn't very clear IMO. 

We could also use explicit Pause and Unpause signals and ignore an illegal one [pausing when already paused] but this would require two RPC endpoints exposed. Another consideration is that the UI can prevent illegal state changes by reading the current state of the workflow, but our query layer has no such ability.

flow/workflows/qrep_flow.go
heavycrystal: The data structures associated with `CDCFlow` and `QRepFlow` are different enough for sharing the logic to be not trivial.

I don't think there's a way to transition from `ShutdownSignal` to `PauseSignal`, as we only process one signal at a time and the return from shutdown happens immediately.

serprex: Could have utility function in shared that takes signal value + active signal value & returns new active signal value

flow/workflows/cdc_flow.go
serprex: In the minute while it's asleep one can imagine someone sending redundant pause signals

serprex (approved):

PR #607, opened 2023-11-02
Basic optimizations

serprex:
Prefer `as_ref` over `clone` when possible
Prefer `eq_ignore_ascii_case` over calling `to_lowercase` before comparison
Executor: convert `Arc<Box<T>>` to `Arc<T>`
Prefer passing `&[u8]` instead of `Vec<u8>` to get_config. The postgres library is able to lend out slices from rows, avoiding a copy. Use that to avoid intermediate string before `to_lowercase` too

Found one bug: `t_lower == "Seconds"` would never be true


iskakaushik (approved):
LGTM!

heavycrystal:
good catches!

heavycrystal (approved):

PR #615, opened 2023-11-06
nexus: update sqlparser-rs & pgwire

serprex:
BINARY(_, op, ANY) was changed to { left, op, right } binary op in sqlparser-rs

Cast had a format field added, ignore for now

portal.parameter had param_type added as a parameter: https://github.com/sunng87/pgwire/commit/38f361d9201a19d487dbe37a83add60d55d62ca4

Follow up to #611

### DIFF
--- a/nexus/peer-bigquery/src/ast.rs
+++ b/nexus/peer-bigquery/src/ast.rs
@@ -239,6 +239,26 @@ impl BigqueryAst {
             ControlFlow::<()>::Continue(())
         });
 
+        // flatten ANY to IN operation overall.
+        visit_expressions_mut(query, |node| {
+            if let Expr::AnyOp { left, compare_op, right } = node {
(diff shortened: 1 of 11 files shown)


nexus/peer-bigquery/src/ast.rs
```diff
@@ -233,37 +239,6 @@ impl BigqueryAst {
             ControlFlow::<()>::Continue(())
         });
 
-        // flatten ANY operand in BINARY to IN operation overall.
-        visit_expressions_mut(query, |node| {
-            if let Expr::BinaryOp { left, op, right } = node {
```
serprex, line 238: Maybe this should be rewritten
```
if let Expr::AnyOp { left, op, expr } {
  if let BinaryOperator::Eq = op {
...
```

iskakaushik (approved):

PR #621, opened 2023-11-07
Lint: Better state management in TableMapping

Amogh-Bharadwaj:
Uses one state value throughout the table list component

### DIFF
--- a/ui/app/mirrors/create/handlers.ts
+++ b/ui/app/mirrors/create/handlers.ts
@@ -58,15 +58,17 @@ interface TableMapping {
   partitionKey: string;
 }
 const reformattedTableMapping = (tableMapping: TableMapRow[]) => {
-  const mapping = tableMapping.map((row) => {
-    if (row.selected === true) {
-      return {
-        sourceTableIdentifier: row.source,
-        destinationTableIdentifier: row.destination,
-        partitionKey: row.partitionKey,
-      };
-    }
-  });
+  const mapping = tableMapping
+    .map((row) => {
+      if (row.selected === true) {
+        return {
+          sourceTableIdentifier: row.source,
+          destinationTableIdentifier: row.destination,
+          partitionKey: row.partitionKey,
+        };
+      }
+    })
+    .filter(Boolean);
   return mapping;
 };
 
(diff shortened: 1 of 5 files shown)


ui/app/mirrors/create/tablemapping.tsx
```diff
@@ -34,51 +34,40 @@ const TableMapping = ({
...
-            source: `${schema}.${tableName}`,
-            destination: `${schema}.${tableName}`,
-            partitionKey: '',
-          };
-        }
-      );
-      setRows(tableNames ?? []);
-    } else setRows([]);
+    const newRows = [...rows];
+    newRows.forEach((_, i) => {
```
serprex, line 60:
```js
for (const row of newRows) {
  row.selected = e.currentTarget.checked;
}
```

ui/app/mirrors/create/tablemapping.tsx
```diff
@@ -34,51 +34,40 @@ const TableMapping = ({
...
 
   const handleAddRow = (source: string) => {
-    setRows([...rows, { source, destination: source, partitionKey: '' }]);
-  };
-
-  const handleRemoveRow = (source: string) => {
     const newRows = [...rows];
     const index = newRows.findIndex((row) => row.source === source);
-    if (index >= 0) newRows.splice(index, 1);
+    if (index >= 0) newRows[index].selected = true;
```
serprex, line 45: Problem is `[...rows]` is a shallow copy. So instead of `selected = true` you need to do `newRows[index] = { ...newRows[index], selected: true }`

Alternatively, use `WeakSet`. Instead of setting newRows, you'd have a `WeakSet<Row>` & at these points you'd create a fresh one & have `setSelectedRows(newWeakSet)` instead of `setRows`. Then code would check `weakset.has(row)` to know what's selected

ui/app/mirrors/create/handlers.ts
```diff
@@ -52,13 +52,20 @@ const validateQRepFields = (
...
+  partitionKey: string;
+}
 const reformattedTableMapping = (tableMapping: TableMapRow[]) => {
   const mapping = tableMapping.map((row) => {
-    return {
-      sourceTableIdentifier: row.source,
-      destinationTableIdentifier: row.destination,
-      partitionKey: row.partitionKey,
-    };
+    if (row.selected === true) {
```
serprex, line 62: Need to avoid returning `undefined`, either with filter or by having `mapping = []` & pushing elements in a `for-of`

ui/app/mirrors/create/tablemapping.tsx
Amogh-Bharadwaj: Decided to go with the first option and performing setRows(newRows)

serprex (approved):

PR #622, opened 2023-11-07
[eventhubs] Add more logs and retries

iskakaushik:


flow/connectors/eventhub/eventhub.go
```diff
@@ -203,10 +199,6 @@ func (c *EventHubConnector) SyncRecords(req *model.SyncRecordsRequest) (*model.S
 		shutdown <- true
 	}()
 
-	eventsPerBatch := int(req.PushBatchSize)
```
serprex, line 206: Should `PushBatchSize` be marked deprecated?

iskakaushik: yup, will do!

iskakaushik: but only for eventhub

iskakaushik:
landing for customer mirage

PR #629, opened 2023-11-08
Experiment: replace testify suite.Suite with our own that supports parallelism

serprex:
while we run the packages in parallel, all of that time is stuck on bigquery/snowflake, limiting parallelism to 2

Running testify suites in parallel is trouble: https://github.com/stretchr/testify/issues/187

So remove suite & run tests in parallel

For now testing snowflake. Doesn't include delta test

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -93,7 +93,7 @@ jobs:
 
       - name: run tests
         run: |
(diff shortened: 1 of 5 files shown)


serprex:
So sharing suite was failing because it'd read 20 rows when expecting 10, moved to suite-per-test, but need postgres schema cleanup etc to be per-suite

serprex:
Close in favor of #640

PR #631, opened 2023-11-09
Tests for types

Amogh-Bharadwaj:
- Adds more data types in the types tests for BQ, SF and GCS.  Such as: ENUM, HSTORE, CIDR, CITEXT, LTREE
- Tested Eventhub and S3/GCS manually and checked the hub and AVRO file respectively


serprex:
Maybe failure rooted in part where tests generate a 20mb map of random uuid:uuid pairs

Issue #616, opened 2023-11-06
Support QRep schema changes for BQ as well


iskakaushik:
Similar to how we do it for snowflake.

PR #633, opened 2023-11-09
schema changes for QRep BigQuery

heavycrystal:
Also fixes a logic bug where the loop for processing SchemaDeltas would end processing early if an empty SchemaDelta was found.


flow/connectors/bigquery/bigquery.go
```diff
@@ -217,7 +217,7 @@ func (c *BigQueryConnector) ReplayTableSchemaDeltas(flowJobName string,
 	schemaDeltas []*protos.TableSchemaDelta) error {
 	for _, schemaDelta := range schemaDeltas {
 		if schemaDelta == nil || len(schemaDelta.AddedColumns) == 0 {
-			return nil
+			continue
```
serprex, line 220: It'd be useful to explain why `return nil` becomes `continue` in PR description

flow/connectors/bigquery/bigquery.go
```diff
@@ -1088,6 +1088,16 @@ func (c *BigQueryConnector) SetupNormalizedTables(
 ) (*protos.SetupNormalizedTableBatchOutput, error) {
 	tableExistsMapping := make(map[string]bool)
 	for tableIdentifier, tableSchema := range req.TableNameSchemaMapping {
+		table := c.client.Dataset(c.datasetID).Table(tableIdentifier)
```
serprex, line 1091: Is this code movement only for optimization purposes & not correctness purposes?

flow/connectors/bigquery/bigquery.go
heavycrystal: added in the description, it is a bugfix

flow/connectors/bigquery/bigquery.go
heavycrystal: only for optimization

serprex (approved):

iskakaushik (approved):

iskakaushik:
@heavycrystal  why can there be an empty schema delta?

heavycrystal:
currently there is no possibility for an empty SchemaDelta to be generated, but changed defensively

heavycrystal:
closes #616

PR #645, opened 2023-11-10
QRep UI Improvements

Amogh-Bharadwaj:
- Searchable dropdowns for table, watermark column, schema selection
- Source peers in create mirror are only postgres peers
- Fixes mirror status screen

Partially implements #646


serprex (approved):

ui/package.json
```diff
@@ -49,6 +49,7 @@
     "protobufjs": "^7.2.5",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "react-select": "^5.8.0",
```
serprex, line 52: Why is this necessary given existing dependency on @radix-ui/react-select ?

Amogh-Bharadwaj: We want the select dropdowns to be searchable - which radix ui's react select does not seem to have

serprex: A separate PR should be made making the switch, removing @radix-ui/react-select too

PR #649, opened 2023-11-13
Add validation for mirror name in UI

### LINKED ISSUE #648: Dashes are not allowed accepted in MIRROR names
I kicked off a CDC MIRROR from PG to PG with the mirror name as "sai-test" It goes through via UI but error is in temporal.

image

Either have validation from the UI or allow MIRRORs with `-`

Amogh-Bharadwaj:
Fixes #648 
Validates that mirror name must contain only letters, numbers and underscores

### DIFF
--- a/ui/app/mirrors/create/handlers.ts
+++ b/ui/app/mirrors/create/handlers.ts
@@ -7,7 +7,12 @@ import {
 import { QRepConfig, QRepWriteType } from '@/grpc_generated/flow';
 import { Dispatch, SetStateAction } from 'react';
 import { CDCConfig, TableMapRow } from '../../dto/MirrorsDTO';
-import { cdcSchema, qrepSchema, tableMappingSchema } from './schema';
+import {
+  cdcSchema,
+  flowNameSchema,
+  qrepSchema,
+  tableMappingSchema,
+} from './schema';
 
 const validateCDCFields = (
   tableMapping: TableMapRow[],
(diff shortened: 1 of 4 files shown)


ui/app/mirrors/create/handlers.ts
```diff
@@ -125,10 +133,13 @@ export const handleCreateQRep = async (
...
   xmin?: boolean
 ) => {
-  if (!flowJobName) {
-    setMsg({ ok: false, msg: 'Mirror name is required' });
-    return;
+  const flowNameValid = flowNameSchema.safeParse(flowJobName);
+  if (!flowNameValid.success) {
+    const flowNameErr = flowNameValid.error.issues[0].message;
+    setMsg({ ok: false, msg: flowNameErr });
+    return false;
```
serprex, line 140: why return false?

ui/app/mirrors/create/handlers.ts
```diff
@@ -85,10 +90,13 @@ export const handleCreateCDC = async (
...
   route: RouteCallback
 ) => {
-  if (!flowJobName) {
-    setMsg({ ok: false, msg: 'Mirror name is required' });
-    return;
+  const flowNameValid = flowNameSchema.safeParse(flowJobName);
+  if (!flowNameValid.success) {
+    const flowNameErr = flowNameValid.error.issues[0].message;
+    setMsg({ ok: false, msg: flowNameErr });
+    return false;
```
serprex, line 97: re return false, also here

serprex:
Update your prettier to 3.1 for the ternary indentation to be correct

PR #653, opened 2023-11-13
Allow configurable column names for soft-delete and synced-at

iskakaushik:
In `soft_delete` mode, PeerDB creates a new `soft_delete_col_name` column in each table in the destination. This mode is currently only supported on Snowflake destination. Rows deleted that are deleted (based on the primary key) are not deleted on the destination, rather the column value is set to `true` in the `soft_delete` mode.


protos/flow.proto
```diff
@@ -206,6 +209,10 @@ message SetupNormalizedTableInput {
 message SetupNormalizedTableBatchInput {
   peerdb_peers.Peer peer_connection_config = 1;
   map<string, TableSchema> table_name_schema_mapping = 2;
+
+  // migration related columns
+  string soft_delete_col_name = 4;
```
heavycrystal, line 214: consider using `optional string` instead so we can distinguish between unset and set to empty string easier

go.work
```diff
@@ -0,0 +1,3 @@
+go 1.21.4
```
heavycrystal, line 1: is this supposed to be committed?

flow/connectors/snowflake/snowflake.go
```diff
@@ -1142,7 +1156,8 @@ and updating the other columns.
 6. Repeat steps 1-5 for each unique set of unchanged toast column groups.
 7. Return the list of generated update statements.
 */
-func (c *SnowflakeConnector) generateUpdateStatement(allCols []string, unchangedToastCols []string) []string {
+func (c *SnowflakeConnector) generateUpdateStatement(
+	syncedAtCol string, allCols []string, unchangedToastCols []string) []string {
```
heavycrystal, line 1160: prefer `*string` instead of `string`

flow/connectors/snowflake/snowflake.go
```diff
@@ -903,8 +905,18 @@ func generateCreateTableSQLForNormalizedTable(
...
+		createTableSQLArray = append(createTableSQLArray,
+			fmt.Sprintf(`"%s" BOOLEAN DEFAULT FALSE,`, softDeleteColName))
+	}
+
+	// add a _peerdb_synced column to the normalized table
+	// this is a timestamp column that is used to mark records as synced
+	// default value is the current timestamp (snowflake)
+	if syncedAtColName != "" {
+		createTableSQLArray = append(createTableSQLArray,
+			fmt.Sprintf(`"%s" TIMESTAMP DEFAULT CURRENT_TIMESTAMP,`, syncedAtColName))
```
heavycrystal, line 918: could we use this default instead of explicitly setting in the merge statement?

heavycrystal (changes requested):

flow/e2e/snowflake/snowflake_helper.go
```diff
@@ -139,3 +141,37 @@ func (s *SnowflakeTestHelper) ExecuteAndProcessQuery(query string) (*model.QReco
 func (s *SnowflakeTestHelper) CreateTable(tableName string, schema *model.QRecordSchema) error {
 	return s.testClient.CreateTable(schema, s.testSchemaName, tableName)
 }
+
+// runs a query that returns an int result
+func (s *SnowflakeTestHelper) RunIntQuery(query string) (int, error) {
```
heavycrystal, line 146: could be generalized better as `CountRowsInQuery`

flow/connectors/snowflake/snowflake.go
iskakaushik: in merge statement we have to set it because when we update the column we have to update the timestamp.

serprex:
Please include description of what soft deletion means in PR description

iskakaushik:
> Please include description of what soft deletion means in PR description

Done!

go.work
iskakaushik: apparently this shouldn't be committed: https://go.googlesource.com/proposal/+/master/design/45713-workspace.md

heavycrystal (approved):

saisrirampur:
@iskakaushik - a couple of items:

1. Populate `_PEERDB_SYNCED_AT`/ `synced_at_col_name` during  initial load.
2.Populate `_PEERDB_SYNCED_AT`/ `synced_at_col_name` when `DELETE` (soft delete)

iskakaushik:
After discussing with @saisrirampur , initial load can be moved to its own PR. Doesn't need to block this PR.

PR #662, opened 2023-11-15
[snowflake] Run merges in parallel during normalize flow

iskakaushik:
Required by customer


serprex:
Is this purely perf related? Was this inspired by a customer experience or is this preemptive?

iskakaushik:
> Is this purely perf related? Was this inspired by a customer experience or is this preemptive?

This was identified as a blocker for a customer. They needed lower latencies than the current state of the world. We also identified `MERGE` as the root cause for the problem.

heavycrystal:
decreases overall durability since normalize is not transactional anymore, should be revisited in the future

heavycrystal:
also consider later if this should be configurable on a peer/mirror basis

heavycrystal (approved):

PR #663, opened 2023-11-15
peer-server: connect lazily

serprex:
While running `dev-peerdb.sh` I saw peer-server failing due to grpc connection taking too long. Replace connect_with_retries with tonic's connect_lazy

Also remove health check, if commandline includes grpc endpoint that endpoint should be used, regardless of health

The underlying principle of this PR is that grpc server should be able to come up & down as it pleases & peer-server will use it when it's up (& ideally fail loudly but non-terminally while it's down). I'm not sure this PR achieves that principle tho

### DIFF
--- a/nexus/flow-rs/src/grpc.rs
+++ b/nexus/flow-rs/src/grpc.rs
@@ -1,6 +1,4 @@
-use std::time::Duration;
 
-use anyhow::Context;
 use catalog::WorkflowDetails;
 use pt::{
     flow_model::{FlowJob, QRepFlowJob},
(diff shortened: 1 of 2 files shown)


heavycrystal:
verified by manually killing flow and restarting peerdb-server, all statements that don't need flow API work as expected.

minor issue with misleading logs, this appears even if MIRROR commands are not enabled:

```
peerdb-server         | 2023-11-15T17:11:40.672939Z  INFO Listening on 0.0.0.0:9900
peerdb-server         | 2023-11-15T17:11:40.672955Z  INFO MIRROR commands enabled
```

serprex:
Enabled/Disabled message is whether the argument was passed in. Maybe makes sense to remove logging that since it's just telling the user what command line arguments they passed in

iskakaushik (approved):

PR #668, opened 2023-11-16
S3 Peer UI

Amogh-Bharadwaj:
Screenshot 2023-11-16 at 1 33 12 PM

- Upon choosing the GCS option, the `Region` and `Role ARN` fields will not be shown as they have fixed values for GCS

### DIFF
--- a/flow/connectors/s3/s3.go
+++ b/flow/connectors/s3/s3.go
@@ -3,11 +3,13 @@ package conns3
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	metadataStore "github.com/PeerDB-io/peer-flow/connectors/external_metadata"
 	"github.com/PeerDB-io/peer-flow/connectors/utils"
 	"github.com/PeerDB-io/peer-flow/generated/protos"
 	"github.com/PeerDB-io/peer-flow/model"
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/s3"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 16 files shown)


flow/connectors/s3/s3.go
```diff
@@ -85,6 +93,47 @@ func (c *S3Connector) Close() error {
 	return nil
 }
 
+func ValidCheck(s3Client *s3.S3, bucketURL string, metadataDB *metadataStore.PostgresMetadataStore) error {
+	_, listErr := s3Client.ListBuckets(nil)
+	if listErr != nil {
+		return fmt.Errorf("failed to list buckets: %w", listErr)
+	}
+
+	reader := strings.NewReader("hello world")
```
iskakaushik, line 102: put something like a timestamp string

flow/connectors/s3/s3.go
```diff
@@ -85,6 +93,47 @@ func (c *S3Connector) Close() error {
...
+	if parseErr != nil {
+		return fmt.Errorf("failed to parse bucket url: %w", parseErr)
+	}
+
+	// Write an empty file and then delete it
+	// to check if we have write permissions
+	bucketName := aws.String(bucketPrefix.Bucket)
+	_, putErr := s3Client.PutObject(&s3.PutObjectInput{
+		Bucket: bucketName,
+		Key:    aws.String("peerdb_check"),
```
iskakaushik, line 113: make this a const in the file and call it `__peerdb_check`

ui/app/peers/create/[peerType]/helpers/s3.ts
```diff
@@ -0,0 +1,63 @@
...
+  region: undefined,
+  endpoint: '',
+  metadataDb: {
+    host: '',
+    port: 5432,
+    user: 'postgres',
+    password: '',
+    database: 'postgres',
+    transactionSnapshot: '',
+  },
```
Comment (iskakaushik) [inline ui/app/peers/create/[peerType]/helpers/s3.ts:62]: Don't expose metadata db to the user, just use catalog.

PR #669, opened 2023-11-16
Wait for at least one record or PKM before returning even after timeout

iskakaushik:
Earlier the flow was on connection timeout we would return the currently accumulated records. Now the `continue` and resetting the standby deadline, means that we will continue waiting for newer records in case we don't have any records or PKMs.

The reason we exit at all from `PullRecords` is to checkpoint. We store metadata on the destination corresponding to committed LSN and we do it in batches because data warehouses operate better that way. Without records there's no need to checkpoint and we can wait indefinitely.

fixes: https://github.com/PeerDB-io/peerdb/issues/670


serprex (approved):

flow/connectors/utils/env.go
```diff
@@ -45,3 +45,23 @@ func GetEnvInt(name string, defaultValue int) int {
...
+func GetEnvString(name string, defaultValue string) string {
+	val, ok := GetEnv(name)
+	if !ok {
+		return defaultValue
+	}
+
+	return val
+}
+
+func IsTestEnv() bool {
```
serprex, line 60: This is generally pretty dangerous unless used carefully. Since it means you aren't testing when `!IsTestEnv` & that's the case that matters

iskakaushik: 100% agreed - filed: https://github.com/PeerDB-io/peerdb/issues/672 and will follow-up with fixing the tests. The urgency in landing the PR is for a ongoing issue.

PR #671, opened 2023-11-16
Disable CI linting sqlparser-rs

serprex:
sqlparser-rs linting belongs at the source: https://github.com/PeerDB-io/sqlparser-rs

That way new clippy rules flagging old code in sqlparser-rs don't cause lint fails for this repo


iskakaushik (approved):
seems like there is no other easy way: https://github.com/rust-lang/cargo/issues/5034

PR #680, opened 2023-11-17
Use catalog as external store

Amogh-Bharadwaj:
If metadataDB is set to empty string/undefined, we now use catalog as the metadata postgres instance for S3/GCS and Eventhub CDC

Screenshot 2023-11-17 at 11 14 29 PM

### DIFF
--- a/ui/components/S3Form.tsx
+++ b/ui/components/S3Form.tsx
@@ -1,11 +1,5 @@
 'use client';
-import { PeerConfig } from '@/app/dto/PeersDTO';
-import { postgresSetting } from '@/app/peers/create/[peerType]/helpers/pg';
-import {
-  blankS3Setting,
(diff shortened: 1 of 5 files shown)


ui/components/S3Form.tsx
```diff
@@ -130,75 +113,6 @@ const S3ConfigForm = ({ setter }: S3Props) => {
             />
           );
       })}
-
-      <Label
```
serprex, line 134: This PR makes this optional, why is it removing it from UI?

Amogh-Bharadwaj: Yup good point. Restored the UI for metadata DB which is now toggled via a switch (Screenshot in PR)

iskakaushik (approved):
Looks good!

PR #681, opened 2023-11-17
set synced at for resync table

iskakaushik:


iskakaushik:
landing for customer:rabbit

PR #683, opened 2023-11-17
Use Selector in BoundSelector to wait in parallel

serprex:
BoundSelector is only used by table cloning

Selector does some juggling so that Select waits on all futures, changing BoundSelector to use Selector avoids pathological cases

Imagine parallelism is 4. The time to clone tables is as follows: 15m 1m 2m 3m 4m 5m

Currently, cloning will take 20m (waits for 15m at 4m, then immediately adds 4m & 5m)
Optimally, cloning would take 15m. 4m would start after 1m & another minute later 5m would start

### DIFF
--- a/flow/concurrency/bound_selector.go
+++ b/flow/concurrency/bound_selector.go
@@ -27,8 +27,7 @@ func (s *BoundSelector) SpawnChild(ctx workflow.Context, w interface{}, args ...
 
 	future := workflow.ExecuteChildWorkflow(ctx, w, args...)
 	s.selector.AddFuture(future, func(f workflow.Future) {
-		err := f.Get(ctx, nil)
-		if err != nil {
+		if err := f.Get(ctx, nil); err != nil {
 			s.ferrors = append(s.ferrors, err)
 		}
 	})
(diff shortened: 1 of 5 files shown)


serprex:
This PR is not based on evidence. It could be argued [Selector's complexity](https://github.com/temporalio/sdk-go/blob/8068455da7292612b527f406b7b25c27ce439d97/internal/internal_workflow.go#L1228) isn't worth dealing with. I'm not sure how much parallelism generally occurs. This change is mostly useful if cx table size distribution follows 80/20 rule

heavycrystal:
This is a known performance holdback. The issue is that Selector makes the order of table cloning non-deterministic [tables could take different times to clone across runs] and this makes Temporal unhappy and fails the workflow in some situations.

serprex:
I don't understand why selector/activity would cause indeterminism based on how temporal is supposed to record their effects for replay

serprex:
Reading old code before bound selector, it was iterating a map, which was likely the actual cause of indeterminism errors

serprex:
@iskakaushik asked this to receive an e2e test merging a few tables in parallel before merging

Amogh-Bharadwaj (approved):
Functionally tested with 80+ table initial load:
- Single threaded
- 4 tables in parallel

PR #686, opened 2023-11-20
Send the events as soon as we have hit the batch size limit

iskakaushik:
This is for performance, we want to keep `START_REPLICATION` connection alive for as long as we can. This was observed to reduce the latency overall significantly.


Amogh-Bharadwaj (approved):

serprex:
Is this to reduce latency or reduce failures or both?

PR #697, opened 2023-11-21
nexus/analyzer: remove unnecessary lazy_static

serprex:
Use of Vec can be replaced with slices in QREP_OPTIONS, making it const


iskakaushik (approved):
love this!

PR #704, opened 2023-11-22
patching FullTablePartition for pkey-less tables

heavycrystal:
This is to enable copying of tables which do not have a primary key. The expectation here is that this will be invoked with `initial_copy_only = true`.


iskakaushik (approved):

flow/model/qrecord_stream.go
```diff
@@ -1,6 +1,8 @@
 package model
 
-import "fmt"
```
serprex, line 3: this file doesn't need to be changed

serprex:
I don't understand what this PR is doing. Is it adding a feature (ie, making tables without a primary key work in more places)? Is it cleaning up code?

iskakaushik:
> I don't understand what this PR is doing. Is it adding a feature (ie, making tables without a primary key work in more places)? Is it cleaning up code?

clarified the usecase in the description

PR #709, opened 2023-11-24
[eventhubs] Update last commited LSN on metadata after we flush

iskakaushik:
If the sync flow fails this would make it so that the metadata is more up-to data. Prior to the change the metadata would only be updated once per sync flow, now this is done every 10s, so its more updated.


serprex:
Could the existing metadata update that we were previously doing be removed now that this is handling it more frequently? Thinking in terms of SyncFlow potentially never returning

PR #712, opened 2023-11-24
Remove sync mode fields from UI

Amogh-Bharadwaj:
- Since Sync Mode fields are now defaulted (see: https://github.com/PeerDB-io/peerdb/pull/707), those fields in the UI are now removed.
- Staging path fields appear depending on if the destination is AVRO type.

### DIFF
--- a/ui/app/mirrors/create/cdc/cdc.tsx
+++ b/ui/app/mirrors/create/cdc/cdc.tsx
@@ -3,11 +3,10 @@ import { RequiredIndicator } from '@/components/RequiredIndicator';
 import { QRepSyncMode } from '@/grpc_generated/flow';
 import { DBType } from '@/grpc_generated/peers';
 import { Label } from '@/lib/Label';
-import { RowWithSelect, RowWithSwitch, RowWithTextField } from '@/lib/Layout';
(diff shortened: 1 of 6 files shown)


ui/app/login/page.tsx
```diff
@@ -21,41 +23,90 @@ export default function Login() {
     })
       .then((res) => res.json())
       .then((res) => {
-        if (res.error) setError(res.error);
-        else router.push('/');
+        if (res.error) {
+          if (res.error === 'wrong password')
+            setError(
+              'Your password is incorrect. Please check your password and try again.'
```
serprex, line 29: Can update middleware to return this string as error

ui/components/SidebarComponent.tsx
```diff
@@ -60,7 +60,7 @@ export default function SidebarComponent(props: { logout?: boolean }) {
               />
             </div>
           </div>
-          {props.logout && <Logout />}
+          {<Logout />}
```
serprex, line 63:
```suggestion
          <Logout />
```
The goal here was to only render logout button if auth is enabled

serprex:
Login page cleanup should be split into a separate PR

PR #725, opened 2023-11-27
Reads cert and key as base64 for Temporal Cloud

Amogh-Bharadwaj:


iskakaushik:
landing this to test deployability.

PR #733, opened 2023-11-28
quote the columns for exclusion to preserve case sensitivity

iskakaushik:


iskakaushik:
@heavycrystal  i'm opportunistically landing this for a customer. Can you please verify that this works with column exclusion and camelcase column names like "colNameTest" please?

PR #745, opened 2023-12-01
moved watermark table creation before QRep initialization

heavycrystal:


iskakaushik (approved):

serprex:
What was the rationale for this? I'm guessing it's relevant to xmin-rep branch

PR #747, opened 2023-12-04
Xmin rep

serprex:
QRep based xmin replication has a fault: xmin isn't monotonic & has wraparound issues

Wraparound workaround: grab all records with `0 < age(xmin) <= age(last snapshot xmin)` which'll include some records from before wraparound but so it goes

Logic is copied from qrep code

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -537,11 +537,12 @@ func (a *FlowableActivity) replicateQRepPartition(ctx context.Context,
 					"flowName": config.FlowJobName,
 				}).Errorf("failed to pull records: %v", err)
 				goroutineErr = err
-			}
-			err = a.CatalogMirrorMonitor.UpdatePullEndTimeAndRowsForPartition(ctx, runUUID, partition, numRecords)
-			if err != nil {
-				log.Errorf("%v", err)
-				goroutineErr = err
+			} else {
+				err = a.CatalogMirrorMonitor.UpdatePullEndTimeAndRowsForPartition(ctx, runUUID, partition, numRecords)
+				if err != nil {
+					log.Errorf("%v", err)
+					goroutineErr = err
+				}
 			}
 			wg.Done()
 		}
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -467,7 +467,7 @@ func (a *FlowableActivity) GetQRepPartitions(ctx context.Context,
 	}, nil
 }
 
-// ReplicateQRepPartition replicates a QRepPartition from the source to the destination.
+// ReplicateQRepPartitions spawns multiple ReplicateQRepPartition
 func (a *FlowableActivity) ReplicateQRepPartitions(ctx context.Context,
 	config *protos.QRepConfig,
 	partitions *protos.QRepPartitionBatch,
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -846,12 +846,6 @@ func (a *FlowableActivity) CreateTablesFromExisting(ctx context.Context, req *pr
 	return nil, fmt.Errorf("create tables from existing is only supported on snowflake and bigquery")
 }
 
-func (a *FlowableActivity) XminWaitUntilNewRows(ctx context.Context,
-	config *protos.QRepConfig) error {
-	// TODO
-	return nil
-}
-
 // ReplicateXminPartition replicates a XminPartition from the source to the destination.
 func (a *FlowableActivity) ReplicateXminPartition(ctx context.Context,
 	config *protos.QRepConfig,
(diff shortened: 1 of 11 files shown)


flow/connectors/postgres/qrep.go
```diff
@@ -596,6 +549,42 @@ func (c *PostgresConnector) SetupQRepMetadataTables(config *protos.QRepConfig) e
...
 }
 
+func (c *PostgresConnector) PullXminRecordStream(
+	config *protos.QRepConfig,
+	currentTxid *int64,
+	partition *protos.QRepPartition,
+	stream *model.QRecordStream,
+) (int, error) {
+	query := config.Query
+	// TODO if initial replication uses xmin with empty partition id, need to do full scan,
```
serprex, line 559: TODO

flow/activities/flowable.go
```diff
@@ -845,3 +845,93 @@ func (a *FlowableActivity) CreateTablesFromExisting(ctx context.Context, req *pr
...
+	}
+	defer connectors.CloseConnector(srcConn)
+
+	dstConn, err := connectors.GetQRepSyncConnector(ctx, config.DestinationPeer)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get qrep destination connector: %w", err)
+	}
+	defer connectors.CloseConnector(dstConn)
+
+	log.Info("replicating xmin\n")
```
iskakaushik, line 872: let's log the flowJobName here as well. Should be in the config object.

flow/activities/flowable.go
```diff
@@ -845,3 +845,93 @@ func (a *FlowableActivity) CreateTablesFromExisting(ctx context.Context, req *pr
...
+
+	var stream *model.QRecordStream
+	bufferSize := shared.FetchAndChannelSize
+	var wg sync.WaitGroup
+
+	var goroutineErr error = nil
+
+	stream = model.NewQRecordStream(bufferSize)
```
iskakaushik, line 880: this could just be `stream := `

iskakaushik:
> `0 <= age(xmin) < lastbatchtxid`

@serprex  should'nt this be `0 <= age(xmin) < age(lastbatchtxid)` ?

serprex:
Potential followups:
1. include configurable wait time between scans for small tables
2. be able to have most scans be since last current txid (like original implementation) & only do snapshot scan every few scans

flow/workflows/xmin_flow.go
```diff
@@ -239,7 +238,10 @@ func XminFlowWorkflow(
 		return fmt.Errorf("xmin replication failed: %w", err)
 	}
 
-	state.LastPartition = &protos.QRepPartition{PartitionId: strconv.FormatInt(lastPartition&0xffffffff, 10)}
+	state.LastPartition = &protos.QRepPartition{
+		PartitionId: uuid.New().String(),
```
iskakaushik, line 242: generating a uuid in a workflow is not repeatable and causes state issues, you have to do it as a side-effect, see https://github.com/PeerDB-io/peerdb/blob/main/flow/workflows/cdc_flow.go#L126 for an example side effect.

serprex: Went with reusing q.runUUID which is generated a bit earlier in this code

flow/activities/flowable.go
```diff
@@ -845,3 +845,93 @@ func (a *FlowableActivity) CreateTablesFromExisting(ctx context.Context, req *pr
...
+	}
+	defer connectors.CloseConnector(dstConn)
+
+	log.Info("replicating xmin\n")
+
+	var stream *model.QRecordStream
+	bufferSize := shared.FetchAndChannelSize
+	var wg sync.WaitGroup
+
+	var goroutineErr error = nil
```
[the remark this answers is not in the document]
serprex: The code in question seems like it shouldn't progress when an error is caught earlier on

flow/workflows/xmin_flow.go
iskakaushik: lol, i just fixed it. feel free to force overwrite.

flow/connectors/postgres/qrep_query_executor.go
```diff
@@ -321,6 +321,51 @@ func (qe *QRepQueryExecutor) ExecuteAndProcessQueryStream(
 		return 0, fmt.Errorf("[pg_query_executor] failed to begin transaction: %w", err)
 	}
 
+	totalRecordsFetched, err := qe.ExecuteAndProcessQueryStreamWithTx(tx, stream, query, args...)
+	return totalRecordsFetched, err
+}
+
+func (qe *QRepQueryExecutor) ExecuteAndProcessQueryStreamGettingCurrentTxid(
```
iskakaushik, line 328: nit: consider renaming the method

PR #754, opened 2023-12-05
Decouple sync/normalize flows

serprex:
Previously after each sync we'd pause reading slot to process table schema deltas & normalize

It turns out table schema deltas don't need to be immediately processed before next sync, so instead use a channel to create one normalize flow per sync batch which let's normalize process concurrently, only waiting at end of every 32 syncs


serprex:
This was done in the totally wrong way. #893 is simpler & correct

PR #756, opened 2023-12-05
CDC: Check for existing table in PG, SF, BQ

### LINKED ISSUE #256: Table matching needs to be uniform across streaming query and cdc

Amogh-Bharadwaj:
We now check if an existing destination table with the same name as the source table has the same schema as the source (ignoring _PEERDB columns we create for snowflake mirrors)
Fixes #256


Amogh-Bharadwaj:
This feature is more complicated than originally pipelined due to newer features related to PeerDB columns, schema changes and so on. Closing with the aim of building this check on top of a newer PeerDB

PR #762, opened 2023-12-05
Metrics For XMIN

### LINKED ISSUE #761: XMIN metrics aren't showing up in UI
After XMIN flow migration pg -> sf rows are moving but no metrics, might've gotten lost in the refactor.

Amogh-Bharadwaj:
Restores metrics for XMIN mirror.
Since the first partition (first sync) of an XMIN mirror is going to have a nil range, here we make a name-sake range with end as number of records pulled and storing that in stats table.

Fixes #761

### DIFF
--- a/flow/connectors/utils/monitoring/monitoring.go
+++ b/flow/connectors/utils/monitoring/monitoring.go
@@ -306,14 +306,14 @@ func (c *CatalogMirrorMonitor) UpdateStartTimeForPartition(
 	ctx context.Context,
 	runUUID string,
 	partition *protos.QRepPartition,
-	specificTime *time.Time,
+	overrideStartTime *time.Time,
 ) error {
 	if c == nil || c.catalogConn == nil {
 		return nil
 	}
 	startTime := time.Now()
-	if specificTime != nil {
-		startTime = *specificTime
+	if overrideStartTime != nil {
+		startTime = *overrideStartTime
 	}
 
 	_, err := c.catalogConn.Exec(ctx, `UPDATE peerdb_stats.qrep_partitions SET start_time=$1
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):

iskakaushik (approved):

serprex (approved):

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -306,13 +306,18 @@ func (c *CatalogMirrorMonitor) UpdateStartTimeForPartition(
...
 	partition *protos.QRepPartition,
+	overrideStartTime *time.Time,
 ) error {
 	if c == nil || c.catalogConn == nil {
 		return nil
 	}
+	startTime := time.Now()
+	if overrideStartTime != nil {
+		startTime = *overrideStartTime
+	}
```
serprex, line 317:
```suggestion
        var startTime Time
	if overrideStartTime == nil {
		startTime := time.Now()
	} else {
		startTime = *overrideStartTime
	}
```
avoids calling `time.Now()` unnecessarily

serprex (changes requested):

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -306,13 +306,18 @@ func (c *CatalogMirrorMonitor) UpdateStartTimeForPartition(
 	ctx context.Context,
 	runUUID string,
 	partition *protos.QRepPartition,
+	overrideStartTime *time.Time,
```
serprex, line 309:
```suggestion
	overrideStartTime time.Time,
```
alternatively, drop the nil check, caller can call `time.Now()`

serprex (changes requested):

flow/connectors/utils/monitoring/monitoring.go
Amogh-Bharadwaj: Opted for this

serprex (approved):

PR #764, opened 2023-12-06
Migrate to Go 1.21's native slog

Amogh-Bharadwaj:
Uses `slog` and gets rid of `logrus`.
`flowName`, `deploymentUid` and for the most part `partitionId` are now set at the top level function/workflow calls and are accessed by children for logs. They do not need to be specified when logging. The context values are read via logging middleware in `logger/handler.go`.


flow/activities/flowable.go
```diff
@@ -173,20 +173,13 @@ func (a *FlowableActivity) StartFlow(ctx context.Context,
 		return nil, fmt.Errorf("failed to get destination connector: %w", err)
 	}
 	defer connectors.CloseConnector(dstConn)
-
-	log.WithFields(log.Fields{
-		"flowName": input.FlowConnectionConfigs.FlowJobName,
-	}).Infof("initializing table schema...")
+	slog.Info("initializing table schema...", shared.FlowNameKey, ctx.Value(shared.FlowNameKey))
```
serprex, line 176: Seems a bit annoying having to do `x, ctx.Value(x)`

serprex (approved):

flow/activities/flowable.go
iskakaushik: can we have a `logging` package?
It would be great to do: `log.with(ctx).Infof(...)`, and the `with` method takes care of adding the additional fields

serprex:
What made you choose slog over zerolog/zap? Preference for stdlib?

flow/workflows/snapshot_flow.go
```diff
@@ -98,19 +104,13 @@ func (s *SnapshotFlowExecution) cloneTable(
 
 	var childWorkflowID string
 	if err := childWorkflowIDSideEffect.Get(&childWorkflowID); err != nil {
-		logrus.WithFields(logrus.Fields{
-			"flowName":     flowName,
-			"snapshotName": snapshotName,
-		}).Errorf("failed to get child id for source table %s and destination table %s",
-			srcName, dstName)
+		slog.Error(fmt.Sprintf("failed to get child id for source table %s and destination table %s",
```
serprex, line 107:
```suggestion
		slog.ErrorContext(childCtx, fmt.Sprintf("failed to get child id for source table %s and destination table %s",
```

flow/workflows/snapshot_flow.go
```diff
@@ -98,19 +104,13 @@ func (s *SnapshotFlowExecution) cloneTable(
...
+			srcName, dstName), slog.Any("error", err), cloneLog)
 		return fmt.Errorf("failed to get child workflow ID: %w", err)
 	}
 
-	logrus.WithFields(logrus.Fields{
-		"flowName":     flowName,
-		"snapshotName": snapshotName,
-	}).Infof("Obtained child id %s for source table %s and destination table %s",
-		childWorkflowID, srcName, dstName)
+	slog.Info(fmt.Sprintf("Obtained child id %s for source table %s and destination table %s",
```
serprex, line 112: `InfoContext`

flow/workflows/snapshot_flow.go
```diff
@@ -196,8 +193,8 @@ func (s *SnapshotFlowExecution) cloneTables(
 	slotInfo *protos.SetupReplicationOutput,
 	maxParallelClones int,
 ) {
-	logrus.Infof("cloning tables for slot name %s and snapshotName %s",
-		slotInfo.SlotName, slotInfo.SnapshotName)
+	slog.Info(fmt.Sprintf("cloning tables for slot name %s and snapshotName %s",
```
serprex, line 196: `workflow.Context` no good for `InfoContext`?

Amogh-Bharadwaj: Nope. The only time passing `workflow.Context` to children works is when giving it to `ExecuteActivity()` (which I think makes it compatible with `context.Context` internally). Otherwise the types are not the same and `context.WithValue(aWorkflowContext, ..,...)` will throw a type error.

PR #767, opened 2023-12-06
cleanup local Avro files, and fixed directory

heavycrystal:


flow/connectors/snowflake/qrep_avro_sync.go
```diff
@@ -138,28 +139,15 @@ func (s *SnowflakeAvroSyncMethod) SyncQRepRecords(
...
-			if err != nil {
-				log.WithFields(log.Fields{
-					"flowName":         config.FlowJobName,
-					"partitionID":      partition.PartitionId,
-					"destinationTable": dstTableName,
-				}).Errorf("failed to remove temp file %s: %v", localFilePath, err)
-			}
-		}()
-	}
+	defer avroFile.Cleanup()
```
iamKunalGupta, line 146: can there be an error in the middle of writing the file and lead to a file still being left around?
in that case we'll return the error, but not cleanup.

Similar cases below

flow/connectors/bigquery/qrep_avro_sync.go
```diff
@@ -325,58 +328,59 @@ func (s *QRepAvroSyncMethod) writeToStage(
...
+			return 0, fmt.Errorf("failed to write records to Avro file on GCS: %w", err)
+		}
+		avroFile = &avro.AvroFile{
+			NumRecords:      numRecords,
+			StorageLocation: avro.AvroGCSStorage,
+			FilePath:        avroFilePath,
+		}
+	} else {
+		tmpDir := fmt.Sprintf("%s/peerdb-avro-%s", os.TempDir(), s.flowJobName)
+		err := os.MkdirAll(tmpDir, os.ModePerm)
```
serprex, line 351: Not deferring directory cleanup

PR #771, opened 2023-12-07
Central slot collecting function for UI and monitoring

Amogh-Bharadwaj:
## Keeping Track Of The WAL
Screenshot 2023-12-07 at 7 54 39 PM

This PR does the following:
1. Moves the API function which queries `pg_replication_slots` to a PullConnector function for Postgres (in `client.go`).
2. Introduces a table - `peer_slot_size` in catalog, and a function in `monitoring.go` to upsert data into this table.
3. The API function for UI to get slot changes now calls the Postgres function for the same.
4. In `StartFlow` in `flowable.go`, we call the same GetSlot function and put it's results into the `peer_slot_size` table in catalog. Here, we store the slot information **once every 10 minutes or at the end of every sync flow - whichever comes first**.

### What is stored ?
- Slot name
- Peer name 
- Redo LSN
- Restart LSN
- Confirmed Flush LSN
- Size of slot AKA Lag
- Timestamp of recording

### Lag calculation
Lag is now calculated (in MB) as:
```sql
round(pg_current_wal_lsn() - confirmed_flush_lsn) -- (divided by 1024 couple of times)
```

**What are those terms ?**
`pg_current_wal_lsn()` is the current LSN the WAL is at. 
`confirmed_flush_lsn` is where the client has told Postgres that it has read till.

 so our "lag" is the difference between where the WAL is vs where we are at as PeerDB reading the WAL.


nexus/catalog/migrations/V5__monitoring_init.sql
```diff
@@ -49,5 +49,3 @@ CREATE TABLE IF NOT EXISTS peerdb_stats.qrep_partitions (
     metadata JSONB
 );
 
-
```
iskakaushik, line 52: revert these, refinery does a dumb checksum on the files including spaces.

flow/connectors/postgres/client.go
```diff
@@ -197,6 +197,44 @@ func (c *PostgresConnector) checkSlotAndPublication(slot string, publication str
 	}, nil
 }
 
+// GetSlotInfo gets the information about the replication slot size and LSNs
+func (c *PostgresConnector) GetSlotInfo(slotName string) ([]*protos.SlotInfo, error) {
```
heavycrystal, line 201: why does this return an array if it's matching on only one name

iskakaushik: if slotName is empty its returns info for all slots. Add a comment Amogh.

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -246,6 +246,33 @@ func (c *CatalogMirrorMonitor) UpdateEndTimeForQRepRun(ctx context.Context, runU
...
+	ctx context.Context,
+	peerName string,
+	slotInfo *protos.SlotInfo,
+) error {
+	if c == nil || c.catalogConn == nil || slotInfo == nil {
+		return nil
+	}
+
+	_, err := c.catalogConn.Exec(ctx,
+		"INSERT INTO peerdb_stats.peer_slot_size"+
```
heavycrystal, line 259: add an `ON CONFLICT DO NOTHING` since we are adding a pkey

iskakaushik (approved):

PR #778, opened 2023-12-07
basic alerting, for internal use primarily

heavycrystal:
For now, an alerting function logs an entry in an alerts table in catalog, and also dispatches notifications to one or more slack channels. Events are staggered to be 15 minutes apart at least, as long as the same "alert key" is used.

Only logs alerts for slot lag increasing a threshold, configured via the environment variable `PEERDB_SLOT_LAG_MB_ALERT_THRESHOLD`


flow/connectors/postgres/postgres.go
```diff
@@ -79,6 +81,11 @@ func NewPostgresConnector(ctx context.Context, pgConfig *protos.PostgresConfig)
 		metadataSchema = *pgConfig.MetadataSchema
 	}
 
+	vigil, err := evervigil.NewVigil()
```
serprex, line 84: should `NewVigil` take ctx given it's calling `Context.Background()`?

flow/utils/evervigil/ever_vigil.go
```diff
@@ -0,0 +1,99 @@
...
+			notifier.UseServices(slackService)
+		default:
+			return fmt.Errorf("unknown service type: %s", serviceType)
+		}
+		return nil
+	})
+
+	return &EverVigil{
+		notifier:    notifier,
+		catalogPool: catalogPool,
```
serprex, line 63: If this is only happening every 15 minutes is there a point to keeping open a connection? Pool will keep it open. Could reconnect every 15 minutes

iskakaushik:
@heavycrystal  could you limit this PR to just implement the alerting service for slot growth. Lets split out the temporal activity failure alerts.

flow/utils/evervigil/ever_vigil.go
serprex: Yes, but that connection stays open afterwards during these 15 minute intervals between polling

serprex: The point of a pool is to keep connections open to share between callers to avoid connection handshake overhead, some links:
1. https://github.com/jackc/pgx/issues/494
2. https://github.com/jackc/pgx/issues/205
3. `MaxConnIdleTime` defaults to 30 minutes https://github.com/jackc/pgx/blob/ba05097642a802eb5a76343f27a9a3515ed86498/pgxpool/pool.go#L134

flow/utils/evervigil/ever_vigil.go
```diff
@@ -0,0 +1,99 @@
...
+	catalogPool *pgxpool.Pool
+}
+
+type slackServiceConfig struct {
+	AuthToken  string   `json:"auth_token"`
+	ChannelIDs []string `json:"channel_ids"`
+}
+
+func NewVigil() (*EverVigil, error) {
+	catalogPool, catalogErr := catalog.GetCatalogConnectionPoolFromEnv()
```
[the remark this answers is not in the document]
heavycrystal: I'm not sure if pools can be shared across workflows/activities due to the implicit serializing/de-serializing that happens. But worth investigating.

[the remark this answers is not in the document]
serprex: https://community.temporal.io/t/go-sdk-dependency-injection/6823

If you create pool before `RegisterActivity` you can pass pool in there

flow/connectors/postgres/postgres.go
serprex: No, if you want that level of isolation you'll want the monitor to be a separate container

flow/utils/evervigil/ever_vigil.go
```diff
@@ -0,0 +1,120 @@
...
+	return &EverVigil{
+		notifier:    notifier,
+		catalogPool: catalogPool,
+	}, nil
+}
+
+func (ev *EverVigil) Close() {
+	if ev.catalogPool != nil {
+		ev.catalogPool.Close()
+	}
```
serprex, line 79: Catalog pool should never be closed

flow/utils/evervigil/ever_vigil.go
```diff
@@ -0,0 +1,120 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/nikoksr/notify"
+	"github.com/nikoksr/notify/service/slack"
+	"github.com/sirupsen/logrus"
```
serprex, line 13: Should prefer slog over logrus in prep for #764

flow/utils/evervigil/ever_vigil.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/nikoksr/notify"
```
serprex, line 11: This library doesn't seem to support emailing as a service, so why not skip it & use `github.com/slack-go/slack` directly?

flow/connectors/core.go
```diff
@@ -43,9 +43,6 @@ type CDCPullConnector interface {
 	// PullFlowCleanup drops both the Postgres publication and replication slot, as a part of DROP MIRROR
 	PullFlowCleanup(jobName string) error
 
-	// SendWALHeartbeat allows for activity to progress restart_lsn on postgres.
```
serprex, line 46: Is this being removed because it's unused? Should be split out to another PR

iskakaushik:
please rename EverVigil to something like `AlertingService` or something more generic. I like the name, but it isn't obvious what it does.

flow/utils/evervigil/ever_vigil.go
heavycrystal: it does support e-mail, but switched to `slack-go` anyway

heavycrystal:
reworked in #866

PR #784, opened 2023-12-08
Re-add versions.go

iskakaushik:
I missed adding this in a previous commit

### DIFF
--- a/flow/cmd/version.go
+++ b/flow/cmd/version.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"context"
+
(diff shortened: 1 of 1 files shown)


serprex (approved):
when you develop do you have files which you can't stage so that `git add .` can't be used? If so, I've avoided that by putting those sorts of files into `private/`

iskakaushik:
> when you develop do you have files which you can't stage so that `git add .` can't be used? If so, I've avoided that by putting those sorts of files into `private/`

great idea!

PR #787, opened 2023-12-10
Update go modules

serprex:
gosnowflake has handled aws sdk breaking change in a pretty gross manner: https://github.com/snowflakedb/gosnowflake/pull/971

I sure do miss semver


serprex:
@heavycrystal I'm wondering if `putFileToStage` should have `OVERWRITE=true` added to its `PUT` command. gosnowflake 1.7.0 has changed default from OVERWRITE=true to OVERWRITE=false. If the file being PUT would always be the same no overwrite is fine, but if restarts could cause different avro content to be staged, then OVERWRITE=true should be added

https://docs.snowflake.com/en/release-notes/clients-drivers/golang-2023#bcr-behavior-change-release-change
https://docs.snowflake.com/en/sql-reference/sql/put

Amogh-Bharadwaj (approved):

PR #789, opened 2023-12-11
Default types replaced with pgtype wherever possible

Amogh-Bharadwaj:
Also set connection pool limit to 3 for our postgres client

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -20,6 +20,7 @@ import (
 	"github.com/PeerDB-io/peer-flow/shared"
 	"github.com/jackc/pglogrepl"
 	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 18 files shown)


iskakaushik (approved):

serprex:
What's the benefit of this? It seems like an unnecessary intermediary compared to using go's primitive types

iskakaushik:
@serprex  the reason was null handling is handled better with these types. When scanning a null string, we encountered an NPE. So pre-emptively moving to this. It might make more sense to inspect each call to see if the extra type is needed, but given that most of these calls are to secondary to mirror perf, I am inclined to be ok with using these types.

serprex (approved):
should explain more in PR description so information is in git history

PR #790, opened 2023-12-11
Track wal_status in slot info

### LINKED ISSUE #786: Track wal_status also in slot info logging
Maybe alert if it becomes `extended` or `lost`

Amogh-Bharadwaj:
Fixes #786


heavycrystal:
change the migration file to `V14` because another PR is using `V13` or vice versa

heavycrystal (approved):

PR #793, opened 2023-12-11
Avoid having multiple catalog connection pools

serprex:
Connection pools are meant to be shared, so remove pool from monitoring

Also go one step further: cache connection pool from env.go into global This requires never closing pool returned by GetCatalogConnectionPoolFromEnv

Run migrations for CI because `monitoring` logic is now non optional

Opens up reusing pool for #778

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -80,13 +80,14 @@ jobs:
           name: "gcs_creds.json"
           json: ${{ secrets.GCS_CREDS }}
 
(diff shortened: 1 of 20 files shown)


serprex:
Storing pool in global variable might be going too far, open to rolling that change back

iskakaushik:
In `postgres.go` we also have a `replPool` and another pool IIRC. I don't think those need to be pools as well. We can refactor them later though.

Amogh-Bharadwaj (approved):

PR #800, opened 2023-12-12
Allow the ability to connect to Postgres via an SSH tunnel

iskakaushik:


flow/connectors/postgres/ssh_wrapped_pool.go
```diff
@@ -0,0 +1,153 @@
...
+	var err error
+
+	// Establish an SSH connection
+	swpp.sshClient, err = ssh.Dial("tcp", swpp.sshServer, swpp.sshConfig)
+	if err != nil {
+		return err
+	}
+	logrus.Info("SSH connection established")
+	// Automatically pick an available local port
+	localListener, err := net.Listen("tcp", "localhost:0")
```
serprex, line 98:
```suggestion
	localListener, err := net.Listen("unix", tmpFile)
```
Did you evaluate unix domain sockets? They have better perf: https://redis.io/docs/management/optimization/benchmarks

> When the server and client benchmark programs run on the same box, both the TCP/IP loopback and unix domain sockets can be used. Depending on the platform, unix domain sockets can achieve around 50% more throughput than the TCP/IP loopback (on Linux for instance). The default behavior of redis-benchmark is to use the TCP/IP loopback.

iskakaushik: @serprex I reworked this to just modify the `Dial` func, this seems to be what the author recommends as well: https://github.com/jackc/tern/blob/4ca45ec386dd3b7c2b174a16d9a9522bac392b26/ssh_tunnel.go#L21

flow/connectors/utils/ssh.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	if len(authMethods) == 0 {
+		return nil, fmt.Errorf("no authentication methods provided")
+	}
+
+	return &ssh.ClientConfig{
+		User:            user,
+		Auth:            authMethods,
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
```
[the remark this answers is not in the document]
serprex: This should be configured. We can get host key from our existing ssh connection with them. Callback only needs to check host key, not ip

flow/connectors/postgres/ssh_wrapped_pool.go
serprex: Dial's first parameter can be `"unix"`

https://github.com/jackc/tern/blob/4ca45ec386dd3b7c2b174a16d9a9522bac392b26/ssh_tunnel.go#L47

iskakaushik: Got it!

PR #813, opened 2023-12-13
qvalue: pass by value

serprex:
QValue is a string/interface{} pair
string is a pointer/int pair
interface{} is a pointer/pointer pair

Copying four words is cheap, especially when for the most part qvalues are stored in a QRecord which will itself be passed around by pointer

Include a `QValueKindEmpty` which is used in places where previously pointer would be nil

### DIFF
--- a/flow/model/model.go
+++ b/flow/model/model.go
@@ -63,7 +63,7 @@ type Record interface {
 // encoding/gob cannot encode unexported fields
 type RecordItems struct {
 	ColToValIdx map[string]int
-	Values      []*qvalue.QValue
+	Values      []qvalue.QValue
 }
 
 func NewRecordItems() *RecordItems {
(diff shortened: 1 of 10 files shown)


flow/model/model.go
```diff
@@ -63,19 +63,19 @@ type Record interface {
...
+	Values      []qvalue.QValue
 }
 
 func NewRecordItems() *RecordItems {
 	return &RecordItems{
 		ColToValIdx: make(map[string]int),
 		// create a slice of 32 qvalues so that we don't have to allocate memory
 		// for each record to reduce GC pressure
-		Values: make([]*qvalue.QValue, 0, 32),
+		Values: make([]qvalue.QValue, 0, 32),
```
serprex, line 74: This is most worrisome part, where we go from preallocating 256 bytes to 1024 bytes. Given this PR's goal is to reduce overall GC pressure, it might make sense to dial this preallocation down to 8 or 16 elements as a compromise

serprex:
<details>
<summary>No benchmarks, but...</summary>
image</details>

flow/model/model.go
serprex: Looking over with Kaushik, we worked out that it's straight forward to have the length, so no need to pick a hopeful capacity

iskakaushik (approved):
approved modulo the pre-allocation change we discussed!

PR #814, opened 2023-12-13
CDCRecordStream: replace mutex with atomics

serprex:


flow/model/model.go
```diff
@@ -339,39 +337,31 @@ func NewCDCRecordStream() *CDCRecordStream {
...
-
-	return r.firstCheckPointID
+	return r.firstCheckPointID.Load()
 }
 
 func (r *CDCRecordStream) GetLastCheckpoint() (int64, error) {
-	r.checkpointMutex.Lock()
-	defer r.checkpointMutex.Unlock()
-
 	if !r.lastCheckpointSet {
```
serprex, line 361: `lastCheckpointSet` seems kind of funny since it gets written to by `Close` without synchronization. Seems like this struct has a mode between collection/read which would be cleaner as a channel which gets sent first/last checkpoint id at end of pull or something

iskakaushik (approved):

flow/model/model.go
iskakaushik: I agree that channels to communicate most fields in this struct would be a good idea. I only wish channels had a GetOrWait primitive :-)

PR #832, opened 2023-12-16
Remove multierror to fix error in error handling

serprex:
multierror struct has a formatting closure, making it unserializable
Thus, multierror cannot be stored in temporal state

Replace with jsonerr, with fallback logic for non serializable errors


iskakaushik (approved):

serprex:
This is probably overkill, simpler would be to store errors as `[]string`

serprex:
Closing in favor of #833

PR #834, opened 2023-12-17
test enabling all golangci lints

serprex:


serprex:
I learnt how to use golangci locally & enable-all seems like a bad idea based on these experiments, best to apply it piecemeal based on how much sense a linter makes

PR #842, opened 2023-12-18
logs schema deltas to catalog as soon as they are read

heavycrystal:
Currently stores the following info:

```
CREATE TABLE IF NOT EXISTS peerdb_stats.schema_deltas_audit_log (
    id BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
    flow_job_name TEXT NOT NULL,
    read_timestamp TIMESTAMP DEFAULT now(),
    workflow_id TEXT NOT NULL,
    run_id TEXT NOT NULL,
    delta_info JSONB NOT NULL
);
```

delta_info is the marshaled version of the `RelationRecord`. Store happens in the same function where `RelationMessages` are processed.

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -516,6 +516,9 @@ func (p *PostgresCDCSource) processMessage(batch *model.CDCRecordStream, xld pgl
 		if p.relationMessageMapping[msg.RelationID] == nil {
 			p.relationMessageMapping[msg.RelationID] = convertRelationMessageToProto(msg)
 		} else {
+			// RelationMessages don't contain an LSN, so we use current clientXlogPos instead.
+			//nolint:lll
+			// https://github.com/postgres/postgres/blob/8b965c549dc8753be8a38c4a1b9fabdb535a4338/src/backend/replication/logical/proto.c#L670
 			return p.processRelationMessage(currentClientXlogPos, convertRelationMessageToProto(msg))
 		}
 
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -43,6 +43,7 @@ type PostgresCDCSource struct {
 
 	// for storing chema delta audit logs to catalog
 	catalogPool *pgxpool.Pool
+	flowJobName string
 }
 
 type PostgresCDCConfig struct {
@@ -53,7 +54,8 @@ type PostgresCDCConfig struct {
 	SrcTableIDNameMapping  map[uint32]string
 	TableNameMapping       map[string]model.NameAndExclude
 	RelationMessageMapping model.RelationMessageMapping
-	CatalogConnection      *pgxpool.Pool
+	CatalogPool            *pgxpool.Pool
+	FlowJobName            string
 }
 
 // Create a new PostgresCDCSource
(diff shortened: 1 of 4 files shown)


flow/connectors/postgres/cdc.go
```diff
@@ -48,6 +53,7 @@ type PostgresCDCConfig struct {
 	SrcTableIDNameMapping  map[uint32]string
 	TableNameMapping       map[string]model.NameAndExclude
 	RelationMessageMapping model.RelationMessageMapping
+	CatalogConnection      *pgxpool.Pool
```
iskakaushik, line 56: can we call it `CatalogPool` (i'm seeing it as pool in `PostgresCDCSource` but connection here)

flow/connectors/postgres/cdc.go
```diff
@@ -504,7 +513,7 @@ func (p *PostgresCDCSource) processMessage(batch *model.CDCRecordStream, xld pgl
 		if p.relationMessageMapping[msg.RelationID] == nil {
 			p.relationMessageMapping[msg.RelationID] = convertRelationMessageToProto(msg)
 		} else {
-			return p.processRelationMessage(xld.WALStart, convertRelationMessageToProto(msg))
+			return p.processRelationMessage(currentClientXlogPos, convertRelationMessageToProto(msg), flowJobName)
```
iskakaushik, line 516: why did this change from `xld.WALStart` to `currentClientXlogPos` ?

flow/connectors/postgres/cdc.go
iskakaushik: can we store `flowJobName` here as well? It will make it easy to access everywhere downstream.

flow/connectors/postgres/cdc.go
```diff
@@ -747,10 +756,31 @@ func convertRelationMessageToProto(msg *pglogrepl.RelationMessage) *protos.Relat
 	}
 }
 
-// processRelationMessage processes a delete message and returns a TableSchemaDelta
+func (p *PostgresCDCSource) genSchemaDeltaAuditLog(flowJobName string, rec *model.RelationRecord) error {
```
iskakaushik, line 759: call this `auditSchemaDelta` instead.

iskakaushik:
can you split out the clientXLogPos change from the audit change and add tests for the audit change, can be as simple as taking the existing schema changes tests and checking the counts in the audit table.

flow/connectors/postgres/cdc.go
```diff
@@ -747,10 +756,31 @@ func convertRelationMessageToProto(msg *pglogrepl.RelationMessage) *protos.Relat
...
+	runID := activityInfo.WorkflowExecution.RunID
+	recJSON, err := json.Marshal(rec)
+	if err != nil {
+		return fmt.Errorf("failed to marshal schema delta to JSON: %w", err)
+	}
+
+	_, err = p.catalogPool.Exec(p.ctx,
+		`INSERT INTO
+		 peerdb_stats.schema_deltas_audit_log(flow_job_name,workflow_id,run_id,delta_info)
+		 VALUES($1,$2,$3,$4) ON CONFLICT DO NOTHING`,
```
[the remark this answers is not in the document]
heavycrystal: removed

flow/connectors/postgres/cdc.go
heavycrystal: `xld.WALStart` is 0 for `RelationMessages`, so using this instead.

iskakaushik (approved):

serprex (approved):

flow/connectors/postgres/cdc.go
```diff
@@ -504,7 +516,7 @@ func (p *PostgresCDCSource) processMessage(batch *model.CDCRecordStream, xld pgl
 		if p.relationMessageMapping[msg.RelationID] == nil {
 			p.relationMessageMapping[msg.RelationID] = convertRelationMessageToProto(msg)
 		} else {
-			return p.processRelationMessage(xld.WALStart, convertRelationMessageToProto(msg))
+			return p.processRelationMessage(currentClientXlogPos, convertRelationMessageToProto(msg))
```
serprex, line 519: Maybe comment that xld.WALStart is not available to relation message. Are there any docs to reference? Can always link to pg code

PR #846, opened 2023-12-18
Replace QValueKindAvroSchema with interface{}

serprex:
Instead of being either a string or a map[string]string,
create structs for the different avro schema definitions

Remove no longer applicable comments,
along with unused nullable parameter


flow/model/qvalue/avro_converter.go
```diff
@@ -23,104 +35,64 @@ type QValueKindAvroSchema struct {
...
-			},
-		}, nil
+		return "string", nil
 	case QValueKindHStore, QValueKindJSON, QValueKindStruct:
-		return &QValueKindAvroSchema{
-			AvroLogicalSchema: map[string]interface{}{
-				"type":   "string",
-				"values": "string",
-			},
+		return AvroSchemaMap{
```
serprex, line 64: This is the only change in behavior, should be split out to a separate PR or reverted

serprex: https://github.com/PeerDB-io/peerdb/issues/847

iskakaushik (approved):

PR #853, opened 2023-12-19
Remove direct dependency on pkcs1/pkcs8

serprex:
Made while attempting to remove pkcs1 to get around https://people.redhat.com/~hkario/marvin

Unfortunately this'll require changing keys. So I'll look to implement support for pkcs8 encoded keys too so pkcs1 can be avoided

### DIFF
--- a/nexus/peer-snowflake/src/auth.rs
+++ b/nexus/peer-snowflake/src/auth.rs
@@ -6,9 +6,9 @@ use std::{
 use anyhow::Context;
 use base64::prelude::{Engine as _, BASE64_STANDARD};
 use jsonwebtoken::{encode as jwt_encode, Algorithm, EncodingKey, Header};
-use pkcs1::EncodeRsaPrivateKey;
-use pkcs8::{DecodePrivateKey, EncodePublicKey};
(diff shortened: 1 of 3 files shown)


nexus/peer-snowflake/src/auth.rs
```diff
@@ -6,9 +6,9 @@ use std::{
 use anyhow::Context;
 use base64::prelude::{Engine as _, BASE64_STANDARD};
 use jsonwebtoken::{encode as jwt_encode, Algorithm, EncodingKey, Header};
-use pkcs1::EncodeRsaPrivateKey;
-use pkcs8::{DecodePrivateKey, EncodePublicKey};
 use rsa::{RsaPrivateKey, RsaPublicKey};
+use rsa::pkcs1::EncodeRsaPrivateKey;
```
Amogh-Bharadwaj, line 10: Seems like this isn't used

Amogh-Bharadwaj (approved):

nexus/peer-snowflake/src/auth.rs
serprex: Trait needs to be imported to call trait methods

PR #856, opened 2023-12-19
Cancel PullCtx when sync fails

iskakaushik:


flow/activities/flowable.go
```diff
@@ -631,6 +631,7 @@ func (a *FlowableActivity) replicateQRepPartition(ctx context.Context,
 
 	rowsSynced, err := dstConn.SyncQRepRecords(config, partition, stream)
 	if err != nil {
+		pullCancel()
```
serprex, line 634: Why isn't the defer call to pullCancel good enough?

PR #858, opened 2023-12-20
Replication should now work from PG 16 Standbys

saisrirampur:
I tried PeerDB on Read Replicas in PG 16:
1. Initial Load worked as expected
2. SyncFlows were erroring out:
```
ERROR:  recovery is in progress
HINT:  WAL control functions cannot be executed during recovery.
```
The reason was `pg_current_wal_lsn()` cannot be run on a standby. I changed that function to `pg_last_wal_replay_lsn()` if Replica.

### DIFF
--- a/flow/connectors/postgres/client.go
+++ b/flow/connectors/postgres/client.go
@@ -209,7 +209,8 @@ func (c *PostgresConnector) GetSlotInfo(slotName string) ([]*protos.SlotInfo, er
 	}
 	rows, err := c.pool.Query(c.ctx, "SELECT slot_name, redo_lsn::Text,restart_lsn::text,wal_status,"+
 		"confirmed_flush_lsn::text,active,"+
-		"round((pg_current_wal_lsn() - confirmed_flush_lsn) / 1024 / 1024) AS MB_Behind"+
+		"round((CASE WHEN pg_is_in_recovery() THEN pg_last_wal_receive_lsn() ELSE pg_current_wal_lsn() END"+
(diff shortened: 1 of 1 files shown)


saisrirampur:
@Amogh-Bharadwaj I am still seeing issues on the UI because to get slot information because we use `pg_current_wal_lsn() ` in the [GetSlotInfo](https://github.com/PeerDB-io/peerdb/blob/531e1c224a7762a58611eb51ad6d9cac95373a7b/flow/connectors/postgres/client.go#L212) function. That needs separate fixing.

serprex:
Can you try keeping it down to a single query roundtrip using
```sql
CASE WHEN pg_is_in_recovery()
  THEN pg_last_wal_receive_lsn()
  ELSE pg_current_wal_lsn()
END
```

saisrirampur:
Below works @serprex. Just checked on a standby server. This PR introduces `isReplica()` function which might be useful to be present in code otherwise. However I'll leave it to you on deciding the right approach.
```
CASE WHEN pg_is_in_recovery()
  THEN pg_last_wal_receive_lsn()
  ELSE pg_current_wal_lsn()
END
```

serprex:
Tested. Problem: `CREATE PUBLICATION` has to be run on primary, not standby. So a followup will need to be made where mirror can be configured to create publication on a different peer than source. But in the meantime one can create the publication slot for peerdb based on error message & things work after that

serprex (approved):

PR #860, opened 2023-12-20
Refactor replica identity type and primary key column retrieval in Postgres

iskakaushik:
This is to support `USING INDEX` replica identity types in Postgres. Treating the index columns as primary key columns for now as it is the fastest way for us to support replica identity index.

### DIFF
--- a/flow/connectors/postgres/client.go
+++ b/flow/connectors/postgres/client.go
@@ -15,6 +15,7 @@ import (
 	"github.com/jackc/pglogrepl"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/lib/pq/oid"
 	"golang.org/x/exp/maps"
 )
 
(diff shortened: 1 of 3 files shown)


flow/connectors/postgres/client.go
```diff
@@ -92,54 +102,99 @@ func (c *PostgresConnector) getRelIDForTable(schemaTable *utils.SchemaTable) (ui
...
 	var replicaIdentity rune
 	err := c.pool.QueryRow(c.ctx,
 		`SELECT relreplident FROM pg_class WHERE oid = $1;`,
 		relID).Scan(&replicaIdentity)
 	if err != nil {
-		return false, fmt.Errorf("error getting replica identity for table %s: %w", schemaTable, err)
+		return ReplicaIdentityDefault, fmt.Errorf("error getting replica identity for table %s: %w", schemaTable, err)
+	}
+
+	var replicaIdentityMap = map[rune]ReplicaIdentityType{
```
serprex, line 119: Why not make ReplicaIdentityType rune instead of string?

iskakaushik: great idea!

serprex (approved):

PR #862, opened 2023-12-20
Fix: Soft Delete for Bigquery Snapshot

Amogh-Bharadwaj:
BigQuery uses destination table schema to derive AVRO schema. We create two columns on the destination which makes the AVRO writing fail. This PR fixes that by skipping the soft delete column (the synced_at column was already being skipped) for AVRO writing


iskakaushik (approved):

iskakaushik:
landing for customer

PR #866, opened 2023-12-21
basic alerting, refactored to use slack-go instead

heavycrystal:
For now, an alerting function logs an entry in an alerts table in catalog, and also dispatches notifications to one or more slack channels. Events are staggered to be atleast 15 minutes apart by default, as long as the same "alert key" is used. This delay can be configured via the `PEERDB_ALERTING_GAP_MINUTES` environment variable, and alerts can be disabled by setting it to 0.

Logs alerts in the following 2 cases for now:
1) When computed slot size of a peer exceeds a threshold, which is `5000MB` by default. This limit is configurable via the environment variable `PEERDB_SLOT_LAG_MB_ALERT_THRESHOLD`. Setting it to 0 should disable this type of alerts entirely.
2) When the number of connections from the configured user for a peer exceed a threshold, which is 5 by default. This limit is configurable by the environment variable `PEERDB_PGPEER_OPEN_CONNECTIONS_ALERT_THRESHOLD`. Setting it to 0 should disable this type of alerts entirely.

To configure alerting, insert a row into the `peerdb_stats.alerting_config` table like this:

```
INSERT INTO peerdb_stats.alerting_config(service_type,service_config) VALUES('slack','{"auth_token": "SLACK_AUTH_TOKEN", "channel_ids": ["SLACK_CHANNEL","IDS_IN_ARRAY"]}');
```


flow/activities/flowable.go
```diff
@@ -190,7 +200,13 @@ func (a *FlowableActivity) recordSlotSizePeriodically(
 	slotName string,
 	peerName string,
 ) {
-	timeout := 10 * time.Minute
+	// ensures slot info is logged atleast once per SyncFlow
```
serprex, line 203:
```suggestion
	// ensures slot info is logged at least once per SyncFlow
```

flow/shared/alerting/slack_alert_sender.go
```diff
@@ -0,0 +1,38 @@
...
+
+func newSlackAlertSender(config *slackAlertConfig) *slackAlertSender {
+	return &slackAlertSender{
+		client:     slack.New(config.AuthToken),
+		channelIDs: config.ChannelIDs,
+	}
+}
+
+func (s *slackAlertSender) sendAlert(ctx context.Context, alertTitle string, alertMessage string) error {
+	for _, channelID := range s.channelIDs {
```
iskakaushik, line 28: can you send `PEERDB_DEPLOYMENT_UID`  as well?

flow/shared/alerting/alerting.go
```diff
@@ -0,0 +1,96 @@
...
+	return &Alerter{
+		catalogPool: catalogPool,
+		logger:      slog.Default(),
+	}
+}
+
+// Only raises an alert if another alert with the same key hasn't been raised
+// in the past 15 minutes
+func (a *Alerter) AlertIf(ctx context.Context, alertKey string, alertMessage string) {
+	if a.catalogPool != nil {
```
iskakaushik, line 60: when can it be nil?

flow/activities/flowable.go
```diff
@@ -178,6 +180,14 @@ func (a *FlowableActivity) handleSlotInfo(
 		return err
 	}
 
+	slotLagInMBThreshold := peerdbenv.GetPeerDBSlotLagMBAlertThreshold()
+	if uint32(slotInfo[0].LagInMb) >= slotLagInMBThreshold {
```
iskakaushik, line 184: add a check for if the number of open connections for the peerdb user on postrges is also greater that configurable threshold, make it 5 to start with.

iskakaushik (approved):

flow/shared/alerting/alerting.go
heavycrystal: shouldn't be, but writing it defensively

flow/shared/alerting/slack_alert_sender.go
heavycrystal: done

PR #867, opened 2023-12-21
Soft Delete: Patch update statements

Amogh-Bharadwaj:
Fixes the update statements we generate during normalize merge for bigquery and postgres. Also adds comments for SF,PG and BQ for generating update statements

### DIFF
--- a/flow/connectors/bigquery/merge_statement_generator.go
+++ b/flow/connectors/bigquery/merge_statement_generator.go
@@ -229,6 +229,15 @@ func (m *mergeStmtGenerator) generateUpdateStatements(
 		(_peerdb_deduped._peerdb_record_type != 2) AND _peerdb_unchanged_toast_columns='%s'
 		THEN UPDATE SET %s `, cols, ssep)
 		updateStmts = append(updateStmts, updateStmt)
+		if peerdbCols.SoftDelete && (peerdbCols.SoftDeleteColName != "") {
+			tmpArray = append(tmpArray[:len(tmpArray)-1],
+				fmt.Sprintf("`%s` = TRUE", peerdbCols.SoftDeleteColName))
+			ssep := strings.Join(tmpArray, ", ")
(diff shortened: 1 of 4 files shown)


flow/connectors/bigquery/merge_statement_generator.go
```diff
@@ -229,6 +229,15 @@ func (m *mergeStmtGenerator) generateUpdateStatements(
 		(_peerdb_deduped._peerdb_record_type != 2) AND _peerdb_unchanged_toast_columns='%s'
 		THEN UPDATE SET %s `, cols, ssep)
 		updateStmts = append(updateStmts, updateStmt)
+		if peerdbCols.SoftDelete && (peerdbCols.SoftDeleteColName != "") {
+			tmpArray = append(tmpArray[:len(tmpArray)-1],
```
iskakaushik, line 233: seems like it should be `if else` where if is for soft delete and else is when there is no soft delete.

flow/connectors/postgres/client.go
```diff
@@ -771,6 +771,19 @@ func (c *PostgresConnector) generateUpdateStatement(allCols []string,
 			src._peerdb_record_type=1 AND _peerdb_unchanged_toast_columns='%s'
 			THEN UPDATE SET %s `, cols, ssep)
 		updateStmts = append(updateStmts, updateStmt)
+
+		// generates update statements for the case where updates and deletes happen in the same branch
+		// the backfill has happened from the pull side already, so treat the DeleteRecord as an update
+		// and then set soft-delete to true.
+		if peerdbCols.SoftDelete && (peerdbCols.SoftDeleteColName != "") {
```
iskakaushik, line 778: @heavycrystal  we should fix it for fallback here as well.

iskakaushik: as a future PR.

iskakaushik (approved):

PR #874, opened 2023-12-22
running PG e2e tests on Postgres 12 too

heavycrystal:
to test fallback statements and other workarounds


iskakaushik:
Love this change!

serprex:
This should be done with a matrix to reduce duplication

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

PR #879, opened 2023-12-22
BQ: avoid generating SQL while holding metadata lock

serprex:
Also optimized code to reduce allocations


serprex:
Better than moving work out of a lock: removing the lock. #895

PR #887, opened 2023-12-22
Fix flow err log and cleanup flowable.go

Amogh-Bharadwaj:
A few places we were logging peer names instead of flow name for the `peerdb_stats.flow_errors` table. Also some long lines have been split in this PR


iskakaushik (approved):

iskakaushik:
@Amogh-Bharadwaj  I think the tests might be failing because flow name param is null when inserting in catalog, can you check if context has flow name set in tests, if not can you set it?

PR #891, opened 2023-12-23
Make it a non-fatal error if we are unable to get slot information

iskakaushik:


serprex (approved):
Is `warning: ` needed when it's slog WarnContext?

PR #893, opened 2023-12-24
Normalize concurrently with sync flows

serprex:
Previously after each sync we'd pause reading slot to process table schema deltas & normalize
This has two problems:
1. we want to always be reading slot, we aren't reading slot during normalize
2. merging multiple batches at once can be less expensive

Now NormalizeFlow is created as a child workflow at start of cdc flow & a signal is sent after each sync flow with schema updates. Normalize consumes all signals since it last checked, merging their processing in parallel with sync flows

NormalizeFlow only reads up to the signal's batch id to avoid potentially syncing a batch without its schema. This creates a range `(normid..syncid]` in which normid is always catching up to syncid as we normalize `normid+1` to `syncid`. Normalize logic already handled this, so it goes untouched in this change

`PEERDB_ENABLE_PARALLEL_SYNC_NORMALIZE` needs to be set to true, for now keep this change behind feature flag to avoid potentially increasing data warehouse costs


flow/connectors/postgres/client.go
```diff
@@ -443,24 +443,21 @@ func (c *PostgresConnector) GetLastSyncBatchID(jobName string) (int64, error) {
 	return result.Int64, nil
 }
 
-func (c *PostgresConnector) GetLastSyncAndNormalizeBatchID(jobName string) (*model.SyncAndNormalizeBatchID, error) {
-	var syncResult, normalizeResult pgtype.Int8
+func (c *PostgresConnector) GetLastNormalizeBatchID(jobName string) (int64, error) {
+	var syncBatchID pgtype.Int8
```
Amogh-Bharadwaj, line 447: `normalizeBatchID`?

flow/connectors/postgres/postgres.go
```diff
@@ -514,8 +517,8 @@ func (c *PostgresConnector) NormalizeRecords(req *model.NormalizeRecordsRequest)
 
 	return &model.NormalizeResponse{
 		Done:         true,
-		StartBatchID: batchIDs.NormalizeBatchID + 1,
-		EndBatchID:   batchIDs.SyncBatchID,
+		StartBatchID: normBatchID + 1,
```
Amogh-Bharadwaj, line 520: could we have a comment above briefly explaining why start and end batch IDs are set to these

flow/connectors/postgres/client.go
serprex: renamed to `result` matching `GetLastSyncBatchID`

flow/connectors/postgres/postgres.go
Amogh-Bharadwaj: Or having it in the PR description would help too

serprex: tbh it doesn't really matter, these responses aren't fed back into the system, only used for display

Amogh-Bharadwaj:
Cancelling a SyncFlow (sometimes done for an upgrade) won't affect the normalize flow right?

flow/connectors/postgres/postgres.go
Amogh-Bharadwaj: thanks

serprex:
> Cancelling a SyncFlow (sometimes done for an upgrade) won't affect the normalize flow right?
right

flow/workflows/normalize_flow.go
```diff
@@ -6,63 +6,88 @@ import (
...
+			tableNameSchemaMapping = s.TableNameSchemaMapping
+		}
+	})
+	for !stopLoop {
+		selector.Select(ctx)
+		for !canceled && selector.HasPending() {
+			selector.Select(ctx)
+		}
+		if canceled || (stopLoop && lastSyncBatchID == syncBatchID) {
+			break
```
iskakaushik, line 56: add logs with the exit reason.

flow/workflows/normalize_flow.go
```diff
@@ -6,63 +6,88 @@ import (
...
-	var normalizeResponse *model.NormalizeResponse
-	if err := fStartNormalize.Get(normalizeFlowCtx, &normalizeResponse); err != nil {
-		return nil, fmt.Errorf("failed to flow: %w", err)
+		if !peerdbenv.PeerDBEnableParallelSyncNormalize() {
+			parent := workflow.GetInfo(ctx).ParentWorkflowExecution
+			workflow.SignalExternalWorkflow(
+				ctx,
+				parent.ID,
+				parent.RunID,
+				"SyncDone",
```
iskakaushik, line 83: move this to a const?

iskakaushik (approved):

PR #894, opened 2023-12-24
Cleanup normalize

serprex:
Clean up code, fetch `syncBatchID`/`normalizeBatchID` together

Reduce redundantly setting FlowName on context in cdc_flow


flow/connectors/bigquery/bigquery.go
```diff
@@ -382,29 +382,32 @@ func (c *BigQueryConnector) GetLastSyncBatchID(jobName string) (int64, error) {
 	}
 }
 
-func (c *BigQueryConnector) GetLastNormalizeBatchID(jobName string) (int64, error) {
-	query := fmt.Sprintf("SELECT normalize_batch_id FROM %s.%s WHERE mirror_job_name = '%s'",
+func (c *BigQueryConnector) GetLastSyncAndNormalizeBatchID(jobName string) (int64, int64, error) {
```
iskakaushik, line 385: can we return a struct instead? will be nice to have names.

serprex: Added. Seems a bit wordy when these methods are only called in one place which immediately stores them to a local with a proper name

iskakaushik (approved):

PR #889, opened 2023-12-23
making BQ merge a single statement without temp tables

heavycrystal:

PR #895, opened 2023-12-24
BQ merge: don't be transactional

serprex:
Discussed with Kevin BQ lock contention. He made #889 to remove temp table

Merge is idempotent, so no need to have transaction, which removes need to have advisory lock on catalog


heavycrystal:
LGTM, but deferring to @iskakaushik in case we want the metadata update to be done in a txn as well

heavycrystal (approved):

iskakaushik (approved):

PR #905, opened 2023-12-26
Ignore unique constraint violation during CREATE IF NOT EXISTS

serprex:
https://stackoverflow.com/questions/29900845/create-schema-if-not-exists-raises-duplicate-key-error

Ran into this while looking to make all tests run in parallel


serprex:
We'll probably want to move `isUniqueError` to a utility package eventually, but not in a rush to do so

Amogh-Bharadwaj (approved):

PR #909, opened 2023-12-27
[breaking] BQ SyncRecords now streams properly, code cleanup

### LINKED ISSUE #908: BigQuery: SyncRecords - Have a parallel goroutine like in Snowflake
In bigquery, we are synchronously pushing to a channel in SyncRecordsViaAvro. The channel has a fixed size limit (~1 million)
If the pull records stream has more records than that, we will be blocked at this stage indefinitely.

To solve this, let's follow Snowflake's sync approach where this step is done via a goroutine.

heavycrystal:
### ⚠️ This change can break existing CDC mirrors from Postgres to BigQuery!

Fixes a bug in BigQuery CDC where a batch with number of records greater than `2 ** 20` causes the Avro file generation part to hang. This is because all records were being written to a bounded channel first and then consumed by the Avro writer instead of the 2 operations happening in parallel. With a large number of records, the channel would fill up and block before the records finished writing, leading to the loop deadlocking itself.

Fixed by switching BigQuery record generation to the mechanism used by Snowflake, where the record generation happens in another goroutine and therefore the channel consumption happens in parallel. As part of this change, some code was cleaned up and the BigQuery raw table schema was changed in a breaking manner to be similar to the SF/PG equivalent. Specifically, the column `_peerdb_timestamp` of type `TIMESTAMP` was removed and the column `_peerdb_timestamp_nanos` of type `INTEGER` was renamed to the former. Existing raw tables will need to be fixed up to match this new, simpler schema.

```
ALTER TABLE <...> DROP COLUMN _peerdb_timestamp;
ALTER TABLE <...> RENAME COLUMN _peerdb_timestamp_nanos TO _peerdb_timestamp;
``` 

Closes #908

### DIFF
--- a/flow/connectors/bigquery/bigquery.go
+++ b/flow/connectors/bigquery/bigquery.go
@@ -18,7 +18,6 @@ import (
 	"github.com/PeerDB-io/peer-flow/model"
 	"github.com/PeerDB-io/peer-flow/model/qvalue"
 	"github.com/PeerDB-io/peer-flow/shared"
-	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5/pgxpool"
 
 	"go.temporal.io/sdk/activity"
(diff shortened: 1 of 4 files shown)


Amogh-Bharadwaj (approved):
lgtm

PR #914, opened 2023-12-27
Restore tabs, set default to true for snapshot, link loader

Amogh-Bharadwaj:
- Clicking on the mirror name links in the mirror tables in the mirror page sometimes has a small delay so this PR displays a loader there.
- Sets default snapshot setting to true
- Restores tab persistence on refresh functionality in specific mirror page

### DIFF
--- a/ui/app/mirrors/create/helpers/cdc.ts
+++ b/ui/app/mirrors/create/helpers/cdc.ts
@@ -6,10 +6,11 @@ export const cdcSettings: MirrorSetting[] = [
     stateHandler: (value, setter) =>
       setter((curr: CDCConfig) => ({
         ...curr,
-        doInitialCopy: (value as boolean) || false,
+        doInitialCopy: (value as boolean) || true,
       })),
     tips: 'Specify if you want initial load to happen for your tables.',
     type: 'switch',
+    default: true,
   },
   {
     label: 'Pull Batch Size',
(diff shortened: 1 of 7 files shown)


ui/app/mirrors/create/helpers/cdc.ts
```diff
@@ -6,10 +6,11 @@ export const cdcSettings: MirrorSetting[] = [
     stateHandler: (value, setter) =>
       setter((curr: CDCConfig) => ({
         ...curr,
-        doInitialCopy: (value as boolean) || false,
+        doInitialCopy: (value as boolean) || true,
```
serprex, line 9: This will always be true. Do you mean `?? true`?

serprex (changes requested):

serprex (approved):

ui/app/mirrors/edit/[mirrorId]/cdc.tsx
```diff
@@ -284,12 +289,18 @@ export function CDCMirror({
 
   useEffect(() => {
     if (typeof window !== 'undefined') {
-      setSelectedTab(localStorage?.getItem('mirrortab') || 'tab1');
+      setSelectedTab(
+        parseInt(localStorage?.getItem(LocalStorageTabKey) || '0')
```
Comment (serprex) [inline ui/app/mirrors/edit/[mirrorId]/cdc.tsx:293]:
```suggestion
        parseInt(localStorage?.getItem(LocalStorageTabKey)) | 0
```
This'll avoid being NaN if localStorage has some non-integer truthy string

PR #919, opened 2023-12-28
partial mixed case support for SF

heavycrystal:
Still some edge cases, but should work for basic mixed case tables.
Will revisit this later, need to spend some time with cases.

Some outstanding issues:
1. Destination tables with the same name but different cases can break our `COPY` statement generation.
2. Proper refactor of the QRep code in general.
3. Some issues with columns specified in pure upper case (like `"INFINITE"`) which is the default casing for Snowflake but the opposite for Postgres.


flow/connectors/snowflake/qrep.go
```diff
@@ -296,10 +301,10 @@ func (c *SnowflakeConnector) getColsFromTable(tableName string) (*model.ColumnIn
 	}
 	defer rows.Close()
 
+	var colName pgtype.Text
```
iskakaushik, line 304: why were these moved outside the loop?

serprex: I'd guess the goal is to avoid clearing an empty stack slot each loop, which may also be boxed depending how go analyzes the Scan. This refactor has been appearing in a few previous PRs

serprex:
Could use test cases demonstrating what now wirks

serprex (approved):

PR #921, opened 2023-12-28
[pg] pkeys mixed case support?

heavycrystal:


iskakaushik:
@heavycrystal  can we add a test for this please.

heavycrystal:
fixed in another PR, tests will be added in a forthcoming PR

PR #926, opened 2023-12-28
nexus: fix tests breaking unless single threaded

serprex:
Kind of silly fixing this right after reducing nexus CI down to 2 cores,
but this'll avoid new hires hitting this error when onboarding like I did

### DIFF
--- a/nexus/catalog/src/lib.rs
+++ b/nexus/catalog/src/lib.rs
@@ -41,7 +41,9 @@ impl<'a> Migration<'a> {
             return Err(anyhow!("migration filename must contain __"));
         };
         let Ok(version) = f[..__idx].parse() else {
-            return Err(anyhow!("migration filename must have number between V & __"));
+            return Err(anyhow!(
+                "migration filename must have number between V & __"
+            ));
         };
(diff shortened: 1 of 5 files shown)


nexus/catalog/src/lib.rs
```diff
@@ -24,18 +24,24 @@ pub struct Catalog {
...
+            for migration in migration_report.applied_migrations() {
+                tracing::info!(
+                    "Migration Applied -  Name: {}, Version: {}",
+                    migration.name(),
+                    migration.version()
+                );
+            }
+            Ok(())
+        }
+        Err(err) if err.to_string().contains("duplicate key value violates unique constraint") => Ok(()),
```
serprex, line 42: Problem: this might be too wide in case migration file itself ran into unique constraint violation

Error message I'm looking at:
```
error applying migration V1__initial`, `db error: ERROR: duplicate key value violates unique constraint "refinery_schema_history_pkey"
       DETAIL: Key (version)=(1) already exists.
```
Options:
- more precise string check
- create PR upstream to detect

PR #930, opened 2023-12-29
Retry when WAL segment has not been found

iskakaushik:
[Debezium](https://github.com/debezium/debezium/blob/db0cfea550308e7386413db25115d531a4c0a61c/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java#L584) also does something similar. Awaiting confirmation from Postgres experts on this but so far what we have seen from the field indicates that this error is OK.

This PR also does some minor refactors and cleaning up of error messages that looked very similar.

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"regexp"
 	"time"
 
 	"github.com/PeerDB-io/peer-flow/connectors/utils"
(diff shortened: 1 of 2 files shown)


flow/connectors/postgres/cdc.go
```diff
@@ -324,6 +353,13 @@ func (p *PostgresCDCSource) consumeStream(
 		}
 
 		if errMsg, ok := rawMsg.(*pgproto3.ErrorResponse); ok {
+			if errMsg.Severity == "ERROR" && errMsg.Code == "XX000" {
```
serprex, line 356: No retry limit?

iskakaushik: added retries

serprex (approved):

heavycrystal:
can we also log the segment that is erroring?

iskakaushik:
> can we also log the segment that is erroring?

it's getting logged higher up in the stack

PR #932, opened 2023-12-29
Replace refinery with our own backwards compatible migration code

serprex:
Currently server_tests must be run with --test-threads=1 because of two things:
1. port collision
2. concurrent migrations failing

I explored fixing these in #926. Ports are easily solved, but migrations are a pain

This migration code works around concurrency with two changes from refinery:
1. ignore unique constraint violation from CREATE TABLE IF NOT EXISTS
2. lock migration table so concurrent migrations are serialized

Considered submitting a PR to refinery with these two fixes,
but this simple change was non trivial since they support multiple async/sync database drivers

### DIFF
--- a/nexus/Cargo.lock
+++ b/nexus/Cargo.lock
@@ -429,7 +429,7 @@ dependencies = [
  "serde_json",
  "tar",
  "tempfile",
- "toml 0.8.2",
+ "toml",
  "xz2",
  "zopfli",
 ]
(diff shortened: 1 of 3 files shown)


serprex:
Two things missing here:
1. making sure versions are contiguous (we only check that the applied migrations have matching order)
2. comparing checksum, version/name should be enough

serprex:
@iskakaushik is `refinery-cli` being used to manage customer migrations? if so, I can look to get this upstreamed

serprex:
#926 passes after being rebased onto this PR

serprex:
Spoke with Kaushik, rolling our own database migration code is too far into yak shaving, so will have to get this moved upstream

PR #935, opened 2023-12-29
UI: add tables as unique rows

Amogh-Bharadwaj:
Removes the business of checking if we've expanded a schema and then inserting rows or not
Simpler: uses lodash functions to attain a unique array of objects while inserting table rows on schema expand

### DIFF
--- a/ui/app/mirrors/create/cdc/schemabox.tsx
+++ b/ui/app/mirrors/create/cdc/schemabox.tsx
@@ -7,7 +7,6 @@ import { Label } from '@/lib/Label';
 import { RowWithCheckbox } from '@/lib/Layout';
 import { SearchField } from '@/lib/SearchField';
 import { TextField } from '@/lib/TextField';
-import _ from 'lodash';
 import {
   Dispatch,
   SetStateAction,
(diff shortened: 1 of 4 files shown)


iskakaushik (approved):

serprex:
Use `lodash-es` instead so that tree shaking can reduce bundle size impact

ui/app/mirrors/create/cdc/schemabox.tsx
```diff
@@ -136,14 +134,11 @@ const SchemaBox = ({
...
-        setTablesLoading(true);
-        setSchemaLoadedSet((loaded) => new Set(loaded).add(schemaName));
-        fetchTables(sourcePeer, schemaName, peerType).then((tableRows) => {
-          setRows((value) => [...value, ...tableRows]);
-          setTablesLoading(false);
-        });
-      }
+      setTablesLoading(true);
+      fetchTables(sourcePeer, schemaName, peerType).then((tableRows) => {
+        setRows((value) => _.uniqWith([...value, ...tableRows], _.isEqual));
```
serprex, line 139: Should be straight forward to roll a utility function to achieve this same effect without bringing in lodash dependency. Uniqueness could be keyed on a generated string (maybe join fields with `\n`) avoiding `isEqual` cost. At worst sort & dedupe

serprex: https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L8441
https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L4249

By passing `_.isEqual` lodash is forced to use `O(N*N)` algo

serprex (changes requested):
more efficient algo should be used

serprex (approved):

PR #937, opened 2023-12-29
Maintain column ordering

### LINKED ISSUE #888: Don't mangle column ordering
Currently column order gets corrupted because go maps are not ordered maps

serprex:
Go maps are not ordered maps,
ie iterating a map has non determinstic order

This is causes us to lose ordering info when using using TableSchema.Columns

Instead store columns as two arrays of column names & column types,
which is used when TableSchema.Columns is nil

This way we're backwards compatible while new mirrors will have correct ordering

Fixes #888


protos/flow.proto
```diff
@@ -195,9 +195,11 @@ message CreateRawTableOutput { string table_identifier = 1; }
 
 message TableSchema {
   string table_identifier = 1;
+  // DEPRECATED: eliminate when breaking changes are allowed.
   map<string, string> columns = 2;
   repeated string primary_key_columns = 3;
   bool is_replica_identity_full = 4;
+  repeated string column_name_type = 5;
```
iskakaushik, line 202: create a new message `TypedColumn` with two fields, name and type. That will be more verbose but makes the code more readable IMO.

iskakaushik:
also adding a test would be beneficial here IMO

protos/flow.proto
serprex: Also won't those be boxed per pair? Would rather have two string arrays than that

iskakaushik: two arrays sound good to me, but FYI, this config is passed around and isn't in the performance critical parts of the code.

serprex: Updated with two arrays

serprex:
already being tested by updated tests, there was initially a failure because I hadn't updated getting schema from snowflake to maintain column order

iskakaushik (approved):

saisrirampur:
@serprex I did functional testing for PG, BQ and SF. Looks good!

saisrirampur (approved):

PR #938, opened 2023-12-29
More initial copy only support

iskakaushik:
When InitialCopyOnly is set, we don't need to check for primary keys as well as replication identity.
We also do not need to create a slot.


serprex (approved):

Amogh-Bharadwaj (approved):
Tested functionality and slot existence after

PR #943, opened 2023-12-30
Don't create metadata schema in a transaction

serprex:
When a command errors during a transaction the transaction can no longer progress, it must be rolled back. This is the case when IF NOT EXISTS errors over unique violation

We do not need schema/table creation to be transactional, so remove transaction

Corrects #942 & #905 to be effective

### DIFF
--- a/flow/connectors/postgres/qrep.go
+++ b/flow/connectors/postgres/qrep.go
@@ -477,18 +477,7 @@ func (c *PostgresConnector) SyncQRepRecords(
 
 // SetupQRepMetadataTables function for postgres connector
 func (c *PostgresConnector) SetupQRepMetadataTables(config *protos.QRepConfig) error {
-	createQRepMetadataTableTx, err := c.pool.Begin(c.ctx)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


flow/connectors/postgres/qrep.go
```diff
@@ -477,6 +477,11 @@ func (c *PostgresConnector) SyncQRepRecords(
 
 // SetupQRepMetadataTables function for postgres connector
 func (c *PostgresConnector) SetupQRepMetadataTables(config *protos.QRepConfig) error {
+	err := c.createMetadataSchema()
+	if err != nil {
+		return fmt.Errorf("error creating metadata schema: %w", err)
+	}
+
 	createQRepMetadataTableTx, err := c.pool.Begin(c.ctx)
```
iskakaushik, line 485: same for this as well?

serprex: Updated. Raw table setup remains in a transaction

serprex:
#941 is now passing when rebased onto this PR

iskakaushik (approved):

PR #830, opened 2023-12-15
fixes merge statement casting for JSONB and array types

heavycrystal:
1) Replaces the legacy `JSON_EXTRACT_*` functions with `JSON_VALUE_*` equivalents which fixes quoting issues in certain types.
2) Changes some logging and Pyroscope to be less noisy.

PR #948, opened 2024-01-01
SF JSON (Variant): Remove escaped quotes

### LINKED ISSUE #828: PG to BQ, CDC messes up format of JSONB columns, whereas initial load doesn't
### Setup steps in Postgres for JSONB
```
CREATE TABLE test_jsonb(id serial primary key, j jsonb);
INSERT INTO test_jsonb(j) VALUES ('{"isDemo":false}');
--Kick off mirror now and after initial load insert another row on Postgres
INSERT INTO test_jsonb(j) VALUES ('{"isDemo":false}');
```

### Data in BQ for JSONB
Screen Shot 2023-12-14 at 7 19 48 PM

### Setup steps in Postgres for ARRAY
```
CREATE TABLE test_array(id serial primary key, t text[], i int[]);
INSERT INTO  test_array(t,i) VALUES ('{sai,krishna}'::text[],'{1,2}'::int[]);
--Kick off mirror now and after initial load insert another row on Postgres
INSERT INTO  test_array(t,i) VALUES ('{sai,krishna}'::text[],'{1,2}'::int[]);
```

### Data in BQ for JSONB

Screen Shot 2023-12-14 at 8 25 09 PM

Amogh-Bharadwaj:
Similar to how #830 fixes JSON structure for BigQuery, this one does it for Snowflake by performing a [PARSE_JSON transformation](https://docs.snowflake.com/en/sql-reference/functions/parse_json) in Copy (for initial load) and Merge (CDC) steps

Screenshot 2024-01-01 at 5 23 52 PM

You can now query synced Postgres JSONs by PeerDB to Snowflake Variants using:
```sql
SELECT j:tags FROM mysftable
``` 
Adds a test for this. Also tested manually

Fixes #828

### DIFF
--- a/flow/connectors/sql/query_executor.go
+++ b/flow/connectors/sql/query_executor.go
@@ -410,7 +410,7 @@ func toQValue(kind qvalue.QValueKind, val interface{}) (qvalue.QValue, error) {
 
 	case qvalue.QValueKindJSON:
 		vraw := val.(*interface{})
-		vstring := (*vraw).(string)
+		vstring, _ := (*vraw).(string)
 
 		if strings.HasPrefix(vstring, "[") {
 			// parse the array
--- a/flow/connectors/sql/query_executor.go
+++ b/flow/connectors/sql/query_executor.go
@@ -410,7 +410,11 @@ func toQValue(kind qvalue.QValueKind, val interface{}) (qvalue.QValue, error) {
 
 	case qvalue.QValueKindJSON:
 		vraw := val.(*interface{})
-		vstring, _ := (*vraw).(string)
+		vstring, ok := (*vraw).(string)
+		if !ok {
+			return qvalue.QValue{},
+				fmt.Errorf("failed to obtain string from json, possibly due to null field value")
+		}
 
 		if strings.HasPrefix(vstring, "[") {
 			// parse the array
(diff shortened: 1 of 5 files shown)


flow/connectors/sql/query_executor.go
```diff
@@ -410,7 +410,7 @@ func toQValue(kind qvalue.QValueKind, val interface{}) (qvalue.QValue, error) {
 
 	case qvalue.QValueKindJSON:
 		vraw := val.(*interface{})
-		vstring := (*vraw).(string)
+		vstring, _ := (*vraw).(string)
```
iskakaushik, line 413: should this be `vstring, ok :=`

When this is not ok, should be log something? Also when is this not ok?

iskakaushik:
@Amogh-Bharadwaj  this works for CDC as well right? Does the test cover CDC, if not can we add an assert in one of the existing CDC tests as well?

flow/connectors/sql/query_executor.go
Amogh-Bharadwaj: It can be not ok when you try performing 
```golang
s.sfHelper.ExecuteAndProcessQuery("SELECT j:field_which_is_not_there FROM sf_table")
```
I've pushed a commit which returns an error with explanation

Amogh-Bharadwaj: Changed it to a log instead. We can't error out there in the case of null jsons

Amogh-Bharadwaj:
Added assertions for QRep and CDC @iskakaushik

iskakaushik (approved):

PR #945, opened 2023-12-31
Merge postgres_repl_test into e2e/postgres/qrep_flow_pg_test

serprex:

PR #953, opened 2024-01-01
PostgresConnector replPool: initialize lazily

serprex:
Alternative to #872

Allows #945 to test replPool code path without losing testing of codepaths without replPool

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log/slog"
 	"regexp"
+	"sync"
 	"time"
 
 	"github.com/PeerDB-io/peer-flow/connectors/utils"
@@ -26,6 +27,8 @@ type PostgresConnector struct {
 	ctx                context.Context
 	config             *protos.PostgresConfig
 	pool               *SSHWrappedPostgresPool
+	replConfig         *pgxpool.Config
+	replMutex          sync.Mutex
 	replPool           *SSHWrappedPostgresPool
 	tableSchemaMapping map[string]*protos.TableSchema
 	customTypesMapping map[uint32]string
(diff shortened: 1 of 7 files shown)


flow/connectors/postgres/postgres.go
```diff
@@ -93,18 +83,36 @@ func NewPostgresConnector(ctx context.Context, pgConfig *protos.PostgresConfig,
...
+		replConfig:         replConfig,
+		replPool:           nil,
 		customTypesMapping: customTypeMap,
 		metadataSchema:     metadataSchema,
 		logger:             *flowLog,
 	}, nil
 }
 
-// GetPool returns the connection pool.
+// nil returns the connection pool.
```
heavycrystal, line 94: revert change to comment

flow/connectors/postgres/postgres.go
```diff
@@ -93,18 +83,36 @@ func NewPostgresConnector(ctx context.Context, pgConfig *protos.PostgresConfig,
...
 func (c *PostgresConnector) GetPool() *SSHWrappedPostgresPool {
 	return c.pool
 }
 
+func (c *PostgresConnector) GetReplPool(ctx context.Context) (*SSHWrappedPostgresPool, error) {
+	c.replMutex.Lock()
```
heavycrystal, line 100: why do we need a mutex? afaik connectors are not shared across mirrors and I don't see a way this can be invoked simultaneously by the same mirror either.

flow/connectors/postgres/postgres.go
```diff
@@ -93,18 +83,36 @@ func NewPostgresConnector(ctx context.Context, pgConfig *protos.PostgresConfig,
 		ctx:                ctx,
 		config:             pgConfig,
 		pool:               pool,
-		replPool:           replPool,
+		replConfig:         replConfig,
+		replPool:           nil,
```
heavycrystal, line 87: explicitly set the mutex to nil if needed

heavycrystal (changes requested):

flow/connectors/postgres/postgres.go
serprex: I didn't know if connectors needed to be threadsafe, so I assumed yes. If we're fine explicitly declaring that connectors shouldn't be used concurrently across goroutines or anything this can be pared down

flow/connectors/postgres/postgres.go
serprex: the zero mutex is documented to be an unlocked mutex

heavycrystal (approved):

PR #961, opened 2024-01-02
fixed interactions with soft delete and CDC resync

heavycrystal:
Resync works by performing initial load to a table with the suffix `_resync`, renaming it to the destination table name and then resuming CDC. This ensures maximum availability of the table for the customer as the downtime is scoped to the renaming phase.

Since soft-deleted rows are not present on the source, this can lead to such rows disappearing after resync, as initial load just copies the source tables over.

This is fixed by adding an additional step to `INSERT` rows from the destination table with primary keys not present in the `_resync` table, before the renaming. This ensures all soft-deleted rows are captured and retained after resync. The synced at column is also updated for these rows.

### DIFF
--- a/flow/connectors/bigquery/bigquery.go
+++ b/flow/connectors/bigquery/bigquery.go
@@ -878,31 +878,75 @@ func (c *BigQueryConnector) getRawTableName(flowJobName string) string {
 }
 
 func (c *BigQueryConnector) RenameTables(req *protos.RenameTablesInput) (*protos.RenameTablesOutput, error) {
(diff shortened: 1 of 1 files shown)


serprex:
PR description should explain how interaction was wrong

iskakaushik (approved):
looks good to me, but could you add the description about what behavior is changed here. Something along the lines of:

When renaming a table, we need to preserve soft deletes which wouldn't be recorded .... etc.

serprex (approved):

PR #963, opened 2024-01-02
Geospatial Support for BigQuery

Amogh-Bharadwaj:
Similar to #516 

Leverages https://cloud.google.com/bigquery/docs/geospatial-data to implement syncing of Postgres' POSTGIS types to BigQuery's GEOGRAPHY data type - for both QRep, Initial Load and CDC.
In `qrep_avro_sync`, we now have a function where we can perform transformations of the data on staging table before copying to destination - a feature which was needed here, and makes it easier to support datatypes in QRep for BQ

Tests added for QRep and CDC

### DIFF
--- a/flow/connectors/bigquery/qrep_avro_sync.go
+++ b/flow/connectors/bigquery/qrep_avro_sync.go
@@ -117,7 +117,8 @@ func getTransformedColumns(dstTableMetadata *bigquery.TableMetadata, syncedAtCol
 			continue
 		}
 		if col.Type == bigquery.GeographyFieldType {
-			transformedColumns = append(transformedColumns, fmt.Sprintf("ST_GEOGFROMTEXT(%s) AS `%s`", col.Name, col.Name))
+			transformedColumns = append(transformedColumns,
+				fmt.Sprintf("ST_GEOGFROMTEXT(`%s`) AS `%s`", col.Name, col.Name))
 		} else {
 			transformedColumns = append(transformedColumns, fmt.Sprintf("`%s`", col.Name))
 		}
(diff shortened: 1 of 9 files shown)


iskakaushik (approved):

flow/connectors/bigquery/qrep_avro_sync.go
```diff
@@ -110,6 +110,21 @@ func (s *QRepAvroSyncMethod) SyncRecords(
...
 }
 
+func getTransformedColumns(dstTableMetadata *bigquery.TableMetadata, syncedAtCol string, softDeleteCol string) []string {
+	transformedColumns := make([]string, 0, len(dstTableMetadata.Schema))
+	for _, col := range dstTableMetadata.Schema {
+		if col.Name == syncedAtCol || col.Name == softDeleteCol {
+			continue
+		}
+		if col.Type == bigquery.GeographyFieldType {
+			transformedColumns = append(transformedColumns, fmt.Sprintf("ST_GEOGFROMTEXT(%s) AS `%s`", col.Name, col.Name))
```
serprex, line 120:
```suggestion
			transformedColumns = append(transformedColumns, fmt.Sprintf("ST_GEOGFROMTEXT(`%s`) AS `%s`", col.Name, col.Name))
```

serprex (changes requested):

flow/connectors/bigquery/qrep_avro_sync.go
iskakaushik: nice catch @serprex , @Amogh-Bharadwaj  can we add a test with mixed case geospatial column please?

iskakaushik (changes requested):
nice catch @serprex , @Amogh-Bharadwaj can we add a test with mixed case geospatial column please?

iskakaushik (approved):

serprex (approved):

PR #968, opened 2024-01-03
Clickhouse add protos

pankaj-peerdb:
add protos as a first step for clickhouse connectors

### DIFF
--- a/nexus/analyzer/src/lib.rs
+++ b/nexus/analyzer/src/lib.rs
@@ -11,7 +11,7 @@ use pt::{
     flow_model::{FlowJob, FlowJobTableMapping, QRepFlowJob},
     peerdb_peers::{
         peer::Config, BigqueryConfig, DbType, EventHubConfig, MongoConfig, Peer, PostgresConfig,
-        S3Config, SnowflakeConfig, SqlServerConfig,
+        S3Config, SnowflakeConfig, SqlServerConfig,ClickhouseConfig
     },
 };
 use qrep::process_options;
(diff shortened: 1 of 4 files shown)


nexus/analyzer/src/lib.rs
```diff
@@ -803,6 +803,37 @@ fn parse_db_options(
             let config = Config::EventhubGroupConfig(eventhub_group_config);
             Some(config)
         }
+        DbType::Clickhouse => {
+            let s3_int = opts
```
heavycrystal, line 807: is there a reason we went with an empty string being the default instead of `optional` like with metadata_schema?

protos/peers.proto
```diff
@@ -89,6 +89,16 @@ message S3Config {
   PostgresConfig metadata_db = 7;
 }
 
+message ClickhouseConfig{
+  string host = 1;
+  uint32 port = 2;
+  string user = 3;
+  string password = 4;
+  string database = 5;
+  optional string metadata_schema = 6;
```
heavycrystal, line 98: would appreciate comments to tell what the last 2 fields specify

nexus/analyzer/src/lib.rs
pankaj-peerdb: I have replicated from the snowflake config. As there would be common code dealing with s3 config it is easier to keep it same as that of snowflake.

protos/peers.proto
pankaj-peerdb: sure, i will remove metadata_schema field, its not needed

nexus/analyzer/src/lib.rs
heavycrystal: makes sense

heavycrystal (approved):

PR #972, opened 2024-01-03
Remove go-geos

serprex:
Replace go-geos with go-geom, compare wkb, handle server/client endian mismatch


serprex:
we rely on go-geos stricter validation

PR #975, opened 2024-01-03
Rename SlotSignal to SnapshotSignal and use errgroup for better error handling

iskakaushik:
Also move errors to a channel to make it clearer to wait for errors.

### DIFF
--- a/flow/connectors/postgres/client.go
+++ b/flow/connectors/postgres/client.go
@@ -1,7 +1,6 @@
 package connpostgres
 
 import (
-	"errors"
 	"fmt"
 	"log"
 	"regexp"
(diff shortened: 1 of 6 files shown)


flow/connectors/postgres/client.go
```diff
@@ -353,27 +353,17 @@ func (c *PostgresConnector) createSlotAndPublication(
...
-		if doInitialCopy {
-			e = errors.New("slot already exists")
-		}
-		slotDetails := SlotCreationResult{
-			SlotName:     slot,
-			SnapshotName: "",
-			Err:          e,
-		}
-		signal.SlotCreated <- slotDetails
+		signal.Error <- errors.New(fmt.Sprintf("Replication slot '%s' already exists", slot))
```
serprex, line 366:
```suggestion
		signal.Error <- fmt.Errorf("Replication slot '%s' already exists", slot)
```

serprex (changes requested):

PR #980, opened 2024-01-04
WaitFor

serprex:
`NormalizeFlowCountQuery` is stunting implementing decoupled sync/normalize workflows. So replace it with `WaitFor`

Besides, I just don't like this `ExitAfterRecords` way of doing things
e2e tests are integration tests: implementation should be treated as a black box as much as possible
Temporal has a bunch of capabilities to mock activities so that we can create unit tests for the more intrusive tests that'd be necessary to raise branch coverage etc

`WaitFor` presents the ideal mechanism for testing convergent processes: update source, wait for destination to reflect change

In order to make this change work, however, I needed to use `env.CancelWorkflow` after completing tests since I now want the workflow running indefinitely. It turns out our code doesn't adequately handle cancellation, so implemented that


flow/activities/flowable.go
```diff
@@ -222,6 +222,12 @@ func (a *FlowableActivity) StartFlow(ctx context.Context,
 
 	go a.recordSlotSizePeriodically(errCtx, srcConn, slotNameForMetrics, input.FlowConnectionConfigs.Source.Name)
 
+	shutdown := utils.HeartbeatRoutine(ctx, 10*time.Second, func() string {
```
serprex, line 225: Moved to make sure we start heartbeats before a potential `errGroup.Wait()`

flow/connectors/postgres/cdc.go
```diff
@@ -297,17 +297,19 @@ func (p *PostgresCDCSource) consumeStream(
 			}
 		}
 
-		if (cdcRecordsStorage.Len() >= int(req.MaxBatchSize)) && !p.commitLock {
-			return nil
-		}
+		if !p.commitLock {
+			if cdcRecordsStorage.Len() >= int(req.MaxBatchSize) {
```
serprex, line 301: Cleaned up this to potentially add [activity.GetWorkerStopChannel](https://pkg.go.dev/go.temporal.io/sdk/activity#GetWorkerStopChannel) handling, but in this ended up not helping for e2e testing & so integrating that can be left as something for the future

flow/connectors/postgres/cdc.go
```diff
@@ -340,9 +342,15 @@ func (p *PostgresCDCSource) consumeStream(
 		} else {
 			ctx, cancel = context.WithDeadline(p.ctx, nextStandbyMessageDeadline)
 		}
-
 		rawMsg, err := conn.ReceiveMessage(ctx)
 		cancel()
+
+		utils.RecordHeartbeatWithRecover(p.ctx, "consumeStream ReceiveMessage")
```
serprex, line 348: This is put here to improve speed at which we react to context cancellation, since the error handling below is looking at `p.commitLock` but our context being cancelled isn't something we can ignore

flow/connectors/snowflake/snowflake.go
```diff
@@ -248,8 +248,7 @@ func (c *SnowflakeConnector) getTableSchemaForTable(tableName string) (*protos.T
 		return nil, fmt.Errorf("error querying Snowflake peer for schema of table %s: %w", tableName, err)
 	}
 	defer func() {
-		// not sure if the errors these two return are same or different?
-		err = errors.Join(rows.Close(), rows.Err())
+		err = rows.Close()
```
serprex, line 251: I checked the code, they're the same

flow/e2e/congen.go
```diff
@@ -137,7 +138,15 @@ func SetupPostgres(suffix string) (*pgxpool.Pool, error) {
 func TearDownPostgres(pool *pgxpool.Pool, suffix string) error {
 	// drop the e2e_test schema
 	if pool != nil {
-		err := cleanPostgres(pool, suffix)
+		deadline := time.Now().Add(time.Minute)
```
serprex, line 141: I was unable to figure out how to make `ExecuteWorkflow` wait on activity stopping (not just cancelling), I was unable to figure out any means to poll for activities, but immediately getting to teardown while consumeStream is still active would fail over replication slot still being in use so it can't be dropped

Solution: `WaitFor` everywhere. teardown has a 1 minute retry loop

flow/workflows/cdc_flow.go
```diff
@@ -216,7 +216,8 @@ func CDCFlowWorkflowWithConfig(
 			RetryPolicy: &temporal.RetryPolicy{
 				MaximumAttempts: 20,
 			},
-			SearchAttributes: mirrorNameSearch,
+			SearchAttributes:    mirrorNameSearch,
+			WaitForCancellation: true,
```
serprex, line 220: I set `WaitForCancellation` on everything to try get `CancelWorkflow` to pause longer, but pretty sure it didn't help, need to test with removing this now that rest is working. Not sure if we want this set to true or not now

serprex:
https://github.com/PeerDB-io/peerdb/pull/980/commits/b9843c7d050cbdbc80e48576c8333424881f59b8 passed, but rerunning to check for flakiness

flow/e2e/test_utils.go
```diff
@@ -129,58 +128,78 @@ func EnvEqualTables(env *testsuite.TestWorkflowEnvironment, suite e2eshared.RowS
...
+				if state.CurrentFlowState == protos.FlowStatus_STATUS_RUNNING {
+					return
+				}
+			} else {
 				slog.Error(err.Error())
-			} else if len(state.NormalizeFlowStatuses) >= minCount {
-				break
 			}
-		} else {
+		} else if counter > 15 {
```
Amogh-Bharadwaj, line 193: Is it worth having a parameter called `setupTimeout` or something instead of having a hard value of 15 seconds?

flow/workflows/cdc_flow.go
Amogh-Bharadwaj: Tested it out with cancelling a few sync flows and it's fine to keep it (or remove it - whichever helps more).

Amogh-Bharadwaj (approved):
lgtm

flow/e2e/test_utils.go
serprex: No, because setup should always take the same amount of time

PR #983, opened 2024-01-04
Remove the need for metadata db in eventhub group peer

iskakaushik:
We already use catalog in cases where this isn't passed, but validation failed for eventhub group peer.

### DIFF
--- a/nexus/analyzer/src/lib.rs
+++ b/nexus/analyzer/src/lib.rs
@@ -800,11 +800,15 @@ fn parse_db_options(
     Ok(config)
 }
 
-fn parse_metadata_db_info(conn_str: Option<&str>) -> anyhow::Result<Option<PostgresConfig>> {
+fn parse_metadata_db_info(conn_str: Option<&&str>) -> anyhow::Result<Option<PostgresConfig>> {
     let conn_str = match conn_str {
         Some(conn_str) => conn_str,
         None => return Ok(None),
     };
+
+    if conn_str.is_empty() {
+        return Ok(None);
+    }
 
     let mut metadata_db = PostgresConfig::default();
     let param_pairs: Vec<&str> = conn_str.split_whitespace().collect();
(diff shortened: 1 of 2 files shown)


nexus/analyzer/src/lib.rs
```diff
@@ -808,10 +800,11 @@ fn parse_db_options(
     Ok(config)
 }
 
-fn parse_metadata_db_info(conn_str: &str) -> anyhow::Result<Option<PostgresConfig>> {
-    if conn_str.is_empty() {
-        return Ok(None);
-    }
+fn parse_metadata_db_info(conn_str: Option<String>) -> anyhow::Result<Option<PostgresConfig>> {
```
serprex, line 803:
```suggestion
fn parse_metadata_db_info(conn_str: Option<&str>) -> anyhow::Result<Option<PostgresConfig>> {
```
Then remove `map(|s| s.to_string())` from callers

serprex (changes requested):

serprex (approved):

nexus/analyzer/src/lib.rs
```diff
@@ -800,12 +800,16 @@ fn parse_db_options(
     Ok(config)
 }
 
-fn parse_metadata_db_info(conn_str: Option<&str>) -> anyhow::Result<Option<PostgresConfig>> {
+fn parse_metadata_db_info(conn_str: Option<&&str>) -> anyhow::Result<Option<PostgresConfig>> {
```
serprex, line 803: Caller can change `Option<&&str>` to `Option<&str>` with `.copied()`

iskakaushik: is it better than taking in `Option<&str>` ?

serprex: Yes, because `&&str` isn't really what you want, it's just a side effect of HashMap returning a reference to its `&str` value

iskakaushik: done!

PR #995, opened 2024-01-05
HStore Support for BigQuery

Amogh-Bharadwaj:
Maps HStore to JSON in BigQuery. 
Achieves this by transforming `'"a"=>"b"'` to `{"a":"b"}` via string functions, in both qrep/initial load and cdc
Test added

Relies on the fact hstore key and value must be quoted strings (although the key can be empty and this case is supported)
And follows the syntax in the example above

```sql
postgres=# select 'a'::hstore;
ERROR:  Unexpected end of string
LINE 1: select 'a'::hstore;
               ^
postgres=# select 'a=>b'::hstore;
-[ RECORD 1 ]----
hstore | "a"=>"b"

postgres=# select 'a=>'::hstore;
ERROR:  Unexpected end of string
LINE 1: select 'a=>'::hstore;

postgres=# select 'a=>3434'::hstore;
-[ RECORD 1 ]-------
hstore | "a"=>"3434"
```

### DIFF
--- a/flow/connectors/bigquery/merge_stmt_generator.go
+++ b/flow/connectors/bigquery/merge_stmt_generator.go
@@ -53,11 +53,11 @@ func (m *mergeStmtGenerator) generateFlattenedCTE() string {
 				colName, bqType, shortCol)
 		case qvalue.QValueKindHStore:
 			// PARSE_JSON doesn't work for HSTORE with \" in the value, so use SAFE.PARSE_JSON and coalesce with TO_JSON
-			castStmt = fmt.Sprintf("CAST(COALESCE("+
-				"SAFE.PARSE_JSON(JSON_VALUE(_peerdb_data, '$.%s'))"+
+			castStmt = fmt.Sprintf("CASE WHEN `%s` IS NULL THEN NULL ELSE CAST(COALESCE("+
+				"SAFE.PARSE_JSON(JSON_VALUE(_peerdb_data, '$.%s'), wide_number_mode=>'round')"+
 				",TO_JSON(JSON_VALUE(_peerdb_data, '$.%s'))"+
 				") AS %s) AS `%s`",
-				colName, colName, bqType, shortCol)
+				colName, colName, colName, bqType, shortCol)
 		// expecting data in BASE64 format
 		case qvalue.QValueKindBytes, qvalue.QValueKindBit:
 			castStmt = fmt.Sprintf("FROM_BASE64(JSON_VALUE(_peerdb_data,'$.%s')) AS `%s`",
--- a/flow/connectors/bigquery/merge_stmt_generator.go
+++ b/flow/connectors/bigquery/merge_stmt_generator.go
@@ -47,7 +47,7 @@ func (m *mergeStmtGenerator) generateFlattenedCTE() string {
 		var castStmt string
 		shortCol := m.shortColumn[colName]
 		switch qvalue.QValueKind(colType) {
-		case qvalue.QValueKindJSON:
+		case qvalue.QValueKindJSON, qvalue.QValueKindHStore:
 			// if the type is JSON, then just extract JSON
 			castStmt = fmt.Sprintf("CAST(PARSE_JSON(JSON_VALUE(_peerdb_data, '$.%s'),wide_number_mode=>'round') AS %s) AS `%s`",
 				colName, bqType, shortCol)
--- a/flow/connectors/bigquery/merge_stmt_generator.go
+++ b/flow/connectors/bigquery/merge_stmt_generator.go
@@ -47,17 +47,10 @@ func (m *mergeStmtGenerator) generateFlattenedCTE() string {
 		var castStmt string
 		shortCol := m.shortColumn[colName]
 		switch qvalue.QValueKind(colType) {
-		case qvalue.QValueKindJSON:
+		case qvalue.QValueKindJSON, qvalue.QValueKindHStore:
 			// if the type is JSON, then just extract JSON
 			castStmt = fmt.Sprintf("CAST(PARSE_JSON(JSON_VALUE(_peerdb_data, '$.%s'),wide_number_mode=>'round') AS %s) AS `%s`",
 				colName, bqType, shortCol)
-		case qvalue.QValueKindHStore:
-			// PARSE_JSON doesn't work for HSTORE with \" in the value, so use SAFE.PARSE_JSON and coalesce with TO_JSON
-			castStmt = fmt.Sprintf("CASE WHEN `%s` IS NULL THEN NULL ELSE CAST(COALESCE("+
-				"SAFE.PARSE_JSON(JSON_VALUE(_peerdb_data, '$.%s'), wide_number_mode=>'round')"+
-				",TO_JSON(JSON_VALUE(_peerdb_data, '$.%s'))"+
-				") AS %s) AS `%s`",
-				colName, colName, colName, bqType, shortCol)
 		// expecting data in BASE64 format
 		case qvalue.QValueKindBytes, qvalue.QValueKindBit:
 			castStmt = fmt.Sprintf("FROM_BASE64(JSON_VALUE(_peerdb_data,'$.%s')) AS `%s`",
--- a/flow/connectors/bigquery/merge_stmt_generator.go
+++ b/flow/connectors/bigquery/merge_stmt_generator.go
@@ -47,10 +47,17 @@ func (m *mergeStmtGenerator) generateFlattenedCTE() string {
 		var castStmt string
 		shortCol := m.shortColumn[colName]
 		switch qvalue.QValueKind(colType) {
-		case qvalue.QValueKindJSON, qvalue.QValueKindHStore:
+		case qvalue.QValueKindJSON:
 			// if the type is JSON, then just extract JSON
 			castStmt = fmt.Sprintf("CAST(PARSE_JSON(JSON_VALUE(_peerdb_data, '$.%s'),wide_number_mode=>'round') AS %s) AS `%s`",
 				colName, bqType, shortCol)
+		case qvalue.QValueKindHStore:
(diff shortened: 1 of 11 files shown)


serprex (approved):

flow/connectors/bigquery/merge_statement_generator.go
```diff
@@ -47,6 +47,9 @@ func (m *mergeStmtGenerator) generateFlattenedCTE() string {
 			// if the type is JSON, then just extract JSON
 			castStmt = fmt.Sprintf("CAST(PARSE_JSON(JSON_VALUE(_peerdb_data, '$.%s'),wide_number_mode=>'round') AS %s) AS `%s`",
 				colName, bqType, colName)
+		case qvalue.QValueKindHStore:
+			castStmt = fmt.Sprintf("CAST(PARSE_JSON(CONCAT('{', REPLACE(JSON_VALUE(_peerdb_data, '$.%s'), '=>', ':'), '}')) as %s) AS `%s`",
```
serprex, line 51: This will break if people use `=>` in their hstore. Also `NULL` needs to be changed to `null`

flow/model/qvalue/avro_converter.go
```diff
@@ -96,6 +98,30 @@ type QValueAvroConverter struct {
 	Nullable  bool
 }
 
+func HStoreToJSON(hstore string) (string, error) {
+	re := regexp.MustCompile(`"(.*?)"=>(?:"([^"]*)"|NULL)`)
+	matches := re.FindAllStringSubmatch(hstore, -1)
```
serprex, line 103: Don't use regex for this, reference the postgres hstore parser

https://github.com/postgres/postgres/blob/bea18b1c949145ba2ca79d4765dba3cc9494a480/contrib/hstore/hstore_io.c#L476

Granted, the hstore_out will always have quotes, but you need to process `\\` & `\"`

serprex (changes requested):

flow/hstore/hstore.go
```diff
@@ -0,0 +1,240 @@
...
+		// must be the exact string NULL: use consumeExpected2 twice
+		err := p.consumeExpected2('N', 'U')
+		if err != nil {
+			return text{}, err
+		}
+		err = p.consumeExpected2('L', 'L')
+		if err != nil {
+			return text{}, err
+		}
+		return text{String: "", Valid: false}, nil
```
serprex, line 168:
```suggestion
		return text{String: nil, Valid: false}, nil
```

flow/hstore/hstore_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	if result != expected {
+		t.Errorf("Unexpected result. Expected: %v, but got: %v", expected, result)
+	}
+}
+
+func TestHStoreNullCase(t *testing.T) {
+	testCase := `"a"=>NULL`
+	expected := `{"a":""}`
```
serprex, line 51:
```suggestion
	expected := `{"a":null}`
```

flow/hstore/hstore.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+	}
+
+	// Convert to JSON string
+	jsonString := "{"
+	for k, v := range result {
+		if v != nil {
+			jsonString += fmt.Sprintf("\"%s\":\"%s\",", k, *v)
+		} else {
+			jsonString += fmt.Sprintf("\"%s\":\"\",", k)
```
serprex, line 234:
```suggestion
			jsonString += fmt.Sprintf("\"%s\":null,", k)
```

flow/hstore/hstore.go
```diff
@@ -0,0 +1,240 @@
...
+			return "", err
+		}
+
+		value, err := p.consumeDoubleQuotedOrNull()
+		if err != nil {
+			return "", err
+		}
+		if value.Valid {
+			valueStrings = append(valueStrings, value.String)
+			result[key] = &valueStrings[len(valueStrings)-1]
```
serprex, line 222:
```suggestion
			result[key] = &value.String
```

flow/hstore/hstore.go
```diff
@@ -0,0 +1,240 @@
...
+		} else {
+			result[key] = nil
+		}
+	}
+
+	// Convert to JSON string
+	jsonString := "{"
```
serprex, line 229: Better to use `encoding/json` here. While it means you need to unescape string val when parsing, it also means you don't have to worry about "or control character" in json spec for strings

https://www.json.org/json-en.html

I assume control characters are https://en.wikipedia.org/wiki/Unicode_control_characters
See also https://en.wikipedia.org/wiki/Control_character

flow/hstore/hstore.go
```diff
@@ -0,0 +1,240 @@
...
+		return unexpectedByteErr(p.str[p.pos+1], two)
+	}
+	p.pos += 2
+	return nil
+}
+
+var errEOSInQuoted = errors.New(`found end before closing double-quote ('"')`)
+
+// consumeDoubleQuoted consumes a double-quoted string from p. The double quote must have been
+// parsed already. This copies the string from the backing string so it can be garbage collected.
```
serprex, line 83: I don't think it's important to copy the string like that; the map we construct doesn't escape the function after being encoded to json, so we aren't creating a long lived reference

flow/connectors/bigquery/merge_stmt_generator.go
```diff
@@ -51,6 +51,13 @@ func (m *mergeStmtGenerator) generateFlattenedCTE() string {
 			// if the type is JSON, then just extract JSON
 			castStmt = fmt.Sprintf("CAST(PARSE_JSON(JSON_VALUE(_peerdb_data, '$.%s'),wide_number_mode=>'round') AS %s) AS `%s`",
 				colName, bqType, shortCol)
+		case qvalue.QValueKindHStore:
+			// PARSE_JSON doesn't work for HSTORE with \" in the value, so use SAFE.PARSE_JSON and coalesce with TO_JSON
```
serprex, line 55: Is this because you're parsing `"` as a string into `\\"` which should be `\"`?

flow/hstore/hstore_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	if result != expected {
+		t.Errorf("Unexpected result. Expected: %v, but got: %v", expected, result)
+	}
+}
+
+func TestHStoreEscapedQuotes(t *testing.T) {
+	testCase := `"a\"b"=>"c\"d"`
+	expected := `{"a"b":"c"d"}`
```
serprex, line 23:
```suggestion
	expected := `{"a\"b":"c\"d"}`
```

flow/hstore/hstore_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	if result != expected {
+		t.Errorf("Unexpected result. Expected: %v, but got: %v", expected, result)
+	}
+}
+
+func TestHStoreEscapedBackslashes(t *testing.T) {
+	testCase := `"a\\b"=>"c\\d"`
+	expected := `{"a\b":"c\d"}`
```
serprex, line 37:
```suggestion
	expected := `{"a\\b":"c\\d"}`
```

flow/hstore/hstore.go
```diff
@@ -0,0 +1,240 @@
+/*
+This is in reference to PostgreSQL's hstore:
+https://github.com/postgres/postgres/blob/bea18b1c949145ba2ca79d4765dba3cc9494a480/contrib/hstore/hstore_io.c
+
+This package is an implementation based on the above code.
```
serprex, line 5:
```suggestion
This package is an implementation based on the above code.
It's simplified to only parse the subset which `hstore_out` outputs.
```

flow/hstore/hstore.go
```diff
@@ -0,0 +1,233 @@
...
+// parsed already. This copies the string from the backing string so it can be garbage collected.
+func (p *hstoreParser) consumeDoubleQuoted() (string, error) {
+	// fast path: assume most keys/values do not contain escapes
+	nextDoubleQuote := strings.IndexByte(p.str[p.pos:], '"')
+	if nextDoubleQuote == -1 {
+		return "", errEOSInQuoted
+	}
+	nextDoubleQuote += p.pos
+	if p.nextBackslash == -1 || p.nextBackslash > nextDoubleQuote {
+		s := strings.Clone(p.str[p.pos:nextDoubleQuote])
```
serprex, line 94:
```suggestion
		s := p.str[p.pos:nextDoubleQuote]
```

flow/hstore/hstore.go
```diff
@@ -0,0 +1,233 @@
...
+		return unexpectedByteErr(p.str[p.pos+1], two)
+	}
+	p.pos += 2
+	return nil
+}
+
+var errEOSInQuoted = errors.New(`found end before closing double-quote ('"')`)
+
+// consumeDoubleQuoted consumes a double-quoted string from p. The double quote must have been
+// parsed already. This copies the string from the backing string so it can be garbage collected.
```
serprex, line 85:
```suggestion
// consumeDoubleQuoted consumes a double-quoted string from p.
// The double quote must have been parsed already.
```

flow/e2e/bigquery/peer_flow_bq_test.go
```diff
@@ -767,6 +768,14 @@ func (s PeerFlowE2ETestSuiteBQ) Test_Types_BQ() {
 	// check if JSON on bigquery side is a good JSON
 	err = s.checkJSONValue(dstTableName, "c17", "sai", "-8.021390374331551")
 	require.NoError(s.t, err)
+
+	// check if HSTORE on bigquery side is a good JSON
+	err = s.checkJSONValue(dstTableName, "c46", "key1", "\"value1\"")
+	require.NoError(s.t, err)
+	err = s.checkJSONValue(dstTableName, "c46", "key2", "\"\"")
```
serprex, line 775: `checkJSONValue` needs to be updated to accept nils, maybe should take `interface{}` a value instead of `string`

PR #1011, opened 2024-01-06
Normalize wait for

serprex:
Testing #980 & #893


serprex:
closing since #980 merged so #893 can go through CI on its own

PR #1020, opened 2024-01-07
Revert "HeartbeatRoutine: return context.CancelFunc instead of chan s…

Amogh-Bharadwaj:
Reverts the change made in #997 
This reverts commit 081eb4065c551570498c193c96f2a10b890d8267.


iskakaushik:
@serprex this was causing issues after partitioning the first partition in qrep.

PR #1023, opened 2024-01-08
refactor(ui): migrate sign in to use next-auth

iamKunalGupta:
Screenshot 2024-01-08 at 7 13 34 PM
Screenshot 2024-01-08 at 7 13 51 PM


Amogh-Bharadwaj (approved):

ui/app/config/config.ts
```diff
@@ -0,0 +1,7 @@
+export var Configuration = {
+  authentication: {
+    PEERDB_PASSWORD: process.env.PEERDB_PASSWORD,
+    // Set this in production to a static value
+    NEXTAUTH_SECRET: process.env.NEXTAUTH_SECRET,
```
iskakaushik, line 5: does this need to be set in `docker-compose.yaml` and `docker-compose-dev.yaml`?

ui/app/auth/options.ts
```diff
@@ -0,0 +1,38 @@
...
+        password: { label: 'Password', type: 'password' },
+      },
+      async authorize(credentials, req) {
+        if (
+          credentials == null ||
+          credentials.password != Configuration.authentication.PEERDB_PASSWORD
+        ) {
+          return null;
+        }
+        return { id: '1', name: 'Admin' };
```
iskakaushik, line 20: what is this for?

ui/components/Logout.tsx
```diff
@@ -1,17 +1,22 @@
...
-          location.assign('/login')
-        )
-      }
-    >
-      Log out
-    </Button>
-  );
+  const { data: session } = useSession();
+  const router = useRouter();
+  if (session) {
```
iskakaushik, line 9: should there be an else branch?

iamKunalGupta: We don't want to show the logout button in case we don't detect a session (is equivalent to not having auth enabled)

ui/app/auth/options.ts
iamKunalGupta: We need to return a `User` which requires an `id` in order to tell next-auth that we have a valid user. I just added a name just in case.

ui/app/config/config.ts
iamKunalGupta: Yes

iskakaushik: can we add static values to those files as apart of this PR? Something like `changeme` or something?

iamKunalGupta: added

PR #1026, opened 2024-01-08
refactored SF and PG to standalone merge generator

heavycrystal:
to allow better testing in the future, also soft-delete only happens when SoftDelete is set to true and a SoftDeleteColName is set.

Added merge statement tests for PG, SF and BQ.
Removed some whitespace in Postgres merge statements.


flow/connectors/bigquery/merge_stmt_generator.go
```diff
@@ -231,9 +226,9 @@ func (m *mergeStmtGenerator) generateUpdateStatements(
 		// generates update statements for the case where updates and deletes happen in the same branch
 		// the backfill has happened from the pull side already, so treat the DeleteRecord as an update
 		// and then set soft-delete to true.
-		if peerdbCols.SoftDelete && (peerdbCols.SoftDeleteColName != "") {
+		if m.peerdbCols.SoftDelete && (m.peerdbCols.SoftDeleteColName != "") {
```
serprex, line 229: Calculate this bool at start, if it's true `updateStmts` should allocate twice as much capacity

serprex (approved):

serprex (approved):

iskakaushik (approved):

PR #1027, opened 2024-01-09
Skip NaNs in Float Arrays

Amogh-Bharadwaj:
- json.Marshal cannot support NaN or INF values
- BigQuery cannot support NULL values in arrays
- Solution: Skip NaN values in float arrays we get


heavycrystal:
approving conditionally on #1028, users should be aware that we're pruning invalid values from their arrays

heavycrystal (approved):

PR #1031, opened 2024-01-09
BQ: Handle NaN in Merge

### LINKED ISSUE #1029: BigQuery: Add tests for NaNs in Floats and Float Arrays
Follow up issue to #1027

Amogh-Bharadwaj:
json.Marshall does not support NaN , so we have to nullfy them, but:
NULLs are aren't allowed when inserting arrays in bigquery. 
So in the Merge statement we skip nulls
Fixes #1029


iskakaushik:
@Amogh-Bharadwaj  can you file an issue to check if this is a problem for Snowflake as well?

iskakaushik (approved):

PR #1034, opened 2024-01-09
UI: Remove force-dynamic

Amogh-Bharadwaj:
Remove force-dynamic in the UI mirrors status file


serprex (approved):

flow/cmd/mirror_status.go
```diff
@@ -17,6 +17,7 @@ func (h *FlowRequestHandler) MirrorStatus(
 	ctx context.Context,
 	req *protos.MirrorStatusRequest,
 ) (*protos.MirrorStatusResponse, error) {
+	slog.Info("Mirror status endpoint called")
```
iskakaushik, line 20: can you add something from the request as well, like the mirror name. Make the log something like: `Getting mirror status for flow - {}`, also add for qrep.

PR #1038, opened 2024-01-09
split BQ merge statements and run them individually

heavycrystal:
For tables with a lot of columns in general and especially many toast columns, some batches can trigger a MERGE statement so complex that BigQuery is unable to process them, with errors like

`The query is too large. The maximum standard SQL query length is 1024.00K characters, including comments and white space characters.`
`Error 400: Resources exceeded during query execution: The query is too complex., resourcesExceeded`

For now, the fix is splitting these complex MERGE statements into smaller ones that act on different subsets of a raw table [partitioning on the basis of `_peerdb_unchanged_toast_columns`]. This can lead to tables need 10+ MERGE statements in a single batch, but this is a compromise with our current design. Instead of sending MERGEs for all tables at once, we do it per table now and update metadata at the end, to avoid exceeding SQL query length limits.


flow/connectors/utils/array.go
```diff
@@ -16,3 +16,16 @@ func ArrayMinus(first []string, second []string) []string {
 	}
 	return result
 }
+
+func ArrayPartition[T any](slice []T, size int) [][]T {
```
serprex, line 20: This is generally referred to as chunking; partition tends to imply some key to group by

flow/connectors/utils/array.go
```diff
@@ -16,3 +16,16 @@ func ArrayMinus(first []string, second []string) []string {
...
+	var partitions [][]T
+
+	for size < len(slice) {
+		slice, partitions = slice[size:], append(partitions, slice[0:size])
```
serprex, line 24:
```suggestion
		partitions = append(partitions, slice[:size])
		slice = slice[size:]
```

flow/connectors/utils/array.go
```diff
@@ -16,3 +16,16 @@ func ArrayMinus(first []string, second []string) []string {
...
+	}
+
+	// Add the last remaining values
+	partitions = append(partitions, slice)
+
+	return partitions
```
serprex, line 30:
```suggestion
	return append(partitions, slice)
```

flow/connectors/utils/array.go
```diff
@@ -16,3 +16,16 @@ func ArrayMinus(first []string, second []string) []string {
 	}
 	return result
 }
+
+func ArrayPartition[T any](slice []T, size int) [][]T {
+	var partitions [][]T
```
serprex, line 21:
```suggestion
	var chunks [][]T
```

serprex (changes requested):

flow/connectors/bigquery/merge_statement_generator.go
```diff
@@ -181,6 +181,20 @@ func (m *mergeStmtGenerator) generateMergeStmt() string {
 		pkeySelectSQL, insertColumnsSQL, insertValuesSQL, updateStringToastCols, deletePart)
 }
 
+func (m *mergeStmtGenerator) generateMergeStmts() []string {
+	// TODO (kaushik): This is so that the statement size for individual merge statements
+	// doesn't exceed the limit. We should make this configurable.
+	const batchSize = 8
+	partitions := utils.ArrayPartition(m.unchangedToastColumns, batchSize)
```
serprex, line 188: This could instead be an iterator,
```go
type ChunkIter[T] struct {
  slice: T[]
  index: int
  chunkSize: int
}
```
then
```go
for chunk in chunks.Next() {
  mergeStmts = append(mergeStmts, g.generateMergeStmt(chunk))
}
```

heavycrystal: From current experience it seems to be 10-30, doesn't seem worth it to implement an iterator for this

flow/connectors/utils/array.go
```diff
@@ -16,3 +16,15 @@ func ArrayMinus(first []string, second []string) []string {
 	}
 	return result
 }
+
+func ArrayChunksGen[T any](slice []T, size int) [][]T {
```
serprex, line 20:
```suggestion
func ArrayChunks[T any](slice []T, size int) [][]T {
```

serprex (approved):

flow/connectors/utils/array.go
```diff
@@ -16,3 +16,15 @@ func ArrayMinus(first []string, second []string) []string {
...
+	var chunks [][]T
+
+	for size < len(slice) {
+		chunks = append(chunks, slice[:size])
+		slice = slice[size:]
+	}
+
+	// Add the last remaining values
+	return append(chunks, slice)
```
serprex, line 29: This shouldn't append slice `if len(slice) == 0`

serprex (approved):

PR #1041, opened 2024-01-09
Clean xmin flow, add query handlers, fix typo

Amogh-Bharadwaj:
Realised we can just create a qrep execution object for many of the activities we're calling in xmin_flow.go
Fixes a typo in qrep_flow
Adds temporal query handlers in xmin_flow.go (same ones as in qrep_flow - these have been refactored to a common function). Without these, XMIN mirror statuses cannot be viewed in UI


serprex (approved):
nice

PR #1048, opened 2024-01-10
Fix nextauth url in docker compose files

Amogh-Bharadwaj:
Must specify http or else the callback upon login fails


iamKunalGupta (approved):
Yep, I found the same when testing for cloud

PR #1051, opened 2024-01-11
make logs a little better

iskakaushik:


ui/app/api/mirrors/errors/route.ts
```diff
@@ -0,0 +1,48 @@
...
+    orderBy: {
+      error_timestamp: 'desc',
+    },
+    select: {
+      id: false,
+      flow_name: true,
+      error_message: true,
+      error_type: true,
+      error_timestamp: true,
+      ack: true,
```
Amogh-Bharadwaj, line 26: I think we can omit this from the select

PR #1052, opened 2024-01-11
Eventhub: Associate partition key value with batch

Amogh-Bharadwaj:
## Eventhubs: Enable Partitioning Via Column Value
This PR aims to achieve a feature where PG to Eventhub CDC now supports partition-level routing.  Suppose there is a column in the source table T1 called `id`. We want T1 to be mapped to one eventhub E1, and the rows with the same `id` value V1 in T1 to be mapped to the same partition P1 of E1. And similarly, V2 or T1 to P2 of E1 and so on. 

**Note**: The limit of 32 on the number of partitions is being accounted for by Eventhub's internal assignment. V100 to P100 is not happening

### Mirror definition
- `ScopedEventhub`'s definition is now :
```golang
// Scoped eventhub is of the form peer_name.eventhub_name.partition_column.partition_key_value
// partition_column is the column in the table that is used to determine
// the partition key for the eventhub. Partition value is one such value of that column.
type ScopedEventhub struct {
	PeerName           string
	Eventhub           string
	PartitionKeyColumn string
	PartitionKeyValue  string
}
```

Therefore now the semantics of the table mapping in the mirror command is:
```sql
WITH TABLE MAPPING (
-- src_table:<eventhub (not group) peer>.<eventhub name>.<column name>
public.oss1:ehpeer1.oss1_destination_eventhub.id 
)
```

### Routing
When creating a batch, we specify the partition key of the batch:
```golang
opts := &azeventhubs.EventDataBatchOptions{
		PartitionKey: &destination.PartitionKeyValue,
	}
batch, err := hub.NewEventDataBatch(ctx, opts)
```

According to azeventhubs docs:
```
PartitionKey is hashed to calculate the partition assignment. Messages and message
batches with the same PartitionKey are guaranteed to end up in the same partition.
```


flow/connectors/eventhub/eventhub.go
```diff
@@ -164,13 +165,25 @@ func (c *EventHubConnector) processBatch(
...
 				return 0, err
 			}
 
-			err = batchPerTopic.AddEvent(ctx, topicName, json, false)
+			// Scoped eventhub is of the form peer_name.eventhub_name.partition_column
+			// partition_column is the column in the table that is used to determine
+			// the partition key for the eventhub.
+			partitionColumn := destination.PartitionKeyColumn
+			partitionValue := record.GetItems().GetColumnValue(partitionColumn).Value
+			partitionKey := fmt.Sprintf("%v", partitionValue)
```
iskakaushik, line 179: do this in the else branch

flow/connectors/eventhub/eventhub.go
```diff
@@ -164,13 +165,25 @@ func (c *EventHubConnector) processBatch(
...
+			if partitionValue == nil {
+				partitionKey = ""
+			}
+
+			destination.SetPartitionValue(partitionKey)
+			c.logger.Info("Partition value has been set", slog.String("destination", destination.ToString()))
```
iskakaushik, line 185: this will be a very verbose log

iskakaushik (approved):

PR #1053, opened 2024-01-11
Eventhub: Fix rows synced log

Amogh-Bharadwaj:
We weren't returning the current sync batch ID as part of eventhub's SyncRecords. As a result, we weren't able to get CDC logs for Eventhub. This PR fixes that


serprex:
PR description should be revised since you seem to've redesigned PR

iskakaushik (approved):
LGTM once you update description

PR #1057, opened 2024-01-11
Add the ability to dynamically configure some thresholds

iskakaushik:


flow/dynamicconf/dynamicconf.go
```diff
@@ -0,0 +1,69 @@
...
+	var value pgtype.Text
+	query := "SELECT config_value FROM alerting_settings WHERE config_name = $1"
+	err = conn.QueryRow(ctx, query, key).Scan(&value)
+	if err != nil {
+		slog.Error("Failed to get key: %v", err)
+		return defaultValue
+	}
+
+	var result uint64
+	result, err = strconv.ParseUint(value.String, 10, 32)
```
heavycrystal, line 46:
```suggestion
	result, err := strconv.ParseUint(value.String, 10, 32)
```

flow/dynamicconf/dynamicconf.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		slog.Error("Failed to check if key exists: %v", err)
+		return false
+	}
+
+	return exists.Bool
+}
+
+func dynamicConfUint32(ctx context.Context, key string, defaultValue uint32) uint32 {
+	conn, err := utils.GetCatalogConnectionPoolFromEnv()
```
heavycrystal, line 27: how often do we expect this to be called across all variables? we might either want to open and close a single connection or open a single pool and cache it.

Also pool isn't being closed

iskakaushik: this is a global pool, it isn't closed.

serprex (approved):

PR #1062, opened 2024-01-11
replace reviewdog action with official golangci action

serprex:
fix lints

### DIFF
--- a/flow/connectors/bigquery/merge_stmt_generator.go
+++ b/flow/connectors/bigquery/merge_stmt_generator.go
@@ -121,7 +121,7 @@ func (m *mergeStmtGenerator) generateDeDupedCTE() string {
 }
 
 // generateMergeStmt generates a merge statement.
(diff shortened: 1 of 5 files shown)


flow/connectors/bigquery/merge_stmt_generator.go
```diff
@@ -121,7 +121,7 @@ func (m *mergeStmtGenerator) generateDeDupedCTE() string {
 }
 
 // generateMergeStmt generates a merge statement.
-func (m *mergeStmtGenerator) generateMergeStmt(unchangedToastColumns []string) string {
+func (m *mergeStmtGenerator) generateMergeStmt(_unchangedToastColumns []string) string {
```
serprex, line 124: This parameter isn't used, so probably this is really bad for BQ where we'll run the same query 8 times. But that can be fixed in another PR

iskakaushik: this was causing issues for a customer: https://github.com/PeerDB-io/peerdb/pull/1066
fixed in the linked PR

PR #1068, opened 2024-01-12
Make all heartbeats 15 seconds, make all heartbeat timeouts 1 minute

serprex:
Recording heartbeat is the time for context to be cancelled,
effectively making pulse rate the expected p99 cancel latency


flow/activities/flowable.go
```diff
@@ -834,7 +834,21 @@ func (a *FlowableActivity) QRepWaitUntilNewRows(ctx context.Context,
 	attemptCount := 1
 	for {
 		activity.RecordHeartbeat(ctx, fmt.Sprintf("no new rows yet, attempt #%d", attemptCount))
-		time.Sleep(waitBetweenBatches)
+		waitUntil := time.Now().Add(waitBetweenBatches)
```
serprex, line 837: Maybe this can be made into a utility function in `connectors/utils/heartbeat.go`

flow/connectors/utils/heartbeat.go
```diff
@@ -26,11 +25,11 @@ func HeartbeatRoutine(
 				return
 			case <-ctx.Done():
 				return
-			case <-time.After(interval):
+			case <-time.After(15 * time.Second):
```
serprex, line 28: Looks like I made it 15s

iskakaushik:
@serprex  before we land this, i'd like us to validate if the context cancel propagation is handled in a reasonable way in the most common scenarios, one bug that I ran into recently was: https://github.com/PeerDB-io/peerdb/pull/1076, the fix also seems like a stop-gap fix. Can you audit the most common situations and make sure context cancelation is handled in a reasonable way?

1. Qrep pull records on Postgres.
2. CDC pull records - postgres.
3. CDC Push records - postgres, snowflake, bq.
4. QRep Sync records - postgres, snowflake, bq.

I'd be much more comfortable landing the change, knowing context cancelation is handled in these scenarios.

flow/activities/flowable.go
```diff
@@ -831,7 +830,21 @@ func (a *FlowableActivity) QRepWaitUntilNewRows(ctx context.Context,
 	attemptCount := 1
 	for {
```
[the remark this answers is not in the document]
serprex: Yes

iskakaushik (approved):

serprex:
In all those cases we're using `utils.HeartbeatRoutine`

PR #1077, opened 2024-01-13
Move ReplayTableSchemaDeltas into SyncRecords

serprex:
A sync batch should not be considered complete until its schema changes are processed,
this avoids failures after commit causing schema changes to be dropped,
& when decoupling normalize/sync in #893 was causing normalization to be missing values

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -278,16 +278,24 @@ func (a *FlowableActivity) StartFlow(ctx context.Context,
 			return nil, fmt.Errorf("failed in pull records when: %w", err)
 		}
 		slog.InfoContext(ctx, "no records to push")
-		syncResponse := &model.SyncResponse{}
-		syncResponse.RelationMessageMapping = <-recordBatch.RelationMessageMapping
-		syncResponse.TableSchemaDeltas = recordBatch.WaitForSchemaDeltas(input.FlowConnectionConfigs.TableMappings)
-		return syncResponse, nil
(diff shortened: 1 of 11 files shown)


serprex:
TODO: schema changes should be idempotent; need to ignore when column already exists

flow/activities/flowable.go
```diff
@@ -283,16 +283,24 @@ func (a *FlowableActivity) StartFlow(ctx context.Context,
 			return nil, fmt.Errorf("failed in pull records when: %w", err)
 		}
 		slog.InfoContext(ctx, "no records to push")
-		syncResponse := &model.SyncResponse{}
-		syncResponse.RelationMessageMapping = <-recordBatch.RelationMessageMapping
-		syncResponse.TableSchemaDeltas = recordBatch.WaitForSchemaDeltas(input.FlowConnectionConfigs.TableMappings)
-		return syncResponse, nil
+		tableSchemaDeltas := recordBatch.WaitForSchemaDeltas(input.FlowConnectionConfigs.TableMappings)
+		if normConn, ok := dstConn.(connectors.CDCNormalizeConnector); ok {
```
iskakaushik, line 287: can we move replay schema deltas to CDCSyncConnector?

serprex: Not without creating nop methods on CDCSyncConnectors for s3/eventhub

serprex: Discussed oob; preference is for nop methods

iskakaushik (approved):

PR #1084, opened 2024-01-15
dynamically add new tables to CDC mirrors

heavycrystal:
1. Call an activity to add the new tables to the publication.
2. Kick off a child `CDCFlow` with only the new tables as input along with the `InitialCopyOnly` and `ForcePkeyChecks` flags set. This is because `InitialCopyOnly` skips pkey checks and slot creation since it was written to help with just snapshotting multiple tables. The second flag is to reenable enough of the pkey checks for the CDC mirror to work.
3. Patch the config of the parent workflow with the new tables and then resume.

Currently there is no way to interface with this feature, will add a way to signal this in a different PR.


flow/connectors/postgres/postgres.go
```diff
@@ -932,3 +938,23 @@ func (c *PostgresConnector) GetOpenConnectionsForUser() (*protos.GetOpenConnecti
 		CurrentOpenConnections: result.Int64,
 	}, nil
 }
+
+func (c *PostgresConnector) AddTablesToPublication(req *protos.AddTablesToPublicationInput) error {
+	// don't modify custom publications
+	if req == nil || req.PublicationName != "" || len(req.AdditionalTables) == 0 {
+		return nil
+	}
```
iskakaushik, line 946: validate if the publication has these new tables.

flow/workflows/cdc_flow.go
```diff
@@ -145,46 +154,134 @@ func (w *CDCFlowWorkflowExecution) receiveAndHandleSignalAsync(ctx workflow.Cont
 	}
 }
 
+func additionalTablesHasOverlap(currentTableMappings []*protos.TableMapping,
```
iskakaushik, line 157: move this to a utils file.

flow/workflows/cdc_flow.go
```diff
@@ -145,46 +154,134 @@ func (w *CDCFlowWorkflowExecution) receiveAndHandleSignalAsync(ctx workflow.Cont
...
+) error {
+	for _, flowConfigUpdate := range state.FlowConfigUpdates {
+		if len(flowConfigUpdate.AdditionalTables) == 0 {
+			continue
+		}
+		if additionalTablesHasOverlap(cfg.TableMappings, flowConfigUpdate.AdditionalTables) {
+			return fmt.Errorf("duplicate source/destination tables found in additionalTables")
+		}
+
+		alterPublicationAddAdditionalTablesCtx := workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
```
iskakaushik, line 190: set heartbeat time.

flow/workflows/cdc_flow.go
```diff
@@ -224,7 +321,9 @@ func CDCFlowWorkflowWithConfig(
 		if err := setupFlowFuture.Get(setupFlowCtx, &cfg); err != nil {
 			return state, fmt.Errorf("failed to execute child workflow: %w", err)
 		}
-		state.CurrentFlowState = protos.FlowStatus_STATUS_SNAPSHOT
+		state.SrcTableIdNameMapping = cfg.SrcTableIdNameMapping
```
iskakaushik, line 324: lets remove these from Config, its making it hard to reason about. It's ok for a breaking change.

flow/connectors/postgres/postgres.go
```diff
@@ -789,17 +803,11 @@ func (c *PostgresConnector) EnsurePullability(
 		}
 
 		// we only allow no primary key if the table has REPLICA IDENTITY FULL
+		// this is ok for replica identity idex as we populate the primary key columns
```
serprex, line 806:
```suggestion
		// this is ok for replica identity index as we populate the primary key columns
```

flow/workflows/cdc_flow.go
```diff
@@ -145,46 +154,134 @@ func (w *CDCFlowWorkflowExecution) receiveAndHandleSignalAsync(ctx workflow.Cont
...
+	for _, currentTableMapping := range currentTableMappings {
+		currentSrcTables = append(currentSrcTables, currentTableMapping.SourceTableIdentifier)
+		currentDstTables = append(currentDstTables, currentTableMapping.DestinationTableIdentifier)
+	}
+	for _, additionalTableMapping := range additionalTableMappings {
+		currentSrcTables = append(currentSrcTables, additionalTableMapping.SourceTableIdentifier)
+		currentDstTables = append(currentDstTables, additionalTableMapping.DestinationTableIdentifier)
+	}
+
+	return utils.ArraysHaveOverlap[string](currentSrcTables, additionalSrcTables) ||
```
serprex, line 174:
```suggestion
	return utils.ArraysHaveOverlap(currentSrcTables, additionalSrcTables) ||
```
use type inference

flow/connectors/utils/array.go
```diff
@@ -29,3 +29,19 @@ func ArrayChunks[T any](slice []T, size int) [][]T {
 
 	return partitions
 }
+
+func ArraysHaveOverlap[T comparable](first, second []T) bool {
+	lookup := make(map[T]struct{})
+
+	for _, element := range first {
```
serprex, line 36: I generally expect the first parameter will be larger than the second (see, for example, how callers have first parameter as existing tables, & second parameter as new tables). Better off constructing set that's the size of second slice

flow/workflows/cdc_flow.go
heavycrystal: done as part of #1087

nexus/server/src/main.rs
```diff
@@ -748,16 +750,17 @@ impl NexusBackend {
...
                             )
                             .await
                             .map_err(|err| {
                                 PgWireError::ApiError(
                                     format!("unable to resume flow job: {:?}", err).into(),
                                 )
                             })?;
-                        let drop_mirror_success = format!("RESUME MIRROR {}", flow_job_name);
+                        let resume_mirror_success = format!("RESUME MIRROR {}", flow_job_name);
                         Ok(vec![Response::Execution(Tag::new(
```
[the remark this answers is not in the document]
heavycrystal: done

flow/connectors/utils/array.go
heavycrystal: swapped around

flow/connectors/postgres/postgres.go
```diff
@@ -947,16 +947,33 @@ func (c *PostgresConnector) GetOpenConnectionsForUser() (*protos.GetOpenConnecti
...
+		if err != nil {
+			return fmt.Errorf("failed to check tables in publication: %w", err)
+		}
+
+		tableNames, err := pgx.CollectRows[string](rows, pgx.RowTo)
+		if err != nil {
+			return fmt.Errorf("failed to check tables in publication: %w", err)
+		}
+		if len(utils.ArrayMinus(tableNames, additionalSrcTables)) > 0 {
+			return fmt.Errorf("some additional tables not present in custom publication")
```
iskakaushik, line 972: include the missing tables in the erorr as well.

iskakaushik (approved):

PR #1087, opened 2024-01-16
[breaking] pruning and reorganizing fields from protos, moving stuff to state

heavycrystal:
### ⚠️ This change can break existing CDC mirrors!

Performs several breaking changes to our protobufs:

1) Removes unused fields from `FlowConnectionConfigs`: `table_schema, src_table_id_name_mapping, table_name_schema_mapping, snapshot_sync_mode, cdc_sync_mode, push_batch_size, push_parallelism, metadata_peer`. `src_table_id_name_mapping` and `table_name_schema_mapping` are now moved to state, setup from `SetupFlow` and passed accordingly via more proto changes. Order shuffled around to group similar fields together.
2) Removed unnecessary generalization of `table_identifier`
3) Removed unused messages `EnsurePullabilityInput` and `EnsurePullabilityOutput`
4) Changed wording of `do_initial_copy` and `initial_copy_only` to better align with other options.
5) Removed `columns` from `TableSchema` and fixed related methods up.
6) `batch_size` had 3 different datatypes and was being modified in multiple places, making sure `batch_size` and `idle_timeout_seconds` has one source of truth.


flow/workflows/normalize_flow.go
```diff
@@ -31,18 +31,20 @@ func NewNormalizeFlowExecution(ctx workflow.Context, state *NormalizeFlowState)
 
 func NormalizeFlowWorkflow(ctx workflow.Context,
 	config *protos.FlowConnectionConfigs,
+	options *protos.NormalizeFlowOptions,
```
serprex, line 34: normalize changes should be held off for #893

nexus/flow-rs/src/grpc.rs
```diff
@@ -155,7 +155,7 @@ impl FlowGrpcClient {
             destination: Some(dst),
             flow_job_name: job.name.clone(),
             table_mappings,
-            do_initial_copy,
+            do_initial_snapshot: do_initial_copy,
```
serprex, line 158: rename variable?

flow/connectors/utils/columns.go
```diff
@@ -4,51 +4,27 @@ import (
 	"slices"
 
 	"github.com/PeerDB-io/peer-flow/generated/protos"
-	"golang.org/x/exp/maps"
 )
 
 func TableSchemaColumns(schema *protos.TableSchema) int {
```
serprex, line 9: This file should probably be removed entirely

protos/flow.proto
```diff
@@ -114,10 +101,12 @@ message SyncFlowOptions {
...
+  map<uint32, string> src_table_id_name_mapping = 4;
+  map<string, TableSchema> table_name_schema_mapping = 5;
 }
 
 message NormalizeFlowOptions {
-  int32 batch_size = 1;
+  map<string, TableSchema> table_name_schema_mapping = 1;
 }
 
 message LastSyncState {
```
serprex, line 112: can get rid of `LastSyncState` we only use it to pass around `int64`

iskakaushik: better to have it around in case we add more foelds and make it backwards compatible

serprex: Can remove `last_synced_at`

flow/workflows/normalize_flow.go
heavycrystal: needed to pass `TableNameSchemaMapping` in, was previously contained in config

flow/connectors/utils/columns.go
heavycrystal: will do in a later PR, will require a lot of changes

protos/flow.proto
heavycrystal: removed `last_synced_at`

nexus/flow-rs/src/grpc.rs
heavycrystal: done

iskakaushik (approved):

PR #1090, opened 2024-01-17
GetSlotInfo: only list replication slots in peer's database

serprex:
Also introduce escape.go to help with properly escaping strings/identifiers

Update all cases of "%s" in connectors/postgres,
along with a few strings such as in GetSlotInfo


flow/connectors/postgres/client.go
```diff
@@ -248,7 +248,9 @@ func (c *PostgresConnector) checkSlotAndPublication(slot string, publication str
 func (c *PostgresConnector) GetSlotInfo(slotName string) ([]*protos.SlotInfo, error) {
 	specificSlotClause := ""
```
Amogh-Bharadwaj, line 249: maybe we can now rename this to `whereClause` or similar

Amogh-Bharadwaj (approved):

PR #1092, opened 2024-01-17
[snowflake] Cache information schema queries on snowflake

iskakaushik:
In cases there are thousands of tables, PeerDB runs A LOT of INFORMATION_SCHEMA queries to Snowflake’s INFORMATION_SCHEMA which counts as Cloud Service usage. This results in very high credit consumption w/ ~1:25 ratio of regular credit consumed to cloud services credits consumed (see screenshot) — so even if an XSMALL or SMALL warehouse is used (1, 2 credits/hour respectively) we still end up consuming ~50 credits an hour.

image

### DIFF
--- a/flow/connectors/snowflake/information_schema_cache.go
+++ b/flow/connectors/snowflake/information_schema_cache.go
@@ -1,0 +1,246 @@
+package connsnowflake
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"log/slog"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/PeerDB-io/peer-flow/connectors/utils"
+	"github.com/PeerDB-io/peer-flow/generated/protos"
+	"github.com/PeerDB-io/peer-flow/model/qvalue"
+	"github.com/PeerDB-io/peer-flow/peerdbenv"
+
+	"github.com/hashicorp/golang-lru/v2/expirable"
+)
+
+const (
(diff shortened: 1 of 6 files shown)


flow/connectors/snowflake/information_schema_cache.go
```diff
@@ -0,0 +1,250 @@
...
+		database: database,
+		logger:   logger,
+	}
+}
+
+// SchemaExists returns true if the schema exists in the database.
+func (c *SnowflakeInformationSchemaCache) SchemaExists(schemaName string) (bool, error) {
+	schemaName = strings.ToUpper(schemaName)
+
+	if cachedExists, ok := cache.schemaExistsCache.Get(schemaName); ok {
```
serprex, line 79: bit unfortunate that they made an LRU expirable cache without an atomic `GetOrUpsert` method

serprex:
should probably generalize this for bigquery

flow/connectors/snowflake/information_schema_cache.go
```diff
@@ -0,0 +1,250 @@
...
+	database *sql.DB
+	logger   slog.Logger
+}
+
+func NewSnowflakeInformationSchemaCache(
+	ctx context.Context,
+	database *sql.DB,
+	logger slog.Logger,
+) *SnowflakeInformationSchemaCache {
+	cacheInit.Do(func() {
```
serprex, line 65: global cache will fail when schema name is shared across different databases with mirrors on each schema

quite likely for a customer with a multi tenant app

iskakaushik: good catch, added a dbCacheKey

serprex (approved):

PR #1094, opened 2024-01-17
snowflake getTableSchemaForTable: error when table does not exist

serprex:
Was returning empty schema, which got as far as

COPY INTO "PUBLIC2"."SequelizeMeta"() FROM (...)

failing because an empty set of columns is a syntax error


heavycrystal:
is this method only invoked as a part of tests?

serprex:
Turns out I ran into the issue because we had the same code in `getColsForTable` which is used. Revised `getTableSchemaForTable` to use `getColsForTable` & fixed bug in `getColsForTable`

iskakaushik (approved):

PR #1101, opened 2024-01-18
drop_flow: drop destination/source concurrently

serprex:
Avoids destination issues blocking source cleanup, along with being a bit faster

Also cleanup identifier escaping in external_metadata/store.go


flow/workflows/drop_flow.go
```diff
@@ -1,43 +1,37 @@
...
-	}
+	var sourceError, destinationError error
+	selector := workflow.NewNamedSelector(ctx, fmt.Sprintf("%s-drop", req.FlowJobName))
+	selector.AddFuture(dropSourceFuture, func(f workflow.Future) {
+		sourceError = f.Get(ctx, nil)
+	})
+	selector.AddFuture(dropDestinationFuture, func(f workflow.Future) {
+		destinationError = f.Get(ctx, nil)
+	})
+	selector.Select(ctx)
```
iskakaushik, line 33: is there a better way to wait for all pending futures?

serprex: No, `Selector` is temporal's `select`

iskakaushik (approved):

PR #1107, opened 2024-01-19
remove SkipPkeyAndReplicaCheck

heavycrystal:
UI and EnsurePullability end up doing the same checks and current implementation also leads to loss of information even when pkeys are present


flow/connectors/postgres/postgres.go
```diff
@@ -600,27 +600,19 @@ func (c *PostgresConnector) GetTableSchema(
...
-		if err != nil {
-			return nil, fmt.Errorf("[getTableSchema]:error getting primary key column for table %s: %w", schemaTable, err)
-		}
+	replicaIdentityType, err := c.getReplicaIdentityType(schemaTable)
+	if err != nil {
+		return nil, fmt.Errorf("[getTableSchema] error getting replica identity for table %s: %w", schemaTable, err)
+	}
+	pKeyCols, err := c.getPrimaryKeyColumns(replicaIdentityType, schemaTable)
+	if err != nil {
+		return nil, fmt.Errorf("[getTableSchema] error getting primary key column for table %s: %w", schemaTable, err)
```
Amogh-Bharadwaj, line 615: When a user kicks off an initial_copy_only mirror via query layer, the mirror will fail here for tables with no pkey/replica identity

flow/workflows/qrep_flow.go
```diff
@@ -108,10 +108,9 @@ func (q *QRepFlowExecution) SetupWatermarkTableOnDestination(ctx workflow.Contex
 		})
 
 		tableSchemaInput := &protos.GetTableSchemaBatchInput{
-			PeerConnectionConfig:    q.config.SourcePeer,
-			TableIdentifiers:        []string{q.config.WatermarkTable},
-			FlowName:                q.config.FlowJobName,
-			SkipPkeyAndReplicaCheck: true,
```
Amogh-Bharadwaj, line 114: Won't `setup_watermark_table_on_destination` fail now if there's no pkey/replica identity ?

Amogh-Bharadwaj (changes requested):

flow/connectors/postgres/postgres.go
heavycrystal: changed getPkeyColumns to not error out

flow/workflows/qrep_flow.go
heavycrystal: same as above

Amogh-Bharadwaj (approved):

PR #1114, opened 2024-01-19
Nexus bq: refactor run tracked

Amogh-Bharadwaj:
We close the tracker after getting the result set but if we error while obtaining the result set, the token wouldn't get closed. This PR fixes that.

### DIFF
--- a/nexus/peer-bigquery/src/lib.rs
+++ b/nexus/peer-bigquery/src/lib.rs
@@ -89,12 +89,10 @@ impl BigQueryQueryExecutor {
             PgWireError::ApiError(err.into())
         })?;
 
-        let result_set = result_set.map_err(|err| {
+        result_set.map_err(|err| {
             tracing::error!("error running query: {}", err);
             PgWireError::ApiError(err.into())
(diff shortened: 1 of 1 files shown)


iskakaushik (approved):

serprex (approved):

nexus/peer-bigquery/src/lib.rs
```diff
@@ -82,17 +82,18 @@ impl BigQueryQueryExecutor {
...
             tracing::error!("error closing tracking token: {}", err);
             PgWireError::ApiError(err.into())
         })?;
 
+        let result_set = result_set.map_err(|err| {
+            tracing::error!("error running query: {}", err);
+            PgWireError::ApiError(err.into())
+        })?;
+
         Ok(result_set)
```
serprex, line 97:
```suggestion
        result_set.map_err(|err| {
            tracing::error!("error running query: {}", err);
            PgWireError::ApiError(err.into())
        })
```

iskakaushik (approved):

PR #1123, opened 2024-01-22
Validate mirror: account for mixed case

Amogh-Bharadwaj:
Forgot to quote the source table identifiers we use in validate mirror's source table check

### DIFF
--- a/flow/connectors/postgres/client.go
+++ b/flow/connectors/postgres/client.go
@@ -596,7 +596,7 @@ func (c *PostgresConnector) CheckSourceTables(tableNames []string, pubName strin
 			return fmt.Errorf("invalid source table identifier: %s", table)
 		}
 
-		tableArr = append(tableArr, fmt.Sprintf(`('%s', '%s')`, schemaName, tableName))
+		tableArr = append(tableArr, fmt.Sprintf(`(%s::text, %s::text)`, QuoteLiteral(schemaName), QuoteLiteral(tableName)))
 		err := c.pool.QueryRow(c.ctx,
 			fmt.Sprintf("SELECT * FROM %s.%s LIMIT 0;", QuoteIdentifier(schemaName), QuoteIdentifier(tableName))).Scan(&row)
 		if err != nil && err != pgx.ErrNoRows {
@@ -609,7 +609,7 @@ func (c *PostgresConnector) CheckSourceTables(tableNames []string, pubName strin
 	if pubName != "" {
 		var pubTableCount int
 		err := c.pool.QueryRow(c.ctx, fmt.Sprintf(`
-		with source_table_components (sname, tname) as ( values %s)
+		with source_table_components (sname, tname) as (values %s)
 		select COUNT(DISTINCT(schemaname,tablename)) from pg_publication_tables 
 		INNER JOIN source_table_components stc 
 		ON schemaname=stc.sname and tablename=stc.tname where pubname=$1;`, tableStr), pubName).Scan(&pubTableCount)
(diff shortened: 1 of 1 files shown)


serprex (dismissed):

flow/connectors/postgres/client.go
```diff
@@ -588,22 +588,26 @@ func (c *PostgresConnector) CheckSourceTables(tableNames []string, pubName strin
...
+		schemaName, tableName, found := strings.Cut(table, ".")
+		if !found {
+			return fmt.Errorf("invalid source table identifier: %s", table)
+		}
+
+		quotedTableIdentifier := fmt.Sprintf("%s.%s",
+			QuoteLiteral(schemaName),
+			QuoteLiteral(tableName))
+		tableArr = append(tableArr, quotedTableIdentifier)
+		err := c.pool.QueryRow(c.ctx, fmt.Sprintf("SELECT * FROM %s LIMIT 0;", quotedTableIdentifier)).Scan(&row)
```
serprex, line 603:
```suggestion
		err := c.pool.QueryRow(c.ctx, fmt.Sprintf("SELECT * FROM %s.%s LIMIT 0;", QuoteIdentifier(schemaName), QuoteIdentifier(tableName))).Scan(&row)
```

flow/connectors/postgres/client.go
```diff
@@ -588,22 +588,26 @@ func (c *PostgresConnector) CheckSourceTables(tableNames []string, pubName strin
...
-		err := c.pool.QueryRow(c.ctx, fmt.Sprintf("SELECT * FROM %s LIMIT 0;", tableName)).Scan(&row)
+		schemaName, tableName, found := strings.Cut(table, ".")
+		if !found {
+			return fmt.Errorf("invalid source table identifier: %s", table)
+		}
+
+		quotedTableIdentifier := fmt.Sprintf("%s.%s",
+			QuoteLiteral(schemaName),
+			QuoteLiteral(tableName))
+		tableArr = append(tableArr, quotedTableIdentifier)
```
serprex, line 602: This should be unescaped given SQL its used in later, `fmt.Sprintf("%s.%s", schemaName, tableName)` but ideally there wouldn't be string building in sql, maybe if you built this as a CTE for query
```sql
with checknames_comeupwithabettername (sname, tname) as ( values (
  %s
))
```
Where %s is replaced by `strings.Join(tableArr, ",")`
```go
tableArr := make([]string, 0, len(tableNames))
for _, table in tableNames {
  ...
  tableArr = append(tableArr, fmt.Sprintf("(%s, %s)", QuoteLiteral(schemaName), QuoteLiteral(tableName)))
}
```
then in query join the CTE with `pg_publication_tables`

flow/connectors/postgres/client.go
```diff
@@ -588,26 +588,31 @@ func (c *PostgresConnector) CheckSourceTables(tableNames []string, pubName strin
...
+	tableArr := make([]string, 0, len(tableNames))
+	for _, table := range tableNames {
 		var row pgx.Row
-		err := c.pool.QueryRow(c.ctx, fmt.Sprintf("SELECT * FROM %s LIMIT 0;", tableName)).Scan(&row)
+		schemaName, tableName, found := strings.Cut(table, ".")
+		if !found {
+			return fmt.Errorf("invalid source table identifier: %s", table)
+		}
+
+		tableArr = append(tableArr, fmt.Sprintf(`('%s', '%s')`, schemaName, tableName))
```
serprex, line 599:
```suggestion
		tableArr = append(tableArr, fmt.Sprintf(`(%s, %s)`, QuoteLiteral(schemaName), QuoteLiteral(tableName)))
```

serprex (approved):

PR #1125, opened 2024-01-22
Support specifying host key for ssh config

### LINKED ISSUE #804: SSH Hostkey should be configured. We can get host key from our existing ssh connection with them. Callback only needs to check host key, not ip
This should be configured. We can get host key from our existing ssh connection with them. Callback only needs to check host key, not ip

_Originally posted by @serprex in https://github.com/PeerDB-io/peerdb/pull/800#discussion_r1424089953_

serprex:
Empty host key interpreted as accept-all

Fixes #804

UX would be greatly improved by implementing a "load host key" button which has server request host key from remote server & fill in ui field

Host keys can be acquired with
```sh
ssh-keyscan $HOST
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,6 +10,6 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 4
 
-[{package.json,*.yml,*.yaml}]
+[{package.json,*.yml,*.yaml,*.proto}]
 indent_style = space
 indent_size = 2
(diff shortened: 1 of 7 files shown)


serprex:
As an aside, I don't like how we take private key as input. Granted, this works for when the user wants to generate everything on their end, but in the end the ideal would be that the user doesn't even have access to the private key: we would generate a private key in handler, & give back public key to user. That'd also enforce private key never traversing internet

Amogh-Bharadwaj (approved):

PR #1130, opened 2024-01-23
Numeric: adjust precision handling

Amogh-Bharadwaj:
`200` was being synced as `199.99999999` in initial load for PostgreSQL -> Snowflake. It resided in a numeric column.

Noticed that in ToJSON where we transform numerics and other types for raw table insertion, we set the scale to 9 to conform with the scale value we've chosen for Snowflake for Numeric values (Number(38,9)). And this issue isn't present in CDC.

This PR does the same at `processNumeric` in the Avro transform step where we write to Avro files

### DIFF
--- a/flow/model/qvalue/avro_converter.go
+++ b/flow/model/qvalue/avro_converter.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	hstore_util "github.com/PeerDB-io/peer-flow/hstore"
+	"github.com/PeerDB-io/peer-flow/model/numeric"
 	"github.com/google/uuid"
 	"github.com/linkedin/goavro/v2"
 )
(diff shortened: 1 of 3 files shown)


serprex (approved):

flow/model/qvalue/avro_converter.go
```diff
@@ -323,6 +323,9 @@ func (c *QValueAvroConverter) processNumeric() (interface{}, error) {
 		return nil, fmt.Errorf("invalid Numeric value: expected *big.Rat, got %T", c.Value.Value)
 	}
 
+	scale := 9
```
serprex, line 326: This should be a constant somewhere else, used in same places we use 9 for scale elsewhere

PR #1134, opened 2024-01-23
Remove CDCRecordStream SchemaDeltas

serprex:
Instead `RelationRecord` is passed in with rest of `Records`,
where connectors handle collecting schema deltas


serprex:
Little silly passing these records down when we could append to a slice held in `CDCRecordStream`

serprex:
Closing in favor of #1135

PR #1138, opened 2024-01-24
Avoid log.Fatalf in connector code, return error instead

serprex:
We don't want temporal activities calling `os.Exit`

Also prefer `fmt.Sprint(x)` to `fmt.Sprintf("%v", x)`

### DIFF
--- a/flow/connectors/utils/avro/avro_writer.go
+++ b/flow/connectors/utils/avro/avro_writer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"log"
 	"log/slog"
 	"os"
 	"sync/atomic"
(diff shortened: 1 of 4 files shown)


flow/connectors/utils/avro/avro_writer.go
```diff
@@ -186,14 +186,13 @@ func (p *peerDBOCFWriter) WriteOCF(w io.Writer) (int, error) {
...
+	var wg sync.WaitGroup
+	wg.Add(1)
 	go func() {
-		defer w.Close()
-		numRows, err := p.WriteOCF(w)
-		if err != nil {
-			log.Fatalf("%v", err)
-		}
-		numRowsWritten <- numRows
+		numRows, writeOcfError = p.WriteOCF(w)
```
serprex, line 194: I could've had `defer wg.Done()` at top, but opted not to. The difference is that now if `WriteOCF` panics the surrounding goroutine will be stuck, rather than returning 0. The ideal would be to return an error, which would mean detecting that the goroutine ended in a panic

Another way would be to introduce a `var noPanic bool` which gets set to true at the end of the goroutine so that if after `wg.Wait()` `noPanic` is false then we can report an error

serprex: reading over, we can rely on `w.Close` to synchronize return values

iskakaushik (approved):

PR #1140, opened 2024-01-24
Fix date and time handling

### LINKED ISSUE #1127: [Bug] Dates are malformed using CDC mirror from PostgreSQL (12.14) to Snowflake
PostgreSQL table DDL:

```sql
CREATE TABLE my_table(id varchar primary key, mydate date);

INSERT INTO public.my_table (id, mydate) VALUES ('1'::varchar, '1970-06-02'::date);
INSERT INTO public.my_table (id, mydate) VALUES ('2'::varchar, '1970-07-07'::date);
INSERT INTO public.my_table (id, mydate) VALUES ('3'::varchar, '1970-07-05'::date);
INSERT INTO public.my_table (id, mydate) VALUES ('4'::varchar, '1970-03-10'::date);
INSERT INTO public.my_table (id, mydate) VALUES ('5'::varchar, '1970-01-10'::date);
INSERT INTO public.my_table (id, mydate) VALUES ('6'::varchar, '1970-01-11'::date);
```

Gets replicated as the following to Snowflake via CDC initial copy mirror:

| ID (Varchar) | MYDATE (Date) |
|--------|--------|
| 1 | 2386-03-01 |
| 2 | 2481-12-27 |
| 3 | 2476-07-06 |
| 4 | 2156-03-06 |
| 5 | 1994-08-23 | 
| 6 | 1997-05-19 | 

**Issue:** The original dates from source table did not get carried over correctly to the destination
**Git Hash** used for testing: 6997d5e8daceb18173678cc4e6a843d408092002

Amogh-Bharadwaj:
Using `time.UnixMicro()` not a good idea for Snowflake since Snowflake decides on its own whether to interpret a number as milliseconds, microseconds and so on. For a date such as 1970-03-02 which is much smaller than 2023-09-08 or something, Snowflake interpets it as milliseconds, leading to wildly incorrect Date values on the target.
This PR removes the usage of Unix.Micro() for Snowflake AVRO

This PR also adds tests for time for QRep 

Fixes #1127


flow/connectors/snowflake/qrep_avro_consolidate.go
```diff
@@ -114,7 +120,6 @@ func (s *SnowflakeAvroConsolidateHandler) getCopyTransformation(copyDstTable str
 	copyOpts := []string{
 		"FILE_FORMAT = (TYPE = AVRO)",
 		"PURGE = TRUE",
-		"ON_ERROR = 'CONTINUE'",
```
serprex, line 117: Should this be reverted?

flow/e2e/bigquery/bigquery_helper.go
```diff
@@ -220,6 +220,8 @@ func toQValue(bqValue bigquery.Value) (qvalue.QValue, error) {
 		return qvalue.QValue{Kind: qvalue.QValueKindBoolean, Value: v}, nil
 	case civil.Date:
 		return qvalue.QValue{Kind: qvalue.QValueKindDate, Value: bqValue.(civil.Date).In(time.UTC)}, nil
```
serprex, line 222:
```suggestion
		return qvalue.QValue{Kind: qvalue.QValueKindDate, Value: v.In(time.UTC)}, nil
```

flow/e2e/test_utils.go
```diff
@@ -310,7 +312,7 @@ func PopulateSourceTable(pool *pgxpool.Pool, suffix string, tableName string, ro
 							'{"2026-01-17 10:00:00","2026-01-18 13:45:00"}',
 							'{true, false}',
 							'{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}',
-							 CURRENT_DATE, 'happy', '"a"=>"b"','POINT(1 2)','POINT(40.7128 -74.0060)',
+							'1970-06-02', CURRENT_TIME, CURRENT_TIME,'happy', '"a"=>"b"','POINT(1 2)','POINT(40.7128 -74.0060)',
```
serprex, line 315: Instead of `CURRENT_TIME` twice we should do `CURRENT_TIME` & `CURRENT_DATE`

Amogh-Bharadwaj:
Removed timetz from tests. From Postgres docs:

```
We do not recommend using the type time with time zone 
(though it is supported by PostgreSQL for legacy applications and for compliance with the SQL standard). 
PostgreSQL assumes your local time zone for any type containing only date or time.
```

serprex (approved):

PR #1142, opened 2024-01-24
Replace model.QRecord with []qvalue.QValue

serprex:


flow/connectors/postgres/qrep_query_executor.go
```diff
@@ -130,7 +130,7 @@ func (qe *QRepQueryExecutor) ProcessRows(
 	fieldDescriptions []pgconn.FieldDescription,
 ) (*model.QRecordBatch, error) {
 	// Initialize the record slice
-	records := make([]model.QRecord, 0)
+	records := make([][]qvalue.QValue, 0)
```
heavycrystal, line 133: will help readability if we do something like this

```
type QRecordA []qvalue.QValue
```

serprex: I disagree, moving `[]` to the end as `Arr` adds little & requires wrapping the type while passing it around, when we just want `[]qvalue.QValue` anyways

heavycrystal: For me it's that double slices read weird, and having an alias for that would help. Fair point on requiring type wrapping though, would a type alias work better? https://stackoverflow.com/questions/61247864/what-is-the-difference-between-type-alias-and-type-definition-in-go

Approving anyway as this is primarily a stylistic difference.

heavycrystal (approved):

flow/connectors/postgres/qrep_query_executor.go
serprex: Type alias is type wrapping, it's specifically what I want to avoid since it makes a type that you can use slice operators on without being an explicit slice type, so you'll end up casting it to slice when you want to do other things like call functions from `slices` on it

PR #1144, opened 2024-01-24
Avoid mutating react state, change exclude from string[] to Set<string>

serprex:


ui/app/mirrors/create/cdc/schemabox.tsx
```diff
@@ -125,17 +123,16 @@ const SchemaBox = ({
     e: React.MouseEvent<HTMLInputElement, MouseEvent>,
     schemaName: string
   ) => {
-    setHandlingAll(true);
```
serprex, line 128: removed `setHandlingAll` because this function is synchronous. If this should wait for resolution of `addTableColumns/removeTableColumns` then their promises need to be collected & then `new Promise(promises).then(() => setHandlingAll(false))`

ui/app/mirrors/create/mirrorcards.tsx
```diff
@@ -58,7 +58,7 @@ const MirrorCards = ({
                 borderRadius: '1rem',
               }}
             >
-              <div>
+              <label style={{ display: 'block' }}>
```
serprex, line 61: Experiment: does this let me click mirror type box without having to click on content / radio button in box?

Amogh-Bharadwaj (approved):
functionally tested

PR #1150, opened 2024-01-24
nexus: impl QueryExecutor for Catalog

serprex:
To share code, Catalog was storing a postgres query executor
Unfortunately, that caused two connections for every nexus connection

Remove is_connection_valid since it isn't used, & for pg was wrong
(see deadpool-postgres for how they check if connection is valid)

Alternative fix would've been to use more connection pools,
that could be added in the future if necessary


nexus/catalog/src/lib.rs
```diff
@@ -602,3 +594,80 @@ impl Catalog {
         })
     }
 }
+
+// This is mostly copied from peer-postgres
+#[async_trait::async_trait]
+impl QueryExecutor for Catalog {
+    #[tracing::instrument(skip(self, stmt), fields(stmt = %stmt))]
+    async fn execute(&self, stmt: &Statement) -> PgWireResult<QueryOutput> {
```
iskakaushik, line 602: is there a way to refactor to share this logic between postgres peer impl and this?

iskakaushik (approved):

PR #1155, opened 2024-01-25
EventHub processBatch: handle `ctx.Done()`

serprex:
Based on #1151,
the ticker.Reset doesn't need to be changed since ticker isn't shared between goroutines

### DIFF
--- a/flow/connectors/eventhub/hubmanager.go
+++ b/flow/connectors/eventhub/hubmanager.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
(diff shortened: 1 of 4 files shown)


flow/connectors/eventhub/hubmanager.go
```diff
@@ -104,6 +104,7 @@ func (m *EventHubManager) closeProducerClient(ctx context.Context, pc *azeventhu
 
 func (m *EventHubManager) Close(ctx context.Context) error {
 	var allErrors error
+	var hubsClosed int
```
iskakaushik, line 107: @serprex I've noticed the close of a single hub taking over 30 seconds, should we do a heart beat routine here?

serprex: wow, yeah

iskakaushik (approved):

PR #1160, opened 2024-01-25
Remove ShutdownSignal

serprex:
The idea behind ShutdownSignal is that we can use signals to gracefully stop workflows,
in practice this doesn't work

1. DROP MIRROR is a 1 minute timeout for shutdown to stop workflow, in cdc I had us cut off early, but even waiting for normalize to finish is too much for 1 minute
2. Signal isn't checked by flows until they reach main loop, in theory they could check, but in effect you're doubling the existing context error code

Solution: go straight to workflow cancelation. If graceful shutdown is desired (ie, finishing current syncs, finishing normalization), pause mirror before drop


flow/shared/constants.go
```diff
@@ -34,8 +34,8 @@ type (
 )
 
 const (
-	NoopSignal CDCFlowSignal = iota
-	ShutdownSignal
+	NoopSignal     CDCFlowSignal = iota
+	ReservedSignal               // previously ShutdownSignal
```
serprex, line 38: Not sure best way to do this, ideally it'd be removed but then it messes up things for existing paused mirrors

iskakaushik (approved):

Issue #929, opened 2023-12-29
go-routine leak detection in CI


iskakaushik:
cc: @serprex

serprex:
related #927
related #977

PR #1161, opened 2024-01-26
add leak detection

iskakaushik:


heavycrystal:
are we adding this to other test suites?

serprex:
#929

PR #1163, opened 2024-01-26
postgres connector: replace pgxpool.Pool with pgx.Conn

serprex:


iskakaushik:
for connections established by PeerDB, can we also set `application_name` to `peerdb` as a part of this.

iskakaushik (approved):

PR #1166, opened 2024-01-29
cdc: remove SetLastOffset

serprex:
This was being experimented with to help reduce slot size, however it's conflicting with reducing pool use

To have the two work together some kind of clone-for-new-connection mechanism needs to be made. But this code isn't used anymore, so defer implementing that mechanism for now by removing this code


iskakaushik (approved):

serprex:
While this removes passing SetLastOffset around between goroutines, I'm leaving it in the interface since it's a useful operation we were already implementing

PR #1167, opened 2024-01-29
external_metadata: use pgx.Conn when using non-catalog config

serprex:


iskakaushik (approved):

flow/connectors/external_metadata/store.go
```diff
@@ -19,32 +20,39 @@ const (
 	lastSyncStateTableName = "last_sync_state"
 )
 
+type Querier interface {
```
serprex, line 23: writing your own interface for being polymorphic over `pgx.Conn`/`pgxpool.Pool` is recommended by pgx author: https://github.com/jackc/pgx/issues/1188

PR #1173, opened 2024-01-30
Fix numeric datatype

Amogh-Bharadwaj:
**Pull side:** The way we were converting pgtype numeric to big.Rat was incorrect. Resulted in integers being written as 200 -> 199.99999 for PG->PG Qrep. This was caught by setting better values for numerics in our qrep tests. This PR adapts the logic from [jackc's numeric.go](https://github.com/jackc/pgtype/blob/31dd0e442de8bbd1fbf0865b3c76d40708d3e429/numeric.go#L422)

**compareNumeric:** For numbers like 1.123 or 1.12345678 (less than 9 digits to the right), we now assert that they should be exactly equal. If one of the values has more than 9 digits to the right, we fall back to the difference less than 1e9 check. 

This change was tested with the old `processNumeric` code, and it caught the suboptimal conversion of 200 -> 199.9999...

### DIFF
--- a/flow/connectors/postgres/qvalue_convert.go
+++ b/flow/connectors/postgres/qvalue_convert.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"math"
 	"math/big"
 	"strings"
 	"time"
@@ -525,12 +524,15 @@ func numericToRat(numVal *pgtype.Numeric) (*big.Rat, error) {
 			return nil, errors.New("numeric value is infinity")
 		}
 
-		rat := new(big.Rat)
-
-		rat.SetInt(numVal.Int)
-		divisor := new(big.Rat).SetFloat64(math.Pow10(int(-numVal.Exp)))
-		rat.Quo(rat, divisor)
-
+		rat := new(big.Rat).SetInt(numVal.Int)
+		big10 := big.NewInt(10)
+		if numVal.Exp > 0 {
+			mul := new(big.Int).Exp(big10, big.NewInt(int64(numVal.Exp)), nil)
+			rat.Mul(rat, new(big.Rat).SetInt(mul))
+		} else if numVal.Exp < 0 {
+			mul := new(big.Int).Exp(big10, big.NewInt(int64(-numVal.Exp)), nil)
+			rat.Quo(rat, new(big.Rat).SetInt(mul))
+		}
 		return rat, nil
 	}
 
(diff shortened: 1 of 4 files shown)


flow/connectors/postgres/qvalue_convert.go
```diff
@@ -525,12 +524,15 @@ func numericToRat(numVal *pgtype.Numeric) (*big.Rat, error) {
 			return nil, errors.New("numeric value is infinity")
 		}
 
-		rat := new(big.Rat)
-
-		rat.SetInt(numVal.Int)
-		divisor := new(big.Rat).SetFloat64(math.Pow10(int(-numVal.Exp)))
-		rat.Quo(rat, divisor)
-
+		rat := new(big.Rat).SetInt(numVal.Int)
```
serprex, line 527: pgtype exposes this with [AssignTo](https://pkg.go.dev/github.com/jackc/pgtype#ArrayType.AssignTo)
```go
var rat big.Rat
err := numVal.AssignTo(&rat)
if err != nil {
  return new(big.Rat), nil
}
return rat, nil
```

Amogh-Bharadwaj: We're using `github.com/jackc/pgx/v5/pgtype` which does not seem to expose this: https://github.com/jackc/pgx/blob/master/pgtype/numeric.go

iskakaushik: We could do something like:

```
iv := numVal.toBigInt()
var rat big.Rat
rat.SetInt(iv)
```

iskakaushik:
Please wait for @serprex  to approve as well.

flow/connectors/postgres/qvalue_convert.go
serprex: https://pkg.go.dev/github.com/jackc/pgx/pgtype#Decimal.AssignTo

serprex: Looks like they decided to prefer shopspring-decimal https://github.com/jackc/pgx/wiki/Numeric-and-decimal-support

serprex: But we want `*big.Rat` because that's what bigquery takes, not a proper decimal type

serprex (approved):

serprex:
Discussed on slack: passing numeric around as big.Rat isn't preferable to passing around as string in the absence of using it for calcs. Since `big.Rat` is two `big.Int` which is `(bool, []Word)` so in terms of memory pressure you're better off using `string`

Amogh-Bharadwaj:
Landing this as a needed short-term fix for this data type. Issued #1175 to address [@serprex comment](https://github.com/PeerDB-io/peerdb/pull/1173#issuecomment-1917332927)

PR #1179, opened 2024-01-30
Snowflake: stop storing metadata on warehouse; store in catalog

serprex:


flow/connectors/external_metadata/store.go
```diff
@@ -118,29 +121,49 @@ func (p *PostgresMetadataStore) SetupMetadata() error {
...
+
+	_, err = p.conn.Exec(p.ctx,
+		`CREATE INDEX IF NOT EXISTS ix_qrep_metadata_partition_id ON `+
+			p.QualifyTable(qrepTableName)+
+			` USING hash (partition_id)`)
+	if err != nil && !utils.IsUniqueError(err) {
+		p.logger.Error("failed to create qrep metadata index", slog.Any("error", err))
+		return err
+	}
+
```
Amogh-Bharadwaj: Eventhub and S3 don't need this I think. Can we have a new `SetupDWHMetadata()` method which sets these tables up or just pass a flag

serprex: Would rather have empty tables than differing schemas, we're considering moving metadata table setup into migration files

Amogh-Bharadwaj: Got it. Makes sense

Amogh-Bharadwaj (approved):

iskakaushik (approved):

PR #1182, opened 2024-01-31
e2e: remove Test_Toast_Nochanges

serprex:
This test doesn't really make sense,
since the goroutine can finish before or after the workflow runs


serprex:
this was also causing flakiness because it was possible for goroutine to use `s.conn` at same time as table equality check

heavycrystal (approved):

PR #1186, opened 2024-01-31
utils.RecordHeartbeat: use activity.IsActivity

serprex:
Ideally e2e tests would pass context with mock activity,
but in meantime this'll reduce noise in test logs
& is a bit more efficient

### DIFF
--- a/flow/connectors/utils/heartbeat.go
+++ b/flow/connectors/utils/heartbeat.go
@@ -3,7 +3,6 @@ package utils
 import (
 	"context"
 	"fmt"
-	"log/slog"
 	"time"
 
 	"go.temporal.io/sdk/activity"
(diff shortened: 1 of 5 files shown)


Amogh-Bharadwaj (approved):

flow/connectors/utils/heartbeat.go
```diff
@@ -37,12 +36,8 @@ func HeartbeatRoutine(
 
 // if the functions are being called outside the context of a Temporal workflow,
 // activity.RecordHeartbeat panics, this is a bandaid for that.
```
heavycrystal, line 38: outdated comment

serprex: Not quite outdated. This function is to avoid panic from `activity.RecordHeartbeat` still

PR #1200, opened 2024-02-02
Eventhubs: Refactor Destination Table Semantics

Amogh-Bharadwaj:
The destination table name in table mapping of Create CDC Mirror now expected to look like: 
```
<eventhub_name>.<eventhub_namespace>.<unique_identifier>.<partition_key>
```
This is to allow multiple source tables to be mapped to same eventhub partition.
`unique identifier` can be the source table name or anything else. It's purely to distinguish destination table names

### DIFF
--- a/flow/connectors/eventhub/scoped_eventhub.go
+++ b/flow/connectors/eventhub/scoped_eventhub.go
@@ -9,26 +9,30 @@ import (
 // partition_column is the column in the table that is used to determine
 // the partition key for the eventhub. Partition value is one such value of that column.
 type ScopedEventhub struct {
-	PeerName           string
+	EventhubNamespace  string
 	Eventhub           string
+	DestinationTable   string
 	PartitionKeyColumn string
 	PartitionKeyValue  string
 }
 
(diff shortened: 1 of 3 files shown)


flow/connectors/eventhub/scoped_eventhub.go
```diff
@@ -9,25 +9,30 @@ import (
...
 
-	// support eventhub name and partition key with hyphens etc.
-	eventhubPart := strings.Trim(parts[1], `"`)
-	partitionPart := strings.Trim(parts[2], `"`)
+	// support eventhub namespace, eventhub name, partition key with hyphens etc.
+	// part[2] will be some table identifier.
+	// It's just so that we have distinct destination table names
+	// in create mirror's table mapping.
+	// We can ignore it.
+	eventhubPart := strings.Trim(parts[0], `"`)
```
serprex, line 31: These parts would ideally be sent as `[]string` in protocol, avoid this incorrect parsing

iskakaushik (approved):

iskakaushik:
landing for customer urgency

PR #1202, opened 2024-02-02
nexus/catalog: fully qualify tables in queries

serprex:
Recently combined catalog connections for nexus connections,
now `SET search_path` queries change search_path for our backend connection, breaking things


serprex:
Alternative fix would be to keep catalog executor connections separate from our internal use of catalog, having a connection pooler shared by all nexus connections

iskakaushik (approved):

PR #1204, opened 2024-02-05
Postgres streaming using CopyFromFunc

heavycrystal:
In the Postgres connector, the sync side doesn't take advantage of record streaming over Go channels, loading all records into an array and then processing when the channel is closed from the pull side. This PR addresses that and now Postgres copies from a function invocation instead of an array, therefore copying records while they're being read from the channel. 

This leads to lower memory utilization (no massive buffer) and improved parallelism.

also changing some `Info` logs back to `Debug` to prevent spam

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -336,7 +336,7 @@ func (c *PostgresConnector) SyncRecords(req *model.SyncRecordsRequest) (*model.S
 				return nil, fmt.Errorf("unsupported record type for Postgres flow connector: %T", typedRecord)
 			}
 
-			numRecords++
+			numRecords += 1
 			tableNameRowsMapping[record.GetDestinationTableName()] += 1
 			return row, nil
 		}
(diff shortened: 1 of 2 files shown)


flow/connectors/postgres/postgres.go
```diff
@@ -257,80 +257,88 @@ func (c *PostgresConnector) SyncRecords(req *model.SyncRecordsRequest) (*model.S
...
-				itemsJSON,
-				2,
-				itemsJSON,
-				req.SyncBatchID,
-				"",
-			})
-			tableNameRowsMapping[typedRecord.DestinationTableName] += 1
-		default:
-			return nil, fmt.Errorf("unsupported record type for Postgres flow connector: %T", typedRecord)
+			numRecords++
```
serprex, line 339:
```suggestion
			numRecords += 1
```

serprex:
Include in PR description perf improvement I suspect this has. Is it that this makes reading & copying concurrent while also reducing memory usage?

serprex (approved):

PR #1209, opened 2024-02-06
UI: SSH fields in Create PG: change field to file upload

Amogh-Bharadwaj:
Screenshot 2024-02-08 at 11 58 32 PM


iamKunalGupta (commented):
We can do:
1. `Private Key` -> `SSH Private Key`
2. `Host Key` -> `Host's Public Key`. Also can be changed to a file upload instead (similar to private key). Tip can additionally tell it generally resides at `/etc/ssh/ssh_host_[algo]_key.pub`

serprex:
Being able to paste host key is fine; see github interface where it's a textarea. Azure devops was same

serprex (approved):

Amogh-Bharadwaj:
This has been functionally tested. Screenshot updated

PR #1211, opened 2024-02-06
Maintain replication connection between sync flows

serprex:
Currently we reconnect with each sync flow, requiring repeatedly starting replication. This can take an exceedingly long time for some workloads on some databases

Fix: use temporal session to share state between activities, use a single source connector throughout cdc flow, & move replication connection back into source connection

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -42,6 +42,8 @@ type SlotSnapshotSignal struct {
 type FlowableActivity struct {
 	CatalogPool *pgxpool.Pool
 	Alerter     *alerting.Alerter
+	CdcCacheRw  sync.RWMutex
+	CdcCache    map[string]connectors.CDCPullConnector
 }
 
 func (a *FlowableActivity) CheckConnection(
@@ -204,10 +206,71 @@ func (a *FlowableActivity) CreateNormalizedTable(
 	}, nil
 }
 
+func (a *FlowableActivity) MaintainPull(
+	ctx context.Context,
+	config *protos.FlowConnectionConfigs,
+	sessionID string,
+) error {
+	srcConn, err := connectors.GetCDCPullConnector(ctx, config.Source)
+	if err != nil {
+		return err
+	}
+	defer connectors.CloseConnector(ctx, srcConn)
+
+	if err := srcConn.SetupReplConn(ctx); err != nil {
+		return err
+	}
+
+	a.CdcCacheRw.Lock()
+	a.CdcCache[sessionID] = srcConn
+	a.CdcCacheRw.Unlock()
+
+	ticker := time.NewTicker(15 * time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			activity.RecordHeartbeat(ctx, "keep session alive")
+			if err := srcConn.ReplPing(ctx); err != nil {
+				activity.GetLogger(ctx).Error("Failed to send keep alive ping to replication connection", slog.Any("error", err))
+			}
+		case <-ctx.Done():
+			a.CdcCacheRw.Lock()
+			delete(a.CdcCache, sessionID)
+			a.CdcCacheRw.Unlock()
+			return nil
+		}
+	}
+}
+
+func (a *FlowableActivity) WaitForSourceConnector(ctx context.Context, sessionID string) error {
+	logger := activity.GetLogger(ctx)
+	attempt := 0
+	for {
+		a.CdcCacheRw.RLock()
+		_, ok := a.CdcCache[sessionID]
+		a.CdcCacheRw.RUnlock()
+		if ok {
+			return nil
+		}
+		activity.RecordHeartbeat(ctx, "wait another second for source connector")
+		attempt += 1
+		if attempt > 2 {
+			logger.Info("waiting on source connector setup", slog.Int("attempt", attempt))
+		}
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		time.Sleep(time.Second)
+	}
+}
+
 func (a *FlowableActivity) SyncFlow(
 	ctx context.Context,
 	config *protos.FlowConnectionConfigs,
 	options *protos.SyncFlowOptions,
+	sessionID string,
 ) (*model.SyncResponse, error) {
 	flowName := config.FlowJobName
 	ctx = context.WithValue(ctx, shared.FlowNameKey, flowName)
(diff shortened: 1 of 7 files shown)


flow/activities/flowable.go
```diff
@@ -186,7 +187,86 @@ func (a *FlowableActivity) CreateNormalizedTable(
...
+	defer connectors.CloseConnector(srcConn)
+	a.CdcCacheRw.Lock()
+	a.CdcCache[sessionID] = srcConn
+	a.CdcCacheRw.Unlock()
+
+	ticker := time.NewTicker(15 * time.Second)
+	defer ticker.Stop()
+
+	/*
+			scan := make(chan struct{}, 1)
```
serprex, line 208: This doesn't seem necessary, particularly full scan. Maybe move this to creating activity as a constant goroutine. Idea is that it'll detect connectors which have failed & remove them from map. ofc, then we need to handle in StartFlow when connector isn't in cache. This would make more sense then to get rid of MaintainPull having source connector tear down. But for drop purposes we'll want that anyway so we'd still need to remove here & there

serprex:
TODO: actually move start replication into setup activities

serprex:
Kevin pointed out a problem: in non parallel sync-normalize long normalize will have connection silent for too long & postgres can drop connection. Need to have keepalive logic. MaintainPull could work, but then we need to make sure to use synchronization between MaintainPull & StartFlow

flow/workflows/cdc_flow.go
```diff
@@ -493,37 +535,48 @@ func CDCFlowWorkflowWithConfig(
...
-		// execute the sync flow as a child workflow
-		childSyncFlowOpts := workflow.ChildWorkflowOptions{
-			WorkflowID:        syncFlowID,
-			ParentClosePolicy: enums.PARENT_CLOSE_POLICY_REQUEST_CANCEL,
-			RetryPolicy: &temporal.RetryPolicy{
-				MaximumAttempts: 20,
-			},
-			SearchAttributes:    mirrorNameSearch,
-			WaitForCancellation: true,
+		lastSyncFuture := workflow.ExecuteLocalActivity(syncMetaCtx, flowable.GetLastSyncedID, lastSyncInput)
```
iskakaushik, line 549: @serprex  should we move `GetLastSyncedId` into `StartFlow`, I think this moves us closer to the world where we flush to destination more frequently and retries pickup from a newer point.

flow/activities/flowable.go
```diff
@@ -45,9 +45,10 @@ type SlotSnapshotSignal struct {
 type FlowableActivity struct {
 	CatalogPool *pgxpool.Pool
 	Alerter     *alerting.Alerter
+	CdcCacheRw  sync.RWMutex
```
iskakaushik, line 48: Maybe have a replication connection manager struct that takes care of:
1. This map and locking.
2. Keeping track of the connection health and lifecycle.
3. Any other additional metadata pertaining to the replication connection.

flow/activities/flowable.go
```diff
@@ -186,7 +184,64 @@ func (a *FlowableActivity) CreateNormalizedTable(
...
+	for {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		a.CdcCacheRw.RLock()
+		srcConn, ok := a.CdcCache[sessionID]
+		a.CdcCacheRw.RUnlock()
+		if ok {
+			if err := srcConn.SetupReplConn(); err != nil {
+				activity.GetLogger(ctx).Error("Failed to setup replication connection", "Error", err)
```
iskakaushik, line 233: can you add a comment here explaining why we are trying this multiple times?

flow/activities/flowable.go
serprex: Tracking connection health & lifecycle are part of `MaintainPull` which needs to exist either way to keep heartbeating session

flow/workflows/cdc_flow.go
serprex: Done, it's cleaner, but not sure it helps in that regard. Better would be to have getoffset/setoffset managed by source connector

flow/activities/flowable.go
```diff
@@ -207,7 +218,93 @@ func (a *FlowableActivity) CreateNormalizedTable(
...
+	sessionID string,
+) error {
+	for {
+		a.CdcCacheRw.RLock()
+		_, ok := a.CdcCache[sessionID]
+		a.CdcCacheRw.RUnlock()
+		if ok {
+			return nil
+		}
+		activity.RecordHeartbeat(ctx, "wait another second for source connector")
```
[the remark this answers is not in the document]
serprex: done, avoiding noise by not logging first 2 attempts

iskakaushik:
does using `OriginalRunID` over uuid have any unexpected consequences, if we were to reset the state of the workflow, would that re-create the new child workflow or would it complain that the run-id matches?

flow/activities/flowable.go
```diff
@@ -206,7 +209,96 @@ func (a *FlowableActivity) CreateNormalizedTable(
...
+	}
+
+	a.CdcCacheRw.Lock()
+	a.CdcCache[sessionID] = srcConn
+	a.CdcCacheRw.Unlock()
+
+	ticker := time.NewTicker(15 * time.Second)
+	defer ticker.Stop()
+
+	slotNameForMetrics := fmt.Sprintf("peerflow_slot_%s", config.FlowJobName)
```
iskakaushik, line 234: @serprex lets move slot size collection to `HeartBeatFlow` the way its structured right now slot size is recorded only during sync flow runs which is awkward. This is not specific to this PR.

flow/connectors/postgres/postgres.go
```diff
@@ -95,19 +108,124 @@ func (c *PostgresConnector) CreateReplConn(ctx context.Context) (*pgx.Conn, erro
...
+		err = c.startReplication(ctx, slotName, opts)
+		if err != nil {
+			return fmt.Errorf("error starting replication: %w", err)
+		}
+
+		c.logger.Info(fmt.Sprintf("started replication on slot %s at startLSN: %d", slotName, startLSN))
+		c.replState = &ReplState{
+			Slot:        slotName,
+			Publication: publicationName,
+			Offset:      req.LastOffset,
```
heavycrystal, line 180: shouldn't this be startLSN, otherwise we can send `StandbyMessages` off by 1

serprex: StandbyMessage is supposed to be the offset up to which we've read. startLSN still gets +1 because that's the LSN after which we've read. We track the former

flow/activities/flowable.go
serprex: Makes sense, HeartbeatFlow could also use some improvement so it doesn't cancel activity after 1 week of uptime

serprex: #1328

flow/workflows/cdc_flow.go
```diff
@@ -352,6 +353,55 @@ func CDCFlowWorkflowWithConfig(
...
+	}
+	syncSessionCtx, err := workflow.CreateSession(ctx, sessionOptions)
+	if err != nil {
+		return nil, err
+	}
+	defer workflow.CompleteSession(syncSessionCtx)
+	sessionInfo := workflow.GetSessionInfo(syncSessionCtx)
+
+	syncCtx := workflow.WithActivityOptions(syncSessionCtx, workflow.ActivityOptions{
+		StartToCloseTimeout: 72 * time.Hour,
```
iskakaushik, line 369: initial copy can take longer than 72 hours, can we make it 2 weeks for now.

PR #1212, opened 2024-02-06
tests for mixed case + fixes/cleanup

heavycrystal:
1) Added test for mixed case column/pkey/table names in PG
2) BatchSize is now taken directly from config instead of from limits which is derived from config, Limits is now empty so removed.
3) table name in addpub should be using `SchemaTable.String()` instead of `QuoteIdentifier`

dynamic properties test isn't working in CI, will debug and raise it separately


serprex (approved):

serprex:
Might be better to split the new tests from the misc improvements

PR #1215, opened 2024-02-07
drop_flow: fix hanging bug

serprex:
Somehow between testing & merging I screwed up adding initial activities to selector,
causing drop flow to indefinitely hang

Also need explicit listen on ctx.Done in selector since `selector.Select(ctx)` doesn't return if ctx is canceled, so if the activities hung in selector that'd cause our workflow to ignore cancelation


heavycrystal (approved):

serprex:
Amogh tested this fixes things, replication slot was cleaned up, but there's still some connections hanging

PR #1223, opened 2024-02-07
Fix normalize heartbeats

Amogh-Bharadwaj:
Was seeing activity heartbeat timeouts during destination table setup for a large number of tables. This PR attempts to fix that


heavycrystal:
should be sending a heartbeat every table instead of a fixed duration heartbeat

iskakaushik:
some times we have seen each table creation taking longer than 30s.

iskakaushik (approved):

PR #1226, opened 2024-02-07
Numeric: NaN and INF support

Amogh-Bharadwaj:
Sets NaNs and Infs to nil for numeric


iskakaushik:
@Amogh-Bharadwaj  bigquery seems to support NaN and inf types: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types

same with Snowflake: https://docs.snowflake.com/en/sql-reference/data-types-numeric

can we move them as the same types?

Amogh-Bharadwaj:
> @Amogh-Bharadwaj bigquery seems to support NaN and inf types: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types
> 
> same with Snowflake: https://docs.snowflake.com/en/sql-reference/data-types-numeric
> 
> can we move them as the same types?

@iskakaushik 
The issue is that Avro (initial load) and JSON (cdc) does not support NaN and INF for decimal schema.
Also:
BigQuery doesn't seem to support NaN and INF for BigNumerics/Numerics:
```sql
CREATE TABLE bignantest(bn BIGNUMERIC);
INSERT INTO bignantest VALUES(CAST('NaN' as  BIGNUMERIC)); -- same with INF
```
```
Query error: Invalid BIGNUMERIC value: NaN at [2:1]
```
Seems like it's only supported for float

Amogh-Bharadwaj:
Seems like Snowflake doesn't support it either: 

```sql
create table nantest(n numeric);
insert into nantest values('NaN'::numeric);
insert into nantest values('INF');
```

```
Invalid expression [TO_NUMBER('NaN')] in VALUES clause
Numeric value 'INF' is not recognized
```

iskakaushik:
got it, in that case this is fine.

iskakaushik (approved):

PR #1231, opened 2024-02-08
spill to disk based on flow-worker memory usage

heavycrystal:
the existing number of records based threshold is still present, both can be active at the same time.

Best used with `GOMEMLIMIT` to keep heap more aligned with set memory limit.

### DIFF
--- a/flow/connectors/utils/cdc_records/cdc_records_storage.go
+++ b/flow/connectors/utils/cdc_records/cdc_records_storage.go
@@ -85,7 +85,7 @@ func (c *cdcRecordsStore) initPebbleDB() error {
 	return nil
 }
 
-func (c *cdcRecordsStore) thresholdsExceeded() bool {
+func (c *cdcRecordsStore) diskSpillThresholdsExceeded() bool {
 	if len(c.inMemoryRecords) >= c.numRecordsSwitchThreshold {
 		c.thresholdReason = fmt.Sprintf("more than %d primary keys read, spilling to disk",
 			c.numRecordsSwitchThreshold)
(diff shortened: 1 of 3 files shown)


flow/peerdbenv/config.go
```diff
@@ -41,9 +41,19 @@ func PeerDBCDCIdleTimeoutSeconds(providedValue int) time.Duration {
...
+}
+
+// PEERDB_CDC_DISK_SPILL_RECORDS_THRESHOLD, negative numbers means memory threshold disabled
+func PeerDBCDCDiskSpillMemPercentThreshold() int {
+	return getEnvInt("PEERDB_CDC_DISK_SPILL_MEM_PERCENT_THRESHOLD", -1)
+}
+
+// PEERDB_FLOW_WORKER_MAX_MEM_BYTES, 0 means no maximum
+func PeerDBFlowWorkerMaxMemBytes() uint64 {
+	return getEnvUint[uint64]("PEERDB_FLOW_WORKER_MAX_MEM_BYTES", 0)
```
iamKunalGupta, line 56: Let's do `GOMEMLIMIT` itself, makes more sense.

flow/connectors/utils/cdc_records/cdc_records_storage.go
```diff
@@ -72,15 +85,32 @@ func (c *cdcRecordsStore) initPebbleDB() error {
 	return nil
 }
 
+func (c *cdcRecordsStore) thresholdsExceeded() bool {
```
iamKunalGupta, line 88: `diskSpillThresholdsExceeded` can be a better name

iamKunalGupta (commented):
small naming changes, rest looks good

iamKunalGupta (approved):

PR #1241, opened 2024-02-09
BQ merge fixes

serprex:
BQ `it.Next(&row)` seems to mutate the target,
https://github.com/googleapis/google-cloud-go/blob/f049c9751415f9fc4c81c1839a8371782cfc016c/bigquery/value.go#L490
so reallocate a fresh row each iteration

ArrayChunks: replace with ArrayIterChunks. Document behavior when input is empty slice
Also reduces allocations, since we've seen people having 1000+ columns

go 1.22 has a rangefunc experiment, so eventually we can replace this with that https://go.dev/wiki/RangefuncExperiment

### DIFF
--- a/flow/connectors/utils/array.go
+++ b/flow/connectors/utils/array.go
@@ -18,17 +18,23 @@ func ArrayMinus[T comparable](first, second []T) []T {
 	return result
 }
 
-func ArrayChunks[T any](slice []T, size int) [][]T {
-	var partitions [][]T
-
-	for size < len(slice) {
-		slice, partitions = slice[size:], append(partitions, slice[0:size])
+// Call f with subslices of slice. An empty slice will call f once with nil.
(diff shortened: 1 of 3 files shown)


flow/connectors/utils/array.go
```diff
@@ -18,17 +18,11 @@ func ArrayMinus[T comparable](first, second []T) []T {
 	return result
 }
 
-func ArrayChunks[T any](slice []T, size int) [][]T {
-	var partitions [][]T
-
-	for size < len(slice) {
-		slice, partitions = slice[size:], append(partitions, slice[0:size])
+func IterChunks[T any](slice []T, size int, f func(chunk []T)) {
+	slicelen := len(slice)
```
heavycrystal, line 22: defensive: exit early if size <= 0

heavycrystal: otherwise will hang indefinitely

heavycrystal (approved):

PR #1256, opened 2024-02-12
Filter out invalid BigQuery timestamps

### LINKED ISSUE #1254: BigQuery Timestamp: Don't accept values less than 0 AD and more than 9999AD
This is the [range for BigQuery's timestamp](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp_type) and it will not allow values outside the range to be inserted.

Log the values which are rejected.

Amogh-Bharadwaj:
Fixes #1254 
Adds a test for it in QRep

Such values aren't accepted by pgtype's timestamp so there's no question of supporting it via CDC (verified via automated test I wrote for CDC)


flow/model/qvalue/avro_converter.go
```diff
@@ -249,6 +250,17 @@ func (c *QValueAvroConverter) ToAvroValue() (interface{}, error) {
 				return t.(int64), nil
 			}
 		}
+
+		// Bigquery will not allow timestamp if it is less than 1AD and more than 9999AD
+		// So, we make such timestamps as null
```
serprex, line 255:
```suggestion
		// So make such timestamps null
```

flow/model/qvalue/avro_converter.go
```diff
@@ -397,7 +409,20 @@ func (c *QValueAvroConverter) processGoTimestampTZ() (interface{}, error) {
 	if c.TargetDWH == QDWHTypeSnowflake {
 		return t.Format("2006-01-02 15:04:05.999999-0700"), nil
 	}
-	return t.UnixMicro(), nil
+
+	tMicro := t.UnixMicro()
+	// Bigquery will not allow timestamp if it is less than 1AD and more than 9999AD
+	// So, we make such timestamps as null
```
serprex, line 415:
```suggestion
	// So make such timestamps null
```

flow/model/qvalue/avro_converter.go
```diff
@@ -415,7 +440,19 @@ func (c *QValueAvroConverter) processGoTimestamp() (interface{}, error) {
 	if c.TargetDWH == QDWHTypeSnowflake {
 		return t.Format("2006-01-02 15:04:05.999999"), nil
 	}
-	return t.UnixMicro(), nil
+
+	tMicro := t.UnixMicro()
+	// Bigquery will not allow timestamp if it is less than 1AD and more than 9999AD
+	// So, we make such timestamps as null
```
serprex, line 446:
```suggestion
	// So make such timestamps null
```

flow/model/qvalue/avro_converter.go
serprex: Seems like this should be a helper function since it's repeated 3 times

flow/model/qvalue/avro_converter.go
```diff
@@ -415,7 +440,19 @@ func (c *QValueAvroConverter) processGoTimestamp() (interface{}, error) {
...
+	if c.TargetDWH == QDWHTypeBigQuery {
+		if tMicro < 0 || tMicro > 253402300799999999 {
+			slog.Info("Unlimited Timestamp value for BigQuery",
```
serprex, line 449: Don't use global slog

serprex (dismissed):

flow/model/qvalue/timestamp.go
```diff
@@ -0,0 +1,18 @@
+package qvalue
+
+import (
+	"time"
+
+	"go.temporal.io/sdk/log"
+)
+
+// Bigquery will not allow timestamp if it is less than 1AD and more than 9999AD
+func DisallowedTimestamp(dwh QDWHType, t time.Time, logger log.Logger) bool {
```
serprex, line 10: not checking `dwh` so this function isn't only applying to BQ

flow/model/qvalue/avro_converter.go
```diff
@@ -414,7 +419,17 @@ func (c *QValueAvroConverter) processGoTimestampTZ() (interface{}, error) {
...
-	return t.UnixMicro(), nil
+
+	tMicro := t.UnixMicro()
+	// Bigquery will not allow timestamp if it is less than 1AD and more than 9999AD
+	// So make such timestamps null
+	if c.TargetDWH == QDWHTypeBigQuery {
+		if DisallowedTimestamp(c.TargetDWH, t, c.logger) {
+			return nil, nil
+		}
+	}
```
serprex, line 430:
```suggestion
	if DisallowedTimestamp(c.TargetDWH, t, c.logger) {
		return nil, nil
	}
```

flow/model/qvalue/avro_converter.go
```diff
@@ -414,7 +419,17 @@ func (c *QValueAvroConverter) processGoTimestampTZ() (interface{}, error) {
 	if c.TargetDWH == QDWHTypeSnowflake {
 		return t.Format("2006-01-02 15:04:05.999999-0700"), nil
 	}
-	return t.UnixMicro(), nil
+
+	tMicro := t.UnixMicro()
```
serprex, line 423:
```suggestion
```
can move this back to `return t.UnixMicro(), nil`

Amogh-Bharadwaj:
updated description

serprex (approved):

PR #1275, opened 2024-02-13
better CDC error logging

iskakaushik:


flow/connectors/utils/avro/avro_writer.go
```diff
@@ -190,11 +192,14 @@ func (p *peerDBOCFWriter) WriteRecordsToS3(ctx context.Context, bucketName, key
...
+	buf := buffer.New(32 * 1024 * 1024) // 32MB in memory Buffer
+	r, w := nio.Pipe(buf)
+
 	defer r.Close()
 	var writeOcfError error
 	var numRows int
 	var noPanic bool
+
 	go func() {
 		defer w.Close()
```
serprex, line 204: We should use recover here to catch panicErr & set `writeOcfError` to panicErr

serprex (approved):

PR #1280, opened 2024-02-13
CreateNormalizedTable: lift loop, make it based on separate connector

serprex:
Snowflake was sending a heartbeat for each table, BQ was using HeartbeatRoutine, clickhouse wasn't sending heartbeats. Unify on HeartbeatRoutine

Move to a separate connector interface so s3/eventhubs etc don't need to implement the start/abort/finish methods as stubs too

Also introduce a generalized GetConnectorAs, use internally for existing functions

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -163,7 +163,7 @@ func (a *FlowableActivity) CreateNormalizedTable(
 	if err != nil {
 		return nil, fmt.Errorf("failed to setup normalized tables tx: %w", err)
 	}
-	defer conn.AbortSetupNormalizedTables(ctx, tx)
+	defer conn.CleanupSetupNormalizedTables(ctx, tx)
 
 	numTablesSetup := atomic.Uint32{}
 	totalTables := uint32(len(config.TableNameSchemaMapping))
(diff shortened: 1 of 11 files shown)


flow/activities/flowable.go
```diff
@@ -141,25 +142,69 @@ func (a *FlowableActivity) GetTableSchema(
...
 	defer connectors.CloseConnector(ctx, conn)
 
-	setupNormalizedTablesOutput, err := conn.SetupNormalizedTables(ctx, config)
+	tx, err := conn.StartSetupNormalizedTables(ctx)
 	if err != nil {
-		a.Alerter.LogFlowError(ctx, config.FlowName, err)
-		return nil, fmt.Errorf("failed to setup normalized tables: %w", err)
+		return nil, fmt.Errorf("failed to setup normalized tables tx: %w", err)
 	}
+	defer conn.AbortSetupNormalizedTables(ctx, tx)
```
iskakaushik, line 166: defer aborts feels like we will always Abort. I'm wondering if there is a way to say this RollbackIfError or something,

serprex: Considered `MaybeRollback` but went with `Cleanup` to keep it a general "defer cleanup of whatever your tx object is"

PR #1288, opened 2024-02-14
Misc cleanup of qvalue code

serprex:
Split out from #1285


flow/model/qvalue/avro_converter.go
```diff
@@ -169,6 +169,10 @@ func NewQValueAvroConverter(value QValue, targetDWH QDWHType, nullable bool) *QV
 }
 
 func (c *QValueAvroConverter) ToAvroValue() (interface{}, error) {
+	if c.Nullable && c.Value.Value == nil {
```
serprex, line 172: This makes a lot of the code below dead, but cleaning up that can be done later

PR #1289, opened 2024-02-14
fixes to handling of custom publications for AddTablesToPublication

heavycrystal:
1. need to fetch tables schema-qualified for comparison
2. order of ArrayMinus was wrong
3. also add validation for custom publication existing

### DIFF
--- a/flow/connectors/postgres/client.go
+++ b/flow/connectors/postgres/client.go
@@ -629,7 +629,7 @@ func (c *PostgresConnector) getDefaultPublicationName(jobName string) string {
 
 func (c *PostgresConnector) CheckSourceTables(ctx context.Context, tableNames []string, pubName string) error {
 	if c.conn == nil {
-		return fmt.Errorf("check tables: conn is nil")
+		return errors.New("check tables: conn is nil")
 	}
 
 	// Check that we can select from all tables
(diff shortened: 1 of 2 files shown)


serprex (approved):

serprex:
Instead of `schemaname||'.'||tablename` you can have postgresql address identifier safety with %I: `format('%I.%I', schemaname, tablename)`

heavycrystal:
> Instead of `schemaname||'.'||tablename` you can have postgresql address identifier safety with %I: `format('%I.%I', schemaname, tablename)`
> 
> https://www.postgresql.org/docs/current/plpgsql-statements.html#PLPGSQL-QUOTE-LITERAL-EXAMPLE
> 
> Then get rid of `schemaTable, err := utils.ParseSchemaTable(additionalSrcTable)`
> 
> Alternative would be to `select schemaname, tablename` & then quote the identifiers together in code

We're getting `additionalSourceTables` completely unescaped from the UI, and we're comparing it with the unescaped identifiers being generated through simple string concat. Proper escaping is necessary, but it'll have to happen at both ends. Will handle this in the future.

iskakaushik (approved):
approving to unblock a customer.

PR #1293, opened 2024-02-14
BigQuery: Skip merge if no primary keys present

Amogh-Bharadwaj:
Merge statements currently are syntactically incorrect for tables without a primary key. This case is ideally not seen as tables without pkey/replica identity full are filtered out the UI level but in branches where we rely on signals for adding tables, this case can go unchecked and normalize flow fails.

This PR skips a table during merge if there are no primary keys for it.

### DIFF
--- a/flow/connectors/bigquery/bigquery.go
+++ b/flow/connectors/bigquery/bigquery.go
@@ -518,7 +518,7 @@ func (c *BigQueryConnector) NormalizeRecords(ctx context.Context, req *model.Nor
 	for _, tableName := range distinctTableNames {
 		normalizeTableSchema := req.TableNameSchemaMapping[tableName]
 		if len(normalizeTableSchema.PrimaryKeyColumns) == 0 {
-			c.logger.Warn(fmt.Sprintf("skipping merge for table %s as it has no primary key", tableName))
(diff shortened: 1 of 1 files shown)


serprex (approved):

flow/connectors/bigquery/bigquery.go
```diff
@@ -511,12 +511,16 @@ func (c *BigQueryConnector) NormalizeRecords(ctx context.Context, req *model.Nor
...
 	}
-
 	// append all the statements to one list
 	c.logger.Info(fmt.Sprintf("merge raw records to corresponding tables: %s %s %v",
 		c.datasetID, rawTableName, distinctTableNames))
 
 	for _, tableName := range distinctTableNames {
+		normalizeTableSchema := req.TableNameSchemaMapping[tableName]
+		if len(normalizeTableSchema.PrimaryKeyColumns) == 0 {
+			c.logger.Warn(fmt.Sprintf("skipping merge for table %s as it has no primary key", tableName))
```
serprex, line 521:
```suggestion
			c.logger.Error(fmt.Sprintf("skipping merge for table %s as it has no primary key", tableName))
```

heavycrystal (changes requested):
If this is needed, should be done for all connectors with a normalize step

PR #1295, opened 2024-02-14
fetch peer configurations in workflow

iskakaushik:
This works towards a world where our entry point workflows can operate with just the peer name. Supporting edit-peer would also require something along these lines.


serprex (approved):

iskakaushik:
@Amogh-Bharadwaj  will tackle this as a part of his EditPeer changes.

PR #1298, opened 2024-02-15
explicitly call EnsurePullability while adding tables

heavycrystal:
The UI for dynamic table addition should prevent tables that we cannot mirror via CDC [no primary key + replident != full] to be selected. However, we end up using the signal for adding tables sometimes, and currently there is no check in this codepath.

Invoking `EnsurePullability` explicitly to guard against this.


flow/workflows/cdc_flow.go
```diff
@@ -152,6 +152,29 @@ func (w *CDCFlowWorkflowExecution) processCDCFlowConfigUpdates(ctx workflow.Cont
...
+					for _, additionalSourceTable := range flowConfigUpdate.AdditionalTables {
+						additionalSourceTables = append(additionalSourceTables, additionalSourceTable.SourceTableIdentifier)
+					}
+					return additionalSourceTables
+				}(),
+				CheckConstraints: true,
+			})
+		if err := ensurePullabilityFuture.Get(ctx, nil); err != nil {
+			w.logger.Error("failed to ensure pullability for additional tables: ", err)
+			return err
```
Amogh-Bharadwaj, line 175: Can this be removed?

flow/workflows/cdc_flow.go
```diff
@@ -152,6 +152,29 @@ func (w *CDCFlowWorkflowExecution) processCDCFlowConfigUpdates(ctx workflow.Cont
...
+			&protos.EnsurePullabilityBatchInput{
+				PeerConnectionConfig: cfg.Source,
+				FlowJobName:          cfg.FlowJobName,
+				SourceTableIdentifiers: func() []string {
+					additionalSourceTables := make([]string, 0, len(flowConfigUpdate.AdditionalTables))
+					for _, additionalSourceTable := range flowConfigUpdate.AdditionalTables {
+						additionalSourceTables = append(additionalSourceTables, additionalSourceTable.SourceTableIdentifier)
+					}
+					return additionalSourceTables
+				}(),
```
serprex, line 170: Compute this above rather than `func(){}()`

flow/workflows/cdc_flow.go
serprex: `processCDCFlowConfigUpdates` return value should not be aborting whole workflow. Any flow config update error should result in an error being reported to user in some way, & then being ignored. Then this can remain `return err`

PR #1299, opened 2024-02-15
Validate mirror: publication creation

Amogh-Bharadwaj:
Validate mirror now also checks: if no publication provided, we can create (and drop) a publication for the selected tables.

### DIFF
--- a/flow/connectors/postgres/validate.go
+++ b/flow/connectors/postgres/validate.go
@@ -127,7 +127,6 @@ func (c *PostgresConnector) CheckPublicationPermission(ctx context.Context, tabl
 	if err != nil {
 		return fmt.Errorf("error starting transaction: %w", err)
 	}
-	defer tx.Conn().Close(ctx)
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):

flow/connectors/postgres/validate.go
```diff
@@ -112,6 +112,53 @@ func (c *PostgresConnector) CheckReplicationPermissions(ctx context.Context, use
...
+	}
+	var pubViaRootString string
+	if supportsPubViaRoot {
+		pubViaRootString = "WITH(publish_via_partition_root=true)"
+	}
+	tx, err := c.conn.Begin(ctx)
+	if err != nil {
+		return fmt.Errorf("error starting transaction: %w", err)
+	}
+	defer tx.Conn().Close(ctx)
```
serprex, line 130: Don't do this, connection should be closed when connector is closed, as is connector breaks after calling this method

PR #1304, opened 2024-02-15
Improve api for UI

Amogh-Bharadwaj:
- Replication connection can't perform queries so removed that
- Exclude dropped columns from being displayed in table picker
- Refactor create pub check to check privileges since transactions don't work on publications and don't want to risk creating and not dropping one. this has been functionally tested

### DIFF
--- a/flow/connectors/postgres/validate.go
+++ b/flow/connectors/postgres/validate.go
@@ -154,6 +154,4 @@ func (c *PostgresConnector) CheckReplicationConnectivity(ctx context.Context) er
 	}
 
 	return conn.Close(ctx)
-
-	return nil
 }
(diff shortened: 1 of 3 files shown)


flow/connectors/postgres/validate.go
```diff
@@ -167,11 +155,5 @@ func (c *PostgresConnector) CheckReplicationConnectivity(ctx context.Context) er
 
 	defer conn.Close(ctx)
```
serprex, line 156:
```suggestion
	return conn.Close(ctx)
```

flow/connectors/postgres/validate.go
```diff
@@ -165,13 +153,7 @@ func (c *PostgresConnector) CheckReplicationConnectivity(ctx context.Context) er
...
-	defer conn.Close(ctx)
-
-	var one int
-	queryErr := conn.QueryRow(ctx, "SELECT 1").Scan(&one)
-	if queryErr != nil {
-		return fmt.Errorf("failed to query replication connection: %v", queryErr)
-	}
+	return conn.Close(ctx)
 
 	return nil
```
serprex, line 158:
```suggestion
```

serprex (approved):

serprex (approved):

PR #1312, opened 2024-02-16
some improvements to drop mirroring

Amogh-Bharadwaj:


flow/workflows/cdc_flow.go
```diff
@@ -506,8 +506,7 @@ func CDCFlowWorkflowWithConfig(
 			RetryPolicy: &temporal.RetryPolicy{
 				MaximumAttempts: 20,
 			},
-			SearchAttributes:    mirrorNameSearch,
-			WaitForCancellation: true,
```
serprex, line 510: Turning off WaitForCancellation will hide children workflows taking too long to cancel themselves

At least for source connector we'll be waiting for the disconnect anyways to drop slot

Amogh-Bharadwaj: What do you mean by hide?

serprex: I mean the workflow will say it's canceled, which seems good, but we shouldn't really consider it canceled until these waits are satisfied

PR #1313, opened 2024-02-16
Improve Close: use 1 minute timeout instead of potentially canceled context

serprex:
Also remove unnecessary nil checks,
& make self nil check in Connector Close methods consistent

This helps gracefully close connections when workflow canceled

### DIFF
--- a/flow/connectors/eventhub/eventhub.go
+++ b/flow/connectors/eventhub/eventhub.go
@@ -53,11 +53,15 @@ func NewEventHubConnector(
 	}, nil
 }
 
-func (c *EventHubConnector) Close(ctx context.Context) error {
-	err := c.hubManager.Close(ctx)
-	if err != nil {
-		c.logger.Error("failed to close event hub manager", slog.Any("error", err))
-		return err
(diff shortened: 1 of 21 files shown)


flow/connectors/eventhub/eventhub.go
```diff
@@ -53,11 +53,15 @@ func NewEventHubConnector(
...
 }
 
-func (c *EventHubConnector) Close(ctx context.Context) error {
-	err := c.hubManager.Close(ctx)
-	if err != nil {
-		c.logger.Error("failed to close event hub manager", slog.Any("error", err))
-		return err
+func (c *EventHubConnector) Close() error {
+	if c != nil {
+		timeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
iskakaushik, line 58: we have seen this take longer than 5 seconds from time to time, can we make this 30 seconds?

iskakaushik (approved):

flow/connectors/eventhub/eventhub.go
serprex: Updated both to 1 minute

PR #1315, opened 2024-02-16
Fixes to potentially aid drop mirror

Amogh-Bharadwaj:
Cancelling drop flow after a minute isn't a good idea. Slot remains active after sync flow dies (i.e it is fully cancelled, not waiting) for more than a minute (for some reason) causing drop flow to cancel and the slot surviving. Bumped it up to 5 minutes.

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -850,9 +850,9 @@ func (c *PostgresConnector) PullFlowCleanup(ctx context.Context, jobName string)
 		return fmt.Errorf("error dropping publication: %w", err)
 	}
 
-	err = c.conn.QueryRow(ctx, `SELECT pg_drop_replication_slot(slot_name) FROM pg_replication_slots
-	 WHERE slot_name=$1`, slotName).Scan(nil)
-	if err != nil && err != pgx.ErrNoRows {
+	_, err = c.conn.Exec(ctx, `SELECT pg_drop_replication_slot(slot_name) FROM pg_replication_slots
+	 WHERE slot_name=$1`, slotName)
+	if err != nil {
 		return fmt.Errorf("error dropping replication slot: %w", err)
 	}
 
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -850,9 +850,9 @@ func (c *PostgresConnector) PullFlowCleanup(ctx context.Context, jobName string)
 		return fmt.Errorf("error dropping publication: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


serprex:
How much of a perf difference are you seeing between Query & Exec? How many samples did you take?

iskakaushik:
> How much of a perf difference are you seeing between QueryRow & Exec? How many samples did you take?

@serprex also I think we weren't closing the `rows` object before?

flow/connectors/postgres/postgres.go
```diff
@@ -850,9 +850,9 @@ func (c *PostgresConnector) PullFlowCleanup(ctx context.Context, jobName string)
 		return fmt.Errorf("error dropping publication: %w", err)
 	}
 
-	_, err = c.conn.Exec(ctx, `SELECT pg_drop_replication_slot(slot_name) FROM pg_replication_slots
-	 WHERE slot_name=$1`, slotName)
-	if err != nil {
+	err = c.conn.QueryRow(ctx, `SELECT pg_drop_replication_slot(slot_name) FROM pg_replication_slots
+	 WHERE slot_name=$1`, slotName).Scan(nil)
```
serprex, line 854:
```suggestion
	err = c.conn.Exec(ctx, `SELECT pg_drop_replication_slot(slot_name) FROM pg_replication_slots
	 WHERE slot_name=$1`, slotName, pgx.QueryExecModeExec)
```
This might better address perf issue

serprex:
> > How much of a perf difference are you seeing between QueryRow & Exec? How many samples did you take?
> 
> @serprex also I think we weren't closing the `rows` object before? Maybe thats why `QueryRow` is helping?

`Exec` has no rows

Amogh-Bharadwaj:
@serprex by performance I mean't that exec wasn't dropping the slot as consistently as queryrow did. less reliable from what I could make out, essentially

heavycrystal:
strictly speaking that query does return a result set so it should be executed under `Query` instead of `Exec` anyway

serprex:
Exec is fine on a select if you aren't going to use the rows

serprex:
> @serprex by performance I mean't that exec wasn't dropping the slot as consistently as queryrow did. less reliable from what I could make out, essentially

This doesn't make sense. These sort of flaky issues need to be worked through by gaining understanding, not applying changes that seem to improve things. Maybe if you set up two simultaneous tests that created/dropped 1000 mirrors then it'd be clearer if there's something here

Amogh-Bharadwaj:
reverted the change in dropping

serprex (approved):

PR #1318, opened 2024-02-16
Clickhouse: use destination columns for qrep sync

Amogh-Bharadwaj:
Clickhouse insert-into-select for QRep used source table's schema, making column exclusion fail. This PR fixes that by using destination columns instead.
This PR has been functionally tested

### DIFF
--- a/flow/connectors/clickhouse/qrep_avro_sync.go
+++ b/flow/connectors/clickhouse/qrep_avro_sync.go
@@ -127,7 +127,8 @@ func (s *ClickhouseAvroSyncMethod) SyncQRepRecords(
 	selector := make([]string, 0, len(dstTableSchema))
 	for _, col := range dstTableSchema {
 		colName := col.Name()
-		if strings.Contains(colName, "_peerdb_") {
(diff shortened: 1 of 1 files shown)


flow/connectors/clickhouse/qrep_avro_sync.go
```diff
@@ -124,13 +124,20 @@ func (s *ClickhouseAvroSyncMethod) SyncQRepRecords(
 	avroFileUrl := fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", s3o.Bucket,
 		s.connector.creds.Region, avroFile.FilePath)
 
-	if err != nil {
-		return 0, err
+	selector := make([]string, 0, len(dstTableSchema))
+	for _, col := range dstTableSchema {
+		colName := col.Name()
+		if strings.Contains(colName, "_peerdb_") {
```
heavycrystal, line 130: what if we set custom column names for soft-delete and synced_at?

iskakaushik (approved):

PR #1320, opened 2024-02-16
drop flow: increase start to close timeout

Amogh-Bharadwaj:
The workflow cancels after start to close timeout. Somehow missed pushing this  in #1312


serprex (approved):

serprex:
There is value to the cmd sending cancel, since workflow has retry logic

serprex (approved):

iskakaushik (approved):

PR #1323, opened 2024-02-16
Clickhouse QRep Insert: use strings.EqualFold

Amogh-Bharadwaj:
The check for skipping peerdb columns in insert-into-select in clickhouse qrep was failing due to case mismatch. this PR fixes this


iskakaushik (approved):

Amogh-Bharadwaj:
functionally tested

PR #1326, opened 2024-02-19
Clickhouse: type mapping change and cleanup

Amogh-Bharadwaj:
- Fix syncing of Timestamp and TimestampTZ
- Map Timestamps and Dates to `Nullable(type)` because epoch in most cases not a sane default.
- Remove dead code


iskakaushik (approved):

flow/connectors/clickhouse/normalize.go
```diff
@@ -81,21 +81,23 @@ func generateCreateTableSQLForNormalizedTable(
...
+		case qvalue.QValueKindNumeric:
 			precision, scale := numeric.ParseNumericTypmod(column.TypeModifier)
 			if column.TypeModifier == -1 || precision > 76 || scale > precision {
 				precision = numeric.PeerDBClickhousePrecision
 				scale = numeric.PeerDBClickhouseScale
 			}
 			stmtBuilder.WriteString(fmt.Sprintf("`%s` DECIMAL(%d, %d), ",
 				colName, precision, scale))
-		} else {
+		case qvalue.QValueKindTimestamp, qvalue.QValueKindTimestampTZ, qvalue.QValueKindDate:
```
serprex, line 93: Can we track nullable in QValue?

Amogh-Bharadwaj: That, and the issue specifically with clickhouse is that [Nullable affects performance and storage](https://clickhouse.com/docs/en/cloud/bestpractices/avoid-nullable-columns) and so even once we wire up nullable in qvalue, clients may not necessarily want columns to be Nullable

flow/connectors/clickhouse/normalize.go
```diff
@@ -176,17 +178,24 @@ func (c *ClickhouseConnector) NormalizeRecords(ctx context.Context, req *model.N
 			if err != nil {
 				return nil, fmt.Errorf("error while converting column type to clickhouse type: %w", err)
 			}
-			if clickhouseType == "DateTime64(6)" || clickhouseType == "UUID" {
+			if clickhouseType == "UUID" {
```
serprex, line 181: Wondering if UUID can be extracted with "String"

serprex: Tested: it can

flow/connectors/clickhouse/normalize.go
```diff
@@ -176,17 +178,24 @@ func (c *ClickhouseConnector) NormalizeRecords(ctx context.Context, req *model.N
...
 				clickhouseType = "String"
 			}
```
serprex, line 183:
```suggestion
```

flow/connectors/clickhouse/normalize.go
serprex: yes, but that should be an option, ideally per-column

in this case it'd only be used for timestamps to be nullable when they're nullable

Amogh-Bharadwaj:
functionally tested

serprex (approved):

PR #1328, opened 2024-02-19
Move slot size recording into global heartbeat workflow

serprex:
Some misc improvements:
1. remove sleeping in activity, instead use temporal cronjobs
2. HandleSlotSize no longer needs to be threadsafe
3. slot size monitored even while sync flows crashing or paused


serprex:
> @serprex can we also move connection related alerts to be here?

Sure, that could be a followup PR

PR #1347, opened 2024-02-21
Use pg_export_snapshot when InitialSnapshotOnly

serprex:
Avoids consistency issues

### DIFF
--- a/flow/activities/snapshot_activity.go
+++ b/flow/activities/snapshot_activity.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"sync"
+	"time"
 
 	"go.temporal.io/sdk/activity"
 
@@ -15,19 +17,20 @@ import (
 )
 
 type SnapshotActivity struct {
-	SnapshotConnections map[string]SlotSnapshotSignal
-	Alerter             *alerting.Alerter
+	SnapshotConnectionsMutex sync.Mutex
+	SnapshotConnections      map[string]SlotSnapshotSignal
+	Alerter                  *alerting.Alerter
 }
 
 // closes the slot signal
 func (a *SnapshotActivity) CloseSlotKeepAlive(ctx context.Context, flowJobName string) error {
-	if a.SnapshotConnections == nil {
-		return nil
-	}
+	a.SnapshotConnectionsMutex.Lock()
+	defer a.SnapshotConnectionsMutex.Unlock()
 
 	if s, ok := a.SnapshotConnections[flowJobName]; ok {
 		close(s.signal.CloneComplete)
 		connectors.CloseConnector(ctx, s.connector)
+		delete(a.SnapshotConnections, flowJobName)
 	}
 
 	return nil
@@ -89,9 +92,8 @@ func (a *SnapshotActivity) SetupReplication(
 		return nil, fmt.Errorf("slot error: %w", slotInfo.Err)
 	}
 
-	if a.SnapshotConnections == nil {
-		a.SnapshotConnections = make(map[string]SlotSnapshotSignal)
-	}
+	a.SnapshotConnectionsMutex.Lock()
+	defer a.SnapshotConnectionsMutex.Unlock()
 
 	a.SnapshotConnections[config.FlowJobName] = SlotSnapshotSignal{
 		signal:       slotSignal,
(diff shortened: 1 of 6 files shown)


flow/connectors/postgres/postgres.go
```diff
@@ -797,6 +797,12 @@ func (c *PostgresConnector) EnsurePullability(
 	return &protos.EnsurePullabilityBatchOutput{TableIdentifierMapping: tableIdentifierMapping}, nil
 }
 
+func (c *PostgresConnector) ExportSnapshot(ctx context.Context) (string, error) {
+	var snapshotName string
+	err := c.conn.QueryRow(ctx, "select pg_export_snapshot()").Scan(&snapshotName)
```
serprex, line 802: need to create tx & commit it at end

flow/connectors/postgres/postgres.go
```diff
@@ -797,6 +797,29 @@ func (c *PostgresConnector) EnsurePullability(
...
 }
 
+func (c *PostgresConnector) ExportSnapshot(ctx context.Context) (string, any, error) {
+	var snapshotName string
+	tx, err := c.conn.Begin(ctx)
+	if err != nil {
+		return "", nil, err
+	}
+
+	err = tx.QueryRow(ctx, "select pg_export_snapshot()").Scan(&snapshotName)
```
heavycrystal, line 807: set `lock_timeout` and `idle_in_transaction_session_timeout` to 0 before returning, just like we do for slots

heavycrystal (changes requested):

flow/connectors/core.go
```diff
@@ -38,6 +38,13 @@ type CDCPullConnector interface {
 	EnsurePullability(ctx context.Context, req *protos.EnsurePullabilityBatchInput) (
 		*protos.EnsurePullabilityBatchOutput, error)
 
+	// For InitialSnapshotOnly correctness without replication slot
+	// `any` is for returning transaction if necessary
+	ExportSnapshot(context.Context) (string, any, error)
```
heavycrystal, line 43: can we leave these methods out of the interface and use them directly from `PostgresConnector`?

heavycrystal: this and a lot of other methods are very PG specific, no point in making them an interface

serprex: Activities should be connector agnostic. I can move this to a separate connector interface tho

heavycrystal (approved):

flow/activities/snapshot_activity.go
```diff
@@ -104,3 +106,54 @@ func (a *SnapshotActivity) SetupReplication(
...
+		return err
+	}
+
+	sss := SlotSnapshotSignal{snapshotName: snapshotName}
+	a.SnapshotConnectionsMutex.Lock()
+	a.SnapshotConnections[sessionID] = sss
+	a.SnapshotConnectionsMutex.Unlock()
+
+	for {
+		activity.RecordHeartbeat(ctx, "maintaining export snapshot transaction")
```
iskakaushik, line 128: can we log something here as well, and make the message for both heartbeat and log something like.

`maintaining export snapshot transaction since X mins`.

flow/workflows/snapshot_flow.go
```diff
@@ -288,7 +277,55 @@ func SnapshotFlowWorkflow(ctx workflow.Context, config *protos.FlowConnectionCon
...
+		sessionSelector.Select(ctx)
+		if sessionError != nil {
+			return sessionError
+		}
+
+		slotInfo := &protos.SetupReplicationOutput{
+			SlotName:     "peerdb_initial_copy_only",
+			SnapshotName: snapshotName,
+		}
+		if err := se.cloneTables(ctx, slotInfo, int(config.SnapshotNumTablesInParallel)); err != nil {
```
iskakaushik, line 325: can we log something here too please?

flow/activities/snapshot_activity.go
```diff
@@ -104,3 +106,54 @@ func (a *SnapshotActivity) SetupReplication(
...
+
+func (a *SnapshotActivity) MaintainTx(ctx context.Context, sessionID string, peer *protos.Peer) error {
+	conn, err := connectors.GetCDCPullConnector(ctx, peer)
+	if err != nil {
+		return err
+	}
+	defer connectors.CloseConnector(ctx, conn)
+
+	snapshotName, tx, err := conn.ExportSnapshot(ctx)
+	if err != nil {
```
iskakaushik, line 118: can we log the error here?

iskakaushik (approved):
minor logging comments, but overall looks good to me.

PR #1359, opened 2024-02-22
BigQuery: More reliable insert-into-select 

Amogh-Bharadwaj:
Not sure how this issue wasn't caught until now, but not explicitly specifying the column order in the QRep insert into select statement can cause it to fail due to column-value insert mismatch (ex: expected int but got bool for _peerdb_is_deleted)

### DIFF
--- a/flow/connectors/bigquery/qrep_avro_sync.go
+++ b/flow/connectors/bigquery/qrep_avro_sync.go
@@ -124,9 +124,15 @@ func (s *QRepAvroSyncMethod) SyncRecords(
 func getTransformedColumns(dstSchema *bigquery.Schema, syncedAtCol string, softDeleteCol string) []string {
 	transformedColumns := make([]string, 0, len(*dstSchema))
 	for _, col := range *dstSchema {
-		if col.Name == syncedAtCol || col.Name == softDeleteCol {
+		if col.Name == syncedAtCol { // PeerDB column
+			transformedColumns = append(transformedColumns, "CURRENT_TIMESTAMP AS `"+col.Name+"`")
 			continue
 		}
+		if col.Name == softDeleteCol { // PeerDB column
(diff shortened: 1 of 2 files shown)


flow/connectors/bigquery/qrep_avro_sync.go
```diff
@@ -184,18 +190,18 @@ func (s *QRepAvroSyncMethod) SyncQRepRecords(
 	)
 	bqClient := s.connector.client
 
+	insertColumns := make([]string, 0, len(dstTableMetadata.Schema))
+	for _, col := range dstTableMetadata.Schema {
+		insertColumns = append(insertColumns, fmt.Sprintf("`%s`", col.Name))
+	}
```
serprex, line 196: Need to include synced_at/deleted columns

iskakaushik:
is there a way to add a test for this?

flow/connectors/bigquery/qrep_avro_sync.go
Amogh-Bharadwaj: for example that's how we're setting their values in getTransformedColumns

iskakaushik:
@Amogh-Bharadwaj  isn't able to reproduce this issue, @serprex do you think we can grant a test exemption here given that this PR qualifies the statements and makes them more robust?

serprex:
> avro_transform_test.go:37: Transform SQL is not correct. Got: [ST_GEOGFROMTEXT(`col1`) AS `col1` PARSE_JSON(`col2`,wide_number_mode=>'round') AS `col2` `camelCol4` CURRENT_TIMESTAMP AS `sync_col`]

If tests are passing then yes, I'd consider this code tested since it's part of our existing test coverage

iskakaushik:
needed for customer.

PR #1364, opened 2024-02-23
UI: Fix slot graph

Amogh-Bharadwaj:
- Fix lag point computation
- Use `pg_current_wal_lsn() - restart_lsn` instead of confirmed_flush_lsn
- Fix placeholder not showing in slot dropdown

Screenshot 2024-02-23 at 1 11 05 PM

### DIFF
--- a/ui/app/api/peers/slots/[name]/route.ts
+++ b/ui/app/api/peers/slots/[name]/route.ts
@@ -6,6 +6,30 @@ export async function GET(
   request: NextRequest,
   context: { params: { name: string } }
 ) {
+  const timeSince = request.nextUrl.searchParams.get('timeSince');
+  console.log('timeSince', timeSince);
+  let forThePastThisMuchTime: number;
+  switch (timeSince) {
+    case 'day':
+      forThePastThisMuchTime = 86400000;
+      break;
+    case 'month':
+      forThePastThisMuchTime = 2592000000;
+      break;
+    case '15min':
+      forThePastThisMuchTime = 900000;
+      break;
+    case '5min':
+      forThePastThisMuchTime = 300000;
+      break;
+    case '1min':
+      forThePastThisMuchTime = 60000;
+      break;
+    default:
+      forThePastThisMuchTime = 3600000;
+      break;
+  }
+  console.log('forThePastThisMuchTime', forThePastThisMuchTime);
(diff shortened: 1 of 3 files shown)


ui/app/api/peers/slots/[name]/route.ts
```diff
@@ -30,5 +57,7 @@ export async function GET(
     };
   });
 
+  console.log('stringedLagPoints', stringedLagPoints);
+
```
Comment (serprex) [inline ui/app/api/peers/slots/[name]/route.ts:61]:
```suggestion
```

ui/app/peers/[peerName]/lagGraph.tsx
```diff
@@ -81,20 +78,26 @@ function LagGraph({ slotNames }: { slotNames: string[] }) {
...
           placeholder='Select a replication slot'
-          options={slotNames.map((slotName) => ({
-            label: slotName,
-            value: slotName,
-          }))}
+          options={
+            slotNames.map((slotName) => ({
+              label: slotName,
+              value: slotName,
+            })) ?? undefined
```
Comment (serprex) [inline ui/app/peers/[peerName]/lagGraph.tsx:85]:
```suggestion
            }))
```
`Array.prototype.map` will always return an array

ui/app/api/peers/slots/[name]/route.ts
```diff
@@ -6,6 +6,30 @@ export async function GET(
...
+      forThePastThisMuchTime = 300000;
+      break;
+    case '1min':
+      forThePastThisMuchTime = 60000;
+      break;
+    default:
+      forThePastThisMuchTime = 3600000;
+      break;
+  }
+  console.log('forThePastThisMuchTime', forThePastThisMuchTime);
```
Comment (iskakaushik) [inline ui/app/api/peers/slots/[name]/route.ts:32]: remove this

ui/app/peers/[peerName]/lagGraph.tsx
Comment (Amogh-Bharadwaj) [reply ui/app/peers/[peerName]/lagGraph.tsx:85]: What I want is if the array is empty then return undefined and not empty array. This is because react-select placeholder doesn't appear otherwise

iskakaushik (approved):

ui/app/peers/[peerName]/lagGraph.tsx
Comment (serprex) [reply ui/app/peers/[peerName]/lagGraph.tsx:85]: You'll have to do `slotNames.length === 0 ? undefined : slotNames.map(...)`

PR #1375, opened 2024-02-25
replace ping with `SELECT 1`

iskakaushik:
Sometimes with replication connections `-- ping` isn't playing well, replace with `SELECT 1`

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -239,7 +239,7 @@ func (c *PostgresConnector) ConnectionActive(ctx context.Context) error {
 	if c.conn == nil {
 		return errors.New("connection is nil")
 	}
-	pingErr := c.conn.Ping(ctx)
+	_, pingErr := c.conn.Exec(ctx, "SELECT 1;")
 	return pingErr
 }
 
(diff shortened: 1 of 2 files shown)


flow/connectors/postgres/postgres.go
```diff
@@ -239,7 +239,7 @@ func (c *PostgresConnector) ConnectionActive(ctx context.Context) error {
 	if c.conn == nil {
 		return errors.New("connection is nil")
 	}
-	pingErr := c.conn.Ping(ctx)
+	_, pingErr := c.conn.Exec(ctx, "SELECT 1;")
```
serprex, line 242:
```suggestion
	_, pingErr := c.conn.Exec(ctx, "SELECT 1")
```

flow/connectors/postgres/ssh_wrapped_pool.go
```diff
@@ -112,7 +112,7 @@ func (tunnel *SSHTunnel) NewPostgresConnFromConfig(
 
 	host := connConfig.Host
 	err = retryWithBackoff(logger, func() error {
-		err = conn.Ping(ctx)
+		_, err := conn.Exec(ctx, "SELECT 1;")
```
serprex, line 115:
```suggestion
		_, err := conn.Exec(ctx, "SELECT 1")
```

serprex (approved):

PR #1379, opened 2024-02-26
Sync flow: don't connect to destination when no incoming rows

Amogh-Bharadwaj:
When we are waiting for the first row, we are still holding on to a connection to the destination data store which we created to fetch the last offset. This may not be ideal in case the destination charges for connection time. 
This PR aims to not have a destination connection maintained if not needed

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -355,7 +355,7 @@ func (a *FlowableActivity) SyncFlow(
 
 		dstConnForSchemaChange, connErr := connectors.GetCDCSyncConnector(ctx, config.Destination)
 		if connErr != nil {
-			return nil, fmt.Errorf("failed to get destination connector for last offset fetch: %w", connErr)
+			return nil, fmt.Errorf("failed to get destination connector for schema changes: %w", connErr)
 		}
 		defer connectors.CloseConnector(ctx, dstConnForSchemaChange)
 		err := dstConnForSchemaChange.ReplayTableSchemaDeltas(ctx, flowName, recordBatch.SchemaDeltas)


flow/activities/flowable.go
```diff
@@ -350,22 +353,34 @@ func (a *FlowableActivity) SyncFlow(
...
+		if connErr != nil {
+			return nil, fmt.Errorf("failed to get destination connector for schema changes: %w", connErr)
+		}
+		defer connectors.CloseConnector(ctx, dstConnForSchemaChange)
+		err := dstConnForSchemaChange.ReplayTableSchemaDeltas(ctx, flowName, recordBatch.SchemaDeltas)
 		if err != nil {
 			return nil, fmt.Errorf("failed to sync schema: %w", err)
 		}
 
+		connectors.CloseConnector(ctx, dstConnForSchemaChange)
```
serprex, line 366:
```suggestion
```
We're returning here, so defer will get it immediately

flow/activities/flowable.go
```diff
@@ -350,22 +353,34 @@ func (a *FlowableActivity) SyncFlow(
...
 			RelationMessageMapping: options.RelationMessageMapping,
 		}, nil
 	}
 
+	dstConnForSync, err := connectors.GetCDCSyncConnector(ctx, config.Destination)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get destination connector for sync records: %w", err)
+	}
+	defer connectors.CloseConnector(ctx, dstConnForSync)
+
```
serprex, line 379:
```suggestion
```
Can move this all into the `errGroup.Go` function

flow/activities/flowable.go
```diff
@@ -350,22 +353,34 @@ func (a *FlowableActivity) SyncFlow(
 		}
 		logger.Info("no records to push")
 
-		err := dstConn.ReplayTableSchemaDeltas(ctx, flowName, recordBatch.SchemaDeltas)
+		dstConnForSchemaChange, connErr := connectors.GetCDCSyncConnector(ctx, config.Destination)
+		if connErr != nil {
+			return nil, fmt.Errorf("failed to get destination connector for schema changes: %w", connErr)
+		}
+		defer connectors.CloseConnector(ctx, dstConnForSchemaChange)
+		err := dstConnForSchemaChange.ReplayTableSchemaDeltas(ctx, flowName, recordBatch.SchemaDeltas)
```
serprex, line 361: nice thing here is with future PR we could move ReplayTableSchemaDeltas into its own interface, get _that_ connector instead here, & then skip when ErrUnsupportedFunctionality

serprex:
This does cause one potential issue: we've recently figured out that offset screwups caused by updating LSN on destination while processing WAL was tied to workflow restart replaying GetLastOffset activity with old offset result. So in theory we could bring back updating offset while no records consumed, but this change prevents that

PR #1388, opened 2024-02-27
Fix & test qrep pause/unpause

serprex:
As is apt to happen when adding tests, also fix pause/unpause for qrep/xmin

1. `waitForNewRows` would prevent pausing; use a selector to wait on signals while waiting on new rows while waiting on context cancelation
2. `CurrentFlowStatus` wasn't being set to RUNNING when exiting from PAUSED

As an aside, it turns out all the qrep/xmin tests are doing `InitialCopyOnly` besides this new test

Also remove `DisableWaitForNewRows` because it's unused & is meant to diverge testing from reality


serprex:
Failures here are showing that temporal testsuite is not suitable for integration testing because ContinueAsNew isn't handled

https://github.com/temporalio/sdk-go/issues/805

iskakaushik:
@serprex  is there a way to prevent `INFO` logs from showing up in tests?

```
***"time":"2024-03-04T15:06:03.6158228Z","level":"INFO","msg":"Sent Standby status message. processed 0 rows","Namespace":"default","TaskQueue":"peer-flow-task-queue","WorkerID":"4459@vm8n9339@","ActivityID":"11","ActivityType":"MaintainPull","Attempt":1,"WorkflowType":"SyncFlowWorkflow","WorkflowID":"sync-flow-test_simple_schema_changes_bq_idwud1dh_20240304150318-f97680e0-504d-438e-8b02-372c948200e3","RunID":"9b958103-daaa-48ad-aa90-0f4093f79338","deploymentUid":""***
```

serprex:
Yes, we'd have to pass a log level when invoking `peer-flow`

iskakaushik (approved):

PR #1368, opened 2024-02-23
supporting Snowflake as a source for full table QRep

heavycrystal:
sample command for reference:
```
CREATE MIRROR watermark3 FROM sf_target TO catalog FOR
$$
  SELECT * FROM public.retwal1
$$
WITH (
        destination_table_name = 'public.retwal2',
        watermark_column = '', -- has to be empty
        watermark_table_name = 'public.retwal1',
        mode = 'overwrite',
        parallelism = 2, -- no effect
        refresh_interval = 10, -- no effect
        num_rows_per_partition = 500, -- no effect
        setup_watermark_table_on_destination = true,
        initial_copy_only=true
);
```
UI:
Screenshot 2024-02-26 at 6 24 33 PM

2. some refactors for `QRepPullConnector`, SQL Server is now not a valid `QRepPullConnector`

PR #1390, opened 2024-02-27
Flowable: Use errgroup in replicateQRep

Amogh-Bharadwaj:
Replaces a usage waitgroup with errgroup in replicateQRepPartitions in flowable.go
This is a preceding PR to #1368

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -637,31 +637,25 @@ func (a *FlowableActivity) replicateQRepPartition(ctx context.Context,
 
 	var stream *model.QRecordStream
 	bufferSize := shared.FetchAndChannelSize
-	var wg sync.WaitGroup
-
-	var goroutineErr error = nil
+	errGroup, errCtx := errgroup.WithContext(ctx)
 	if config.SourcePeer.Type == protos.DBType_POSTGRES {
 		stream = model.NewQRecordStream(bufferSize)
-		wg.Add(1)
-
-		go func() {
+		errGroup.Go(func() error {
 			pgConn := srcConn.(*connpostgres.PostgresConnector)
-			tmp, err := pgConn.PullQRepRecordStream(ctx, config, partition, stream)
+			tmp, err := pgConn.PullQRepRecordStream(errCtx, config, partition, stream)
 			numRecords := int64(tmp)
 			if err != nil {
 				a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
-				logger.Error("failed to pull records", slog.Any("error", err))
-				goroutineErr = err
+				return fmt.Errorf("failed to pull records: %v", err)
 			} else {
-				err = monitoring.UpdatePullEndTimeAndRowsForPartition(ctx,
+				err = monitoring.UpdatePullEndTimeAndRowsForPartition(errCtx,
 					a.CatalogPool, runUUID, partition, numRecords)
 				if err != nil {
 					logger.Error(err.Error())
-					goroutineErr = err
 				}
 			}
-			wg.Done()
-		}()
+			return nil
+		})
 	} else {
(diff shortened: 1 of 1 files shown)


serprex (approved):

flow/activities/flowable.go
```diff
@@ -692,10 +686,10 @@ func (a *FlowableActivity) replicateQRepPartition(ctx context.Context,
 		logger.Info("no records to push for partition " + partition.PartitionId)
```
serprex, line 686: Ideally `dstConn.SyncQRepRecords(ctx, config, partition, stream)` would also run inside errgroup for pg situation, causing either sync or pull failing to cancel the other

Amogh-Bharadwaj: I'm not sure if sync should fail if pull fails, if say there are rows currently being pushed, then we should finish pushing them and then get stuck on pull

flow/activities/flowable.go
```diff
@@ -637,31 +637,25 @@ func (a *FlowableActivity) replicateQRepPartition(ctx context.Context,
...
+		errGroup.Go(func() error {
 			pgConn := srcConn.(*connpostgres.PostgresConnector)
-			tmp, err := pgConn.PullQRepRecordStream(ctx, config, partition, stream)
+			tmp, err := pgConn.PullQRepRecordStream(errCtx, config, partition, stream)
 			numRecords := int64(tmp)
 			if err != nil {
 				a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
-				logger.Error("failed to pull records", slog.Any("error", err))
-				goroutineErr = err
+				return fmt.Errorf("failed to pull records: %v", err)
```
serprex, line 649:
```suggestion
				return fmt.Errorf("failed to pull records: %w", err)
```

flow/activities/flowable.go
serprex: In that case you'd put sync in the errGroup & pass it the non-errctx, so that if sync fails pull stops, removing the need for `pullCancel`

Amogh-Bharadwaj: Looking at it now, I'm not sure what the point of pullCtx is, we only use it to get the source connector. I think it was relevant back when ctx was part of the connector struct and was being used by all the methods

Amogh-Bharadwaj: What is pullCancel cancelling exactly

serprex: Yes, seems like a bug tho, because the idea is that if sync finishes early then `pullCancel()` will cancel the errgroup

flow/activities/flowable.go
```diff
@@ -638,6 +636,7 @@ func (a *FlowableActivity) replicateQRepPartition(ctx context.Context,
 	var stream *model.QRecordStream
 	bufferSize := shared.FetchAndChannelSize
 	errGroup, errCtx := errgroup.WithContext(ctx)
```
serprex, line 638: This can now move into `config.SourcePeer.Type == protos.DBType_POSTGRES` branch

flow/activities/flowable.go
```diff
@@ -635,33 +633,43 @@ func (a *FlowableActivity) replicateQRepPartition(ctx context.Context,
...
 
+	var rowsSynced int
 	var stream *model.QRecordStream
 	bufferSize := shared.FetchAndChannelSize
-	var wg sync.WaitGroup
-
-	var goroutineErr error = nil
 	if config.SourcePeer.Type == protos.DBType_POSTGRES {
+		errGroup, errCtx := errgroup.WithContext(ctx)
 		stream = model.NewQRecordStream(bufferSize)
```
serprex, line 641:
```suggestion
		stream := model.NewQRecordStream(bufferSize)
```

PR #1391, opened 2024-02-27
flow ci: try running temporal, getting out of sdk/testsuite

serprex:


serprex:
Some issues once this is passing:
1. should find some way to cache layers of docker build
2. want logs from containers, as workflow errors / logging aren't happening within context of testing process anymore

docker-compose-ci.yml
```diff
@@ -0,0 +1,122 @@
...
+    container_name: flow-worker
+    build:
+      context: .
+      dockerfile: stacks/flow.Dockerfile
+      target: flow-worker
+    environment:
+      <<: [*catalog-config, *flow-worker-env]
+    depends_on:
+      temporal-admin-tools:
+        condition: service_healthy
```
iskakaushik, line 122: do we need these here?

serprex: Once things are working I'll look to move having these run goroutines with TestMain outside of docker

flow-api as a whole isn't being used

iskakaushik (approved):

PR #1401, opened 2024-02-29
check for partition being synced before starting errgroup

heavycrystal:


serprex:
Why is this necessary since Sync still checks this?

flow/connectors/s3/qrep.go
```diff
@@ -81,3 +81,10 @@ func (c *S3Connector) SetupQRepMetadataTables(_ context.Context, config *protos.
 	c.logger.Info("QRep metadata setup not needed for S3.")
 	return nil
 }
+
+// S3 doesn't appear to check if the partition already exists
```
serprex, line 85: Comment should be updated with reality. It doesn't check. Instead, S3 will always overwrite an existing partition

serprex (approved):

PR #1419, opened 2024-03-01
Remove integer TaskQueueID

serprex:
Removes need for error check
Also previously both task queues had same iota value, so that was wrong


iskakaushik (approved):

serprex:
Reading over, seems golang's iota works out to make these two constants distinct, so there wasn't a bug here, but still good cleanup

PR #1427, opened 2024-03-04
Snowflake normalize: check gCtx.Err() while iterating through tables

serprex:
errgroup.Go unconditionally executes the function / waits, causing pauses to be ignored for awhile


iskakaushik:
@serprex  should we audit other usages of errgroup as well?

heavycrystal (approved):

serprex:
> @serprex should we audit other usages of errgroup as well?

This is the only place in the codebase we use `SetLimit`, other places we only invoke a couple goroutines right away

PR #1428, opened 2024-03-04
Record slot size: select distinct

serprex:
Turns out we create a flows record per source/destination table pair


iskakaushik (approved):

serprex:
@heavycrystal tested query

PR #1434, opened 2024-03-05
CDC Flow: fix state wipe

Amogh-Bharadwaj:
Pause mirror second time onwards was not being respected as the state was not being wiped if only idle timeout / batch size is edited, because when additionTables is empty we just return nil before setting state to nil. This made flowConfigUpdate perenially not nil, which caused the mirror to never enter the pause loop.

This PR moves the setting of state to nil outside of processConfigUpdates and also adds logs


heavycrystal (approved):

iskakaushik:
@heavycrystal -- can you add a test that will prevent this from regressing?

cc: @serprex

PR #1436, opened 2024-03-05
Fix & test cdc rollover

serprex:
Customer workflow recently got stuck after sync-stop,
add test covering cdc_flow returning ContinueAsNew multiple times

cdc_flow had custom retry logic to pass tests before they were switched to integration tests,
get rid of that, it was racy with signals, rely on temporal retries for workflow errors
& restarting cdc entirely if sync/normalize finish for whatever reason


serprex:
CDC code has some custom retry logic to pass tests before they were switched to integration tests. We should get rid of that, relying on temporal retries for the workflows & restarting cdc entirely if sync/normalize finish for whatever reason

As is, it's apt to drop signals

PR #1438, opened 2024-03-05
comparing RelationMessage against TableSchema, stop caching RelationMessages

heavycrystal:
Previously we diffed the previous and current `RelationMessages` received from Postgres during logical replication to ascertain if the schema changed or not. This fails in cases where the schema changed after mirror setup but before we get the first `RelationMessage` which can happen a) during initial load, before CDC and b) during CDC, if no rows are synced before schema changes are made. 

This causes a desync between PeerDB's view of the table and the actual status of the table, causing issues. Changed to diff from the current `TableSchema` instead, which should always be present. Eliminates having to store and pass `RelationMessageMapping` around, it is scoped to reading from the replication connection.

Also fixes bugs with schema changes being handled incorrectly for PG and BQ, and schema audits being logged without schema deltas actually being present.

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -758,7 +758,7 @@ func (p *PostgresCDCSource) processRelationMessage(
 	}
 
 	p.relationMessageMapping[currRel.RelationID] = currRel
-	// only log audit if there is actually is an actionable delta
+	// only log audit if there is actionable delta
 	if len(schemaDelta.AddedColumns) > 0 {
 		rec := &model.RelationRecord{
 			TableSchemaDelta: schemaDelta,
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -28,11 +28,13 @@ import (
 
 type PostgresCDCSource struct {
 	*PostgresConnector
-	SrcTableIDNameMapping  map[uint32]string
-	TableNameMapping       map[string]model.NameAndExclude
+	srcTableIDNameMapping  map[uint32]string
+	tableNameMapping       map[string]model.NameAndExclude
+	tableNameSchemaMapping map[string]*protos.TableSchema
+	// relationMessageMapping is not an input, but populated by RelationMessages received here.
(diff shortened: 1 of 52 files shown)


flow/connectors/postgres/cdc.go
```diff
@@ -721,66 +702,70 @@ func (p *PostgresCDCSource) auditSchemaDelta(ctx context.Context, flowJobName st
...
 				schemaDelta.SrcTableName))
 		}
 	}
 
-	p.relationMessageMapping[currRel.RelationId] = currRel
-	rec := &model.RelationRecord{
-		TableSchemaDelta: schemaDelta,
-		CheckpointID:     int64(lsn),
+	p.relationMessageMapping[currRel.RelationID] = currRel
+	// only log audit if there is actually is an actionable delta
```
serprex, line 760:
```suggestion
	// only log audit if there is actually actionable delta
```

heavycrystal:
Since we now maintain the replication connection across syncs, we don't get a second `RelationMessage` for the same table unless its schema changes so some amount of caching is needed. Storing at the pgConn level instead of global state.

Also added defensive check to only update `TableNameSchemaMapping` of `NormalizeFlow` if not nil, fixes a test.

serprex (approved):

protos/flow.proto
```diff
@@ -99,7 +87,6 @@ message CreateTablesFromExistingOutput {
 
 message SyncFlowOptions {
   uint32 batch_size = 1;
-  map<uint32, RelationMessage> relation_message_mapping = 2;
```
serprex, line 102: if we only deprecate this field then we won't break `pause`/`unpause` upgrade

serprex:
lgtm, good to see a red PR, would want @iskakaushik to take a look over it too

protos/flow.proto
heavycrystal: will need to retain the other Message types for this, also not sure if it's enough since CDC state also loses fields

iskakaushik: please deprecate and not delete

PR #1445, opened 2024-03-07
fix(alerting): add recover for possible panics

iamKunalGupta:
This is for preventing non-critical alerting/telemetry code-flow from taking down the workflow during a panic

### DIFF
--- a/flow/alerting/alerting.go
+++ b/flow/alerting/alerting.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"runtime/debug"
 	"time"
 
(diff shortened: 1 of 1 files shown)


flow/alerting/alerting.go
```diff
@@ -74,7 +75,26 @@ func NewAlerter(ctx context.Context, catalogPool *pgxpool.Pool) *Alerter {
 	}
 }
 
+func withPanicGuard(ctx context.Context, functionName string, f func()) {
```
serprex, line 78: Can compute functionName on error, see second example https://stackoverflow.com/a/38551362

serprex:
Personally I lean towards letting logging panics crash things, all bugs are bugs

PR #1447, opened 2024-03-07
Validate peer: fix snowflake validate

Amogh-Bharadwaj:
Snowflake validate could fail if the user creates an internal schema before hand, in which case permissions for schema creation are not necessary and may not be given. Our validate for snowflake will fail as it tries to create a dummy schema

The change this PR makes is to check if the raw schema exists (see note). If it doesn't then, create it. The remaining steps - table create, insert row etc. remain the same. `create schema if not exists` does not bypass snowflake's permission checks for schema creation

Removes drop schema step as that isn't something we do in mirrors

note - this is also done at raw table creation and qrep metadata table creation where we create schemas

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -12,6 +12,7 @@ import (
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"go.temporal.io/sdk/activity"
+	"go.temporal.io/sdk/log"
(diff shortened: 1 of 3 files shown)


Amogh-Bharadwaj:
Requires functional testing

Amogh-Bharadwaj:
Functionally tested

iskakaushik (approved):

serprex (approved):

PR #1461, opened 2024-03-10
kafka + lua scripting + flatbuffers

serprex:
At the point now that I'm looking to start testing, trying to get ui/nexus working so I can create kafka peer as an entrypoint

Maybe flatbuffers should be split out as a separate PR, but not too important since flatbuffers port is split out into its own files

In theory kafka could be merged without lua, instead having some default json encoding, but that seems a bit pointless


serprex:
Lua docs: https://github.com/PeerDB-io/docs/pull/96

Will also need kafka docs

flow/connectors/kafka/kafka.go
```diff
@@ -0,0 +1,330 @@
...
+func (c *KafkaConnector) ReplayTableSchemaDeltas(_ context.Context, flowJobName string, schemaDeltas []*protos.TableSchemaDelta) error {
+	return nil
+}
+
+func (c *KafkaConnector) SyncFlowCleanup(ctx context.Context, jobName string) error {
+	return c.pgMetadata.DropMetadata(ctx, jobName)
+}
+
+func (c *KafkaConnector) SyncRecords(ctx context.Context, req *model.SyncRecordsRequest) (*model.SyncResponse, error) {
+	// TODO BeginTransaction if transactional
```
iskakaushik, line 162: file an issue and link it here

flow/connectors/kafka/kafka.go
```diff
@@ -0,0 +1,330 @@
...
+
+func (c *KafkaConnector) ReplayTableSchemaDeltas(_ context.Context, flowJobName string, schemaDeltas []*protos.TableSchemaDelta) error {
+	return nil
+}
+
+func (c *KafkaConnector) SyncFlowCleanup(ctx context.Context, jobName string) error {
+	return c.pgMetadata.DropMetadata(ctx, jobName)
+}
+
+func (c *KafkaConnector) SyncRecords(ctx context.Context, req *model.SyncRecordsRequest) (*model.SyncResponse, error) {
```
iskakaushik, line 161: can we refactor this to smaller methods?

serprex: Split out methods for `loadScript`/`lvalueToKafkaRecord`

flow/connectors/kafka/kafka.go
```diff
@@ -0,0 +1,338 @@
...
+func (c *KafkaConnector) ReplayTableSchemaDeltas(_ context.Context, flowJobName string, schemaDeltas []*protos.TableSchemaDelta) error {
+	return nil
+}
+
+func (c *KafkaConnector) SyncFlowCleanup(ctx context.Context, jobName string) error {
+	return c.pgMetadata.DropMetadata(ctx, jobName)
+}
+
+func loadScript(ctx context.Context, script string, print lua.LGFunction) (*lua.LState, error) {
+	if script != "" {
```
iskakaushik, line 162: can we have an early return (invert it)

iskakaushik (approved):

PR #1464, opened 2024-03-11
SF merge: don't return gCtx.Err, defer to g.Wait

serprex:
Returning `gCtx.Err()` caused errors to be logged only as "context canceled"

Still need to check `ctx.Err()` to avoid returning nil with partial merge


iskakaushik:
can we log all these various errors as well @serprex ?

serprex:
Caller logs them

iskakaushik (approved):

PR #1478, opened 2024-03-12
For replica identity full don't create pkey on snowflake

iskakaushik:
We create primary key indexes on the destination table in Snowflake CDC. For Replica Identity Full on a table, we assume all columns to be primary keys and replay that to the destination. 
But if one of those columns on the source have null values, our normalize records step fails saying `NULL result in a non-nullable column` 
This PR fixes this by not creating Primary Keys on the target for replica identity full


serprex (approved):

heavycrystal (approved):

heavycrystal:
comment on why we're doing this would be nice

Amogh-Bharadwaj (approved):
Functionally tested with replica identity full table. No constraints found on the destination

Amogh-Bharadwaj:
> comment on why we're doing this would be nice

Added

PR #1488, opened 2024-03-15
Add environment variable to disable one-sync

serprex:


Amogh-Bharadwaj (approved):

flow/peerdbenv/config.go
```diff
@@ -51,6 +51,11 @@ func PeerDBCDCDiskSpillMemPercentThreshold() int {
 	return getEnvInt("PEERDB_CDC_DISK_SPILL_MEM_PERCENT_THRESHOLD", -1)
 }
 
+// PEERDB_DISABLE_ONE_SYNC
+func PeerDBDisableOneSync() bool {
+	return getEnvBool("PEERDB_DISABLE_ONE_SYNC", false)
```
heavycrystal, line 56: nit: prefer it as enable and set true by default

Amogh-Bharadwaj: disabling it is going to be the rare setting so it makes sense to have that be the name

Amogh-Bharadwaj: kind of like disableTLS

PR #1491, opened 2024-03-16
Replace *big.Rat with shopspring decimal.Decimal

### LINKED ISSUE #1175: Numeric data type: handle it in Go and Avro as string instead of big.Rat
`big.Rat` is causing precision issues for numeric data. Trying to get it to behave is leading to surprises. 
Let's instead handle numerics as strings in code and avro schema
Use BigQuery and Snowflake's text to numeric functions to get the desired result on the target

serprex:
Rational numbers are awkward decimals,
using a decimal type for decimal data has a few benefits:
1. perf-wise decimal representation is a *big.Int & an int32 exponent (same as pgtype.Numeric), instead of 2 *big.Ints
2. decimal has obvious conversion to/from decimal strings,
    whereas *big.Rat can be wonky going through FloatString
4. with scripting most people will want to use their decimal values as decimals, not decimal rationals

A side effect of this is that we now convert NaN/inf/-inf numerics to null. Previously they'd all get converted to NaN. This would only be usable for pg<->pg, only causing trouble with other peers, & even then we'd want inf/-inf to be properly translated

Fixes #1175

### DIFF
--- a/flow/connectors/postgres/qrep_query_executor_test.go
+++ b/flow/connectors/postgres/qrep_query_executor_test.go
@@ -4,12 +4,12 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"math/big"
 	"testing"
 	"time"
 
 	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5"
+	"github.com/shopspring/decimal"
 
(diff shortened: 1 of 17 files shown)


Amogh-Bharadwaj (approved):
LGTM once tests pass

PR #1492, opened 2024-03-16
Simplify pgtype.Time to time.Time conversion

serprex:
pgtime is just microseconds,
https://pkg.go.dev/github.com/jackc/pgtype#Time
https://pkg.go.dev/time#UnixMicro


serprex:
Tests pass without limiting micros, but going to investigate this further & maybe add a test after I get #1491 merged

PR #1494, opened 2024-03-16
Snowflake: remove ON_ERROR=CONTINUE

serprex:
This was added before truncation was implemented

There are still situations where errors can occur,
such as timestamps in the year 20020,
but we shouldn't be dropping those records


Amogh-Bharadwaj (approved):

saisrirampur:
Lets be careful with this during upgrades as it may affect existing customer MIRRORs.

PR #1495, opened 2024-03-16
BigQuery schema changes: fix added float column

Amogh-Bharadwaj:
Currently when we add a float column to a table during a bigquery cdc mirror, it fails saying :
```
Type not found: FLOAT
```

So we need to catch that and use FLOAT64 instead


flow/connectors/bigquery/bigquery.go
```diff
@@ -295,10 +295,14 @@ func (c *BigQueryConnector) ReplayTableSchemaDeltas(
 
 		for _, addedColumn := range schemaDelta.AddedColumns {
 			dstDatasetTable, _ := c.convertToDatasetTable(schemaDelta.DstTableName)
+			addedColumnBigQueryType := qValueKindToBigQueryType(addedColumn.ColumnType)
```
serprex, line 298: qValueKindToBigQueryType should be fixed to handle this instead

Amogh-Bharadwaj: In SetupNormalize it needs the exact `FieldType` so `"FLOAT64"` won't work there:
```Golang
columns = append(columns, &bigquery.FieldSchema{
    Name:     column.Name,
    Type:     qValueKindToBigQueryType(genericColType),
    Repeated: qvalue.QValueKind(genericColType).IsArray(),
})
```

Amogh-Bharadwaj: I've moved this to a new function `qValueKindToBigQueryTypeString`

flow/connectors/bigquery/qvalue_convert.go
```diff
@@ -92,3 +92,13 @@ func BigQueryTypeToQValueKind(fieldType bigquery.FieldType) (qvalue.QValueKind,
...
 }
+
+func qValueKindToBigQueryTypeString(colType string) string {
+	bqType := qValueKindToBigQueryType(colType)
+	bqTypeAsString := string(bqType)
+	// string(bigquery.FloatFieldType) is "FLOAT" which is not a BigQuery type.
+	if bqType == bigquery.FloatFieldType {
+		bqTypeAsString = "FLOAT64"
+	}
+	return bqTypeAsString
```
serprex, line 103:
```suggestion
	// string(bigquery.FloatFieldType) is "FLOAT" which is not a BigQuery type.
	if bqType == bigquery.FloatFieldType {
		return "FLOAT64"
	}
	return string(bqType)
```
should update with other types from https://github.com/googleapis/google-cloud-go/blob/37ef89207989af819106ee4012498e638b8cc195/bigquery/schema.go#L337

serprex (approved):

PR #1499, opened 2024-03-18
[snapshot] fallback to full partitions in <PG12, cleanup

heavycrystal:
separating slot and txn snapshots better, split state and make things clearer
TID scans not supported in PG12, falling back to full table partitions for now, might add support for custom partition keys later


flow/connectors/postgres/postgres.go
```diff
@@ -941,30 +941,48 @@ func (c *PostgresConnector) EnsurePullability(
...
+	ok, _, err := c.MajorVersionCheck(ctx, POSTGRES_13)
 	if err != nil {
-		return "", nil, fmt.Errorf("[export-snapshot] error setting lock_timeout: %w", err)
+		return nil, nil, fmt.Errorf("[export-snapshot] error getting PG version: %w", err)
 	}
 
-	err = tx.QueryRow(ctx, "select pg_export_snapshot()").Scan(&snapshotName)
+	err = tx.QueryRow(ctx, "SELECT pg_export_snapshot()").Scan(&snapshotName)
 	if err != nil {
 		_ = tx.Rollback(ctx)
```
serprex, line 977:
```suggestion
```
since defer handles now

flow/connectors/postgres/postgres.go
```diff
@@ -941,30 +941,48 @@ func (c *PostgresConnector) EnsurePullability(
...
 	var snapshotName string
 	tx, err := c.conn.Begin(ctx)
 	if err != nil {
-		return "", nil, err
+		return nil, nil, err
+	}
+	txNeedsRollback := true
+	defer func() {
+		if txNeedsRollback {
+			err := tx.Rollback(context.Background())
```
serprex, line 953:
```suggestion
			err := tx.Rollback(context.Background())
```
Should have at least a 5 minute timeout or something

flow/activities/snapshot_activity.go
```diff
@@ -16,21 +16,33 @@ import (
...
+	SupportsTIDScans bool
+}
+
 type SnapshotActivity struct {
-	SnapshotConnectionsMutex sync.Mutex
-	SnapshotConnections      map[string]SlotSnapshotSignal
-	Alerter                  *alerting.Alerter
+	SnapshotStatesMutex sync.Mutex
+	SlotSnapshotStates  map[string]SlotSnapshotState
+	TxnSnapshotStates   map[string]*TxnSnapshotState
```
serprex, line 33:
```suggestion
	TxSnapshotStates   map[string]TxnSnapshotState
```
Might as well be consistent with SlotSnapshotStates map. Also rename to `TxSnap` since `Tx` is used throughout codebase as transaction abbreviation

serprex (approved):

PR #1508, opened 2024-03-20
Snapshot: do export snapshot after creating replication 

Amogh-Bharadwaj:
Doing conn.Close() after `CreateReplicationSlot` and then doing export snapshot after, setting snapshot name in our state to that.


Amogh-Bharadwaj:
closing in view of #1509

PR #1509, opened 2024-03-20
Use pg_export_snapshot instead of replication slot connection

iskakaushik:


serprex:
What's the benefit of this? We're still creating slot

PR #1521, opened 2024-03-21
Refactor peer validate check functions

Amogh-Bharadwaj:
Validation functions for most of our peers were inside NewSnowflakeConnector() , NewBigQueryConnector() and so on. 
We instantiate new connectors often during mirrors, where it isn't necessary to validate again and again.

This PR introduces a new method for Connector interface - `ValidateCheck()` which does the validation and this is called only in validate peer. The validation which was done in NewConnector() is now done here.

Functionally tested for all peers where validation is there:
- Snowflake
- BigQuery
- Clickhouse
- S3

### DIFF
--- a/flow/connectors/core.go
+++ b/flow/connectors/core.go
@@ -25,6 +25,13 @@ var ErrUnsupportedFunctionality = errors.New("requested connector does not suppo
 type Connector interface {
 	Close() error
 	ConnectionActive(context.Context) error
+}
+
+type ValidationConnector interface {
+	Connector
+
+	// ValidationCheck performs validation for the connectors,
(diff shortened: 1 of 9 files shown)


flow/connectors/core.go
```diff
@@ -25,6 +25,7 @@ var ErrUnsupportedFunctionality = errors.New("requested connector does not suppo
 type Connector interface {
 	Close() error
 	ConnectionActive(context.Context) error
+	ValidateCheck(context.Context) error
```
serprex, line 28: Instead of stub methods you can make something like `GetTableSchemaConnector` then in validation check if you can cast connector to `ValidationConnector`

iskakaushik (approved):

PR #1528, opened 2024-03-24
Change QValue into an interface

serprex:
Existing QValue was troublesome,
an artifact of wanting tagged unions,
but result was having to dispatch on tag & then dispatch again on type,
not only in terms of perf, but in terms of what can be in Value:
UUID could be string, [16]byte, or uuid.UUID

This problem was not unique to UUID

This isn't an attempt at a final QValue representation,
long term I'd like to get rid of QValueKind altogether

Not happy with duplication around time types,
ideally we'd have 1 time type with a kind tag,
maybe a QTime interface the types share


flow/model/qvalue/qvalue.go
```diff
@@ -21,157 +21,603 @@ import (
...
+	return v.Val
+}
+
+func Equals(qv QValue, other QValue) bool {
+	qvValue := qv.Value()
+	otherValue := other.Value()
+	qvEmpty := qvValue == nil || qvValue == "" || qvValue == "null" ||
+		(reflect.TypeOf(qvValue).Kind() == reflect.Slice && reflect.ValueOf(qvValue).Len() == 0)
+	otherEmpty := otherValue == nil || otherValue == "" || otherValue == "null" ||
+		(reflect.TypeOf(otherValue).Kind() == reflect.Slice && reflect.ValueOf(otherValue).Len() == 0)
```
iskakaushik, line 501: maybe have a IsQValueEmpty method?

flow/model/qvalue/qvalue.go
```diff
@@ -21,157 +21,603 @@ import (
...
+	if qvEmpty && otherEmpty {
+		return true
+	}
+	qvString, ok1 := qvValue.(string)
+	otherString, ok2 := otherValue.(string)
+	if ok1 && ok2 && qvString == otherString {
```
iskakaushik, line 507: why special case string and not just use the switch?

serprex: Because our tests rely on a multiple cases of `qvalue.Equals(QValueINET{Val:"aa"}, QValueString{Val:"aa"})` being true

Amogh-Bharadwaj (approved):

flow/model/qvalue/qvalue.go
```diff
@@ -21,157 +21,589 @@ import (
...
+type QValue interface {
+	Kind() QValueKind
+	Value() any
 }
 
-func (q QValue) Equals(other QValue) bool {
-	if q.Kind == QValueKindJSON {
-		return true // TODO fix
-	} else if q.Value == nil && other.Value == nil {
+type QValueNull QValueKind
```
heavycrystal, line 29: could be split up more, maybe a file for numeric types and helpers, time types and helpers and so on

serprex: In a follow up refactor I'd like to make these types more convenient, in particular I'd like to extend the `QValue` interface. At that time it may make more sense how to divide the code

serprex: moved Equals out, halving size of qvalue.go

heavycrystal:
LGTM, good change!

heavycrystal (approved):

PR #1534, opened 2024-03-25
No flow log entry for replication start blip

Amogh-Bharadwaj:
This PR aims to mitigate flooding of flow logs with :
```
ERROR: replication slot "myslot" is active for PID 1992482 (SQLSTATE 55006)
```
by not adding it to the catalog errors table

This error will always trigger a retry of PullRecords and is not actually fatal
Not sure if there's a better way to achieve this


flow/activities/flowable.go
```diff
@@ -352,7 +354,10 @@ func (a *FlowableActivity) SyncFlow(
 		// wait for the pull goroutine to finish
 		err = errGroup.Wait()
 		if err != nil {
-			a.Alerter.LogFlowError(ctx, flowName, err)
+			var pgErr *pgconn.PgError
+			if !(errors.As(err, &pgErr) && pgErr.SQLState() == pgerrcode.ObjectInUse) {
```
serprex, line 358:
```suggestion
			if !errors.As(err, &pgErr) || pgErr.SQLState() != pgerrcode.ObjectInUse {
```

flow/activities/flowable.go
```diff
@@ -412,7 +417,11 @@ func (a *FlowableActivity) SyncFlow(
 
 	err = errGroup.Wait()
 	if err != nil {
-		a.Alerter.LogFlowError(ctx, flowName, err)
+		var pgErr *pgconn.PgError
+		if !(errors.As(err, &pgErr) && pgErr.SQLState() == pgerrcode.ObjectInUse) {
```
serprex, line 421:
```suggestion
		if !errors.As(err, &pgErr) || pgErr.SQLState() != pgerrcode.ObjectInUse {
```

serprex:
Did you investigate why this is happening? Ideally old connection should be gone to avoid latency on reconnect. This is particularly painful for queues where P99 latency will suffer

serprex (approved):

Amogh-Bharadwaj:
> Did you investigate why this is happening? Ideally old connection should be gone to avoid latency on reconnect. This is particularly painful for queues where P99 latency will suffer

Good point @serprex. I'm closing this for now as there's no evidence that such a change is needed

PR #1535, opened 2024-03-26
lua: add msgpack preload, remove peerdb.RowToJSON

serprex:
Implement json marshaling for `*RecordItems` so `json.encode(row)` works


iskakaushik (approved):

serprex:
Not going with a `v1.0.0` on msgpack yet, waiting a bit until it's evaluated

PR #1537, opened 2024-03-26
configurable maximum concurrency of the temporal activity & workflow workers

yasinzaehringer-paradime:
This allows to configure these temporal worker parameters:
1. `MaxConcurrentActivityExecutionSize` via the `TEMPORAL_MAX_CONCURRENT_ACTIVITIES` environment variable and an equivalent CLI flag
2. `MaxConcurrentWorkflowTaskExecutionSize` via the `TEMPORAL_MAX_CONCURRENT_WORKFLOW_TASKS` environment variable and an equivalent CLI flag


flow/cmd/worker.go
```diff
@@ -4,15 +4,14 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"github.com/grafana/pyroscope-go"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/worker"
```
serprex, line 9: Moving these violates lints

serprex (approved):

PR #1554, opened 2024-03-28
Snowflake: null out numerics >38 digits

serprex:


Amogh-Bharadwaj:
Seems like we also need to do this for CDC? `toMap()` function is where we convert the records to JSON for the raw table. Should null it there as well

flow/model/qvalue/avro_converter.go
```diff
@@ -429,6 +429,10 @@ func (c *QValueAvroConverter) processNullableUnion(
 }
 
 func (c *QValueAvroConverter) processNumeric(num decimal.Decimal) interface{} {
+	if c.TargetDWH == protos.DBType_SNOWFLAKE && num.NumDigits() > 38 {
+		// snowflake only supports 38 digits
```
heavycrystal, line 433: add a log for this, maybe once instead of per record

serprex: Added logging similar to logging for JSON truncation

serprex:
> Seems like we also need to do this for CDC? `toMap()` function is where we convert the records to JSON for the raw table. Should null it there as well

Doesn't seem necessary; test hits both initial load & cdc

Amogh-Bharadwaj:
> > Seems like we also need to do this for CDC? `toMap()` function is where we convert the records to JSON for the raw table. Should null it there as well
> 
> Doesn't seem necessary; test hits both initial load & cdc

Ah this is an avro specific issue. Makes sense

Amogh-Bharadwaj (approved):

PR #1557, opened 2024-03-30
queues: default script json encodes record

serprex:


flow/connectors/utils/lua.go
```diff
@@ -62,13 +57,32 @@ func LoadScript(ctx context.Context, script string, printfn lua.LGFunction) (*lu
...
 	pua.RegisterTypes(ls)
 	ls.Env.RawSetString("print", ls.NewFunction(printfn))
-	err := ls.GPCall(pua.LoadPeerdbScript, lua.LString(script))
-	if err != nil {
-		return nil, fmt.Errorf("error loading script %s: %w", script, err)
-	}
-	err = ls.PCall(0, 0, nil)
-	if err != nil {
-		return nil, fmt.Errorf("error executing script %s: %w", script, err)
+	if script != "" {
```
Amogh-Bharadwaj, line 60: why do we not want to error out if script is empty?

serprex: Kaushik asked that there be a default encoding if script not specified. That way using Kafka/PubSub isn't blocked on adding scripts. This helps in two ways:
1. this json encoding pretty much checks all the boxes if the customer just wants their data to show up in a queue with minimal fuss
2. reduces time to wow. At Microsoft there was often a push for [5 minutes to wow](https://www.microsoft.com/en-us/sql-server/blog/2016/01/14/microsoft-business-intelligence-pass-keynote-five-minutes-to-sign-up-five-minutes-to-wow) which for us is the customer seeing their data on the other side. It compounds here where if we add updating a mirror's script to settings shortly after seeing data transmit, the customer can throw some example script in & quickly see their data in another format. You want a sequence of wins, rather than having to wait for everything to come together. This lines up with importance of initial load timing

Amogh-Bharadwaj: That makes sense. 
Wondering if script == "" can be a thing from another path, in which case there would be no errors for visibility. Maybe we can add a log for that

serprex: It'll end up spamming logs that we'll always ignore since it's quite likely most people will use the default encoding

Amogh-Bharadwaj (approved):

PR #1560, opened 2024-04-01
PostgresMetadata: implement connector interfaces

serprex:
Embed compositions for implicit connector implementations

### DIFF
--- a/flow/connectors/s3/s3.go
+++ b/flow/connectors/s3/s3.go
@@ -23,11 +23,11 @@ const (
 )
 
 type S3Connector struct {
-	logger     log.Logger
-	pgMetadata *metadataStore.PostgresMetadataStore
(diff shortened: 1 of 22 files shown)


flow/connectors/s3/s3.go
```diff
@@ -117,41 +117,13 @@ func (c *S3Connector) ValidateCheck(ctx context.Context) error {
 		return fmt.Errorf("failed to delete from bucket: %w", delErr)
 	}
 
-	// check if we can ping external metadata
-	if c.pgMetadata != nil {
```
serprex, line 121: This check isn't necessary; we ping connection before returning connection from pool

serprex:
Also: replace all `Qrep` with `QRep`

Amogh-Bharadwaj (approved):

PR #1562, opened 2024-04-01
RecordItems: only track values in map[string]qvalue.QValue

serprex:
We never make use of slice ordering

### DIFF
--- a/flow/pua/peerdb.go
+++ b/flow/pua/peerdb.go
@@ -137,15 +137,17 @@ func LuaRowIndex(ls *lua.LState) int {
 
 func LuaRowLen(ls *lua.LState) int {
 	row := LuaRow.StartMethod(ls)
-	ls.Push(lua.LNumber(len(row.Values)))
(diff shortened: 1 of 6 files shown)


flow/pua/peerdb.go
```diff
@@ -137,15 +137,17 @@ func LuaRowIndex(ls *lua.LState) int {
...
 }
 
 func LuaRowColumns(ls *lua.LState) int {
 	row := LuaRow.StartMethod(ls)
-	tbl := ls.CreateTable(len(row.ColToValIdx), 0)
-	for col, idx := range row.ColToValIdx {
-		tbl.RawSetInt(idx+1, lua.LString(col))
+	tbl := ls.CreateTable(len(row.ColToVal), 0)
+	idx := 0
+	for col := range row.ColToVal {
```
serprex, line 148: This makes this functions output non deterministic, but it's only being used in a script to create a table for encoding

I should instead implement `__pairs` on `Row`

Amogh-Bharadwaj (approved):

PR #1566, opened 2024-04-03
UI: Create Eventhub Peer and Eventhub Mirror

Amogh-Bharadwaj:
This PR adds Eventhubs as a peer in UI.
- Includes form validation
- Helper text
Screenshot 2024-04-04 at 2 14 13 AM

It also adapts the CDC create mirror page for Eventhubs.
- Info on target table semantics for Eventhub
- Automatically turns off and does not display initial load related and soft delete settings
Screenshot 2024-04-04 at 2 27 27 AM

### DIFF
--- a/ui/app/mirrors/create/handlers.ts
+++ b/ui/app/mirrors/create/handlers.ts
@@ -22,6 +22,17 @@ import {
   tableMappingSchema,
 } from './schema';
 
+export const IsQueuePeer = (peerType?: DBType): boolean => {
+  if (!peerType) {
+    return false;
+  }
+  return (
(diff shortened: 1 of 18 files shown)


ui/app/mirrors/create/handlers.ts
```diff
@@ -71,6 +71,11 @@ const CDCCheck = (
     config.replicationSlotName = '';
   }
 
+  if (config.destination?.type == DBType.EVENTHUBS) {
+    config.doInitialSnapshot = false;
+    config.softDelete = false;
+  }
```
serprex, line 77: This needs to happen for other queues too

PR #1571, opened 2024-04-04
Add scripting to eventhubs

serprex:
When script is empty stick to existing eventhubs behavior

Script can return quite a bit as a table:
1. value: EventData Body
2. contentType: EventData ContentType
3. messageId: EventData MessageID
4. headers: EventData Properties

Destination can also be overriden:
1. destination: overrides mirror's destination table value. Following values will overwrite values from this string
2. namespace: can pick other namespaces, but pick from hub configs
3. partitionColumn: setting without `key` will use existing partition-by-column logic
4. key: sets partition key value, bypassing hashing & everything
5. hub: sets eventhub name

By having destination apply first, users can
```lua
return { destination = r.target, key = "override" }
```
causing behavior to adapt existing behavior set by their destination table string

### DIFF
--- a/flow/connectors/eventhub/eventhub.go
+++ b/flow/connectors/eventhub/eventhub.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"strings"
 	"sync/atomic"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	azeventhubs "github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs"
+	"github.com/yuin/gopher-lua"
 	"go.temporal.io/sdk/log"
 
 	metadataStore "github.com/PeerDB-io/peer-flow/connectors/external_metadata"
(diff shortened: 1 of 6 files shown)


flow/connectors/eventhub/eventhub.go
```diff
@@ -80,11 +85,104 @@ func (c *EventHubConnector) SetupMetadataTables(_ context.Context) error {
...
+			scoped.Hub.NamespaceName = namespace
+		}
+		partCol, err := utils.LVAsStringOrNil(ls, ls.GetField(v, "partitionColumn"))
+		if err != nil {
+			return scoped, fmt.Errorf("invalid partitionColumn, %w", err)
+		}
+		if partCol != "" {
+			scoped.Hub.PartitionKeyColumn = partCol
+		}
+		key, err := utils.LVAsStringOrNil(ls, ls.GetField(v, "key"))
```
Amogh-Bharadwaj, line 124:
```suggestion
		key, err := utils.LVAsStringOrNil(ls, ls.GetField(v, "partitionKey"))
```
To make it more clear

Amogh-Bharadwaj:
Can we also move all the json fields we are get-ing in eventhub.go to a constants file and then replace the literals with those?

flow/connectors/eventhub/eventhub.go
serprex: Using "key" to be consistent with kafka/pubsub

serprex:
> Can we also move all the json fields we are get-ing in eventhub.go to a constants file and then replace the literals with those?

Going to address this in a more general way in future PR using tagged struct fields

Amogh-Bharadwaj (approved):

PR #1572, opened 2024-04-04
RenameTablesConnector

serprex:


flow/connectors/core.go
```diff
@@ -228,27 +234,27 @@ func GetConnectorAs[T Connector](ctx context.Context, config *protos.Peer) (T, e
 }
 
 func GetCDCPullConnector(ctx context.Context, config *protos.Peer) (CDCPullConnector, error) {
```
heavycrystal, line 236: should we remove these helpers and invoke `GetAs` directly? since we have a lot more interfaces now

serprex: Eventually yes

Amogh-Bharadwaj (approved):

PR #1574, opened 2024-04-04
UI: Also show partitioned tables in table picker

Amogh-Bharadwaj:
We were filtering by `relkind='r'` in the endpoint query to get the tables in a schema for CDC table picker UI, but we also support CDC of partitioned tables, so add ``relkind='p'` as well.
https://www.postgresql.org/docs/current/catalog-pg-class.html


Amogh-Bharadwaj:
Need to implement:
For PG12 don't display partitioned tables as [publish_via_partition_root](https://www.postgresql.org/docs/current/sql-createpublication.html#SQL-CREATEPUBLICATION-WITH-PUBLISH-VIA-PARTITION-ROOT) is not supported there.

flow/cmd/peer_data.go
```diff
@@ -94,7 +94,7 @@ func (h *FlowRequestHandler) GetTablesInSchema(
 	FROM pg_class t
 	LEFT JOIN pg_namespace n ON t.relnamespace = n.oid
 	LEFT JOIN pg_constraint con ON con.conrelid = t.oid
-	WHERE n.nspname = $1 AND t.relkind = 'r'
+	WHERE n.nspname = $1 AND (t.relkind = 'r' OR t.relkind = 'p')
```
serprex, line 97:
```suggestion
	WHERE n.nspname = $1 AND t.relkind IN ('r', 'p')
```

serprex (approved):

flow/connectors/postgres/client.go
```diff
@@ -523,7 +523,7 @@ func (c *PostgresConnector) jobMetadataExists(ctx context.Context, jobName strin
 	return result.Bool, nil
 }
 
-func majorVersion(ctx context.Context, conn *pgx.Conn) (PGVersion, error) {
+func GetMajorVersion(ctx context.Context, conn *pgx.Conn) (PGVersion, error) {
```
serprex, line 526: Could move this function to `shared/postgres.go`

Amogh-Bharadwaj: #1576

PR #1579, opened 2024-04-05
Update dependencies

serprex:
rustls recently switched to aws-lc-rs as their default crypto provider, stay with ring

grpc deprecated DialContext, switching to NewClient will be a followup PR


serprex:
https://github.com/bikeshedder/deadpool/pull/312 could wait on this to avoid bringing in random console dependency

grpc NewClient PR: https://github.com/grpc/grpc-go/pull/7010

iskakaushik (approved):

PR #1580, opened 2024-04-05
Resync for SF and BQ: Quote column names for RenameTables

Amogh-Bharadwaj:
This PR quotes column names for the insert into select command in RenameTables activity for Snowflake and BigQuery. A column with name 'rows' would make the query fail due to it being a reserved keyword. This PR fixes that

Functionally tested


serprex (approved):

serprex:
Should rename a column somewhere in tests to `rows`

Amogh-Bharadwaj:
> Should rename a column somewhere in tests to `rows`

That comes under the mixed case support task which is still in the backburner. I think a test for resync itself would be good.

PR #1583, opened 2024-04-08
Move datatype folders to super folder

Amogh-Bharadwaj:
Moves geo, hstore, interval and numeric files where we do special operations on those types, to a `datatypes` folder


serprex:
These can all share one package name

serprex (approved):

PR #1593, opened 2024-04-10
remove SendBatch in PG normalize

heavycrystal:
seems to be flaky, hangs sometimes

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -14,7 +14,6 @@ import (
 	"github.com/google/uuid"
 	"github.com/jackc/pglogrepl"
 	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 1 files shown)


iskakaushik (approved):

serprex (approved):

flow/connectors/postgres/postgres.go
```diff
@@ -620,18 +618,11 @@ func (c *PostgresConnector) NormalizeRecords(ctx context.Context, req *model.Nor
 	for _, destinationTableName := range destinationTableNames {
 		normalizeStatements := normalizeStmtGen.generateNormalizeStatements(destinationTableName)
 		for _, normalizeStatement := range normalizeStatements {
```
serprex, line 620: How many statements do we expect to generate? should we perhaps batch the normalize statements but not the tables?

PR #1594, opened 2024-04-10
Fix PG-PG QRep Overwrite

Amogh-Bharadwaj:
Overwrite mode was using the sync records branch for Upsert, which expects unique key columns. It failed there.
This PR fixes this by making it use the append branch. It also moves the truncate step to this branch from setup flow, since after the first sync, qrep flow continues as new, enters setup flow again where it would truncate what it just synced even though no new rows were ingested necessarily

Automated test added
Functionally tested

### DIFF
--- a/flow/connectors/postgres/qrep_sql_sync.go
+++ b/flow/connectors/postgres/qrep_sql_sync.go
@@ -88,6 +88,7 @@ func (s *QRepStagingTableSync) SyncQRepRecords(
 
 		if writeMode.WriteType == protos.QRepWriteType_QREP_WRITE_MODE_OVERWRITE {
 			// Truncate the destination table before copying records
+			s.connector.logger.Info(fmt.Sprintf("Truncating table %s for overwrite mode", dstTableName), syncLog)
 			_, err = tx.Exec(ctx,
 				"TRUNCATE TABLE "+dstTableName.String())
 			if err != nil {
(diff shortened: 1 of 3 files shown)


flow/connectors/postgres/qrep_sql_sync.go
```diff
@@ -83,7 +83,18 @@ func (s *QRepStagingTableSync) SyncQRepRecords(
...
 
 	if writeMode == nil ||
-		writeMode.WriteType == protos.QRepWriteType_QREP_WRITE_MODE_APPEND {
+		writeMode.WriteType == protos.QRepWriteType_QREP_WRITE_MODE_APPEND ||
+		writeMode.WriteType == protos.QRepWriteType_QREP_WRITE_MODE_OVERWRITE {
+		if writeMode != nil && writeMode.WriteType == protos.QRepWriteType_QREP_WRITE_MODE_OVERWRITE {
+			// Truncate the destination table before copying records
+			s.connector.logger.Info(fmt.Sprintf("Truncating table %s for overwrite mode", dstTableName), syncLog)
+			_, err = tx.Exec(ctx,
+				"TRUNCATE TABLE "+dstTableName.String())
```
heavycrystal, line 92: `QuoteIdentifier`?

serprex (approved):

Amogh-Bharadwaj:
Note: All other connectors have the same flaw where the destination table will never have any rows. 
This PR puts in place the semantics that at every sync interval the target table will be overwritten with the newly pulled rows.
If this isn't the behaviour we want, then we should probably remove overwrite mode for now
Otherwise, should apply this fix for the other connectors too, where truncation happens at sync and not in setup
cc: @iskakaushik

PR #1604, opened 2024-04-11
Add missing flow name keys to some logs

Amogh-Bharadwaj:
Logs in PullRecords and ConsolidatePartitions were missing the flow name key.
Added values to context in activity for the latter
For PullRecords, `p.logger` (and p.PostgresConnector.logger) for some reason does not print flow name even though it takes from loggerfromctx where ctx should have the flowName key. Using loggerfromctx in the pullrecords function explicitly instead, does print it

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -112,6 +112,7 @@ func GetChildToParentRelIDMap(ctx context.Context, conn *pgx.Conn) (map[uint32]u
 
 // PullRecords pulls records from the cdc stream
 func (p *PostgresCDCSource) PullRecords(ctx context.Context, req *model.PullRecordsRequest) error {
+	logger := logger.LoggerFromCtx(ctx)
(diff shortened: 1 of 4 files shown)


flow/connectors/postgres/cdc.go
```diff
@@ -112,6 +112,7 @@ func GetChildToParentRelIDMap(ctx context.Context, conn *pgx.Conn) (map[uint32]u
 
 // PullRecords pulls records from the cdc stream
 func (p *PostgresCDCSource) PullRecords(ctx context.Context, req *model.PullRecordsRequest) error {
+	logger := logger.LoggerFromCtx(ctx)
```
Amogh-Bharadwaj, line 115: As mentioned in the description, `p.logger` also takes from logger.LoggerFromCtx during getconnector, so not sure why it doesn't print the flow name but this does

serprex (approved):

PR #1605, opened 2024-04-11
UI Mirror page: better overview

Amogh-Bharadwaj:
- Moves pause and resume to actions
- Remove redundant state change onclick functions
- Use tremor's select instead of radix dropdown
- Adds tooltip for edit button saying you need to pause first
- Shows sync interval instead of last synced at, which was a misleading piece of info as it implies data is on target but actually normalise is still running
- Adds line in edit page saying you need to alter publication for custom publications
Screenshot 2024-04-11 at 10 27 23 PM
Screenshot 2024-04-11 at 10 30 38 PM


serprex (approved):
qrep pause when?

PR #1611, opened 2024-04-12
UI Create Peer: replace drop down with columns of buttons

serprex:
Reduces clicks, with dropdown it's:
1. click dropdown
2. scroll list
3. click peer type
4. click create

Now user only needs one click

localhost_3000_peers_create


Amogh-Bharadwaj:
One drawback is that while this reduces clicking, user needs to search for the peer they want, when before they could type one letter and hit enter

iskakaushik:
lets sort them horizontally first and then vertically.

iskakaushik (approved):

PR #1613, opened 2024-04-13
parallelized lua script execution with serialized output

serprex:


flow/connectors/kafka/kafka.go
```diff
@@ -242,38 +241,61 @@ Loop:
...
-			ls.Push(pua.LuaRecord.New(ls, record))
-			err := ls.PCall(1, -1, nil)
-			if err != nil {
-				return nil, fmt.Errorf("script failed: %w", err)
-			}
-			args := ls.GetTop()
-			for i := range args {
-				kr, err := lvalueToKafkaRecord(ls, ls.Get(i-args))
+			var err error
+			chain, err = utils.LuaPoolJob(wgCtx, pool, chain, func(ls *lua.LState, wait <-chan struct{}) {
```
iskakaushik, line 245: ideally be a method on `pool` SubmitJob or something.

serprex: pool is `puddle.Pool[*lua.LState]` so can't add a method, but once this is validated I'll look to move more code into utils for sharing with pubsub

serprex:
This currently spawns a goroutine for each message. It could be optimized to only spawn 1 goroutine per lua.LState

flow/connectors/kafka/kafka.go
serprex: Replaced `puddle` with a finite number of coroutines pulling from message channels while pushing back to channels which a coordinator goroutine reads in proper order. Channel juggling now entirely hidden behind abstraction

flow/connectors/utils/lua.go
```diff
@@ -82,31 +81,95 @@ func DefaultOnRecord(ls *lua.LState) int {
...
+		pool.Close()
+		return nil, err
 	}
-	return pool
+	return pool, nil
 }
 
-func LuaPoolJob(ctx context.Context, pool LPool, wait <-chan struct{}, f func(*lua.LState, <-chan struct{})) (chan struct{}, error) {
-	res, err := pool.Acquire(ctx)
+func (pool *LPool[T]) Spawn() error {
```
serprex, line 128: While it'd be fine to spin up a constant N goroutines, the correct amount will depend on the script, making it better to specify that kind of knob on the mirror instead of environment

Instead Spawn is called when messages aren't being waited on, this way an environment can set parallelism to some good amount for the vcores available & fast scripts will only create as many goroutines as necessary to keep up with record ingestion

flow/connectors/utils/lua.go
```diff
@@ -80,3 +80,96 @@ func DefaultOnRecord(ls *lua.LState) int {
...
+	returns  chan<- (<-chan T)
+	wait     <-chan struct{}
+	cons     func() (*lua.LState, error)
+	maxSize  int
+	size     int
+	closed   bool
+}
+
+func LuaPool[T any](cons func() (*lua.LState, error), merge func(T)) (*LPool[T], error) {
+	maxSize := 4 // TODO env variable
```
serprex, line 99: TODO

iskakaushik (approved):

PR #1618, opened 2024-04-16
Elasticsearch connector for QRep

heavycrystal:
Current limitations:
1. doesn't support upsert with multiple key columns
2. need to evaluate what types are supported as an upsert key
3. basic types tested, need to check the rest
4. no flow tests yet
5. doesn't support index creation, so we need to disable create table on destination


flow/connectors/core.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/PeerDB-io/peer-flow/alerting"
 	connbigquery "github.com/PeerDB-io/peer-flow/connectors/bigquery"
 	connclickhouse "github.com/PeerDB-io/peer-flow/connectors/clickhouse"
+	conn_elasticsearch "github.com/PeerDB-io/peer-flow/connectors/elasticsearch"
```
serprex, line 13:
```suggestion
	connelasticsearch "github.com/PeerDB-io/peer-flow/connectors/elasticsearch"
```
for consistency

flow/connectors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,210 @@
...
+func (esc *ElasticsearchConnector) Close() error {
+	// stateless connector
+	return nil
+}
+
+func (esc *ElasticsearchConnector) IsQRepPartitionSynced(ctx context.Context,
+	req *protos.IsQRepPartitionSyncedInput,
+) (bool, error) {
+	return esc.pgMetadata.IsQRepPartitionSynced(ctx, req)
+}
```
serprex, line 84:
```suggestion
```
Not needed after flattening pgMetadata field

flow/connectors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,210 @@
...
+		qRecordJsonBytes, err := json.Marshal(qRecordJsonMap)
+		if err != nil {
+			esc.logger.Error("[es] failed to json.Marshal record", slog.Any("error", err))
+			return 0, fmt.Errorf("[es] failed to json.Marshal record: %w", err)
+		}
+
+		esBulkIndexer.Add(ctx, esutil.BulkIndexerItem{
+			Action:     "index",
+			DocumentID: docId,
+			Body:       bytes.NewReader(qRecordJsonBytes),
```
serprex, line 177: oh we could straight up pass an actual reader instead of `bytes.NewReader` here & stream the above json encoding (convert appends into `Write` calls) but would have to spin up a goroutine so probably not worth

flow/connectors/elasticsearch/elasticsearch.go
heavycrystal: adds CDC related methods which I find confusing

flow/connectors/elasticsearch/elasticsearch.go
heavycrystal: doesn't interface well with QRep parallelism implemented across multiple workflows currently

flow/connectors/elasticsearch/elasticsearch.go
serprex: don't worry, eventually when we want cdc anyways it'll already be there

flow/connectors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,210 @@
...
+		qRecordJsonMap := make(map[string]any)
+
+		if upsertColIndex >= 0 {
+			docId = fmt.Sprintf("%v", qRecordOrErr.Record[upsertColIndex].Value())
+		} else {
+			docId = uuid.New().String()
+		}
+		for i, field := range schema.Fields {
+			switch field.Type {
+			// JSON is stored as a string, fix that
```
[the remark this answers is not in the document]
serprex: https://pkg.go.dev/encoding/json#RawMessage turns out encoding/json does have the escape hatch included

flow/connectors/connelasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func (esc *ElasticsearchConnector) SyncQRepRecords(ctx context.Context, config *protos.QRepConfig,
+	partition *protos.QRepPartition, stream *model.QRecordStream,
+) (int, error) {
+	startTime := time.Now()
+
+	schema := stream.Schema()
+
+	var bulkIndexHasError bool
```
serprex, line 93:
```suggestion
```
Can check `len(bulkIndexErrors) > 0`

serprex:
Approving since it looks alright & isn't impactful to existing code, but could use an e2e test

heavycrystal:
issue for E2E tests filed as #1621

flow/connectors/connelasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		for i, field := range schema.Fields {
+			switch field.Type {
+			// JSON is stored as a string, fix that
+			case qvalue.QValueKindJSON:
+				qRecordJsonMap[field.Name] = json.RawMessage(shared.
+					UnsafeFastStringToReadOnlyBytes(qRecord[i].(qvalue.QValueJSON).Val))
+			default:
+				qRecordJsonMap[field.Name] = qRecord[i].Value()
+			}
```
serprex, line 148:
```suggestion
			switch r := qRecord[i].(type) {
			// JSON is stored as a string, fix that
			case qvalue.QValueJSON:
				qRecordJsonMap[field.Name] = json.RawMessage(shared.
					UnsafeFastStringToReadOnlyBytes(r.Val))
			default:
				qRecordJsonMap[field.Name] = r.Value()
			}
```
This properly handles `QValueNull`

PR #1620, opened 2024-04-16
Remove QRecordOrError

serprex:
Errors are non recoverable, indicating end of stream
Match existing APIs which have an Err method which should be checked after iteration

### DIFF
--- a/flow/model/qrecord_stream.go
+++ b/flow/model/qrecord_stream.go
@@ -75,6 +75,9 @@ func (s *QRecordStream) Err() error {
 	return s.err
 }
 
+// Set error & close stream. Calling with multiple errors only tracks first error & does not panic.
+// Close(nil) after an error won't panic, but Close after Close(nil) will panic,
+// this is enough to be able to safely `defer stream.Close(nil)`.
 func (s *QRecordStream) Close(err error) {
 	if s.err == nil {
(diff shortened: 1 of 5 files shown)


flow/model/qrecord_stream.go
```diff
@@ -77,3 +70,14 @@ func (s *QRecordStream) IsSchemaSet() bool {
 func (s *QRecordStream) SchemaChan() <-chan struct{} {
 	return s.schemaLatch
 }
+
+func (s *QRecordStream) Err() error {
+	return s.err
+}
+
+func (s *QRecordStream) Close(err error) {
+	if s.err == nil {
```
heavycrystal, line 79: shouldn't this be != nil and an unconditional close?

serprex: No, notice that we're checking `s.err` not `err`. `close` panics when called twice on a channel. So this method is designed such that:
1. Close(err) can be called multiple times, QRecordStream handling only tracking first error & only closing channel once
2. Close(nil) can be called unconditionally once (ie, defer) & it won't panic if an error already closed channel

heavycrystal: thanks, please add comment

heavycrystal (approved):

PR #1630, opened 2024-04-19
UI for opting into PG type system

Amogh-Bharadwaj:
Adds a switch field in create CDC UI  for opting for pg type system - for source and destination set to Postgres peers.
Functionally tested

Screenshot 2024-04-19 at 9 29 08 PM


serprex:
Little unsure about if this needs extra guardrails. Info should specify this is experimental & should only be used with PG<>PG mirrors

serprex (approved):

PR #1507, opened 2024-03-19
feat(aws): add credentials loading mechanism using sdk config

iamKunalGupta:
Important function is `GetAWSCredentialsProvider` in `flow/connectors/utils/aws.go`

Order (in descending preference):
1. Peer Credentials
2. Explicit `PEERDB_<CONNECTOR-NAME>_AWS_CREDENTIALS_<AWS-ENV-NAME>` like `PEERDB_CLICKHOUSE_AWS_CREDENTIALS_AWS_ACCESS_KEY_ID`
3. Default creds loaded by AWS sdk

Pending Items:
- [x] Resolve TODOs
- [x] Clickhouse Test - ONLY WORKS WITH versions after 18th March 2024
- [x] Snowflake Test
- [x] @iskakaushik this is blocked until https://github.com/ClickHouse/ClickHouse/issues/61230 gets to a stable version of clickhouse.
- [x] Retest:
  - [x] incompatible version check
  
    image

  - [x] compatible version check
  
    image
 - [x] Check what version clickhouse cloud is currently on
   - [x] Most customers are on 24.2
- [x] Run a long test for 1 day to make sure creds are refreshed

PR #1634, opened 2024-04-22
In-house S3 stage for Clickhouse

Amogh-Bharadwaj:
This PR:
- Adds an endpoint field for Clickhouse S3 configuration
- Adds a [minio](https://min.io/) container in both docker-compose files
- Wires the above container to be the S3 stage which clickhouse will use by default in PeerDB OSS, if user provided fields are not there.
- Above mentioned user provided fields are now under a cover in UI:

Screenshot 2024-04-26 at 5 17 17 PM

Functionally tested (peer validation + full CDC mirror with initial load) for:
- S3
- GCS
- Remote Minio
- Local minio (no fields inputted in UI)

Tested minio container and bucket creation for both docker compose files


iskakaushik (approved):
@Amogh-Bharadwaj  is it possible to test this somehow.

flow/connectors/utils/aws.go
```diff
@@ -85,13 +86,15 @@ func GetClickhouseAWSSecrets(bucketPathSuffix string) *ClickhouseS3Credentials {
 	awsKey := os.Getenv("PEERDB_CLICKHOUSE_AWS_CREDENTIALS_AWS_ACCESS_KEY_ID")
 	awsSecret := os.Getenv("PEERDB_CLICKHOUSE_AWS_CREDENTIALS_AWS_SECRET_ACCESS_KEY")
 	awsBucketName := os.Getenv("PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME")
+	awsEndpoint := os.Getenv("PEERDB_CLICKHOUSE_AWS_S3_ENDPOINT")
```
iamKunalGupta, line 89: Can you change this to Also can you use `PEERDB_CLICKHOUSE_AWS_ENDPOINT_URL_<SERVICE>` as the suffix instead of `PEERDB_CLICKHOUSE_AWS_S3_ENDPOINT `, would allow us to be in sync with 
 https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-endpoints.html#endpoints-service-specific-table

iamKunalGupta (changes requested):
Should we wait for #1507 ?

flow/connectors/utils/aws.go
Amogh-Bharadwaj: having `<` and `>` in env name is a bit unconventional. maybe 
`PEERDB_CLICKHOUSE_AWS_ENDPOINT_URL_SERVICE` ?

iamKunalGupta: > having `<` and `>` in env name is a bit unconventional. maybe `PEERDB_CLICKHOUSE_AWS_ENDPOINT_URL_SERVICE` ?

Yes by <SERVICE> I mean `PEERDB_CLICKHOUSE_AWS_ENDPOINT_URL_S3`

check the url for more clarity

Amogh-Bharadwaj: As in one environment variable for each service?

iamKunalGupta: just the one for s3 for now

iskakaushik:
@iamKunalGupta  -- can you take this for a spin as well and let me know if this in in-par with what you are thinking.

flow/connectors/s3/s3.go
```diff
@@ -76,29 +77,25 @@ func (c *S3Connector) Close() error {
 	return nil
 }
 
-func (c *S3Connector) ValidateCheck(ctx context.Context) error {
+// Write an empty file and then delete it
+// to check if we have write permissions
+func PutAndRemoveS3(ctx context.Context, client *s3.Client, bucket string, prefix string) error {
```
serprex, line 82: this should go to connectors/utils to avoid connectors depending on each other

flow/connectors/utils/aws.go
```diff
@@ -178,6 +201,20 @@ func GetAWSCredentialsProvider(ctx context.Context, connectorName string, peerCr
 	return NewConfigBasedAWSCredentialsProvider(awsConfig), nil
 }
 
+func FileURLForS3Service(endpoint string, region string, bucket string, filePath string) string {
+	fileUrl := fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", bucket, region, filePath)
+	if strings.Contains(endpoint, "storage.googleapis.com") {
+		fileUrl = fmt.Sprintf("https://storage.googleapis.com/%s/%s", bucket, filePath)
+	}
+	// example: min.io local bucket or GCS
+	match, _ := regexp.MatchString(`^https?://[a-zA-Z0-9.-]+(:\d+)?$`, endpoint)
```
serprex, line 210: move to a global `var x = regexp.MustCompile("^https?://[a-zA-Z0-9.-]+(:\d+)?$")`

flow/connectors/utils/aws.go
```diff
@@ -178,6 +201,20 @@ func GetAWSCredentialsProvider(ctx context.Context, connectorName string, peerCr
...
+	if match {
+		fileUrl = fmt.Sprintf("%s/%s/%s", endpoint, bucket, filePath)
+	}
+
+	return fileUrl
```
serprex, line 215:
```suggestion
	if strings.Contains(endpoint, "storage.googleapis.com") {
		return fmt.Sprintf("https://storage.googleapis.com/%s/%s", bucket, filePath)
	}
	// example: min.io local bucket or GCS
	match, _ := regexp.MatchString(`^https?://[a-zA-Z0-9.-]+(:\d+)?$`, endpoint)
	if match {
		return fmt.Sprintf("%s/%s/%s", endpoint, bucket, filePath)
	}

	return fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", bucket, region, filePath)
```

flow/connectors/utils/aws.go
```diff
@@ -178,6 +201,20 @@ func GetAWSCredentialsProvider(ctx context.Context, connectorName string, peerCr
 	return NewConfigBasedAWSCredentialsProvider(awsConfig), nil
 }
 
+func FileURLForS3Service(endpoint string, region string, bucket string, filePath string) string {
+	fileUrl := fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", bucket, region, filePath)
+	if strings.Contains(endpoint, "storage.googleapis.com") {
```
serprex, line 206: Don't like this, ideally the regex would below would handle if user enters `https://storage.googleapis.com` & we're pretty much ignoring the endpoint when it contains this specific string

flow/go.mod
```diff
@@ -170,7 +170,7 @@ require (
 	github.com/zeebo/xxh3 v1.0.2 // indirect
 	go.opencensus.io v0.24.0 // indirect
 	golang.org/x/exp v0.0.0-20240416160154-fe59bbe5cc7f
-	golang.org/x/mod v0.17.0 // indirect
+	golang.org/x/mod v0.17.0
```
serprex, line 173: move to non-indirect dependencies if this is now a non-indirect dependency

Amogh-Bharadwaj:
Updated title and description. cc @iskakaushik

serprex (approved):

iamKunalGupta (approved):

PR #1637, opened 2024-04-22
MySQL peer

serprex:
Currently testing with [mariadb](https://hub.docker.com/_/mariadb)
```
CREATE PEER my FROM MYSQL WITH (
  host = 'mysql',
  port = '3306',
  user = 'root',
  password = 'example',
  database = 'main',
  compression = 1,
  disable_tls = true
);
```

The scope of this PR is only peer creation & nexus proxy. No mirror support


saisrirampur:
Does the PR handle prepared statements @serprex I know that for postgres_fdw to work, unnamed prepared statements would need to work cc @iskakaushik

serprex:
No, we only support SELECT

nexus/analyzer/src/lib.rs
```diff
@@ -903,5 +915,39 @@ fn parse_db_options(db_type: DbType, with_options: &[SqlOption]) -> anyhow::Resu
...
+                .cloned()
+                .unwrap_or_default()
+                .to_string(),
+            setup: opts
+                .get("setup")
+                .map(|s| s.split(';').map(String::from).collect::<Vec<_>>())
+                .unwrap_or_default(),
+            compression: opts
+                .get("compression")
+                .and_then(|s| s.parse::<i8>().ok())
```
heavycrystal, line 945: u8? also apply any restrictions if possible

nexus/peer-mysql/src/client.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub query: String,
+    pub response: mpsc::Sender<Response>,
+}
+
+#[derive(Clone)]
+pub struct MyClient {
+    pub chan: mpsc::Sender<Message>,
+}
+
+impl MyClient {
```
heavycrystal, line 23: nit: names are a tad confusing

nexus/peer-mysql/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+pub struct MySqlQueryExecutor {
+    peer_name: String,
+    client: client::MyClient,
+    cursor_manager: CursorManager,
+}
+
+impl MySqlQueryExecutor {
+    pub async fn new(peer_name: String, config: &MySqlConfig) -> anyhow::Result<Self> {
+        let mut opts = mysql_async::OptsBuilder::default().prefer_socket(Some(false)); // prefer_socket breaks connecting to StarRocks
+        if !config.user.is_empty() {
```
heavycrystal, line 24: if it is empty should the client even be constructed, similar for other fields

serprex: Yes, it'll use defaults (in this case, `root`)

nexus/peer-mysql/src/client.rs
serprex: I agree, MySQL is a confusing name

nexus/analyzer/src/lib.rs
serprex: probably u32 to match mysql_async

nexus/Cargo.toml
```diff
@@ -19,5 +20,8 @@ resolver = "2"
 
 [workspace.dependencies]
 chrono = { version = "0.4", default-features = false, features = ["serde", "std"] }
-sqlparser = { git = "https://github.com/peerdb-io/sqlparser-rs.git", branch = "ps-es-parsing" }
+dashmap = "5.0"
```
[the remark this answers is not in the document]
serprex: there's a lot of dependency movement, but updating sqlparser-rs to latest main brings in mysql syntax

nexus/peer-mysql/src/lib.rs
heavycrystal: should we make username etc. optional in that case in the analyzer logic?

serprex: Yes

iskakaushik (approved):

PR #1645, opened 2024-04-24
error workflow in specific ES error

heavycrystal:
Currently for the error type `illegal_argument_exception` that contains the error when a index has "too many documents" which is an underlying Lucene limitation. Limit is 2b per Lucene shard.

Needs to be addressed better for ES in the future.

### DIFF
--- a/flow/connectors/connelasticsearch/elasticsearch.go
+++ b/flow/connectors/connelasticsearch/elasticsearch.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


serprex (approved):

flow/connectors/connelasticsearch/elasticsearch.go
```diff
@@ -174,13 +176,20 @@ func (esc *ElasticsearchConnector) SyncQRepRecords(ctx context.Context, config *
 					bulkIndexErrors = append(bulkIndexErrors,
 						fmt.Errorf("id:%s type:%s reason:%s %s", item.DocumentID, res.Error.Type,
 							res.Error.Reason, causeString))
+					if res.Error.Type == "illegal_argument_exception" &&
+						strings.Contains(res.Error.Reason, "Number of documents in the index can't exceed") {
+						bulkIndexHasFatalError = true
```
serprex, line 181:
```suggestion
						bulkIndexFatalError = res.Error
```
Then change type to `error`. Allows for better logging at end

PR #1646, opened 2024-04-25
fix: enable support for TLS 1.2

iamKunalGupta:
- fix for usage with confluent cloud

### DIFF
--- a/flow/connectors/kafka/kafka.go
+++ b/flow/connectors/kafka/kafka.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"github.com/twmb/franz-go/plugin/kslog"
 	"log/slog"
 	"strings"
 	"sync"
@@ -13,7 +14,6 @@ import (
 	"github.com/twmb/franz-go/pkg/kgo"
 	"github.com/twmb/franz-go/pkg/sasl/plain"
 	"github.com/twmb/franz-go/pkg/sasl/scram"
-	"github.com/twmb/franz-go/plugin/kslog"
 	lua "github.com/yuin/gopher-lua"
 	"go.temporal.io/sdk/log"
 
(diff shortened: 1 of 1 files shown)


iskakaushik (approved):

flow/connectors/kafka/kafka.go
```diff
@@ -45,7 +47,7 @@ func NewKafkaConnector(
 		kgo.SoftwareNameAndVersion("peerdb", peerdbenv.PeerDBVersionShaShort()),
 	)
 	if !config.DisableTls {
-		optionalOpts = append(optionalOpts, kgo.DialTLSConfig(&tls.Config{MinVersion: tls.VersionTLS13}))
+		optionalOpts = append(optionalOpts, kgo.DialTLSConfig(&tls.Config{MinVersion: tls.VersionTLS12}))
```
serprex, line 50: This should be a setting. FedRAMP requires 1.3

flow/connectors/kafka/kafka.go
```diff
@@ -10,6 +10,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/twmb/franz-go/plugin/kslog"
+
```
serprex, line 14:
```suggestion
```

flow/connectors/kafka/kafka.go
iamKunalGupta: But this is minimum supported, max supported is still 1.3

iamKunalGupta: https://github.com/golang/go/blob/go1.22.2/src/crypto/tls/common.go#L738-L751

serprex: Yes. Guess we can always switch to an environment variable at some point to block TLS 1.2 if a customer wants that

serprex (approved):

iskakaushik (approved):

PR #1649, opened 2024-04-26
Elasticsearch connector for CDC

heavycrystal:
Current limitations:
1) soft-delete does not work, needs to be handled specifically from the ES side
2) relies on implicit index creation, no custom mapping or other index configuration is possible
3) TOAST columns aren't handled properly

### DIFF
--- a/flow/connectors/connelasticsearch/elasticsearch.go
+++ b/flow/connectors/connelasticsearch/elasticsearch.go
@@ -4,12 +4,14 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
 	"net/http"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/elastic/go-elasticsearch/v8"
@@ -104,12 +106,10 @@ func recordItemsProcessor(items model.RecordItems) ([]byte, error) {
 	qRecordJsonMap := make(map[string]any)
 
 	for key, val := range items.ColToVal {
-		switch r := val.(type) {
-		// JSON is stored as a string, fix that
-		case qvalue.QValueJSON:
-			qRecordJsonMap[key] = json.RawMessage(shared.
-				UnsafeFastStringToReadOnlyBytes(r.Val))
-		default:
+		if r, ok := val.(qvalue.QValueJSON); ok { // JSON is stored as a string, fix that
+			qRecordJsonMap[key] = json.RawMessage(
+				shared.UnsafeFastStringToReadOnlyBytes(r.Val))
+		} else {
 			qRecordJsonMap[key] = r.Value()
 		}
 	}
@@ -124,8 +124,8 @@ func (esc *ElasticsearchConnector) SyncRecords(ctx context.Context,
 	// atomics for counts will be unnecessary in other destinations, using a mutex instead
 	var recordCountsUpdateMutex sync.Mutex
 	// we're taking a mutex anyway, avoid atomic
-	var lastSeenLSN int64
-	var numRecords int64
+	var lastSeenLSN atomic.Int64
+	var numRecords atomic.Int64
 
 	// no I don't like this either
 	esBulkIndexerCache := make(map[string]esutil.BulkIndexer)
@@ -149,7 +149,7 @@ func (esc *ElasticsearchConnector) SyncRecords(ctx context.Context,
 
 	flushLoopDone := make(chan struct{})
 	// we only update lastSeenLSN in the OnSuccess call, so this should be safe even if race
-	// between loop breaking and sending on flushLoopDone
+	// between loop breaking and closing flushLoopDone
 	go func() {
 		ticker := time.NewTicker(peerdbenv.PeerDBQueueFlushTimeoutSeconds())
 		defer ticker.Stop()
@@ -161,7 +161,7 @@ func (esc *ElasticsearchConnector) SyncRecords(ctx context.Context,
 			case <-flushLoopDone:
 				return
 			case <-ticker.C:
-				lastSeen := lastSeenLSN
+				lastSeen := lastSeenLSN.Load()
 				if lastSeen > req.ConsumedOffset.Load() {
 					if err := esc.SetLastOffset(ctx, req.FlowJobName, lastSeen); err != nil {
 						esc.logger.Warn("[es] SetLastOffset error", slog.Any("error", err))
@@ -174,6 +174,7 @@ func (esc *ElasticsearchConnector) SyncRecords(ctx context.Context,
 		}
 	}()
 
+	var docId string
 	var bulkIndexFatalError error
 	var bulkIndexErrors []error
 	var bulkIndexOnFailureMutex sync.Mutex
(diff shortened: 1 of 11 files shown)


serprex:
How do you intend to converge qrep/cdc uuid issue?

flow/connectors/connelasticsearch/elasticsearch.go
```diff
@@ -78,101 +86,168 @@ func (esc *ElasticsearchConnector) Close() error {
...
-	schema := stream.Schema()
+	for key, val := range items.ColToVal {
+		switch r := val.(type) {
+		// JSON is stored as a string, fix that
+		case qvalue.QValueJSON:
+			qRecordJsonMap[key] = json.RawMessage(shared.
+				UnsafeFastStringToReadOnlyBytes(r.Val))
+		default:
+			qRecordJsonMap[key] = r.Value()
+		}
```
serprex, line 114:
```suggestion
		if r, ok := val.(type); ok { // JSON is stored as a string, fix that
			qRecordJsonMap[key] = json.RawMessage(shared.
				UnsafeFastStringToReadOnlyBytes(r.Val))
		} else {
			qRecordJsonMap[key] = r.Value()
		}
```

flow/connectors/connelasticsearch/elasticsearch.go
```diff
@@ -78,101 +86,168 @@ func (esc *ElasticsearchConnector) Close() error {
...
-			Body:       bytes.NewReader(qRecordJsonBytes),
+		err = bulkIndexer.Add(ctx, esutil.BulkIndexerItem{
+			Action:     action,
+			DocumentID: tablePkey.String(),
+			Body:       bytes.NewReader(bodyBytes),
 
+			OnSuccess: func(_ context.Context, _ esutil.BulkIndexerItem, _ esutil.BulkIndexerResponseItem) {
+				recordCountsUpdateMutex.Lock()
+				defer recordCountsUpdateMutex.Unlock()
+				record.PopulateCountMap(tableNameRowsMapping)
```
serprex, line 229: PopulateCountMap could be changed from `map[string]*struct{int,int,int}` to `map[string]struct{*atomicint,*atomicint,*atomicint}` & then no lock would be necessary by using atomics here

flow/connectors/connelasticsearch/elasticsearch.go
```diff
@@ -78,101 +86,168 @@ func (esc *ElasticsearchConnector) Close() error {
...
-		esc.logger.Error("[es] failed to initialize bulk indexer", slog.Any("error", err))
-		return 0, fmt.Errorf("[es] failed to initialize bulk indexer: %w", err)
-	}
-	defer func() {
-		if !bulkIndexerHasShutdown {
-			err := esBulkIndexer.Close(context.Background())
-			if err != nil {
-				esc.logger.Error("[es] failed to close bulk indexer", slog.Any("error", err))
+	flushLoopDone := make(chan struct{})
+	// we only update lastSeenLSN in the OnSuccess call, so this should be safe even if race
```
serprex, line 151: not safe, use atomics or have ticker.C case take lock

flow/connectors/connelasticsearch/elasticsearch.go
```diff
@@ -78,101 +86,168 @@ func (esc *ElasticsearchConnector) Close() error {
...
+	// between loop breaking and sending on flushLoopDone
```
serprex, line 152: Nothing ever sends on flushLoopDone

heavycrystal:
> How do you intend to converge qrep/cdc uuid issue?

converged in latest commit, single pkey/upsert col is used directly as document ID while multiple triggers SHA256 hash.

Snapshot updated to trigger workflows with upsert on the primary key columns of the tables

serprex (approved):

flow/connectors/connelasticsearch/elasticsearch.go
```diff
@@ -78,101 +88,178 @@ func (esc *ElasticsearchConnector) Close() error {
...
+				return nil, fmt.Errorf("[es] failed to process record: %w", err)
+			}
+			docId = fmt.Sprint(qValue.Value())
+		} else {
+			tablePkey, err := model.RecToTablePKey(req.TableNameSchemaMapping, record)
+			if err != nil {
+				esc.logger.Error("[es] failed to process record", slog.Any("error", err))
+				return nil, fmt.Errorf("[es] failed to process record: %w", err)
+			}
+			docId = hex.EncodeToString(tablePkey.PkeyColVal[:])
```
serprex, line 230: why not base64?

flow/connectors/connelasticsearch/qrep.go
```diff
@@ -0,0 +1,175 @@
...
+	return nil
+}
+
+func upsertKeyColsHash(qRecord []qvalue.QValue, upsertColIndices []int) string {
+	upsertColsMerged := make([][]byte, 0, len(upsertColIndices))
+
+	for _, upsertColIndex := range upsertColIndices {
+		upsertColsMerged = append(upsertColsMerged, []byte(fmt.Sprint(qRecord[upsertColIndex].Value())))
+	}
+	hashBytes := sha256.Sum256(slices.Concat(upsertColsMerged...))
```
serprex, line 36: https://pkg.go.dev/crypto/sha256#New returns `hash.Hash` which implements io.Writer so loop could use `hasher.Write(...)` instead of appending & at end `return hex.EncodeToString(hasher.Sum(nil))`

flow/connectors/connelasticsearch/elasticsearch.go
heavycrystal: there is a limit to size of the document ID, so won't work for really wide pkeys

serprex: right, but I'm suggesting using `base64.NoPadding.EncodeToString` instead of `hex.EncodeToString`, as base64 is a more compact encoding (6 bits per character) compared to hex (4 bits per character)

flow/e2e/elasticsearch/qrep_flow_es_test.go
```diff
@@ -55,7 +55,9 @@ func (s elasticsearchSuite) Test_Simple_Qrep() {
 	env := e2e.RunQRepFlowWorkflow(tc, qrepConfig)
 	e2e.EnvWaitForFinished(s.t, env, 3*time.Minute)
 	require.NoError(s.t, env.Error())
-	time.Sleep(2 * time.Second)
+	e2e.EnvWaitFor(s.t, env, 5*time.Second, "waiting for ES to catch up", func() bool {
+		return s.CountDocumentsInIndex(srcTableName) > 0
+	})
 
 	require.EqualValues(s.t, rowCount, s.CountDocumentsInIndex(srcTableName))
```
serprex, line 62:
```suggestion
	e2e.EnvWaitFor(s.t, env, 10*time.Second, "waiting for ES to catch up", func() bool {
		return s.CountDocumentsInIndex(srcTableName) == rowCount
	})
```

flow/connectors/connelasticsearch/elasticsearch.go
serprex: can leave as is for now

flow/connectors/connelasticsearch/elasticsearch.go
heavycrystal: makes sense, changed

PR #1653, opened 2024-04-29
Telemetry for peer: only televise non nil validate peer err

Amogh-Bharadwaj:
Log non flow error only if not nil for validate peer


flow/cmd/validate_peer.go
```diff
@@ -62,10 +62,10 @@ func (h *FlowRequestHandler) ValidatePeer(
 	if ok {
 		validErr := validationConn.ValidateCheck(ctx)
 		displayErr := fmt.Sprintf("failed to validate peer %s: %v", req.Peer.Name, validErr)
```
iamKunalGupta, line 64: we can move this in the if block as well

iamKunalGupta (approved):

serprex (approved):

PR #1666, opened 2024-05-03
Changes for Create Mirror

Amogh-Bharadwaj:
- Fix XMIN validation
- Change label to Create Watermark Table on Destination, and make it false by default
- Catch putting existing mirror name in UI

Screenshot 2024-05-03 at 8 10 21 PM

### DIFF
--- a/flow/cmd/validate_mirror.go
+++ b/flow/cmd/validate_mirror.go
@@ -6,11 +6,12 @@ import (
 	"fmt"
 	"log/slog"
 
+	"github.com/jackc/pgx/v5/pgtype"
+
 	connpostgres "github.com/PeerDB-io/peer-flow/connectors/postgres"
 	"github.com/PeerDB-io/peer-flow/connectors/utils"
 	"github.com/PeerDB-io/peer-flow/generated/protos"
 	"github.com/PeerDB-io/peer-flow/shared/telemetry"
-	"github.com/jackc/pgx/v5/pgtype"
 )
 
 func (h *FlowRequestHandler) ValidateCDCMirror(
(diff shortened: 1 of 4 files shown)


serprex (approved):

flow/cmd/validate_mirror.go
```diff
@@ -98,3 +117,13 @@ func (h *FlowRequestHandler) ValidateCDCMirror(
...
 }
+
+func (h *FlowRequestHandler) CheckIfMirrorNameExists(ctx context.Context, mirrorName string) (bool, error) {
+	var nameCount pgtype.Int4
+	err := h.pool.QueryRow(ctx, "SELECT COUNT(*) FROM flows WHERE name = $1", mirrorName).Scan(&nameCount)
+	if err != nil {
+		return true, fmt.Errorf("failed to check if mirror name exists: %v", err)
+	}
+
+	return nameCount.Int32 > 0, nil
```
serprex, line 128:
```suggestion
	var nameExists bool
	err := h.pool.QueryRow(ctx, "SELECT EXISTS(SELECT * FROM flows WHERE name = $1)", mirrorName).Scan(&nameExists)
	if err != nil {
		return true, fmt.Errorf("failed to check if mirror name exists: %v", err)
	}

	return nameExists, nil
```

ui/app/mirrors/create/helpers/qrep.ts
```diff
@@ -28,15 +28,15 @@ export const qrepSettings: MirrorSetting[] = [
     required: true,
   },
   {
-    label: 'Create Destination Table',
+    label: 'Create Watermark Table On Destination',
```
iskakaushik, line 31: watermark table is confusing right Amogh? that's why we wanted it as destination table.

Amogh-Bharadwaj: I'm wondering if users are thinking that PeerDB will create a table which will work with the query they give. This setting is more niche than it comes across. Maybe can introduce advanced settings for qrep and put this there

Amogh-Bharadwaj: Reverting this change for now. Can tackle it as a separate PR if needed

PR #1672, opened 2024-05-06
PeerDB: For queues, move sync interval to advanced, default 10 mins

Amogh-Bharadwaj:
For queue destination peers:
Moves sync interval field in create cdc mirror UI to advanced settings and defaults to 10 minutes.
Functionally tested


serprex (approved):

ui/app/mirrors/create/cdc/cdc.tsx
```diff
@@ -46,12 +46,37 @@ export default function CDCConfigForm({
...
+          IsQueuePeer(mirrorConfig.destination?.type) &&
+          setting.label === 'Sync Interval (Seconds)'
+        ) {
+          return;
+        }
+        if (setting.advanced != true) {
+          return setting;
+        }
+      })
+      .filter((setting) => setting !== undefined);
```
serprex, line 61:
```suggestion
      .filter((setting) =>
        (
          IsQueuePeer(mirrorConfig.destination?.type) &&
          setting.label === 'Sync Interval (Seconds)'
        ) || setting.advanced
      );
```
Maybe clearer to keep it as multiple statements instead of one expression, but no need for map/filter over a single filter

ui/app/mirrors/create/cdc/cdc.tsx
```diff
@@ -46,12 +46,37 @@ export default function CDCConfigForm({
   };
 
   const normalSettings = useMemo(() => {
-    return settings.filter((setting) => setting.advanced != true);
-  }, [settings]);
+    return settings!
+      .map((setting) => {
+        if (
+          IsQueuePeer(mirrorConfig.destination?.type) &&
+          setting.label === 'Sync Interval (Seconds)'
```
serprex, line 53: rather than hardcoding it here, spec could be `advanced: 'queue'` but maybe makes types annoying

ui/app/mirrors/create/helpers/common.ts
```diff
@@ -1,5 +1,6 @@
 import { FlowConnectionConfigs, TypeSystem } from '@/grpc_generated/flow';
 
+export type AdvancedSettingType = boolean | 'queue';
```
heavycrystal, line 3: make this an enum instead
```
export enum AdvancedSettingType {
  NONE = 0,
  QUEUE_ONLY = 1,
  ALL = 2
}
```

PR #1675, opened 2024-05-06
kafka initial load

serprex:
TODO: other queues
TODO: share code
TODO: make populatecountmap atomic

Turns out waitgroup unnecessary since kgo flush handles wait


flow/connectors/kafka/qrep.go
```diff
@@ -0,0 +1,111 @@
...
+	ctx context.Context,
+	config *protos.QRepConfig,
+	partition *protos.QRepPartition,
+	stream *model.QRecordStream,
+) (int, error) {
+	startTime := time.Now()
+	numRecords := atomic.Int64{}
+	schema := stream.Schema()
+
+	wgCtx, wgErr := context.WithCancelCause(ctx)
```
iskakaushik, line 31: same here about the wg comment.

flow/connectors/kafka/qrep.go
```diff
@@ -0,0 +1,111 @@
...
+	pool, err := c.createPool(wgCtx, config.Script, config.FlowJobName, wgErr)
+	if err != nil {
+		return 0, err
+	}
+	defer pool.Close()
+
+Loop:
+	for {
+		select {
+		case qrecord, ok := <-stream.Records:
```
iskakaushik, line 41: seems like a lot of the code below is shared between pub-sub and Kafka, can we unify it?

flow/connectors/kafka/qrep.go
```diff
@@ -0,0 +1,111 @@
...
+				}
+				record := &model.InsertRecord[model.RecordItems]{
+					BaseRecord:           model.BaseRecord{},
+					Items:                items,
+					SourceTableName:      config.WatermarkTable,
+					DestinationTableName: config.DestinationTableIdentifier,
+					CommitID:             0,
+				}
+
+				lfn := ls.Env.RawGetString("onRecord")
```
iskakaushik, line 60: doesn't cdc have to do something similar as well?

flow/connectors/kafka/qrep.go
serprex: Also share quite a bit between cdc/qrep. It'll be a mess of conditions. Would be better as an isolated PR

flow/connectors/kafka/qrep.go
serprex: Yes, but code is mixed up with tracking LSN/counts

iskakaushik:
makes sense, might be worth following up with a PR on cleaning some of this up :)

iskakaushik (approved):

flow/connectors/kafka/kafka.go
```diff
@@ -166,42 +165,48 @@ func lvalueToKafkaRecord(ls *lua.LState, value lua.LValue) (*kgo.Record, error)
...
+			c.client.Produce(ctx, kr, produceCb)
 		}
 	})
+}
+
+func (c *KafkaConnector) SyncRecords(ctx context.Context, req *model.SyncRecordsRequest[model.RecordItems]) (*model.SyncResponse, error) {
+	numRecords := atomic.Int64{}
+	tableNameRowsMapping := utils.InitialiseTableRowsMap(req.TableMappings)
+
+	wgCtx, wgErr := context.WithCancelCause(ctx)
```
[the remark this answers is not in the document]
serprex: renamed wg to queue, wasn't really sure what to call

PR #1679, opened 2024-05-07
BigQuery: Improve intermediate Avro methods

Amogh-Bharadwaj:
- Refactors `DefineAvroSchema` to now use the source schema instead of the target BigQuery table schema. This is consistent with Snowflake and Clickhouse, albeit not using the single method they both share - this is due to some type behaviour differences in BQ (Ex: timestamps).
- Refactors `qValueKindToBigQueryType`'s return body to contain array (repeated) information, instead of relying on the assumption that source arrays are mapped to BQ arrays

### DIFF
--- a/flow/connectors/bigquery/qrep_avro_sync.go
+++ b/flow/connectors/bigquery/qrep_avro_sync.go
@@ -49,8 +49,9 @@ func (s *QRepAvroSyncMethod) SyncRecords(
 	s.connector.logger.Info(
 		fmt.Sprintf("Obtaining Avro schema for destination table %s and sync batch ID %d",
 			rawTableName, syncBatchID))
+	sourceSchema := stream.Schema()
 	// You will need to define your Avro schema as a string
-	avroSchema, err := DefineAvroSchema(rawTableName, dstTableMetadata, "", "")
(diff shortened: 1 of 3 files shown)


flow/connectors/bigquery/qrep_avro_sync.go
```diff
@@ -280,9 +283,9 @@ func DefineAvroSchema(dstTableName string,
 	}, nil
 }
 
-func GetAvroType(bqField *bigquery.FieldSchema) (interface{}, error) {
-	avroNumericPrecision := int16(bqField.Precision)
-	avroNumericScale := int16(bqField.Scale)
+func GetAvroType(sourceField *qvalue.QField) (interface{}, error) {
+	avroNumericPrecision := int16(sourceField.Precision)
+	avroNumericScale := int16(sourceField.Scale)
```
serprex, line 288:
```suggestion
func GetAvroType(sourceField *qvalue.QField) (interface{}, error) {
	avroNumericPrecision := sourceField.Precision
	avroNumericScale := sourceField.Scale
```

serprex (approved):

serprex:
Does this fix any existing issues?

Amogh-Bharadwaj:
> Does this fix any existing issues?

No, but it allows for data type mapping configuration for bigquery

PR #1682, opened 2024-05-07
PopulateCountMap: use atomics

serprex:
Race existed with parallel queue processing


serprex:
Alternatively, we could adjust populate to happen at start of loop, counting 1 per incoming record, rather than 1 per outgoing record

iskakaushik (approved):

PR #1684, opened 2024-05-07
Better handling for context cancelation while writing to avro

iskakaushik:


flow/connectors/utils/avro/avro_writer.go
```diff
@@ -131,23 +131,29 @@ func (p *peerDBOCFWriter) writeRecordsToOCFWriter(ctx context.Context, ocfWriter
 
 	numRows := 0
 	for qrecord := range p.stream.Records {
-		avroMap, err := avroConverter.Convert(qrecord)
-		if err != nil {
-			logger.Error("failed to convert QRecord to Avro compatible map: ", slog.Any("error", err))
-			return 0, fmt.Errorf("failed to convert QRecord to Avro compatible map: %w", err)
-		}
+		select {
```
heavycrystal, line 134: I think can be better expressed by ensuring context cancellation closes this channel from upstream

serprex, line 134: prefer
```go
if err := ctx.Err(); err != nil {
  return numRows, err
} else {
  // default
}
```

serprex (approved):

PR #1687, opened 2024-05-07
PG to PG: TSTZRange and JSONB

Amogh-Bharadwaj:
PG to PG CDC:
- Adds support for JSONB (now maps to JSONB on target)
- Constructs the string for tstzrange (still text on target) in a way that you can cast to tstzrange like select mytstzrange::tstzrange from table. This would fail before as the string we were inserting was of some json type string. Null lower and upper bounds tested

Functionally tested

### DIFF
--- a/flow/connectors/postgres/qvalue_convert.go
+++ b/flow/connectors/postgres/qvalue_convert.go
@@ -283,12 +283,12 @@ func parseFieldFromQValueKind(qvalueKind qvalue.QValueKind, value interface{}) (
 		tstzrangeObject := value.(pgtype.Range[interface{}])
 		lowerBoundType := tstzrangeObject.LowerType
 		upperBoundType := tstzrangeObject.UpperType
-		lowerTime, err := ConvertTimeRangeBounds(tstzrangeObject.Lower)
+		lowerTime, err := convertTimeRangeBounds(tstzrangeObject.Lower)
 		if err != nil {
 			return nil, fmt.Errorf("[tstzrange]error for lower time bound: %v", err)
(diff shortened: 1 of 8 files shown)


iskakaushik (approved):

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -481,3 +512,23 @@ func customTypeToQKind(typeName string) qvalue.QValueKind {
 		return qvalue.QValueKindString
 	}
 }
+
+// Postgres does not like timestamps of the form 2006-01-02 15:04:05 +0000 UTC
+// in tstzrange.
+// ConvertTimeRangeBounds removes the +0000 UTC part
+func ConvertTimeRangeBounds(timeBound interface{}) (string, error) {
```
heavycrystal, line 519: unexport

serprex:
What's the motivation for this? PG<>PG should use PG type system to have this accuracy & more

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -108,6 +110,8 @@ func (c *PostgresConnector) postgresOIDToQValueKind(recvOID uint32) qvalue.QValu
 		return qvalue.QValueKindArrayString
 	case pgtype.IntervalOID:
 		return qvalue.QValueKindInterval
+	case pgtype.TstzrangeOID:
```
serprex, line 113: should also handle tsrange at least

serprex:
#1733 has merged, making this obsolete

PR #1704, opened 2024-05-08
Query layer: support ssh fields, sync interval, simplify code path

Amogh-Bharadwaj:
- Adds parsing for SSHConfig fields in Create Peer command for Postgres
```sql
CREATE PEER postgres_peer FROM POSTGRES WITH
(
    host = '<hostname>',
    port = 5432,
    user = '<user>',
    password = '<password>',
    database = '<dbname>',
    ssh_config = '{
        "host": "<ssh_host>",
        "port": 22,
        "user": "<ssh_user>",
        "password": "<ssh_password>",
        "private_key": "<ssh_private>"
    }'
);
```

- Also you can now specify `sync_interval` in the create mirror command

- Remove catalog flows table entry and update operations in query layer side, as we are anyways hitting a grpc endpoint which takes care of that. This makes the create_catalog_entry flag in the request to create/validate mirror endpoints redundant, so that is removed from route.proto

- Fixes a bug in validate mirror where select * from sourcetables limit 0 was not checked for empty publication

Functionally tested via query layer and UI


serprex (approved):

serprex (approved):

flow/cmd/validate_mirror.go
```diff
@@ -46,7 +46,9 @@ func (h *FlowRequestHandler) ValidateCDCMirror(
 	sourcePeerConfig := req.ConnectionConfigs.Source.GetPostgresConfig()
 	if sourcePeerConfig == nil {
 		slog.Error("/validatecdc source peer config is nil", slog.Any("peer", req.ConnectionConfigs.Source))
-		return nil, errors.New("source peer config is nil")
+		return &protos.ValidateCDCMirrorResponse{
```
serprex, line 49: this looks like something easy to screw up in future, ideally it'd be a model that returns by value so zero value is correct response. should try make code work with errors returning nil

PR #1709, opened 2024-05-10
kafka/pubsub: fix LSN potentially being updated too early

serprex:
LSN should not be updated before success confirmed,
as intermediate value may now be read before error aborts when queue is flushed

With parallelism, lsn should be updated in critical section so that if earlier invocation is lagging behind LSN doesn't skip

Also fix pubsub needing EnableMessageOrdering explicitly enabled


iskakaushik (approved):
@serprex lets land this ASAP, but quickly follow-up with a test.

flow/connectors/kafka/kafka.go
```diff
@@ -194,25 +194,40 @@ func (c *KafkaConnector) createPool(
...
+			}
+		} else {
+			recordCounter := atomic.Int32{}
+			recordCounter.Store(lenRecords)
+			for _, kr := range result.records {
+				c.client.Produce(ctx, kr, func(_ *kgo.Record, err error) {
+					if err != nil {
+						queueErr(err)
+					} else if recordCounter.Add(-1) == 0 && lastSeenLSN != nil {
+						shared.AtomicInt64Max(lastSeenLSN, result.lsn)
```
serprex, line 211: relying on Produce callbacks executing sequentially

serprex:
> @serprex lets land this ASAP, but quickly follow-up with a test.

testing races is beyond our current testing infra's capabilities

PR #1710, opened 2024-05-10
Move to alpine to reduce vulns

iskakaushik:
reduces the number of high and medium vulns to zero.


serprex (approved):

serprex:
should also update ui to use alpine

PR #1711, opened 2024-05-13
Flow API: Custom resume endpoint, polished mirror status endpoint, auth to all endpoints

Amogh-Bharadwaj:
## Endpoints:
Custom sync:
```
/v1/flows/cdc/sync
```

Status:
```
/v1/mirrors/{flow_job_name}?include_flow_info=true|false
```

## Authentication
Flow API now has authentication on all routes (except for health). It uses Bearer Authentication with token being:
`base64Encode(bcryptHash(PEERDB_PASSWORD))`

It can be generated on Unix based systems via:
`htpasswd -nbB -C 10 "" <PEERDB_PASSWORD_HERE>  | tr -d ":"  | base64`

## API Endpoint to Resume Mirror With Specified Syncs
This PR introduces an endpoint in our Flow API to resume mirrors with a request of custom number of syncs. 
- The number of syncs the client can request is, for now, a number between 1 and the `MaxSyncsPerCDCFlow` which is currently 32.
- Syncs in this case refers to one sync flow - normalise flow pair.
- This endpoint requires the mirror to be paused before hand.
- Request validation checks the above along with whether the mirror name exists, etc.

### Implementation:
Extends `CDCFlowConfigUpdate` to include number of syncs information.

```
message CDCFlowConfigUpdate {
  repeated TableMapping additional_tables = 1;
  uint32 batch_size = 2;
  uint64 idle_timeout = 3;
  int32 number_of_syncs = 4;
}
```
- This is picked up in CDC flow after processing FlowConfigUpdate:
```Golang
if state.FlowConfigUpdate != nil {
	err = processCDCFlowConfigUpdate(ctx, logger, cfg, state, mirrorNameSearch)
        if err != nil {
		return state, err
	}
	syncCountLimit = int(state.SyncFlowOptions.NumberOfSyncs)
```
- Limit check is done when listening for normalise flow finishing:
```Golang
normDoneChan.AddToSelector(mainLoopSelector, func(x struct{}, _ bool) {
	if syncCount == syncCountLimit {
		logger.Info("sync count limit reached, pausing",
		slog.Int("limit", syncCountLimit),
		slog.Int("count", syncCount))
		state.ActiveSignal = model.PauseSignal
	}

```

### Functional testing:
- Tested with a paused mirror with low sync interval via HTTP requests to this endpoint
- Resume mirror via the normal UI button route will stick to existing behaviour of MaxSyncsPerCDCFlow=32 sync flows.

Screenshot 2024-05-10 at 5 25 18 PM
Screenshot 2024-05-10 at 5 25 42 PM
Screenshot 2024-05-10 at 5 27 36 PM

## Mirror Status Endpoint
Existing mirror status endpoint now has a flag in the request to include flow information. The UI sets this flag. If not set, response only contains flow name and running status, along with clean error messages.

This endpoint is also gated by authorization

Functionally tested


flow/workflows/cdc_flow.go
```diff
@@ -221,11 +220,17 @@ func CDCFlowWorkflow(
 		shared.MirrorNameSearchAttribute: cfg.FlowJobName,
 	}
 
+	var syncCountLimit int
```
serprex, line 223: why temporal state instead of property on config where when 0 it doesn't pause after N batches?

Then user only needs to keep invoking `RESUME MIRROR` after setting value

serprex: it'd be in the edit mirror dialog

flow/connectors/core.go
```diff
@@ -224,6 +224,9 @@ type RenameTablesConnector interface {
 }
 
 func GetConnector(ctx context.Context, config *protos.Peer) (Connector, error) {
+	if config == nil {
```
serprex, line 227: why are we passing nil config?

serprex (approved):

flow/connectors/core.go
Amogh-Bharadwaj: not required

serprex: should remove, bugs should be panics not errors

Amogh-Bharadwaj:
Description updated with new status endpoint

protos/route.proto
```diff
@@ -303,6 +330,9 @@ service FlowService {
   rpc MirrorStatus(MirrorStatusRequest) returns (MirrorStatusResponse) {
     option (google.api.http) = { get: "/v1/mirrors/{flow_job_name}" };
   }
+  rpc CDCMirrorStatus(MirrorStatusRequest) returns (CDCStatusResponse) {
```
heavycrystal, line 333: this is unnecessary, fold additional parameters into existing `MirrorStatus` route

protos/route.proto
```diff
@@ -267,6 +287,12 @@ service FlowService {
       body: "*"
      };
   }
+  rpc CustomSyncFlow(CreateCustomFlowRequest) returns (CreateCustomFlowResponse) {
+    option (google.api.http) = {
+      post: "/v1/flows/cdc/sync",
```
heavycrystal, line 292: nit: name and protos aren't matching, change one or the other

flow/cmd/custom_sync.go
```diff
@@ -0,0 +1,171 @@
...
+	peerflow "github.com/PeerDB-io/peer-flow/workflows"
+
+	"crypto/sha256"
+
+	"google.golang.org/grpc/metadata"
+)
+
+const peerdbPauseGuideDocLink = "https://docs.peerdb.io/features/pause-mirror"
+
+func AuthenticateSyncRequest(ctx context.Context) error {
```
heavycrystal, line 21: should be implemented using gRPC middleware ideally: https://shijuvar.medium.com/writing-grpc-interceptors-in-go-bf3e7671fe48

flow/cmd/custom_sync.go
```diff
@@ -0,0 +1,171 @@
...
+		values = md.Get("authorization")
+	}
+
+	if len(values) > 0 {
+		token = values[0]
+	}
+
+	peerdbPassword := peerdbenv.PeerDBPassword()
+	hash := sha256.New()
+	_, err := hash.Write([]byte(peerdbPassword))
```
heavycrystal, line 36: nit: this never errors so we can ignore err https://pkg.go.dev/hash#Hash

flow/cmd/custom_sync.go
```diff
@@ -0,0 +1,171 @@
...
+	if ok {
+		values = md.Get("authorization")
+	}
+
+	if len(values) > 0 {
+		token = values[0]
+	}
+
+	peerdbPassword := peerdbenv.PeerDBPassword()
+	hash := sha256.New()
```
heavycrystal, line 35: can be memoized

flow/cmd/custom_sync.go
```diff
@@ -0,0 +1,171 @@
...
+	_, err := hash.Write([]byte(peerdbPassword))
+	if err != nil {
+		slog.Error("Server error: unable to verify authorization", slog.Any("error", err))
+		return errors.New("server error: unable to verify authorization. Please try again.")
+	}
+
+	passwordHashed := hex.EncodeToString(hash.Sum(nil))
+	if token != "Bearer "+passwordHashed {
+		slog.Error("Unauthorized: invalid authorization token", slog.String("token", token),
+			slog.String("expected_token", "Bearer "+passwordHashed))
```
heavycrystal, line 45: should we be logging the expected token

protos/flow.proto
```diff
@@ -371,6 +372,7 @@ message CDCFlowConfigUpdate {
   repeated TableMapping additional_tables = 1;
   uint32 batch_size = 2;
   uint64 idle_timeout = 3;
+  int32 number_of_syncs = 4;
```
heavycrystal, line 375: any reason this is signed, same below

protos/route.proto
Amogh-Bharadwaj: MirrorStatus route includes peer configuration information, mirror configuration information etc (UI needs) in its response which isn't needed for a status endpoint, and probably not ideal to send this info over the air from a public endpoint. It would also be slower. cc @iskakaushik

protos/flow.proto
Amogh-Bharadwaj: int32 is smallest primitive and number of syncs is <= 32 anyways

flow/cmd/custom_sync.go
Amogh-Bharadwaj: we will be exposing these as http endpoints as well

flow/cmd/custom_sync.go
Amogh-Bharadwaj: removed

protos/route.proto
heavycrystal: no route is 'public', all routes should either be unexposed and/or authenticated
you can add fields to the request objects to return less parameters if needed, but 2 routes for similar purposes will make our API unwieldly

serprex: +1 to consolidating endpoint, long term we should be sending as little peer config info back & forth

Amogh-Bharadwaj:
Updated description and functionally tested post changes

heavycrystal (approved):

docker-compose-dev.yml
```diff
@@ -131,6 +131,7 @@ services:
       - 8113:8113
     environment:
       <<: [*catalog-config, *flow-worker-env, *minio-config]
+      PEERDB_PASSWORD: peerdb
```
iskakaushik, line 134: does this need to change in `peerdb-ui` as well?

flow/middleware/auth.go
```diff
@@ -0,0 +1,45 @@
...
+	md, _ := metadata.FromIncomingContext(ctx)
+	if len(md["authorization"]) == 0 {
+		return nil, status.Errorf(codes.Unauthenticated, "Authorization token is required")
+	}
+	headerValue := md["authorization"][0]
+	bearerPrefix := "Bearer "
+	if !strings.HasPrefix(headerValue, bearerPrefix) {
+		return nil, status.Errorf(codes.Unauthenticated, "Unsupported authorization type")
+	}
+	base64Token := strings.TrimPrefix(headerValue, bearerPrefix)
```
serprex, line 28: use https://pkg.go.dev/strings#CutPrefix

flow/middleware/auth.go
```diff
@@ -0,0 +1,45 @@
...
+		return nil, status.Errorf(codes.Unauthenticated, "Authorization token is required")
+	}
+	// Always a good practice to have the actual token in base64
+	tokenBytes, err := base64.StdEncoding.DecodeString(base64Token)
+	if err != nil {
+		logger.LoggerFromCtx(ctx).Warn("Error decoding token", slog.String("token", base64Token), slog.Any("error", err))
+		return nil, status.Errorf(codes.Unauthenticated, "Authentication failed")
+	}
+	token := string(tokenBytes)
+	err = bcrypt.CompareHashAndPassword([]byte(token), []byte(plaintext))
```
serprex, line 39:
```suggestion
	err = bcrypt.CompareHashAndPassword(tokenBytes, []byte(plaintext))
```

flow/middleware/auth.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		logger.LoggerFromCtx(ctx).Warn("Error validating token", slog.String("token", token), slog.Any("error", err))
```
serprex, line 41:
```suggestion
		logger.LoggerFromCtx(ctx).Warn("Error validating token", slog.String("token", string(tokenBytes)), slog.Any("error", err))
```

flow/middleware/interceptors.go
```diff
@@ -0,0 +1,82 @@
...
+	"google.golang.org/grpc/status"
+
+	"github.com/PeerDB-io/peer-flow/logger"
+)
+
+func CreateAuthServerInterceptor(ctx context.Context, plaintext string, unauthenticatedMethods []string) grpc.UnaryServerInterceptor {
+	if plaintext == "" {
+		logger.LoggerFromCtx(ctx).Warn("Authentication is disabled")
+		//nolint:nonamedreturns
+		return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp any, err error) {
```
serprex, line 19:
```suggestion
		return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
```

flow/middleware/interceptors.go
```diff
@@ -0,0 +1,82 @@
...
+			return handler(ctx, req)
+		}
+	}
+	unauthenticatedMethodsMap := make(map[string]bool)
+	for _, method := range unauthenticatedMethods {
+		unauthenticatedMethodsMap[method] = true
+	}
+	//nolint:nonamedreturns
+	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp any, err error) {
```
serprex, line 28:
```suggestion
	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
```

ui/package.json
```diff
@@ -32,6 +32,7 @@
     "@types/node": "^20.12.12",
     "@types/react": "^18.3.3",
     "@types/react-dom": "^18.3.0",
+    "axios": "^1.7.2",
```
serprex, line 35: No need for axios, just use fetch

flow/middleware/interceptors.go
```diff
@@ -0,0 +1,82 @@
...
+		logger.LoggerFromCtx(ctx).Warn("Authentication is disabled")
+		//nolint:nonamedreturns
+		return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp any, err error) {
+			return handler(ctx, req)
+		}
+	}
+	unauthenticatedMethodsMap := make(map[string]bool)
+	for _, method := range unauthenticatedMethods {
+		unauthenticatedMethodsMap[method] = true
+	}
```
serprex, line 26:
```suggestion
	unauthenticatedMethodsMap := make(map[string]struct{})
	for _, method := range unauthenticatedMethods {
		unauthenticatedMethodsMap[method] = struct{}{}
	}
```

ui/package.json
iamKunalGupta: Using it because now creds need to be injected in every request to flow-api, so just created a new instance of axios with relevant headers. Check `ui/rpc/http.ts`

serprex: np, make a light wrapper around fetch

serprex:
testing on my cloud setup, drop mirror not working

flow/middleware/interceptors.go
```diff
@@ -2,37 +2,51 @@ package middleware
...
+		return nil, fmt.Errorf("error generating hash: %w", err)
+	}
+
 	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
 		if _, ok := unauthenticatedMethodsSet[info.FullMethod]; ok {
 			return handler(ctx, req)
 		}
-		ctx, err := Authorize(ctx, plaintext)
+		ctx, err := authorize(ctx, hashCache)
 		if err != nil {
```
iamKunalGupta, line 45: Why do we need to look at live password changes?
Why not send the hash directly?

iamKunalGupta (changes requested):

flow/middleware/auth.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	// Always a good practice to have the actual token in base64
+	tokenBytes, err := base64.StdEncoding.DecodeString(base64Token)
+	if err != nil {
+		logger.LoggerFromCtx(ctx).Warn("Error decoding token", slog.String("token", base64Token), slog.Any("error", err))
+		return nil, status.Errorf(codes.Unauthenticated, "Authentication failed")
+	}
+
+	if err := bcrypt.CompareHashAndPassword(hash, tokenBytes); err != nil {
+		logger.LoggerFromCtx(ctx).Warn("Error validating token", slog.String("token", string(tokenBytes)), slog.Any("error", err))
```
iamKunalGupta, line 38: remove token from logging now since it is plaintext now.

iamKunalGupta (commented):
Can you check dropmirror flow once @heavycrystal ? @serprex said that it was not working

PR #1712, opened 2024-05-14
v2

serprex:
TODO test
TODO make a config option or detect PG version
TODO toast
TODO don't persist in catalog
TODO move in memory buffer to cdcStore for disk spill


serprex:
#4300

PR #1720, opened 2024-05-14
replace infinite time values from postgres with NULL

serprex:


iskakaushik (approved):

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -247,11 +247,17 @@ func parseFieldFromQValueKind(qvalueKind qvalue.QValueKind, value interface{}) (
 
 	switch qvalueKind {
 	case qvalue.QValueKindTimestamp:
-		timestamp := value.(time.Time)
-		return qvalue.QValueTimestamp{Val: timestamp}, nil
+		if timestamp, ok := value.(time.Time); ok {
+			return qvalue.QValueTimestamp{Val: timestamp}, nil
+		} else { // postgres supports infinite times
+			return qvalue.QValueNull(qvalue.QValueKindTimestamp), nil
```
iamKunalGupta, line 253: Maybe we should explicity check if it is infinite and only then null. There might be other types we missed similar to the infinite?
If not probably log at the very least.

serprex: we don't log when nulling out other values like infinite decimals. Explicit infinity check makes sense

PR #1722, opened 2024-05-15
fix: duplicate metrics pushed via otel sdk

iamKunalGupta:
This is accomplished by clearing out the metric from the MetricsMap once Observed/Collected.

Use a query like:
```
max(peerdb_cdc_slot_lag_MB{}) without(k8s_pod_name, k8s_pod_uid, k8s_pod_ip, k8s_pod_start_time, k8s_node_name)
```
to deduplicate metrics across pods, they are only pushed once in 1 cycle in a pod

TODO:
- [x] Enable GRPC exporter
- [x] Update dockerfile env vars:
  - [x] Set a default export interval (?? in the docker file ??)
  - [x] Set default compression
- [x] Revert cron back to 5 minutes
- [x] Open replication connections

### DIFF
--- a/flow/workflows/scheduled_flows.go
+++ b/flow/workflows/scheduled_flows.go
@@ -62,7 +62,8 @@ func GlobalScheduleManagerWorkflow(ctx workflow.Context) error {
 
 	slotSizeCtx := withCronOptions(ctx,
 		"record-slot-size-"+info.OriginalRunID,
-		"*/5 * * * *")
+		// TODO revert this back to 5 minutes?
+		"*/1 * * * *")
 	workflow.ExecuteChildWorkflow(slotSizeCtx, RecordSlotSizeWorkflow)
 
 	ctx.Done().Receive(ctx, nil)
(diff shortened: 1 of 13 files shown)


flow/workflows/scheduled_flows.go
```diff
@@ -62,7 +62,8 @@ func GlobalScheduleManagerWorkflow(ctx workflow.Context) error {
 
 	slotSizeCtx := withCronOptions(ctx,
 		"record-slot-size-"+info.OriginalRunID,
-		"*/5 * * * *")
+		// TODO revert this back to 5 minutes?
```
serprex, line 65: TODO

iamKunalGupta: On a side note @serprex should we let it be 1 minute?

serprex: prefer to leave it same for now, changing should be separate benchmark, maybe env variable

serprex (approved):

PR #1725, opened 2024-05-15
wal heartbeat: allow overriding query with env variable

serprex:
To enable heartbeating, set these two environment vars:

```yaml
  PEERDB_WAL_HEARTBEAT_QUERY: ${PEERDB_WAL_HEARTBEAT_QUERY:-UPDATE _peerdb_heartbeat SET last_heartbeat = now() WHERE id = 1;}
  PEERDB_ENABLE_WAL_HEARTBEAT: ${PEERDB_ENABLE_WAL_HEARTBEAT:-true}
```


Amogh-Bharadwaj:
Should we also make the frequency of heartbeat an env ? @serprex @iskakaushik

serprex:
I think that would be better as a followup PR, has to touch at temporal cron setup

Amogh-Bharadwaj:
Functionally tested, updated description

Amogh-Bharadwaj (approved):

PR #1733, opened 2024-05-17
qrep pgvalue

serprex:
Ends up being pretty different from cdc pgvalue, but makes sense since pgvalue isn't a type so a much as an idea that intermediate formats should be discarded
Instead of using channels & cursors, qrep pgvalue pipes COPY TO STDOUT from source into COPY FROM STDIN on destination
Values transmit as text protocol for maximum compatibility for now. Future PRs may add options, current PR is large enough already


serprex:
postgis is failing in decode/encode mix up, https://github.com/jackc/pgx/issues/867 outlines using lower level pgconn api to pipe `copy to` into `copy from`

flow/activities/flowable_core.go
```diff
@@ -292,27 +291,38 @@ func (a *FlowableActivity) getPostgresPeerConfigs(ctx context.Context) ([]*proto
 }
 
 // replicateQRepPartition replicates a QRepPartition from the source to the destination.
-func (a *FlowableActivity) replicateQRepPartition(ctx context.Context,
+func replicateQRepPartition[TRead any, TWrite any, TSync connectors.QRepSyncConnectorCore, TPull connectors.QRepPullConnectorCore](
```
heavycrystal, line 294: function signature is a bit complex
a) use a struct to simplify some inputs
b) idx and total are only being used for heartbeat, make it caller side

flow/connectors/postgres/sink.go
```diff
@@ -0,0 +1,239 @@
...
+			str = QuoteLiteral(arg)
+		case time.Time:
+			str = arg.Truncate(time.Microsecond).Format("'2006-01-02 15:04:05.999999999Z07:00:00'")
+		default:
+			return 0, fmt.Errorf("invalid arg type: %T", arg)
+		}
+		query = strings.ReplaceAll(query, fmt.Sprintf("$%d", i+1), str)
+	}
+
+	copyQuery := fmt.Sprintf("COPY (%s) TO STDOUT", query)
```
heavycrystal, line 198: no fetch size here as with the QValue type system, not sure if can lead to issues if sync isn't fast enough

flow/connectors/postgres/sink.go
```diff
@@ -0,0 +1,239 @@
...
+func (p PgCopyReader) GetColumnNames() []string {
+	return p.schema.schema
+}
+
+func (p PgCopyReader) CopyInto(ctx context.Context, c *PostgresConnector, tx pgx.Tx, table pgx.Identifier) (int64, error) {
+	<-p.schema.schemaLatch
+	ct, err := c.conn.PgConn().CopyFrom(
+		ctx,
+		p.PipeReader,
+		fmt.Sprintf("COPY %s (%s) FROM STDIN", table.Sanitize(), strings.Join(p.schema.schema, ",")),
```
heavycrystal, line 236: no chance of `BINARY` copy?

flow/connectors/postgres/sink.go
serprex: if sync is slow writes will block, as they should

will want to make a general benchmark at some point, even if it's only to have some numbers for a blog post

flow/connectors/postgres/sink.go
serprex: Could do it, but I don't expect much benefit. Especially if compression is enabled. Binary has potential compatibility issues, would want to have a bool option to enable it

See https://github.com/antirez/RESP3/blob/master/spec.md where antirez outlines text protocols being fine for wire protocol. Of course if a customer has a bunch of binary blobs they want replicated we'll have trouble then. & uuid columns fall in that category

flow/activities/flowable_core.go
serprex: unfortunately go makes generic function type aliases not a thing. moving heartbeats out would be good. have heartbeat cleanup happening in another pr, #1742, so maybe after as a followup to that

flow/connectors/postgres/sink.go
heavycrystal: like the idea of being an option, file an issue for that to be revisited

flow/activities/flowable.go
```diff
@@ -428,9 +430,41 @@ func (a *FlowableActivity) ReplicateQRepPartitions(ctx context.Context,
...
 		partitions.BatchId, numPartitions),
 	)
-	for i, p := range partitions.Partitions {
+	for _, p := range partitions.Partitions {
 		logger.Info(fmt.Sprintf("batch-%d - replicating partition - %s", partitions.BatchId, p.PartitionId))
-		err := a.replicateQRepPartition(ctx, config, i+1, numPartitions, p, runUUID)
+		var err error
+		switch config.System {
+		case protos.TypeSystem_Q:
+
```
iskakaushik, line 438: nit: empty new line

flow/activities/flowable.go
```diff
@@ -428,9 +430,41 @@ func (a *FlowableActivity) ReplicateQRepPartitions(ctx context.Context,
...
+				}))
+				if err != nil {
+					a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
+					return err
+				}
+				lfn := ls.Env.RawGetString("transformRow")
+				if fn, ok := lfn.(*lua.LFunction); ok {
+					outstream = pua.AttachToStream(ls, fn, stream)
+				}
+			}
```
iskakaushik, line 453: i'm guessing this is done at other places as well? for cdc as well, can we have something like `scripting.LoadLfn` or some unified method?

serprex: Not sure what you're thinking since I already put most of it into utils.LoadScript

PR #1735, opened 2024-05-20
fix: close and recreate destination connections during syncflow when needed

iamKunalGupta:


serprex:
why is this needed?

serprex:
Better would be to lazily open connection after constructor called. First dstConn is only looking up data from catalog

iamKunalGupta:
> Better would be to lazily open connection after constructor called. First dstConn is only looking up data from catalog

That also makes sense

serprex (approved):

serprex (approved):

PR #1737, opened 2024-05-20
Heartbeats for qrep kafka

iskakaushik:


Amogh-Bharadwaj (approved):

serprex:
this should be handled by the activity instead of connectors

PR #1741, opened 2024-05-21
feat: Implement OpenTelemetry tracing

iamKunalGupta:
- Add basic telemetry and examples
- Integrate with temporal's opentelemetry interceptor

To run/test:
1. Run jaeger:
   ```shell
   docker run --rm --name jaeger -p 16686:16686 -p 4317:4317 -e JAEGER_REPORTER_LOG_SPANS=true -e JAEGER_SAMPLER_TYPE=const -e JAEGER_SAMPLER_PARAM=1 -e COLLECTOR_OTLP_ENABLED=true jaegertracing/all-in-one:1.57
   ```
   Open jaeger ui at http://localhost:16686/
2. Run flow worker with the following env:
   ```shell
   ENABLE_OTEL_TRACING=true
   OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc
   OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4317
   ```
Start any activity etc and we should start seeing traces like below:
image


flow/cmd/worker_common.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"go.temporal.io/sdk/contrib/opentelemetry"
+	"go.temporal.io/sdk/interceptor"
+)
+
+func GetTemporalClientInterceptors() []interceptor.ClientInterceptor {
+	// Maybe do this only when otel is actually enabled?
+	tracingInterceptor, err := opentelemetry.NewTracingInterceptor(opentelemetry.TracerOptions{})
+	if err != nil {
+		log.Printf("failed to create tracing interceptor: %v\n", err)
```
serprex, line 14: return nil instead of slice containing nil to properly ignore

flow/instrumentation/common/resource.go
```diff
@@ -0,0 +1,19 @@
+package common
+
+import (
+	"go.opentelemetry.io/otel/sdk/resource"
+	"go.opentelemetry.io/otel/semconv/v1.24.0"
+)
+
+// NewOtelResource newOtelResource returns a resource describing this application.
```
serprex, line 8:
```suggestion
// NewOtelResource returns a resource describing this application.
```

flow/instrumentation/common/resource.go
```diff
@@ -0,0 +1,19 @@
...
+func NewOtelResource(otelServiceName string) (*resource.Resource, error) {
+	r, err := resource.Merge(
+		resource.Default(),
+		resource.NewWithAttributes(
+			semconv.SchemaURL,
+			semconv.ServiceNameKey.String(otelServiceName),
+		),
+	)
+
+	return r, err
```
serprex, line 18:
```suggestion
	return resource.Merge(
		resource.Default(),
		resource.NewWithAttributes(
			semconv.SchemaURL,
			semconv.ServiceNameKey.String(otelServiceName),
		),
	)
```

flow/cmd/worker_common.go
iamKunalGupta: my bad, was testing something out

flow/instrumentation/common/resource.go
```diff
@@ -0,0 +1,17 @@
+package common
```
serprex, line 1: avoid creating many small packages, it doesn't really matter until intersubpackage dependencies get involved. go's naming also makes having common names like util, utils, common, etc a pain

Already had to resolve util/utils split

flow/instrumentation/setup.go
```diff
@@ -0,0 +1,40 @@
...
+		var err error
+		for _, fn := range shutdownFuncs {
+			err = errors.Join(err, fn(ctx))
+		}
+		shutdownFuncs = nil
+		return err
+	}
+
+	var err error
+	handleErr := func(inErr error) {
```
serprex, line 26: shouldn't be a function

iamKunalGupta: added as we'll setup more things in this

flow/instrumentation/setup.go
```diff
@@ -0,0 +1,40 @@
...
+		err = errors.Join(inErr, shutdown(ctx))
+	}
+
+	if config.EnableTracing {
+		traceShutdown, err := tracing.SetupOtelTraceProviderExporter(serviceName)
+		if err != nil {
+			handleErr(err)
```
serprex, line 33: we always return after calling handleErr so we could just return the join, no err variable needed

flow/instrumentation/common/resource.go
iamKunalGupta: the call graph like setup -> tracing/metrics/logging -> common, so had to create a separate one this time.
I'll also change it to otel_common for better visibility

flow/instrumentation/setup.go
iamKunalGupta: my bad the otel example itself is very poorly written

flow/connectors/core.go
```diff
@@ -267,6 +268,8 @@ func GetAs[T Connector](ctx context.Context, config *protos.Peer) (T, error) {
 }
 
 func GetCDCPullConnector(ctx context.Context, config *protos.Peer) (CDCPullConnector, error) {
+	ctx, span := trace.StartSpan(ctx, "connectors.GetCDCPullConnector")
+	defer span.End()
```
serprex, line 272: can move this into GetAs, using `reflect.TypeFor[T]` to put T in span label

flow/go.mod
```diff
@@ -112,12 +117,11 @@ require (
 	github.com/prometheus/common v0.53.0 // indirect
 	github.com/prometheus/procfs v0.15.0 // indirect
 	github.com/rogpeppe/go-internal v1.12.0 // indirect
-	github.com/segmentio/asm v1.2.0 // indirect
+	github.com/segmentio/asm v1.2.0 // indirect; indirectianian
```
heavycrystal, line 120: is this tool generated?

iamKunalGupta: weird

PR #1746, opened 2024-05-22
Numeric: refactoring and truncate for BigQuery

Amogh-Bharadwaj:
Some fixes related to numerics.
- `.Round()` can turn a value like `999999....dot 99999...` to `1000...`. Changed this to `.Truncate`
- Included bigquery too for the snowflake specific steps in `processNumeric`
- Moved code snippets to functions

Functionally tested for BigQuery initial load

### DIFF
--- a/flow/datatypes/bigint.go
+++ b/flow/datatypes/bigint.go
@@ -1,41 +1,12 @@
 package datatypes
 
 import (
-	"math"
 	"math/big"
 )
 
-var tenInt = big.NewInt(10)
-
 func CountDigits(bi *big.Int) int {
-	tenInt := big.NewInt(10)
-	if bi.IsUint64() {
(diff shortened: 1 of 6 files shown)


flow/datatypes/bigint.go
```diff
@@ -0,0 +1,12 @@
+package datatypes
+
+import (
+	"math/big"
+)
+
+func CountDigits(bi *big.Int) int {
```
serprex, line 7: This seems like you found a bug in https://github.com/shopspring/decimal/pull/356 would be better to contribute fix upstream

Amogh-Bharadwaj: Any reason why string route isn't optimal?

serprex (approved):

PR #1756, opened 2024-05-26
CDC scripting

serprex:
script can define either `transformRecord` which can mutate its rows, or else `transformRow` which will be called on non-nil `record.old`/`record.new`


flow/activities/flowable_core.go
```diff
@@ -60,6 +60,7 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	config *protos.FlowConnectionConfigs,
 	options *protos.SyncFlowOptions,
 	sessionID string,
+	adaptStream func(*model.CDCStream[Items]) (*model.CDCStream[Items], error),
```
heavycrystal, line 63: make `adaptStream` nil in case of PgValue or nil script instead of adding the check within `adaptStream`

serprex: go doesn't support nil function values

heavycrystal (approved):

PR #1762, opened 2024-05-28
switching to local protobuf plugins in CI

heavycrystal:
working around frequent errors with protobuf generation with Buf remote plugins


serprex (approved):

buf.gen.yaml
```diff
@@ -1,6 +1,7 @@
 version: v1
 managed:
   enabled: true
+  optimize_for: SPEED
```
serprex, line 4: this is default, seems unnecessary

PR #1763, opened 2024-05-28
feat(cdc): reload peers on flow resume

iskakaushik:
Implement the reloadPeers function to support the edit peer functionality. When a flow is paused and resumed, reload the latest peer details for both the source and destination peers. This ensures that any changes made to the peer configuration during the pause are picked up when the flow resumes execution.

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -801,3 +801,95 @@ func (a *FlowableActivity) AddTablesToPublication(ctx context.Context, cfg *prot
 	}
 	return err
 }
+
+func (a *FlowableActivity) LoadPeer(ctx context.Context, peerName string) (*protos.Peer, error) {
+	row := a.CatalogPool.QueryRow(ctx, `
+		SELECT name, type, options
+		FROM peers
+		WHERE name = $1`, peerName)
+
+	var peer protos.Peer
+	var peerOptions []byte
+	if err := row.Scan(&peer.Name, &peer.Type, &peerOptions); err != nil {
+		return nil, fmt.Errorf("failed to load peer: %w", err)
+	}
+
+	switch peer.Type {
+	case protos.DBType_BIGQUERY:
+		var config protos.BigqueryConfig
+		if err := proto.Unmarshal(peerOptions, &config); err != nil {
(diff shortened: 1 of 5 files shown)


flow/activities/flowable.go
```diff
@@ -801,3 +801,95 @@ func (a *FlowableActivity) AddTablesToPublication(ctx context.Context, cfg *prot
...
+		WHERE name = $1`, peerName)
+
+	var peer protos.Peer
+	var peerOptions []byte
+	if err := row.Scan(&peer.Name, &peer.Type, &peerOptions); err != nil {
+		return nil, fmt.Errorf("failed to load peer: %w", err)
+	}
+
+	switch peer.Type {
+	case protos.DBType_BIGQUERY:
```
serprex, line 818: this switch looks pretty unfortunate. Could probably collapse it using `reflect`, but then you still need to have mapping from protos.DBType to reflect.Type

flow/workflows/cdc_flow.go
```diff
@@ -242,6 +271,13 @@ func CDCFlowWorkflow(
 				return state, err
 			}
 
+			// reload peers this is to support EDIT PEER functionality.
```
serprex, line 274:
```suggestion
			// reload peers to support EDIT PEER functionality
```

flow/workflows/cdc_flow.go
```diff
@@ -181,6 +182,34 @@ func addCdcPropertiesSignalListener(
 	})
 }
 
+func reloadPeers(ctx workflow.Context, logger log.Logger, cfg *protos.FlowConnectionConfigs) error {
+	reloadPeersCtx := workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		StartToCloseTimeout: 5 * time.Minute,
+	})
+
+	logger.Info(fmt.Sprintf("reloading source peer %s", cfg.Source.Name))
```
serprex:
```suggestion
	logger.Info("reloading source peer", slog.String("peerName", cfg.Source.Name))
```
annoying since datadog doesn't render as well but more structural is easier to query logs

serprex (approved):

PR #1769, opened 2024-05-29
Snowflake: use sysdate() for peerdb synced at

Amogh-Bharadwaj:
CURRENT_TIMESTAMP returns timestamp in local timezone which can be misleading when inspecting `peerdb_synced_at` values on Snowflake. SYSDATE() gives it in UTC

https://docs.snowflake.com/en/sql-reference/functions/sysdate


Amogh-Bharadwaj:
Updated description

heavycrystal (approved):

PR #1771, opened 2024-05-29
dynamic config from catalog: groundwork

heavycrystal:
1. Renames `alerting_settings` table and adds columns to make it more useful for dynamic configuration.
2. Some code cleanup and genericization

Default values should only be set from catalog, a migration will be added for existing settings and new settings to be added to this table.

UI to be added in a future PR.


Amogh-Bharadwaj (approved):

serprex (approved):

serprex:
also useful would be converging env configs & dynamic configs

PR #1773, opened 2024-05-30
Add primary key check for validate mirror

### LINKED ISSUE #1623: Add PKey/Repl ident validation in validate mirror
This is to account for mirror creation via SQL Layer

Amogh-Bharadwaj:
Add primary key and replica identity check as a step in validate mirror. This is more relevant for mirrors created via query layer
Functionally tested UI and query layer:

```sql

amogh=> CREATE MIRROR mymirror                                                                                                           
FROM postgres_local TO postgres_target                                                                                                   
WITH TABLE MAPPING(                                                                                                                          
myMixed.nopkMixed: public.mixed1,                                                                                                        
public.nopkMixed: public.mixed2,                                                                                                         
public.nopk:public.mixed3                                                                                                            
)                                                                                                                                        
WITH(                                                                                                                                        
do_initial_copy = false,                                                                                                                 
max_batch_size = 1000                                                                                                                
);
ERROR:  unable to submit job: 
"status: Unknown, message: 
"invalid mirror: provided source tables invalidated: 
tables myMixed.nopkMixed,public.nopkMixed,public.nopk do not have a primary key nor REPLICA IDENTITY FULL",
 details: [], metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\"} }"
```

closes #1623


flow/connectors/postgres/validate.go
```diff
@@ -130,3 +137,35 @@ func (c *PostgresConnector) CheckReplicationConnectivity(ctx context.Context) er
...
+// Check if all tables have either a primary key or REPLICA IDENTITY FULL
+func (c *PostgresConnector) CheckIfTablesAreMirrorable(
+	ctx context.Context,
+	tables []*utils.SchemaTable,
+) error {
+	badTables := make([]string, 0, len(tables))
+	for _, table := range tables {
+		var canMirror sql.NullBool
+		err := c.conn.QueryRow(ctx, `SELECT
+		(con.contype = 'p' OR t.relreplident in ('i', 'f')) AS can_mirror
```
heavycrystal, line 150: nit: replident being index doesn't seem to be enough, we need to check if the index also exists. 

`Records the old values of the columns covered by the named index, that must be unique, not partial, not deferrable, and include only columns marked NOT NULL. If this index is dropped, the behavior is the same as NOTHING.`

heavycrystal: can refer to existing `getReplicaIdentityIndexColumns` in Postgres connector for this

flow/connectors/postgres/validate.go
```diff
@@ -130,3 +137,35 @@ func (c *PostgresConnector) CheckReplicationConnectivity(ctx context.Context) er
 
 	return conn.Close(ctx)
 }
+
+// Check if all tables have either a primary key or REPLICA IDENTITY FULL
+func (c *PostgresConnector) CheckIfTablesAreMirrorable(
+	ctx context.Context,
+	tables []*utils.SchemaTable,
+) error {
+	badTables := make([]string, 0, len(tables))
```
serprex, line 146:
```suggestion
	var badTables []string
```
This doesn't need to allocate anything except in error path

iskakaushik (approved):

serprex (approved):

PR #1775, opened 2024-05-31
SF RenameTables: fix for mixed case

Amogh-Bharadwaj:
This PR accounts for mixed case table name in Snowflake's RenameTables()

### DIFF
--- a/flow/connectors/snowflake/client.go
+++ b/flow/connectors/snowflake/client.go
@@ -99,20 +99,3 @@ func snowflakeSchemaTableNormalize(schemaTable *utils.SchemaTable) string {
 	return fmt.Sprintf(`%s.%s`, SnowflakeIdentifierNormalize(schemaTable.Schema),
 		SnowflakeIdentifierNormalize(schemaTable.Table))
 }
-
-func normalizeSrcAndDst(source, destination string) (string, string, error) {
-	srcTable, err := utils.ParseSchemaTable(source)
-	if err != nil {
-		return "", "", fmt.Errorf("unable to parse source %s: %w", source, err)
-	}
-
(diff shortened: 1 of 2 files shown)


Amogh-Bharadwaj:
Functional tests not passing

flow/connectors/snowflake/snowflake.go
```diff
@@ -739,7 +739,11 @@ func (c *SnowflakeConnector) RenameTables(ctx context.Context, req *protos.Renam
 
 	if req.SyncedAtColName != nil {
 		for _, renameRequest := range req.RenameTableOptions {
-			resyncTblName := renameRequest.CurrentName
+			resyncSchemaTable, err := utils.ParseSchemaTable(renameRequest.CurrentName)
```
heavycrystal, line 742: invert the for loop and if statements so the conditionality happens inside the for loops, we can then reuse the parsed names for the entire loop instead of generating it 3 times per table

(yes this means we need to keep checking for the conditions, but I feel it would be more efficient than parsing the same thing thrice)

flow/connectors/snowflake/client.go
```diff
@@ -99,3 +99,20 @@ func snowflakeSchemaTableNormalize(schemaTable *utils.SchemaTable) string {
 	return fmt.Sprintf(`%s.%s`, SnowflakeIdentifierNormalize(schemaTable.Schema),
 		SnowflakeIdentifierNormalize(schemaTable.Table))
 }
+
+func normalizeSrcAndDst(source, destination string) (string, string, error) {
```
heavycrystal, line 103: if the other thing is fixed this shouldn't be necessary

Amogh-Bharadwaj:
functionally tested post change

PR #1779, opened 2024-06-01
md5 bad, scram

serprex:


heavycrystal (approved):

nexus/server/src/main.rs
```diff
@@ -68,8 +68,8 @@ impl AuthSource for FixedPasswordAuthSource {
         // randomly generate a 4 byte salt
         let salt = rand::thread_rng().gen::<[u8; 4]>();
         let password = &self.password;
-        let hash_password = hash_md5_password(login_info.user().unwrap_or(""), password, &salt);
-        Ok(Password::new(Some(salt.to_vec()), Vec::from(hash_password)))
+        let hash_password = gen_salted_password(password, &salt, 4096);
```
serprex, line 71:
```suggestion
        let hash_password = gen_salted_password(password, &salt, 16384);
```
https://rfc-editor.org/rfc/rfc7677 from 2015 suggests 15000, but CI was failing when I adjusted, maybe CI was broken

PR #1781, opened 2024-06-02
Edit Peer for record slot size

iskakaushik:


flow/activities/flowable.go
```diff
@@ -636,8 +653,15 @@ func (a *FlowableActivity) RecordSlotSizes(ctx context.Context) error {
 		return err
 	}
 
-	logger := activity.GetLogger(ctx)
+	recordedPeers := make(map[string]bool)
```
serprex, line 656:
```suggestion
	recordedPeers := make(map[string]struct{}, len(configs))
```

flow/activities/flowable.go
```diff
@@ -636,8 +653,15 @@ func (a *FlowableActivity) RecordSlotSizes(ctx context.Context) error {
...
+
 	for _, config := range configs {
+		peerName := config.Source.Name
+		if recordedPeers[peerName] {
+			continue
+		}
+		recordedPeers[peerName] = true
```
serprex, line 663:
```suggestion
	recordedPeers := make(map[string]struct{}, len(configs))

	for _, config := range configs {
		peerName := config.Source.Name
		if _, ok := recordedPeers[peerName]; !ok {
			continue
		}
		recordedPeers[peerName] = struct{}{}
```

flow/activities/flowable.go
```diff
@@ -636,8 +653,15 @@ func (a *FlowableActivity) RecordSlotSizes(ctx context.Context) error {
...
+
 		func() {
```
serprex, line 665: Not seeing why this block needs to be wrapped in closure

iskakaushik: it was like that earlier too, i think for the defer

serprex: Yes

serprex (approved):

serprex:
this should wait on #1844

PR #1784, opened 2024-06-03
SetupNormalize for Kafka

Amogh-Bharadwaj:
Implements SetupNormalize for Kafka connector, mainly needed for environments where auto topic creation is not an option, such as Confluent.
Need to figure out how to gate/flag this feature 
For confluent to work, the following things need to be in place:
- We cannot set the software version: apiVersion fetch fails 
- We need to set the number of partitions (5 default for now)
- Replication factor has to be 3

Functionally tested (Initial load + CDC) with Confluent


serprex:
with scripting destination table name may not be topic name. in particular, people may use destination table name to encode json parameters per table for script

serprex:
replaced by #1790
checking in validation for pubsub could work when force topic creation is not enabled & no script is being used

PR #1785, opened 2024-06-04
UI: consolidated logs section

Amogh-Bharadwaj:
Introduces a 'Logs' section in SIdebar where users can see logs across all mirrors.
This is useful as now users don't need to click on logs of each mirror to search for erroring mirrors


iskakaushik (approved):

serprex:
we should group peers/mirrors/scripts as resources & alerts/logs as observability. can be future pr

serprex (approved):

PR #1786, opened 2024-06-04
MultiQRepFlow

heavycrystal:
dogfooding by using for SnapshotFlow

TODO:
- [ ] tests
- [ ] manual testing with xmin
- [ ] creating route and handler
- [ ] UI (in a future PR)
- [ ] find a way to continue as new to prevent history too large
- [ ] figure out catalog for child workflows

### DIFF
--- a/flow/workflows/multi_qrep_flow.go
+++ b/flow/workflows/multi_qrep_flow.go
@@ -1,16 +1,18 @@
 package peerflow
 
 import (
+	"errors"
 	"fmt"
 	"log/slog"
 	"time"
 
-	"github.com/PeerDB-io/peer-flow/generated/protos"
-	"github.com/PeerDB-io/peer-flow/shared"
 	"go.temporal.io/api/enums/v1"
 	"go.temporal.io/sdk/log"
 	"go.temporal.io/sdk/workflow"
 	"google.golang.org/protobuf/proto"
+
+	"github.com/PeerDB-io/peer-flow/generated/protos"
+	"github.com/PeerDB-io/peer-flow/shared"
 )
 
 // getChildWorkflowID returns the child workflow ID for a new sync flow.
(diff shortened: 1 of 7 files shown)


flow/workflows/multi_qrep_flow.go
```diff
@@ -0,0 +1,177 @@
...
+						flowJobName: childWorkflowConfig.FlowJobName,
+						isWait:      true,
+					})
+				}, childWorkflowConfig, childWorkflowState)
+			}
+		}
+	}
+
+	// if initial copy is enabled then this selector will eventually finish waiting
+	// if initial copy is disabled then this selector will keep waiting until an error
```
serprex, line 173:
```suggestion
	// if initial copy is enabled then this selector will eventually finish waiting,
	// otherwise this selector will keep waiting until an error
```

flow/workflows/multi_qrep_flow.go
```diff
@@ -0,0 +1,177 @@
...
+			if childWorkflowError != nil {
+				break
+			}
+
+			syncSelector.DrainAsync(ctx)
+			waitSelector.DrainAsync(ctx)
+			var input workflowChanInput
+			ok, _ := workflowChan.ReceiveWithTimeout(ctx, 5*time.Second, &input)
+			if !ok {
+				continue
```
serprex, line 121: check `ctx.Err()` here?

serprex:
is this a breaking change?

heavycrystal:
> is this a breaking change?

so far it shouldn't be, but while implementing route and UI might have to change existing protos
in the future we can remove the workflow indirection and directly call activities from MultiQRepWorkflow

serprex (approved):

heavycrystal:
not a priority as of now, can revisit later

PR #1793, opened 2024-06-05
feat(cdc): enhance schema box with schema settings and target schema

iskakaushik:
The rationale here is that editing the schema individually is not an ideal experience.

Screenshot 2024-06-05 at 4 46 24 PM

- Added `SchemaSettings` component for modifying target schema.
- Introduced `defaultTargetSchema` state to manage target schema.
- Implemented `fetchTablesForSchema` function to handle table fetching with the target schema.
- Modified `fetchTables` function to accept `targetSchemaName` parameter.
- Updated table row handling to use `value` instead of `defaultValue` for destination inputs.
- Improved schema box initialization with `useEffect` to fetch tables based on `defaultTargetSchema`.

### DIFF
--- a/ui/app/mirrors/create/cdc/schemabox.tsx
+++ b/ui/app/mirrors/create/cdc/schemabox.tsx
@@ -1,4 +1,5 @@
 'use client';
+
 import { TableMapRow } from '@/app/dto/MirrorsDTO';
 import { DBType } from '@/grpc_generated/peers';
 import { Checkbox } from '@/lib/Checkbox';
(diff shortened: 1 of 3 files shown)


ui/app/mirrors/create/cdc/schemabox.tsx
```diff
@@ -61,7 +67,7 @@ const SchemaBox = ({
           row.source.toLowerCase().includes(tableQueryLower)
       )
       .sort((a, b) => a.source.localeCompare(b.source));
-  }, [schema, rows, tableQuery]);
+  }, [rows, tableQuery]);
```
serprex, line 70:
```suggestion
  }, [schema, rows, tableQuery]);
```

serprex (approved):

PR #1796, opened 2024-06-06
dynconf pt.2: more dynconf and UI improvements

heavycrystal:


flow/connectors/utils/lua.go
```diff
@@ -109,8 +109,12 @@ type LPool[T any] struct {
 	closed   bool
 }
 
-func LuaPool[T any](cons func() (*lua.LState, error), merge func(T)) (*LPool[T], error) {
-	maxSize := peerdbenv.PeerDBQueueParallelism()
+func LuaPool[T any](ctx context.Context, cons func() (*lua.LState, error), merge func(T)) (*LPool[T], error) {
+	maxSize, err := peerdbenv.PeerDBQueueParallelism(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get parallelism: %w", err)
+	}
```
serprex, line 116: maybe better to make maxSize parameter instead of taking ctx

nexus/catalog/migrations/V30__more_dynconf_settings.sql
```diff
@@ -0,0 +1,16 @@
+INSERT INTO dynamic_settings (config_name,config_default_value,config_value_type,config_description,config_apply_mode)
```
serprex, line 1: these migration files are why I'd rather have envs & defaults in code, not db. In particular this migration script fails if we add an env & insert it into table causing conflict

ui/app/settings/page.tsx
```diff
@@ -65,7 +68,49 @@ const DynamicSettingItem = ({
...
+        a > Number.MAX_SAFE_INTEGER ||
+        a < Number.MIN_SAFE_INTEGER
+      ) {
+        notifyErr('Invalid value. Please enter a valid 64-bit signed integer.');
+        return false;
+      }
+      return true;
+    } else if (setting.config_value_type === DynconfValueType.UINT) {
+      const a = parseInt(Number(notNullValue).toString());
+      console.log(a);
```
serprex, line 86:
```suggestion
```

ui/app/settings/page.tsx
```diff
@@ -149,6 +194,7 @@ const SettingsPage = () => {
       return 0;
     });
 
+    setTotalPages(Math.ceil(filteredSettings.length / ROWS_PER_PAGE));
```
serprex, line 197: why is this state? should not call setters in useMemo

ui/app/settings/page.tsx
```diff
@@ -134,12 +178,16 @@ const SettingsPage = () => {
...
-    const filteredSettings = settings.filter((setting) =>
+  const filteredSettings = useMemo(() => {
+    return settings.filter((setting) =>
       setting.config_name.toLowerCase().includes(searchQuery.toLowerCase())
     );
+  }, [settings, searchQuery]);
+  const totalPages = useMemo(
+    () => Math.ceil(filteredSettings.length / ROWS_PER_PAGE),
+    [filteredSettings]
+  );
```
serprex, line 189:
```suggestion
  const totalPages = Math.ceil(filteredSettings.length / ROWS_PER_PAGE);
```
not worth overhead of hook

serprex (approved):

ui/app/settings/page.tsx
heavycrystal: won't update with search term then

serprex: yes it will, useMemo is only useful to prevent reexecution

PR #1797, opened 2024-06-06
Drop stats via trigger

Amogh-Bharadwaj:
This PR implements dropping stats of previous mirrors when a new entry is made to the flows table. Essentially, when a user creates (/re-creates) a mirror with the same name, we do not want to display the stats of the previous runs.

Doing this on flow insertion rather than flow deletion has the advantage of the stats of dropped mirrors remaining for our investigative purposes.

This operation is done via a triggered function in Postgres declared in catalog.

### DIFF
--- a/nexus/catalog/migrations/V27__drop_stats_trigger.sql
+++ b/nexus/catalog/migrations/V27__drop_stats_trigger.sql
@@ -1,0 +1,26 @@
+CREATE OR REPLACE FUNCTION clear_mirror_stats() RETURNS TRIGGER AS $$
+DECLARE
+    mirrorName TEXT;
+    cloneMirrorRegex TEXT;
+BEGIN
+    mirrorName := NEW.name;
+    cloneMirrorRegex := 'clone_' || mirrorName || '_%';
+
+    -- Initial Load Stats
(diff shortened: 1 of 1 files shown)


nexus/catalog/migrations/V27__drop_stats_trigger.sql
```diff
@@ -0,0 +1,26 @@
+CREATE OR REPLACE FUNCTION clear_mirror_stats() RETURNS TRIGGER AS $$
+DECLARE
+    mirrorName TEXT;
+    cloneMirrorRegex TEXT;
+BEGIN
+    mirrorName := NEW.name;
+    cloneMirrorRegex := 'clone_' || mirrorName || '_%';
```
serprex, line 7:
```suggestion
    cloneMirrorPattern := 'clone\\_' || replace(replace(mirrorName, '_', '\\_'), '%', '\\%') || '\\_%';
```
_ is a single character wildcard in like patterns. Need to escape. https://www.postgresql.org/docs/12/functions-matching.html

serprex:
This is a valuable feature for upgrades to persist history etc. Ideally this would be an option during mirror creation to scrub conflicting history. Then go code would wipe, no triggers

Amogh-Bharadwaj:
Going to implement a UI option for this on drop mirror instead

Issue #1803, opened 2024-06-07
ui for empty string metadata columns to omit metadata columns


serprex:
particularly useful with pgvalue in HA setups

PR #1805, opened 2024-06-07
Fix soft-delete column creation, rely on checking soft delete setting

Amogh-Bharadwaj:
In `handler.go`, the value of softDeleteColumnName was being set no matter what, and downstream we treated soft delete being on as this column name being non-empty. One issue this was causing was we were creating the soft delete column even if soft delete is off, as we were doing `if softDeleteColumnName is empty`

This PR passes `softDeleteEnabled` as a boolean parameter in code paths where `softDeleteColName` is used, and incorporates `softDeleteEnabled` as well in steps which are conditional on soft delete

All code paths functionally tested:
1. Snowflake - Initial Load + CDC With Soft Delete and Without Soft Delete
2. BigQuery - Initial Load + CDC With Soft Delete and Without Soft Delete
3. Snowflake resync with and without soft delete
4. BigQuery resync with and without soft delete

### DIFF
--- a/protos/flow.proto
+++ b/protos/flow.proto
@@ -87,6 +87,7 @@ message RenameTablesInput {
   repeated RenameTableOption rename_table_options = 3;
   optional string soft_delete_col_name = 4;
   optional string synced_at_col_name = 5;
+  bool soft_delete_enabled = 7;
 }
 
 message RenameTablesOutput {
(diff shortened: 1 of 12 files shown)


protos/flow.proto
```diff
@@ -87,6 +87,7 @@ message RenameTablesInput {
   repeated RenameTableOption rename_table_options = 3;
   optional string soft_delete_col_name = 4;
```
heavycrystal, line 88: this was written so we can indicate soft delete being disabled with nil, use that instead of a seperate boolean

we should use optional everywhere else as well, but will break protos so we can do it at v0.18

Amogh-Bharadwaj:
Closing as we decided deleted_at column is always required

serprex:
reopening because #1803

heavycrystal:
superseded by #1806

PR #1806, opened 2024-06-08
added option to disable all PeerDB columns

### LINKED ISSUE #1803: ui for empty string metadata columns to omit metadata columns

heavycrystal:
closes #1803 

makes all soft delete logic now depend on not empty column name, fasttrack to removing it entirely

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -226,6 +226,14 @@ func CDCFlowWorkflow(
 	logger := log.With(workflow.GetLogger(ctx), slog.String(string(shared.FlowNameKey), cfg.FlowJobName))
 	flowSignalChan := model.FlowSignal.GetSignalChannel(ctx)
 
+	// ugly, find a better way to represent this later
+	if cfg.DisablePeerdbColumns {
+		logger.Info("force disabling peerdb columns")
+		cfg.SoftDelete = false
+		cfg.SoftDeleteColName = ""
+		cfg.SyncedAtColName = ""
+	}
+
 	err := workflow.SetQueryHandler(ctx, shared.CDCFlowStateQuery, func() (CDCFlowWorkflowState, error) {
 		return *state, nil
 	})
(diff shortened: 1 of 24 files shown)


flow/workflows/cdc_flow.go
```diff
@@ -226,6 +226,14 @@ func CDCFlowWorkflow(
 	logger := log.With(workflow.GetLogger(ctx), slog.String(string(shared.FlowNameKey), cfg.FlowJobName))
 	flowSignalChan := model.FlowSignal.GetSignalChannel(ctx)
 
+	// ugly, find a better way to represent this later
+	if cfg.DisablePeerdbColumns {
```
serprex, line 230: this seems like a ui flag instead of a config flag. @iskakaushik was thinking of splitting frontend vs backend protos, which this would benefit from

nexus/flow-rs/src/grpc.rs
```diff
@@ -176,6 +176,7 @@ impl FlowGrpcClient {
             script: job.script.clone(),
             system: system as i32,
             idle_timeout_seconds: job.sync_interval.unwrap_or_default(),
+            disable_peerdb_columns: job.disable_peerdb_columns,
```
serprex, line 179: so like ui flag here is that while bool may belong in model, by time it hits config we should clear fields instead of further propagation of this bool

heavycrystal: makes sense, I had it at the handler level initially but moved it here in case someone tries to set it for existing flows and also simplify testing

iskakaushik (approved):

nexus/flow-rs/src/grpc.rs
serprex: for testing you can set the existing fields. same goes for the bool tracking deleted at being there, that's a ui flag, on proto empty string signals

serprex (approved):

PR #1811, opened 2024-06-10
BigQuery: better type handling

Amogh-Bharadwaj:
Refactors `qvalue_convert.go` to now use `bigquery.FieldSchema` and not `FieldType`. No need to rely on hacks like `IsArray` while creating tables
Functionally tested with cdc and schema change of array

### DIFF
--- a/flow/connectors/bigquery/qvalue_convert.go
+++ b/flow/connectors/bigquery/qvalue_convert.go
@@ -6,75 +6,134 @@ import (
 	"github.com/PeerDB-io/peer-flow/model/qvalue"
 )
 
-func qValueKindToBigQueryType(colType string) bigquery.FieldType {
+func qValueKindToBigQueryType(colType string) bigquery.FieldSchema {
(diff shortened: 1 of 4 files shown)


serprex (approved):

flow/connectors/bigquery/qvalue_convert.go
```diff
@@ -91,23 +126,28 @@ func BigQueryTypeToQValueKind(fieldType bigquery.FieldType) qvalue.QValueKind {
...
 	}
-	return bqTypeAsString
+	return bqType
 }
 
 func BigQueryFieldToQField(bqField *bigquery.FieldSchema) qvalue.QField {
 	return qvalue.QField{
 		Name:      bqField.Name,
-		Type:      BigQueryTypeToQValueKind(bqField.Type),
+		Type:      BigQueryTypeToQValueKind(*bqField),
```
serprex, line 150: might be better to pass param by reference instead of by value

PR #1817, opened 2024-06-11
Avro JSON size limit based truncation: make it empty object

Amogh-Bharadwaj:
When we truncate jsons for abiding by the avro size limit, we set it to empty string, which is not a valid json. Instead set it to empty object. 
An issue was seen due to this in BigQuery merge statement where parse_json fails
Functionally tested


serprex:
may be preferable to replace with `null`

serprex (approved):

PR #1823, opened 2024-06-12
Custom Resume and Cleaned Up Status Endpoint on Flow Api

Amogh-Bharadwaj:


Amogh-Bharadwaj:
Functionally tested both endpoints again

docker-compose-dev.yml
```diff
@@ -129,6 +129,7 @@ services:
       - 8113:8113
     environment:
       <<: [*catalog-config, *flow-worker-env, *minio-config]
+      PEERDB_PASSWORD: peerdb
```
serprex, line 132: this can be omitted. when env not set endpoints are unauthenticated

docker-compose.yml
```diff
@@ -115,6 +115,7 @@ services:
       - 8113:8113
     environment:
       <<: [*catalog-config, *flow-worker-env, *minio-config]
+      PEERDB_PASSWORD: peerdb
```
serprex, line 118:
```suggestion
```

serprex (approved):

PR #1824, opened 2024-06-12
Replace debezium/postgres with upstream postgres

serprex:
Avoids having copy of debezium in image
Avoids user confusion when seeing debezium in docker-compose

Change catalog locale to C.UTF8, which is what the default locale of postgres should be https://pganalyze.com/blog/5mins-postgres-17-builtin-c-utf8-locale


heavycrystal (approved):

volumes/docker-entrypoint-initdb.d/pg-hba-replication.sh
iamKunal, line 1: Maybe move all docker related volumes to a `docker`/`docker-volumes` subdirectory?

serprex: that's what volumes directory is

PR #1835, opened 2024-06-14
feat(iceberg): add streaming support to improve performance and memory footprint

iamKunalGupta:
This uses streams instead of sending all the records in 1 go.


flow-jvm/src/main/java/io/peerdb/flow/jvm/iceberg/resource/IcebergResource.java
```diff
@@ -67,4 +69,28 @@ public Uni<AppendRecordsResponse> appendRecords(AppendRecordsRequest request) {
                         icebergService.processAppendRecordsRequest(request))
                 .build());
     }
+
+    @Blocking
+    @Override
+    public Uni<AppendRecordsStreamResponse> streamingAppendRecords(Multi<AppendRecordsStreamRequest> request) {
+        return Uni.createFrom().item(Unchecked.supplier(() -> {
+//            var firstMessage = Uni.createFrom().multi(request).await().indefinitely();
```
serprex, line 77: commented code?

PR #1837, opened 2024-06-14
error out in unusual normalize conditions

heavycrystal:
primarily for CH


serprex:
this'll happen when batches are empty due to pg_logical_emit_message

flow/connectors/clickhouse/normalize.go
```diff
@@ -121,6 +124,13 @@ func generateCreateTableSQLForNormalizedTable(
 }
 
 func (c *ClickhouseConnector) NormalizeRecords(ctx context.Context, req *model.NormalizeRecordsRequest) (*model.NormalizeResponse, error) {
+	// ClickHouse is great, fix for potential consistency issues
+	select {
+	case <-time.After(3 * time.Second):
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
```
serprex, line 132: fine to sleep without select, let GetLastNormalizeBatchID find out ctx is done

serprex (approved):
want this tested, want Kaushik's approval

long term this ends up being weird logic which makes product more fragile

heavycrystal:
flow will change a lot, can revisit this if needed

PR #1839, opened 2024-06-14
finegrained replLock and making record channel send context aware

heavycrystal:


flow/model/cdc_stream.go
```diff
@@ -39,8 +43,19 @@ func (r *CDCStream[T]) GetLastCheckpoint() int64 {
...
-func (r *CDCStream[T]) AddRecord(record Record[T]) {
-	r.records <- record
+func (r *CDCStream[T]) AddRecord(ctx context.Context, record Record[T]) {
+	logger := logger.LoggerFromCtx(ctx)
+	for {
+		select {
+		case r.records <- record:
+			return
+		case <-time.After(10 * time.Second):
+			logger.Warn("waiting on adding record to stream", slog.Any("record", record))
```
serprex, line 53:
```suggestion
```
Adding check on context cancel is fair as receiver may leave channel blocked, but no need for time.After

flow/model/cdc_stream.go
```diff
@@ -39,8 +43,19 @@ func (r *CDCStream[T]) GetLastCheckpoint() int64 {
...
+		case <-ctx.Done():
+			logger.Warn("context cancelled while adding record to stream", slog.Any("record", record))
```
serprex, line 55: return error

flow/model/cdc_stream.go
heavycrystal: want it for logging purposes mainly, to debug future issues

will change to timer

flow/connectors/postgres/cdc.go
```diff
@@ -334,7 +346,10 @@ func PullCdcRecords[Items model.Items](
 		if err != nil {
 			return err
 		}
-		records.AddRecord(rec)
+		err = records.AddRecord(ctx, rec)
+		if err != nil {
+			return err
+		}
```
serprex, line 352:
```suggestion
		if err := records.AddRecord(ctx, rec); err != nil {
			return err
		}
```

serprex (approved):

PR #1841, opened 2024-06-14
UI: Fix display of partitioned tables

Amogh-Bharadwaj:
```
t.relispartition IS NOT TRUE ORDER BY t.relname, can_mirror DESC;
```
this is so that we only show the root table as we use publish_via_partition_root


Amogh-Bharadwaj:
Functionally tested with CDC + Initial load with a partitioned table

heavycrystal (approved):

PR #1844, opened 2024-06-15
no peer config stored in temporal state

serprex:
Also api uses peer names more too instead of passing configs around

This will need two version bumps to sanitize:
- on 0.13 user pauses their mirrors. This is important, failure to do so will break mirror when upgrading
- after upgrading to 0.14 unpause mirror. Temporal data will be sanitized at this point
- in 0.15 sanitization logic will be removed & protobufs can remove fields. Mirror should be paused in 0.14 to prevent replay using outdated protobufs


iskakaushik (approved):

protos/flow.proto
```diff
@@ -32,16 +31,16 @@ message TableMapping {
 }
 
 message SetupInput {
-  peerdb_peers.Peer peer = 1;
+  string peer = 1;
```
heavycrystal, line 34: what is the upgrade story for this, also if we're going to break protos might as well split it off and do other changes as well

protos/flow.proto
```diff
@@ -32,16 +31,16 @@ message TableMapping {
...
   string flow_name = 2;
 }
 
 message FlowConnectionConfigs {
   string flow_job_name = 1;
 
   // source and destination peer
-  peerdb_peers.Peer source = 2;
-  peerdb_peers.Peer destination = 3;
+  string source = 2;
```
heavycrystal, line 42: `source_peer_name` or something similar? applies to all

flow/connectors/core.go
```diff
@@ -290,6 +397,15 @@ func GetAs[T Connector](ctx context.Context, config *protos.Peer) (T, error) {
 	}
 }
 
+func GetByNameAs[T Connector](ctx context.Context, catalogPool *pgxpool.Pool, name string) (T, error) {
```
heavycrystal, line 400: this wouldn't play well with `Edit Peer`, we need to cache peer data atleast across one sync+normalize or QRep run to avoid peers change in the middle of a flow

heavycrystal (changes requested):

protos/flow.proto
serprex: hoping pause unpause will make it through alright, but have go ahead to make breaking changes for this. Could deprecate fields & add name fields & fallback to name from deprecated config

flow/connectors/core.go
serprex: Guess Edit Peer requires pausing all mirrors for that peer then

flow/workflows/snapshot_flow.go
```diff
@@ -127,11 +127,6 @@ func (s *SnapshotFlowExecution) cloneTable(
 		TaskQueue:           taskQueue,
 	})
 
-	// we know that the source is postgres as setup replication output is non-nil
-	// only for postgres
-	sourcePostgres := s.config.Source
-	sourcePostgres.GetPostgresConfig().TransactionSnapshot = snapshotName
```
serprex, line 133: pass in qrepconfig or something, ideally initial snapshot logic would be restart safe

flow/activities/flowable_core.go
```diff
@@ -186,7 +186,7 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	var res *model.SyncResponse
 	errGroup.Go(func() error {
 		syncBatchID, err := dstConn.GetLastSyncBatchID(errCtx, flowName)
-		if err != nil && config.Destination.Type != protos.DBType_EVENTHUBS {
```
serprex, line 189: @Amogh-Bharadwaj do you know why this was here?

Amogh-Bharadwaj: I think we can remove this @serprex

iskakaushik (approved):

flow/activities/flowable.go
```diff
@@ -841,94 +814,38 @@ func (a *FlowableActivity) AddTablesToPublication(ctx context.Context, cfg *prot
 	return err
 }
 
-func (a *FlowableActivity) LoadPeer(ctx context.Context, peerName string) (*protos.Peer, error) {
-	row := a.CatalogPool.QueryRow(ctx, `
-		SELECT name, type, options
-		FROM peers
-		WHERE name = $1`, peerName)
+// TODO remove in 0.15
```
heavycrystal, line 817: don't remove, could be used for handling config changes from UI

flow/workflows/snapshot_flow.go
```diff
@@ -33,14 +33,6 @@ type SnapshotFlowExecution struct {
 	tableNameSchemaMapping map[string]*protos.TableSchema
 }
 
-type cloneTablesInput struct {
```
heavycrystal, line 36: why removed?

ui/app/mirrors/create/schema.ts
```diff
@@ -38,34 +38,22 @@ export const tableMappingSchema = z
...
+  destinationName: z
+    .string({ required_error: 'Destination peer is required' })
+    .min(1),
   doInitialCopy: z.boolean().optional(),
   publicationName: z
     .string({
       invalid_type_error: 'Publication name must be a string',
     })
-    .max(255, 'Publication name must be less than 255 characters')
+    .max(255, 'Publication name must be less than 256 characters')
```
heavycrystal, line 50: these seem arbitrary, PG only supports 63 byte identifiers without recompiling

heavycrystal (approved):

flow/activities/flowable.go
serprex: must not be called from temporal, UI endpoints must update directly

flow/workflows/snapshot_flow.go
serprex: Unnecessary

ui/app/mirrors/create/schema.ts
serprex: yes, but these changes are to have text match code

flow/workflows/snapshot_flow.go
heavycrystal: had implemented it because getting arguments mixed around is quite possible with this call

PR #1847, opened 2024-06-17
UI cleanup

heavycrystal:
1. redacting more things, and including peer types not covered before. This isn't the ideal approach, have another idea I'll try when I have bandwidth
2. graph generation code refactor and switching to enums


serprex (approved):

Amogh-Bharadwaj (approved):
Nice !

PR #1850, opened 2024-06-19
Revert "Add primary key check for validate mirror (#1773)"

Amogh-Bharadwaj:
This reverts commit 0923a68f3139b3347864bb258c6fd7091fad96b4.

Faced an issue with a customer instance where this check was throwing errors but SetupFlow passed successfully on removing the check. 
This blocks mirror creation so getting the reversion in first and then will re-apply the feature fixed

### DIFF
--- a/flow/connectors/postgres/client.go
+++ b/flow/connectors/postgres/client.go
@@ -171,9 +171,6 @@ func (c *PostgresConnector) getReplicaIdentityIndexColumns(
          WHERE indrelid=$1 AND indisreplident=true`,
 		relID).Scan(&indexRelID)
 	if err != nil {
-		if err == pgx.ErrNoRows {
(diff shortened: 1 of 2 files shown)


serprex (approved):

flow/connectors/postgres/client.go
```diff
@@ -171,9 +171,6 @@ func (c *PostgresConnector) getReplicaIdentityIndexColumns(
          WHERE indrelid=$1 AND indisreplident=true`,
 		relID).Scan(&indexRelID)
 	if err != nil {
-		if err == pgx.ErrNoRows {
-			return nil, fmt.Errorf("no replica identity index for table %s", schemaTable)
-		}
```
serprex, line 176: can keep this

PR #1852, opened 2024-06-19
Drop mirror: Deal with peer names only in client side

Amogh-Bharadwaj:
Fetch the peer from the names in server, avoiding having peer info on client (browser)


serprex:
this is already done in #1844

PR #1859, opened 2024-06-22
don't call getMaxSyncsPerCDCFlow in loop

serprex:
this was spamming temporal history


heavycrystal:
defeats the purpose of it being dynconf

serprex:
it doesn't, dynconf has ui, pause/unpause takes on change. Soft pause should be implemented separately

iskakaushik (approved):

serprex:
alternatively, use https://pkg.go.dev/go.temporal.io/sdk/workflow#MutableSideEffect

serprex:
prodding more, MutableSideEffect is taking workflow.Context & passing down workflow.Context, creating context.Context with deadline which is scoped by calling context.Context isn't straightforward

PR #1871, opened 2024-06-25
Fix Status Link For Initial Load Tables In UI

Amogh-Bharadwaj:
Clicking on a table in initial load table redirects to /mirrors/<mirror_name> page. Here, mirror_name would be clone_mirror_name_..., for which getting the workflow id would error out. We were trying to get the workflow id of this clone job in the mirror page in UI and erroring out
This PR fixes this
Functionally tested with initial load table click and regular qrep mirror

Also add logs for mirror status endpoint and escape string for ILIKE clause


flow/cmd/mirror_status.go
```diff
@@ -185,7 +196,9 @@ func (h *FlowRequestHandler) cloneTableSummary(
 	var numRowsSynced pgtype.Int8
 	var avgTimePerPartitionMs pgtype.Float8
 
-	rows, err := h.pool.Query(ctx, q, "clone_"+flowJobName+"_%")
+	flowNamePart := strings.Replace("clone_"+
+		strings.Replace(flowJobName, "%", "\\%", -1)+"_\\%", "_", "\\_", -1)
```
serprex, line 200:
```suggestion
		strings.Replace(flowJobName, "%", "\\%", -1)+"\\_\\%", "_", "\\_", -1)
```
then move replace more nestedly

ui/app/mirrors/[mirrorId]/page.tsx
```diff
@@ -40,7 +38,7 @@ export default async function ViewMirror({
     return <div>No mirror status found!</div>;
   }
 
-  const mirrorInfo = await prisma.flows.findFirst({
+  let mirrorInfo = await prisma.flows.findFirst({
```
Comment (serprex) [inline ui/app/mirrors/[mirrorId]/page.tsx:41]:
```suggestion
  const mirrorInfo = await prisma.flows.findFirst({
```

ui/app/mirrors/[mirrorId]/page.tsx
```diff
@@ -51,7 +49,7 @@ export default async function ViewMirror({
     },
   });
 
-  const syncs = await prisma.cdc_batches.findMany({
+  let syncs = await prisma.cdc_batches.findMany({
```
Comment (serprex) [inline ui/app/mirrors/[mirrorId]/page.tsx:52]:
```suggestion
  const syncs = await prisma.cdc_batches.findMany({
```

ui/app/mirrors/[mirrorId]/page.tsx
```diff
@@ -115,77 +113,29 @@ export default async function ViewMirror({
...
-            startTime: partition.startTime,
-            endTime: partition.endTime,
-            numRows: partition.pulledRows,
-          }))}
-        />
-        <br></br>
-        <QRepStatusTable flowJobName={mirrorId} partitions={partitions} />
-      </LayoutMain>
-    );
+    redirect(`/mirrors/status/qrep/${mirrorId}`);
```
Comment (serprex) [inline ui/app/mirrors/[mirrorId]/page.tsx:117]: this redirect shouldn't happen, should be updating link instead of reverting change

Amogh-Bharadwaj:
replaced by #1875

PR #1879, opened 2024-06-26
Log chunk index and total chunks to better indicate progress

iskakaushik:


flow/shared/array.go
```diff
@@ -19,17 +19,18 @@ func ArrayMinus[T comparable](first, second []T) []T {
 }
 
 // Call f with subslices of slice. An empty slice will call f once with nil.
-func ArrayIterChunks[T any](slice []T, size int, f func(chunk []T) error) error {
+func ArrayIterChunks[T any](slice []T, size int, f func(chunk []T, chunkIdx, totalChunks int) error) error {
 	if len(slice) == 0 {
-		return f(nil)
+		return f(nil, 0, 1)
```
serprex, line 24:
```suggestion
		return f(nil, 0, 0)
```

flow/shared/array.go
```diff
@@ -19,17 +19,18 @@ func ArrayMinus[T comparable](first, second []T) []T {
...
 	if len(slice) == 0 {
-		return f(nil)
+		return f(nil, 0, 0)
 	}
 	if size <= 0 {
 		return nil
 	}
 	lo := 0
-	for lo < len(slice) {
+	totalChunks := (len(slice) + size - 1) / size
```
serprex, line 30: we have shared function for ceil div no?

serprex (approved):

flow/shared/array.go
iskakaushik: yup!

PR #1880, opened 2024-06-26
deprisma pt.1: API and UI cleanup

heavycrystal:
had started with just Prisma cleanup but turned into a general cleanup so splitting off to prevent code churn

1. removed drop mirror endpoint, switched to using flow state change endpoint with status TERMINATED
2. migration to make some row counts BIGINT
3. eliminated some Prisma calls, moved the rest back to route.ts files under /api
4. Resync for CDC mirrors is now an endpoint, prevent having to read the proto and then trigger multiple endpoints
5. remove some DTO types and replace with protos
6. protos remove fields


ui/app/api/mirrors/state/route.ts
```diff
@@ -1,17 +1,15 @@
...
+import { MirrorStatusResponse } from '@/grpc_generated/route';
 import { GetFlowHttpAddressFromEnv } from '@/rpc/http';
 
 export async function POST(request: Request) {
-  const body: MirrorStatusRequest = await request.json();
+  const body = await request.json();
+
   const flowServiceAddr = GetFlowHttpAddressFromEnv();
   try {
     const res: MirrorStatusResponse = await fetch(
```
serprex, line 9:
```suggestion
    return fetch(
```
See how columns lookup optimized in https://github.com/PeerDB-io/peerdb/pull/1877

ui/app/api/mirrors/state_change/route.ts
```diff
@@ -16,6 +16,9 @@ export async function POST(request: Request) {
 
     return new Response(JSON.stringify(res));
   } catch (e) {
+    if (e instanceof Response) {
```
serprex, line 19: again here just return fetch instead of this

flow/activities/flowable.go
```diff
@@ -544,36 +544,26 @@ func (a *FlowableActivity) CleanupQRepFlow(ctx context.Context, config *protos.Q
...
 }
 
-func (a *FlowableActivity) DropFlowSource(ctx context.Context, config *protos.ShutdownRequest) error {
-	sourcePeerName, err := a.getPeerNameForMirror(ctx, config.FlowJobName, Source)
-	if err != nil {
-		return err
-	}
-
-	ctx = context.WithValue(ctx, shared.FlowNameKey, config.FlowJobName)
+func (a *FlowableActivity) DropFlowSource(ctx context.Context, flowJobName string, sourcePeerName string) error {
```
iskakaushik, line 547: lets use protos as inputs and outputs for activities. Better flexibility to change things.

ui/app/mirrors/[mirrorId]/handlers.ts
```diff
@@ -1,37 +1,38 @@
...
 import {
   FlowStateChangeRequest,
   MirrorStatusResponse,
 } from '@/grpc_generated/route';
 
-export const getMirrorState = async (mirrorId: string) => {
+export const getMirrorState = async (
+  mirrorName: string
+): Promise<MirrorStatusResponse> => {
   return await fetch('/api/mirrors/state', {
```
Comment (serprex) [inline ui/app/mirrors/[mirrorId]/handlers.ts:10]:
```suggestion
export const getMirrorState = (
  mirrorName: string
): Promise<MirrorStatusResponse> => {
  return fetch('/api/mirrors/state', {
```

ui/app/mirrors/[mirrorId]/page.tsx
```diff
@@ -1,115 +1,71 @@
...
-    cache: 'no-store',
-  });
-  const json = await resp.json();
-  return json;
-}
+  const fetchState = useCallback(async () => {
+    await getMirrorState(mirrorId).then((res) => {
+      setMirrorState(res);
+      setMounted(true);
+    });
```
Comment (serprex) [inline ui/app/mirrors/[mirrorId]/page.tsx:29]:
```suggestion
    const res = await getMirrorState(mirrorId);
    setMirrorState(res);
    setMounted(true);
```

nexus/catalog/migrations/V33__more_counts_bigint.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE peerdb_stats.qrep_partitions ALTER COLUMN rows_in_partition TYPE bigint;
```
serprex, line 1: as discussed, V33 migration reserved for encrypted peer config migration

ui/app/mirrors/[mirrorId]/handlers.ts
Comment (heavycrystal) [reply ui/app/mirrors/[mirrorId]/handlers.ts:10]: TS has other ideas:

```This may be converted to an async function. ts(80006)```

File: protos/flow.proto [hunk 7]
```diff
@@ -302,7 +302,7 @@ message QRepParitionResult {
 }
 
 message DropFlowInput {
-  string flow_name = 1;
+  string flow_job_name = 1;
```
serprex, line 305: renaming fields breaks things

heavycrystal: proto wasn't used before this PR for some reason

serprex (approved):

PR #1881, opened 2024-06-26
Improve logging for queueing errors for pub-sub

iskakaushik:


flow/connectors/pubsub/pubsub.go
```diff
@@ -252,7 +252,12 @@ func (c *PubSubConnector) SyncRecords(ctx context.Context, req *model.SyncRecord
 	publish := make(chan publishResult, 32)
 	waitChan := make(chan struct{})
 
-	queueCtx, queueErr := context.WithCancelCause(ctx)
+	queueCtx, queueCtxErr := context.WithCancelCause(ctx)
+	queueErr := func(err error) {
+		c.logger.Error("[pubsub] queue error", slog.Any("error", err))
```
serprex, line 257: may want same thing for kafka

serprex (approved):

flow/connectors/pubsub/pubsub.go
iskakaushik: done!

PR #1882, opened 2024-06-26
Encrypt credentials stored in peers table

iskakaushik:
This commit enhances security by encrypting peer options in the 'peers' table using AES encryption. AES was chosen for its reliability and widespread adoption in securing sensitive data. Changes include:

- Added `enc_key_id` column to the `peers` table.
- Modified `LoadPeer` and `CreatePeerNoValidate` functions to handle encryption/decryption.
- Updated environment configuration to support encryption key management.

### DIFF
--- a/flow/connectors/core.go
+++ b/flow/connectors/core.go
@@ -25,6 +25,7 @@ import (
 	"github.com/PeerDB-io/peer-flow/logger"
 	"github.com/PeerDB-io/peer-flow/model"
 	"github.com/PeerDB-io/peer-flow/otel_metrics/peerdb_guages"
+	"github.com/PeerDB-io/peer-flow/peerdbenv"
 )
 
 type Connector interface {
@@ -264,13 +265,19 @@ func LoadPeerType(ctx context.Context, catalogPool *pgxpool.Pool, peerName strin
 
 func LoadPeer(ctx context.Context, catalogPool *pgxpool.Pool, peerName string) (*protos.Peer, error) {
 	row := catalogPool.QueryRow(ctx, `
-		SELECT type, options
+		SELECT type, options, enc_key_id
 		FROM peers
 		WHERE name = $1`, peerName)
 
 	peer := &protos.Peer{Name: peerName}
-	var peerOptions []byte
-	if err := row.Scan(&peer.Type, &peerOptions); err != nil {
+	var encKeyID string
+	var encPeerOptions []byte
+	if err := row.Scan(&peer.Type, &encPeerOptions, &encKeyID); err != nil {
+		return nil, fmt.Errorf("failed to load peer: %w", err)
+	}
+
+	peerOptions, err := decryptPeerOptions(encKeyID, encPeerOptions)
+	if err != nil {
 		return nil, fmt.Errorf("failed to load peer: %w", err)
 	}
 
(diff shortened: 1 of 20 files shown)


flow/peerdbenv/config.go
```diff
@@ -92,3 +92,17 @@ func PeerDBAlertingEmailSenderRegion() string {
 func PeerDBAlertingEmailSenderReplyToAddresses() string {
 	return GetEnvString("PEERDB_ALERTING_EMAIL_SENDER_REPLY_TO_ADDRESSES", "")
 }
+
+func PeerDBCurrentEncKeyID() string {
+	return GetEnvString("PEERDB_CURRENT_ENC_KEY_ID", "")
+}
+
+func PeerDBEncKeys() shared.PeerDBEncKeys {
+	return GetEnvJSON("PEERDB_ENC_KEYS", make(shared.PeerDBEncKeys, 0))
```
serprex, line 101:
```suggestion
	return GetEnvJSON[shared.PeerDBEncKeys]("PEERDB_ENC_KEYS", nil)
```

iskakaushik: i liked it before because i could invoke `.Get` on it without worrying about NPE.

flow/shared/enc.go
```diff
@@ -0,0 +1,132 @@
...
+	if key == nil {
+		return nil, errors.New("encryption key is nil")
+	}
+
+	decodedKey, err := base64.StdEncoding.DecodeString(key.Value)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode base64 key: %w", err)
+	}
+
+	block, err := aes.NewCipher(decodedKey)
```
serprex, line 47: should error if decodedKey not 32 bytes

serprex:
Need to also add decrypting to `getPostgresPeerConfigs`. Could revise that code to loop over peer names, but that'd have to make N queries instead of just 1

flow/peerdbenv/config.go
serprex: right, was thinking in terms of `map`

flow/connectors/core.go
```diff
@@ -264,13 +265,19 @@ func LoadPeerType(ctx context.Context, catalogPool *pgxpool.Pool, peerName strin
 
 func LoadPeer(ctx context.Context, catalogPool *pgxpool.Pool, peerName string) (*protos.Peer, error) {
 	row := catalogPool.QueryRow(ctx, `
-		SELECT type, options
+		SELECT type, options, enc_key_id
```
heavycrystal, line 268: is there a migration for this

flow/connectors/core.go
```diff
@@ -264,13 +265,19 @@ func LoadPeerType(ctx context.Context, catalogPool *pgxpool.Pool, peerName strin
...
-	if err := row.Scan(&peer.Type, &peerOptions); err != nil {
+	var encKeyID string
+	var encPeerOptions []byte
+	if err := row.Scan(&peer.Type, &encPeerOptions, &encKeyID); err != nil {
+		return nil, fmt.Errorf("failed to load peer: %w", err)
+	}
+
+	peerOptions, err := decryptPeerOptions(encKeyID, encPeerOptions)
+	if err != nil {
 		return nil, fmt.Errorf("failed to load peer: %w", err)
```
heavycrystal, line 281: error should be swallowed at some point

flow/connectors/core.go
iskakaushik: see: https://github.com/PeerDB-io/peerdb/pull/1883

flow/connectors/core.go
serprex: ?

serprex: talked with Kevin, he's suggesting crypto related error messages shouldn't display to user, can review crypto error messages for leaks later

serprex (approved):

heavycrystal (approved):

PR #1891, opened 2024-06-28
UI: Refactor schema qualification check

Amogh-Bharadwaj:
- IsSchemaLessPeer -> changed to be a function ValidSchemaQualifiedTarget 
- Adds a check for this in CDC

Functionally tested for all target peers - CDC and QRep

Screenshot 2024-06-28 at 6 48 51 AM

### DIFF
--- a/ui/app/mirrors/create/handlers.ts
+++ b/ui/app/mirrors/create/handlers.ts
@@ -33,6 +33,16 @@ export const IsQueuePeer = (peerType?: DBType): boolean => {
   );
 };
 
+const IsSchemaLessPeer = (peerType?: DBType): boolean => {
+  const isSchemaLessPeer =
+    peerType === DBType.ELASTICSEARCH ||
+    peerType === DBType.BIGQUERY ||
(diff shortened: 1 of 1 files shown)


iskakaushik (approved):

ui/app/mirrors/create/handlers.ts
```diff
@@ -33,6 +33,19 @@ export const IsQueuePeer = (peerType?: DBType): boolean => {
...
+  peerType: DBType,
+  tableName: string
+): boolean => {
+  const schemaRequiredPeer =
+    peerType === DBType.POSTGRES || peerType === DBType.SNOWFLAKE;
+  if (!schemaRequiredPeer) {
+    return true;
+  }
+
+  return tableName?.includes('.') && tableName?.split('.')[0].length !== 0;
```
serprex, line 46:
```suggestion
  return tableName && tableName.includes('.') && !tableName.startsWith('.');
```
With `?.` calling methods you also need to do `?.(args)` otherwise you end up with error about calling `null` as a function

PR #1892, opened 2024-06-28
remove SoftDelete boolean, rely only on column name

heavycrystal:
needs to be done after the v0.14.1 migration to patch up cases


protos/flow.proto
```diff
@@ -67,11 +67,12 @@ message RenameTableOption {
 }
 
 message RenameTablesInput {
+  reserved 2,4,5;
   string flow_job_name = 1;
   repeated RenameTableOption rename_table_options = 3;
-  optional string soft_delete_col_name = 4;
-  optional string synced_at_col_name = 5;
   string peer_name = 6;
+  string soft_delete_col_name = 7;
```
serprex, line 74: this will break things

heavycrystal: this message is only used during resync and isn't persisted

serprex (approved):

PR #1902, opened 2024-07-03
Query Layer: Add limit 1 for workflow id fetch

Amogh-Bharadwaj:
Pause and resume currently fails in query layer when there are more than 1 entries for flow name.
The error is 
```
query returned an unexpected number of rows
```
because `query_opt()` errors out when the query returns more than 1 row

Add limit 1 at the end to cover for this


heavycrystal:
#1880 removes the need for this query entirely, one remaining call site can be refactored to just check for existence of a mirror with the name

Amogh-Bharadwaj:
Closed in light of https://github.com/PeerDB-io/peerdb/pull/1880

PR #1903, opened 2024-07-04
deprisma alerting config

serprex:
in prep for encrypting alerting configs

### DIFF
--- a/flow/cmd/alerts.go
+++ b/flow/cmd/alerts.go
@@ -1,0 +1,63 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+
+	"github.com/PeerDB-io/peer-flow/generated/protos"
+)
+
+func (h *FlowRequestHandler) GetAlertConfigs(ctx context.Context, req *protos.GetAlertConfigsRequest) (*protos.GetAlertConfigsResponse, error) {
+	rows, err := h.pool.Query(ctx, "select id, service_type, service_config from peerdb_stats.alerting_config")
+	if err != nil {
+		return nil, err
+	}
+
+	configs, err := pgx.CollectRows(rows, func(row pgx.CollectableRow) (*protos.AlertConfig, error) {
+		config := &protos.AlertConfig{}
+		err := row.Scan(&config.Id, &config.ServiceType, &config.ServiceConfig)
+		return config, err
+	})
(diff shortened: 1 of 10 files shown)


ui/app/alert-config/page.tsx
```diff
@@ -20,23 +20,21 @@ const ServiceIcon = ({
...
+  />
+);
 
 const AlertConfigPage: React.FC = () => {
   const {
     data: alerts,
     isLoading,
   }: {
-    data: UAlertConfigResponse[];
+    data: { configs: UAlertConfigResponse[] };
```
heavycrystal, line 37: could replace this with `GetAlertConfigsResponse`

protos/route.proto
```diff
@@ -37,6 +37,31 @@ message CreateCustomSyncResponse {
...
+  int32 id = 1;
+}
+message GetAlertConfigsResponse {
+  repeated AlertConfig configs = 1;
+}
+message PostAlertConfigResponse {
+  int32 id = 3;
+}
+message DeleteAlertConfigResponse {
+}
```
heavycrystal, line 63: should contain `ok` at the very least, same for other routes

flow/cmd/alerts.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (h *FlowRequestHandler) GetAlertConfigs(ctx context.Context, req *protos.GetAlertConfigsRequest) (*protos.GetAlertConfigsResponse, error) {
+	rows, err := h.pool.Query(ctx, "select id, service_type, service_config from alerting_config")
+	if err != nil {
+		return nil, err
+	}
+
+	configs, err := pgx.CollectRows(rows, func(row pgx.CollectableRow) (*protos.AlertConfig, error) {
+		config := &protos.AlertConfig{}
+		err := row.Scan(&config.Id, &config.ServiceType, &config.ServiceConfig)
```
heavycrystal, line 19: could give https://pkg.go.dev/github.com/jackc/pgx/v5#RowToStructByName a shot

protos/route.proto
serprex: seems like errors should rely on grpc error reporting mechanism already there

flow/cmd/alerts.go
serprex: Don't trust it with protos

heavycrystal (approved):

PR #1913, opened 2024-07-09
Unnest columns: edge case fix

Amogh-Bharadwaj:
For a JSON column whose name matches one of its keys in one of its objects, our unnesting logic unnests the fields and then deletes the key with the column name, thereby overwriting and then deleting this one field.

This PR fixes and adds a test for it

### DIFF
--- a/flow/model/model.go
+++ b/flow/model/model.go
@@ -274,12 +274,9 @@ func (r *RecordItems) ToJSONWithOpts(opts *ToJSONOptions) (string, error) {
 					return "", err
 				}
 
+				delete(jsonStruct, col)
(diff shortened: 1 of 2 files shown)


flow/model/model.go
```diff
@@ -277,7 +277,10 @@ func (r *RecordItems) ToJSONWithOpts(opts *ToJSONOptions) (string, error) {
 				for k, v := range unnestStruct {
 					jsonStruct[k] = v
 				}
-				delete(jsonStruct, col)
+
```
serprex, line 280: could move delete above for loop, no need to check field after

serprex (approved):

PR #1916, opened 2024-07-09
Edit Scripts UI: Fetch script by ID

Amogh-Bharadwaj:
This PR changes edit script page in UI to fetch the script by `scriptid` in search params rather than parsing base64

Functionally tested


ui/app/api/scripts/[scriptId]/route.ts
```diff
@@ -0,0 +1,20 @@
+import { ScriptsType } from '@/app/dto/ScriptsDTO';
+import prisma from '@/app/utils/prisma';
+import { NextRequest, NextResponse } from 'next/server';
+
+export async function GET(
+  _: NextRequest,
+  context: { params: { scriptId: string } }
+) {
+  const existingScript = await prisma.scripts.findFirst({
```
Comment (serprex) [inline ui/app/api/scripts/[scriptId]/route.ts:9]: we're getting rid of prisma

Comment (heavycrystal) [reply ui/app/api/scripts/[scriptId]/route.ts:9]: will update this as needed after #1917 lands

serprex (approved):

protos/route.proto
```diff
@@ -88,6 +88,7 @@ message Script {
   string source = 4;
 }
 message GetScriptsRequest {
+  int32 id = 1;
```
serprex, line 91: would probably be cleaner to create a separate endpoint, but doesn't really matter

PR #1921, opened 2024-07-11
UI: no initial copy for eventhubs mirror

Amogh-Bharadwaj:
We do not display initial copy setting for eventhubs in UI yet as it is yet to be supported, but the default of that setting is true. This PR sets it to false in case of an Eventhubs mirror

Functionally tested

### DIFF
--- a/ui/app/mirrors/create/handlers.ts
+++ b/ui/app/mirrors/create/handlers.ts
@@ -33,6 +33,10 @@ export const IsQueuePeer = (peerType?: DBType): boolean => {
   );
 };
 
+const IsEventhubsPeer = (peerType?: DBType): boolean => {
+  return !!peerType && peerType === DBType.EVENTHUBS;
+};
+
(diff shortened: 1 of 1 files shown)


ui/app/mirrors/create/handlers.ts
```diff
@@ -66,6 +70,10 @@ const CDCCheck = (
   config.tableMappings = tableNameMapping as TableMapping[];
   config.flowJobName = flowJobName;
 
+  if (IsEventhubsPeer(destinationType)) {
```
heavycrystal, line 73: if only used once no need to make it a function

heavycrystal (approved):

ui/app/mirrors/create/handlers.ts
Amogh-Bharadwaj: it's fine, more readable

PR #1924, opened 2024-07-11
UI: Hide XMIN mirror type

Amogh-Bharadwaj:
This PR makes create mirror UI not display the XMIN option

Screenshot 2024-07-11 at 9 19 15 PM

### DIFF
--- a/ui/app/mirrors/create/mirrorcards.tsx
+++ b/ui/app/mirrors/create/mirrorcards.tsx
@@ -27,12 +27,12 @@ const MirrorCards = ({
         'Query Replication allows you to specify a set of rows to be synced via a SELECT query.',
       link: 'https://docs.peerdb.io/usecases/Streaming%20Query%20Replication/overview',
     },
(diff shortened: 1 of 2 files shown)


ui/app/mirrors/create/mirrorcards.tsx
```diff
@@ -27,12 +27,12 @@ const MirrorCards = ({
         'Query Replication allows you to specify a set of rows to be synced via a SELECT query.',
       link: 'https://docs.peerdb.io/usecases/Streaming%20Query%20Replication/overview',
     },
-    {
-      title: MirrorType.XMin,
-      description:
```
serprex, line 32: maybe instead of removing we can explicitly advise against its use

Amogh-Bharadwaj: I like that idea too cc @iskakaushik 
Similar to how we have a line for CDC saying it's recommended

PR #1942, opened 2024-07-15
Fix connectors.Close race

serprex:
Recently observed race: https://github.com/jackc/pgx/pull/2067

The issue is on our end, caused by this sequence:
1. create connector, setup defer
2. start errgroup using connector & another task
3. other task fails, before connector task stops errGroup.Wait returns & defer runs

defer may not run while connector task is meant to have exclusive access

Solve this by moving connector create/defer into task

Left dstConn qrep since it's being used to check already synced, can be followup

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -92,12 +92,6 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	})
 	defer shutdown()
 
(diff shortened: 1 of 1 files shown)


heavycrystal (approved):

serprex:
https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Wait

> Wait blocks until all function calls from the Go method have returned, then returns the first non-nil error (if any) from them.

Seems like errgroup shoud be avoiding this issue, but change here simplifies things at least, maybe still improved situation, trouble happened before still when returning before `Wait`

PR #1944, opened 2024-07-16
nexus: mode to only run migrations

heavycrystal:
can update cutomers not exposing nexus to run this as an `InitContainer` or something


flow/connectors/clickhouse/qrep.go
```diff
@@ -71,12 +71,16 @@ func (c *ClickhouseConnector) createMetadataInsertStatement(
 func (c *ClickhouseConnector) getTableSchema(tableName string) ([]*sql.ColumnType, error) {
 	//nolint:gosec
 	queryString := fmt.Sprintf(`SELECT * FROM %s LIMIT 0`, tableName)
-	//nolint:rowserrcheck
 	rows, err := c.database.Query(queryString)
```
serprex, line 74: should be using QueryContext

serprex (approved):

PR #1950, opened 2024-07-17
UI Graphs: convert to Number

Amogh-Bharadwaj:
String concatenation was happening for the rows synced numbers in the rows synced graph
Explicityly convert to Number so that it actually adds


Amogh-Bharadwaj:
functionally tested

PR #1951, opened 2024-07-17
Clickhouse resync: use Go time.now for setting synced_at column

Amogh-Bharadwaj:
In Clickhouse's RenameTables, we need to set the synced_at column of the resynced table to the current time.
Research suggests that using now() along with allowing non deterministic mutations in Clickhouse could be expensive and also lead to inaccuracies

This PR calculates the current timestamp in Go and substitutes it in the alter update command.

### DIFF
--- a/flow/connectors/clickhouse/cdc.go
+++ b/flow/connectors/clickhouse/cdc.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log/slog"
 	"strings"
+	"time"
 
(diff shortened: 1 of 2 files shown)


Amogh-Bharadwaj:
Functionally tested

iskakaushik (approved):

serprex (approved):

heavycrystal:
can benchmark this later

heavycrystal (approved):

PR #1953, opened 2024-07-18
Clickhouse: use mergetree instead in validation

Amogh-Bharadwaj:
Seems like in some versions/setups of Clickhouse, alter add column is altogether not supported for engine memory. Use mergetree instead when validating


Amogh-Bharadwaj:
Functionally tested with clickhouse local and clickhouse cloud

heavycrystal (approved):

PR #1957, opened 2024-07-18
Clickhouse resync: no need to touch peerdb columns

Amogh-Bharadwaj:
- Not sure why we need to set synced at column to now in RenameTables
- Migrating soft delete rows from the old table is not needed and probably not expected

### DIFF
--- a/flow/connectors/clickhouse/cdc.go
+++ b/flow/connectors/clickhouse/cdc.go
@@ -6,8 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"strings"
-	"time"
 
 	_ "github.com/ClickHouse/clickhouse-go/v2"
 	_ "github.com/ClickHouse/clickhouse-go/v2/lib/driver"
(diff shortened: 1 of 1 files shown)


flow/connectors/clickhouse/cdc.go
```diff
@@ -160,38 +158,13 @@ func (c *ClickhouseConnector) ReplayTableSchemaDeltas(ctx context.Context, flowJ
...
-			if err != nil {
-				return nil, fmt.Errorf("unable to set synced at column for table %s: %w",
-					renameRequest.CurrentName, err)
-			}
-		}
-
 		columnNames := make([]string, 0, len(renameRequest.TableSchema.Columns))
 		for _, col := range renameRequest.TableSchema.Columns {
 			columnNames = append(columnNames, col.Name)
 		}
```
serprex, line 165:
```suggestion
```

serprex:
how do other connector's compare?

Amogh-Bharadwaj:
will revisit this decision post some more discussion

PR #1958, opened 2024-07-18
RenameTables: remove synced at setting in rename

Amogh-Bharadwaj:
For PG, SF, and CH, we don't need to set synced_at column to now in RenameTables as that is taken care of by `default now()` in column declaration and also qrep takes care of setting it

CH then does not require alter update permission and validation


Amogh-Bharadwaj:
functionally tested for the mentioned peers

serprex (approved):

heavycrystal (approved):

PR #1961, opened 2024-07-19
UI: remove extra columns in Settings page

Amogh-Bharadwaj:
UI adaptation of https://github.com/PeerDB-io/peerdb/pull/1926

Screenshot 2024-07-19 at 7 39 00 PM

Functionally tested (parallel sync normalize on)


serprex:
if we're removing those columns we can remove it from proto too

PR #1964, opened 2024-07-19
remove /api/peers except for create/validate endpoint

serprex:
these 4 endpoints were all straightforward POST-to-GET proxies,
the last remaining two endpoints will require some movement since they fiddle with protobuf sum type juggling


serprex:
tested

heavycrystal (approved):

PR #1968, opened 2024-07-20
revert dependency removal

heavycrystal:
regression in #1966, package is used by generated proto files


serprex:
Kevin didn't generate protos

PR #1980, opened 2024-08-07
support grpc oauth

serprex:
Enables using OAuth2 for calls to the Flow API

Screenshot 2024-09-09 at 17 36 13

Requires the following env vars:
- `PEERDB_OAUTH_ISSUER_URL` - This is the OAuth Issuer URL of the JWT (Like `https://{AUTH0_DOMAIN}/`)
- `PEERDB_OAUTH_DISCOVERY_ENABLED` - Set this to `true` to enable discovery via `/.well-known/jwks.json` endpoint defined in openID spec
- `PEERDB_OAUTH_KEYSET_JSON` - If custom json keyset is to be provided.
- `PEERDB_OAUTH_JWT_CLAIM_KEY`, `PEERDB_OAUTH_JWT_CLAIM_VALUE` - any custom key-value to be additionally checked while validating the incoming jwt

Health Endpoints are explicitly excluded from auth.


flow/shared/authenticator.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// jwtFromRequest extracts the JWT token from the Authorization header.
+// it truncates the "Bearer" prefix from the header value if exists.
+func jwtFromRequest(authHeader string) ([]byte, error) {
+	if authHeader == "" {
+		return nil, errors.New("missing Authorization header")
+	}
+
+	if len(authHeader) > 7 && authHeader[:7] == "Bearer " {
```
heavycrystal, line 111: how about https://pkg.go.dev/strings#TrimPrefix

flow/shared/authenticator.go
```diff
@@ -0,0 +1,218 @@
...
+
+type identityProvider struct {
+	keySet      jwk.Set
+	validateOpt jwt.ValidateOption
+	issuer      string
+}
+
+func AuthGrpcMiddleware() ([]grpc.ServerOption, error) {
+	devKeySetJSON, _ := os.LookupEnv("PEERDB_OAUTH_KEY_SET_JSON")
+	auth0Domain, _ := os.LookupEnv("PEERDB_OAUTH_AUTH0DOMAIN")
```
heavycrystal, line 32: nit: don't like the digit in the env var
also could use `GetEnvString`

flow/shared/authenticator.go
```diff
@@ -0,0 +1,218 @@
...
+	"os"
+
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/lestrrat-go/jwx/v2/jws"
+	"github.com/lestrrat-go/jwx/v2/jwt"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+type AuthenticationConfig struct {
```
heavycrystal, line 17: nit: don't think we need the struct tags
also could be a proto for consistency

serprex: Kept it like this to match upstream, if we keep it coming from environment we may get rid of it

Want to get feedback from Pete if these split up environment variables work or if infra will prefer the json as a single variable or if we go as far as parsing it. Hope this iteration sticks tho

flow/shared/authenticator.go
serprex: yeah, TODO move to peerdbenv once stabilized

flow/shared/authenticator.go
serprex: Can look to get that implemented upstream

flow/shared/authenticator.go
```diff
@@ -0,0 +1,218 @@
...
+
+	validateOpts := identityProviderValidateOpts(provider)
+	if err := jwt.Validate(token, validateOpts...); err != nil {
+		return nil, fmt.Errorf("failed to validate token: %w", err)
+	}
+
+	if _, err := jws.Verify(payload, jws.WithKeySet(provider.keySet)); err != nil {
+		return nil, fmt.Errorf("failed to verify token: %w", err)
+	}
+
```
iamKunalGupta, line 100: TODO: We need to add additionally validation against some custom claims.
We could take something like `..._DESIRED_CLAIM_KEY` and `..._DESIRED_CLAIM_VALUE` and match it against the jwt

serprex: why isn't this upstream?

iamKunalGupta: no clue, but I think we do need it as per the discussion I had

pjhampton: My opinion is that it doesn't go far enough at the moment. I'm not sure why they didn't implement upstream. It would make me feel more comfortable if there was some data section checking here, but we can wait to see what is in that payload.

iamKunalGupta:
@serprex we need to exclude the health route from auth

iamKunalGupta:
> > @serprex we need to exclude the health route from auth
> 
> https://pkg.go.dev/google.golang.org/grpc#UnaryClientInterceptor
> 
> > method is the RPC name
> 
> can check that for skips

Something I have done in the past:

```go
func CreateAuthServerInterceptor(authConfig *AuthConfig, unauthenticatedMethods []string) grpc.UnaryServerInterceptor {
	unauthenticatedMethodsMap := make(map[string]struct{}, len(unauthenticatedMethods))
	for _, method := range unauthenticatedMethods {
		unauthenticatedMethodsMap[method] = struct{}{}
	}
	if authConfig.Disabled {
		logging.Log().Warn("Authentication is disabled for the current server")
		return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
			return handler(ctx, req)
		}
	}
	// Can add configuration and use auth accordingly
	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
		if _, unauthorized := unauthenticatedMethodsMap[info.FullMethod]; !unauthorized {
			// TODO add recover in case of panics here
			var err error
			ctx, err = Authorize(ctx, authConfig)
			if err != nil {
				return nil, err
			}
		}
		resp, err = handler(ctx, req)
		return resp, err
	}
}

```

pjhampton (approved):
tested. lgtm

flow/shared/authenticator.go
```diff
@@ -0,0 +1,215 @@
...
+type identityProvider struct {
+	keySet      jwk.Set
+	validateOpt jwt.ValidateOption
+	issuer      string
+}
+
+func AuthGrpcMiddleware() ([]grpc.ServerOption, error) {
+	keySetJSON := os.Getenv("PEERDB_OAUTH_KEYSET_JSON")
+	auth0Domain := os.Getenv("PEERDB_OAUTH_AUTH0DOMAIN")
+	auth0ClientID := os.Getenv("PEERDB_OAUTH_AUTH0CLIENTID")
```
iskakaushik, line 34: can we move this to `peerdbenv/env.go` please? Something like `PeerDBGetOAuthConfig` which returns a struct?

iskakaushik (approved):
LGTM modulo a minot comment to better audit env vars.

pjhampton:
LGTM with some additional testing

PR #1982, opened 2024-08-10
UI: Fix validate mirror not showing error message

Amogh-Bharadwaj:
For errors in mirror validation, the actual error doesn't show up. Seems to be a regression
This PR fixes this and adds a log in flow api


ui/app/mirrors/create/handlers.ts
```diff
@@ -446,18 +445,15 @@ export const handleValidateCDC = async (
...
-  const status: ValidateCDCMirrorResponse = await fetch(
-    '/api/v1/mirrors/cdc/validate',
-    {
-      method: 'POST',
-      body: JSON.stringify({
-        connectionConfigs: processCDCConfig(config),
-      }),
-    }
-  ).then((res) => res.json());
+  const status = await fetch('/api/v1/mirrors/cdc/validate', {
```
serprex, line 448:
```suggestion
  const status: ValidateCDCMirrorResponse = await fetch('/api/v1/mirrors/cdc/validate', {
```
I was liking idea of moving away from ok-bool / message-string in favor of checking fetch http code

https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

Amogh-Bharadwaj: `ValidateCDCMirrorResponse` currently does not have an error message field, which is why I removed it here.
The change made checks the http Ok you linked and not the Ok of `ValidateCDCMirrorResponse` which is what you wanted if I'm understanding right ?

serprex: yes, you can keep type info by first checking `.ok` on result of fetch, then when it's true call `res.json()` with typed assignment

serprex (approved):

PR #1983, opened 2024-08-13
go 1.23

serprex:
Replace x/exp/maps dependency with stdlib
Replace shared.ArrayIterChunks with slices.Chunk


heavycrystal (approved):
our new iterator overlords

serprex:
waiting on golangci release, tip supports 1.23

PR #1986, opened 2024-08-15
UI: fix system toggle in UI

Amogh-Bharadwaj:
For non-postgres destinations, type system must never be PG


Amogh-Bharadwaj:
Functionally tested

heavycrystal (approved):

serprex (approved):

PR #1987, opened 2024-08-15
Add env var for scripting

Amogh-Bharadwaj:
`PEERDB_EXPERIMENTAL_ENABLE_SCRIPTING` has been introduced in the docker compose files. Default is true in the docker-compose files.
The code checks if the var is true and true only so if not specified, it will be taken as false


Amogh-Bharadwaj:
Functionally tested

iskakaushik (approved):

PR #1988, opened 2024-08-15
Clickhouse: change defaults for batch sizes

Amogh-Bharadwaj:
Change snapshot num rows per partition and max batch size defaults in UI to 100K instead of 1 Mil


iskakaushik:
1. Change CDC batch size to 500K
2. Change QRep batch size to 250L
3. Have an dyn conf var `PEERDB_CLICKHOUSE_EXPERIMENTAL_S3_IIS_PARTS` default to `1` if 1 do whatever is happening now, if > 2 use that as `numParts` for both CDC and Qrep.

heavycrystal:
probably better to use negative values/0 to indicate disabled and positive values to indicate `numParts`

Amogh-Bharadwaj:
QRep batch size is already 100K @iskakaushik

serprex (approved):

serprex:
https://github.com/ClickHouse/data-plane-application/pull/12971 I'll make PR

heavycrystal:
not necessary right now

PR #1992, opened 2024-08-19
Package generated protos for a CH service to call

pjhampton:
## Summary

This makes the protos available to a ClickHouse service that needs them to talk to the flow-api.

---

For now, I suggest we release this manually via

```sh
git tag flow/flow-api-client/v0.1.0
git push origin flow/flow-api-client/v0.1.0
```

and retrieve it in the internal codebase with

```sh
go get github.com/PeerDB-io/peerdb/flow/flow-api-client@v1.0.0
```

## Future work

I was planning to add a Github action to release this automatically, but this requires state management to maintain the version + bump, and I want to see if this is still the approach we adopt in a couple of weeks when we have everything working e2e.


iskakaushik (approved):

iamKunalGupta (commented):
Recommend adding a `.gitattributes` with the following content:
```
flow/flow-api-client/** linguist-generated=true
```

serprex:
this shows how go's repo-as-module breaks down unless you move this to some go build step. If repo must have build artifacts then there should instead be a github action on commit to main that pushes these generated files to a different repository which is entirely code generated

pjhampton:
> this shows how go's repo-as-module breaks down unless you move this to some go build step. If repo must have build artifacts then there should instead be a github action on commit to main that pushes these generated files to a different repository which is entirely code generated

I agree. Let's do that once we have stabilised the provisioning in the ingest platform api

PR #1994, opened 2024-08-19
ClickHouse: add certificate and private key fields for authentication

heavycrystal:
Tested validate peer with a self-signed CA and local ClickHouse
password can be used to authenticate after the certificates from both sides are validated, but it is optional now
TODO: need to add a way to specify custom root CAs instead of relying on the system store


iskakaushik (approved):

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -209,6 +209,18 @@ func Connect(ctx context.Context, config *protos.ClickhouseConfig) (clickhouse.C
...
 	}
+	if config.Certificate != nil || config.PrivateKey != nil {
+		if config.Certificate == nil || config.PrivateKey == nil {
+			return nil, errors.New("both certificate and private key must be provided if using certificate-based authentication")
+		}
+		cert, err := tls.X509KeyPair([]byte(*config.Certificate), []byte(*config.PrivateKey))
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse provided certificate: %w", err)
+		}
+		// TODO: find a way to modify list of root CAs as well
```
serprex, line 220: This can be left as a followup

serprex (approved):

PR #1997, opened 2024-08-20
Reenable Event Hubs CI

serprex:


heavycrystal:
still broken

serprex:
if we reenable this it'll probably be with eventhubs emulator

PR #1998, opened 2024-08-21
Peer page UI: Allow for editing when connectivity is down

Amogh-Bharadwaj:
Current design of this page where we await slot and stats data on the server side prevents us from editing peers when the connectivity to the peer is down - the page crashes

Refactored the wiring here to still render the page in this scenario and show what can be shown


Amogh-Bharadwaj:
Requires functional test

Amogh-Bharadwaj:
Realised the API calls are redundant here to v1/
Will fix

heavycrystal (approved):

PR #2001, opened 2024-08-21
nullable

serprex:
Allow nullable support to be enabled with env dynconf

Followup will want to allow per-mirror dynconf overrides. That'd make e2e testing easier


flow/peerdbenv/dynamicconf.go
```diff
@@ -76,6 +76,12 @@ END;`, ValueType: protos.DynconfValueType_STRING,
 		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_AFTER_RESUME,
 		TargetForSetting: protos.DynconfTarget_ALL,
 	},
+	{
+		Name: "PEERDB_NULLABLE", DefaultValue: "false", ValueType: protos.DynconfValueType_BOOL,
```
Amogh-Bharadwaj, line 80: I thought we want it be true by default? 
cc @iskakaushik @saisrirampur

heavycrystal: shouldn't be, perf impact
needs to be opt-in

flow/connectors/bigquery/qvalue_convert.go
```diff
@@ -12,7 +12,8 @@ import (
 
 func qValueKindToBigQueryType(columnDescription *protos.FieldDescription) bigquery.FieldSchema {
 	bqField := bigquery.FieldSchema{
-		Name: columnDescription.Name,
+		Name:     columnDescription.Name,
+		Required: columnDescription.NullableEnabled && columnDescription.Nullable,
```
heavycrystal, line 16: should be `!columnDescription.Nullable` I think

flow/peerdbenv/dynamicconf.go
serprex: also prefer to have new options not change existing behavior

flow/connectors/clickhouse/normalize.go
```diff
@@ -83,27 +83,30 @@ func generateCreateTableSQLForNormalizedTable(
...
+		} else {
 			stmtBuilder.WriteString(fmt.Sprintf("`%s` %s, ", colName, clickhouseType))
 		}
 	}
 	// TODO support soft delete
 	// synced at column will be added to all normalized tables
 	if syncedAtColName != "" {
 		colName := strings.ToLower(syncedAtColName)
-		stmtBuilder.WriteString(fmt.Sprintf("`%s` %s, ", colName, "DateTime64(9) DEFAULT now()"))
+		stmtBuilder.WriteString(fmt.Sprintf("`%s` %s, ", colName, "DateTime64(9) DEFAULT now64()"))
```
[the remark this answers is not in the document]
serprex: better to fix new raw tables in meantime so problem doesn't continue to grow

iskakaushik (approved):

PR #2003, opened 2024-08-22
add env to mirror configs, overriding dynconf

serprex:


heavycrystal:
wondering if we can use https://pkg.go.dev/context#WithValue at the top level and then read at the env calls

serprex:
> wondering if we can use https://pkg.go.dev/context#WithValue at the top level and then read at the env calls

we could, but I prefer having explicit data flow, context will be prone to breaking in weird ways

ui/app/mirrors/create/helpers/cdc.ts
```diff
@@ -227,4 +227,18 @@ export const cdcSettings: MirrorSetting[] = [
     tips: 'Disables columns like synced_at, soft_delete etc. from being added to the destination table',
     advanced: AdvancedSettingType.ALL,
   },
+  {
+    label: 'Settings override',
```
heavycrystal, line 231: should be more clear that it's JSON input here

heavycrystal (approved):
the diff hurts but you gotta do what you gotta do

Amogh-Bharadwaj:
Documentation on usage here please :)

PR #2004, opened 2024-08-24
don't alert for some common errors

heavycrystal:
1. `replState changed` is usually after something else happened, so no need to alert on its own
2. `slot is active` is annoyingly common, it autorecovers, don't alert

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -3,6 +3,7 @@ package activities
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"reflect"
@@ -169,7 +170,7 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 		// wait for the pull goroutine to finish
 		if err := errGroup.Wait(); err != nil {
 			// don't log flow error for "replState changed" and "slot is already active"
-			if !(temporal.IsApplicationError(err) || err == connpostgres.ErrSlotActive) {
+			if !(temporal.IsApplicationError(err) || errors.Is(err, connpostgres.ErrSlotActive)) {
 				a.Alerter.LogFlowError(ctx, flowName, err)
 			}
 			if temporal.IsApplicationError(err) {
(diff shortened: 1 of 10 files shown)


serprex (approved):

flow/activities/flowable_core.go
```diff
@@ -167,7 +167,10 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	if !hasRecords {
 		// wait for the pull goroutine to finish
 		if err := errGroup.Wait(); err != nil {
-			a.Alerter.LogFlowError(ctx, flowName, err)
+			// don't log flow error for "replState changed" and "slot is already active"
+			if !(temporal.IsApplicationError(err) || err == connpostgres.ErrSlotActive) {
```
serprex, line 171:
```suggestion
			if !(temporal.IsApplicationError(err) || errors.Is(err, connpostgres.ErrSlotActive)) {
```

flow/connectors/postgres/postgres.go
```diff
@@ -180,6 +182,9 @@ func (c *PostgresConnector) MaybeStartReplication(
 		defer c.replLock.Unlock()
 		if err := pglogrepl.StartReplication(ctx, c.replConn.PgConn(), slotName, startLSN, replicationOpts); err != nil {
 			c.logger.Error("error starting replication", slog.Any("error", err))
+			if strings.Contains(err.Error(), "is active for PID") {
```
serprex, line 185: check with pgerrors.Is(err, pgerrors.ObjectInUse)

But wondering if we can preserve original error & use pgerrors.Is to check at caller. Need to use errors.As to unwrap error into `*pgconn.PgError`

flow/connectors/clickhouse/qrep_avro_sync.go
```diff
@@ -39,7 +39,7 @@ func (s *ClickhouseAvroSyncMethod) CopyStageToDestination(ctx context.Context, a
 		return err
 	}
 
-	endpoint := s.connector.credsProvider.Provider.GetEndpointURL()
+	endpoint := "http://localhost:9001"
```
serprex, line 42: dev change?

flow/shared/postgres.go
```diff
@@ -125,3 +125,8 @@ func UpdateCDCConfigInCatalog(ctx context.Context, pool *pgxpool.Pool,
 	logger.Info("synced state to catalog: updated config_proto in flows", slog.String("flowName", cfg.FlowJobName))
 	return nil
 }
+
+func CheckSQLStateError(err error, sqlState string) bool {
```
serprex, line 129: can replace a lot of existing code to use this function, search pgerrcode

PR #2008, opened 2024-08-26
Clickhouse SetupNormalize: Include version and is_deleted as params

Amogh-Bharadwaj:
This PR:
- Includes the version column and peerdb_is_deleted as parameters of the [ReplacingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree#creating-a-table) engine
- Changes the variable name from signColName/Type to isDeletedColName/Type. Pretty sure signCol is for collapsing merge tree contexts.
- Changes the types of these to columns to UInt*, as that is needed to include them as params

- [ ] Functionally tested


serprex (approved):

serprex:
with #3066 it might be more feasible to do this in a backwards compatible way

serprex (approved):

PR #2010, opened 2024-08-27
Column Settings

serprex:
For the table box in the UI's table picker for CDC create flow:
- Have a field for selecting ReplacingMergeTree vs MergeTree
- Number field for each column. This will be used to order the columns in the ORDER BY clause while creating Clickhouse target tables

Screenshot 2024-08-30 at 2 42 24 AM


flow/connectors/clickhouse/normalize.go
```diff
@@ -67,65 +64,140 @@ func (c *ClickhouseConnector) SetupNormalizedTable(
...
+		if tableMapping != nil {
+			for _, col := range tableMapping.Columns {
+				if col.SourceName == colName {
+					columnSetting = col
+					if columnSetting.DestinationName != "" {
+						dstColName = columnSetting.DestinationName
+						colNameMap[colName] = dstColName
+					}
+					if columnSetting.DestinationType != "" {
+						// TODO can we restrict this to avoid injection?
```
heavycrystal, line 107: `slices.Contains(slices.Collect(maps.Values(qvalue.QValueKindToClickhouseTypeMap)), strings.TrimSpace(columnSetting.DestinationType))`

could further restrict to non-array types

protos/flow.proto
```diff
@@ -226,6 +236,11 @@ message PartitionRange {
   }
 }
 
+enum TableEngine {
```
heavycrystal, line 239: `CHTableEngine`

protos/flow.proto
```diff
@@ -226,6 +236,11 @@ message PartitionRange {
...
+  CH_ENGINE_REPLACING_MERGE_TREE = 0;
```
heavycrystal, line 240: `UNKNOWN` should be 0, to differentiate between unset and explicit set

nexus/flow-rs/src/grpc.rs
```diff
@@ -119,6 +119,8 @@ impl FlowGrpcClient {
                 destination_table_identifier: mapping.destination_table_identifier.clone(),
                 partition_key: mapping.partition_key.clone().unwrap_or_default(),
                 exclude: mapping.exclude.clone(),
+                columns: Default::default(),
+                engine: Default::default(),
```
heavycrystal, line 123: this should be `1` after proto changes are made

flow/connectors/clickhouse/normalize.go
```diff
@@ -67,65 +64,140 @@ func (c *ClickhouseConnector) SetupNormalizedTable(
...
 		stmtBuilder.WriteString(") ")
+	}
+
+	orderby := make([]*protos.ColumnSetting, len(tableMapping.Columns))
+	for _, col := range tableMapping.Columns {
+		if col.Ordering > 0 && !slices.Contains(pkeys, getColName(colNameMap, col.SourceName)) {
+			orderby = append(orderby, col)
+		}
+	}
+	slices.SortStableFunc(orderby, func(a *protos.ColumnSetting, b *protos.ColumnSetting) int {
```
heavycrystal, line 177: do we want to allow multiple columns with the same `.Ordering`?

protos/flow.proto
serprex: unknown is -1

nexus/flow-rs/src/grpc.rs
serprex: no, it would be unknown like existing mirrors

flow/connectors/clickhouse/normalize.go
serprex: Not going to go out the way to not handle it. SortStableFunc gives it a consistent result

protos/flow.proto
heavycrystal: can we make it explicit?

serprex: what benefit does that bring? either unknown is de facto ReplacingMergeTree for backwards compat, or unknown is an error & we break existing mirrors if they resync

flow/connectors/clickhouse/normalize.go
serprex: That will miss out on clickhouse types we don't map to like Int256

protos/flow.proto
heavycrystal: I don't like the idea of the Golang unset value equaling anything because it can mask coding errors later

The issue with existing mirrors is valid, I'd fix it by doing a one-time "correction" like we've done before, but it's iffy and painful in practice

serprex: embrace go's implicit nature, one time corrections are not something I want to make a habit of, they're error prone & put a burden on customers/support

Amogh-Bharadwaj:
With PEERDB_NULLABLE set, we get the following error when creating the table:
```
Sorting key contains nullable columns, but merge tree setting `allow_nullable_key` is disabled
```

Do we want to set `allow_nullable_key` or exclude the column if it is Nullable? Or something else

serprex:
For now we can leave it as unsupported

heavycrystal (approved):

PR #2014, opened 2024-08-28
Replace 32 sync limit with ContinueAsNewSuggested

serprex:
Experiment. Still unsure since long history impacts ui latency, but nice to have something that'll tune restart frequency based on workload


flow/workflows/sync_flow.go
```diff
@@ -195,7 +169,7 @@ func SyncFlowWorkflow(
 			break
 		}
 
-		restart := currentSyncFlowNum >= maxSyncsPerSyncFlow || syncErr
+		restart := syncErr || workflow.GetInfo(ctx).GetContinueAsNewSuggested()
```
serprex, line 172: need to check how this behaves, if cdc flow keeps going until it restarts we could end up with stuttering where reconnect happens twice within a short period (this restarts, then 10 syncs later cdc restarts)

heavycrystal:
another thing: SyncFlow also restarts independently (set by `maxSyncsPerSyncFlow`)
should handle that so this works properly

serprex:
tested with 3s batch timeout: continueasnew after >200 batches over 15m, 708440 history size, 3k events

PR #2017, opened 2024-08-29
Make resync more reliable for warehouse connectors

Amogh-Bharadwaj:
### Overview
Resync is becoming an increasingly utilized feature. Users of our connectors, particularly Clickhouse, sometimes wish to make small table definition changes on target or source and then want to repopulate data. This is alongside the usual recovery use-cases of resync.

### Failure points

`RenameTables` (or resync in general) currently can fail in the following ways:
1. If, between original mirror kick off and resync, a column was added to the table and a row not inserted after, then that column would not have been added to the target via schema changes. In this case, upon resync, the `_resync` table and the original table have different schemas, causing the soft-delete transfer step to fail. This can then lead to:
2. It processes some tables but not all and hits a failure - upon retrying it tries to resync the first table again but the `_resync `table of it is dropped since it succeeded before.
3. Resyncing again midway through a resync (initial load) can result in duplicate data in the `_resync` table if initial load of that `_resync` table was done already.

### Fixes
In light of these scenarios, this PR puts in place the following guards:

- Perform `CREATE OR REPLACE` in `SetupNormalize()` if it is being called in a resync
- If the `_resync` table does not exist, skip rename for the table
- If the original target table does not exist, skip just the soft-delete transfer step

TODO:
- [x] Functionally test the failure points mentioned above for the warehouse peers.
- [x] Also implement clearing of stats after resync : Done by #2029

### DIFF
--- a/flow/connectors/clickhouse/cdc.go
+++ b/flow/connectors/clickhouse/cdc.go
@@ -181,13 +181,10 @@ func (c *ClickhouseConnector) RenameTables(ctx context.Context, req *protos.Rena
 			}
 
 			allCols := strings.Join(columnNames, ",")
-			pkeyCols := strings.Join(renameRequest.TableSchema.PrimaryKeyColumns, ",")
 			c.logger.Info(fmt.Sprintf("handling soft-deletes for table '%s'...", renameRequest.NewName))
 			err = c.execWithLogging(ctx,
-				fmt.Sprintf("INSERT INTO %s(%s) SELECT %s,true AS %s FROM %s WHERE (%s) NOT IN (SELECT %s FROM %s)",
-					renameRequest.CurrentName, fmt.Sprintf("%s,%s", allCols, signColName), allCols,
-					signColName,
-					renameRequest.NewName, pkeyCols, pkeyCols, renameRequest.CurrentName))
+				fmt.Sprintf("INSERT INTO %s(%s,%s) SELECT %s,true FROM %s WHERE %s  = 1",
+					renameRequest.CurrentName, allCols, signColName, allCols, renameRequest.NewName, signColName))
 			if err != nil {
 				return nil, fmt.Errorf("unable to handle soft-deletes for table %s: %w", renameRequest.NewName, err)
 			}
--- a/flow/connectors/clickhouse/cdc.go
+++ b/flow/connectors/clickhouse/cdc.go
@@ -159,21 +159,40 @@ func (c *ClickhouseConnector) ReplayTableSchemaDeltas(ctx context.Context, flowJ
 
 func (c *ClickhouseConnector) RenameTables(ctx context.Context, req *protos.RenameTablesInput) (*protos.RenameTablesOutput, error) {
 	for _, renameRequest := range req.RenameTableOptions {
-		columnNames := make([]string, 0, len(renameRequest.TableSchema.Columns))
-		for _, col := range renameRequest.TableSchema.Columns {
-			columnNames = append(columnNames, col.Name)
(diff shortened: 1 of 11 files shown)


flow/connectors/clickhouse/cdc.go
```diff
@@ -159,21 +159,40 @@ func (c *ClickhouseConnector) ReplayTableSchemaDeltas(ctx context.Context, flowJ
...
+			columnNames := make([]string, 0, len(renameRequest.TableSchema.Columns))
+			for _, col := range renameRequest.TableSchema.Columns {
+				columnNames = append(columnNames, col.Name)
+			}
+
+			allCols := strings.Join(columnNames, ",")
+			pkeyCols := strings.Join(renameRequest.TableSchema.PrimaryKeyColumns, ",")
+			c.logger.Info(fmt.Sprintf("handling soft-deletes for table '%s'...", renameRequest.NewName))
+			err = c.execWithLogging(ctx,
+				fmt.Sprintf("INSERT INTO %s(%s) SELECT %s,true AS %s FROM %s WHERE (%s) NOT IN (SELECT %s FROM %s)",
```
Amogh-Bharadwaj, line 187: The soft delete transfer step for Clickhouse here seems wrong

flow/connectors/clickhouse/normalize.go
```diff
@@ -72,9 +74,16 @@ func generateCreateTableSQLForNormalizedTable(
...
 	syncedAtColName string,
+	isResync bool,
 ) (string, error) {
 	var stmtBuilder strings.Builder
-	stmtBuilder.WriteString(fmt.Sprintf("CREATE TABLE IF NOT EXISTS `%s` (", normalizedTable))
+	stmtBuilder.WriteString("CREATE ")
+	if isResync {
+		stmtBuilder.WriteString("OR REPLACE ")
+	}
+	stmtBuilder.WriteString("TABLE IF NOT EXISTS `")
```
heavycrystal, line 84: if we are doing `CREATE OR REPLACE`, `IF NOT EXISTS` doesn't make sense

flow/connectors/postgres/client.go
```diff
@@ -34,13 +34,14 @@ const (
...
-	getLastNormalizeBatchID_SQL = "SELECT normalize_batch_id FROM %s.%s WHERE mirror_job_name=$1"
-	createNormalizedTableSQL    = "CREATE TABLE IF NOT EXISTS %s(%s)"
-
-	upsertJobMetadataForSyncSQL = `INSERT INTO %s.%s AS j VALUES ($1,$2,$3,$4)
+	getLastOffsetSQL                  = "SELECT lsn_offset FROM %s.%s WHERE mirror_job_name=$1"
+	setLastOffsetSQL                  = "UPDATE %s.%s SET lsn_offset=GREATEST(lsn_offset, $1) WHERE mirror_job_name=$2"
+	getLastSyncBatchID_SQL            = "SELECT sync_batch_id FROM %s.%s WHERE mirror_job_name=$1"
+	getLastNormalizeBatchID_SQL       = "SELECT normalize_batch_id FROM %s.%s WHERE mirror_job_name=$1"
+	createNormalizedTableSQL          = "CREATE TABLE IF NOT EXISTS %s(%s)"
+	createOrReplaceNormalizedTableSQL = "CREATE OR REPLACE TABLE IF NOT EXISTS %s(%s)"
```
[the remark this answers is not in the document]
serprex: https://www.postgresql.org/docs/current/sql-createtable.html syntax isn't here

iskakaushik (approved):

PR #2020, opened 2024-08-29
Adapt validate mirror for initial load only

Amogh-Bharadwaj:
If initial load only is selected in UI, allow non-mirrorable tables to be synced, and validation will not check replication and publication related things.

Functionally tested

One item I tried was also allowing excluding primary key columns when initial load only is selected, but seems like this requires some refactoring on the pull side. Could do it in this PR or follow up

### DIFF
--- a/flow/cmd/validate_mirror.go
+++ b/flow/cmd/validate_mirror.go
@@ -79,7 +79,7 @@ func (h *FlowRequestHandler) ValidateCDCMirror(
 		if err != nil {
 			displayErr := fmt.Errorf("unable to establish replication connectivity: %v", err)
 			h.alerter.LogNonFlowWarning(ctx, telemetry.CreateMirror, req.ConnectionConfigs.FlowJobName,
-				fmt.Sprint(displayErr),
+				displayErr.Error(),
 			)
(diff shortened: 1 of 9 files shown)


flow/cmd/validate_mirror.go
```diff
@@ -72,31 +72,33 @@ func (h *FlowRequestHandler) ValidateCDCMirror(
...
-		}, displayErr
-	}
+	noCDC := req.ConnectionConfigs.DoInitialSnapshot && req.ConnectionConfigs.InitialSnapshotOnly
+	if !noCDC {
+		// Check replication connectivity
+		err = pgPeer.CheckReplicationConnectivity(ctx)
+		if err != nil {
+			displayErr := fmt.Errorf("unable to establish replication connectivity: %v", err)
+			h.alerter.LogNonFlowWarning(ctx, telemetry.CreateMirror, req.ConnectionConfigs.FlowJobName,
+				fmt.Sprint(displayErr),
```
serprex, line 82:
```suggestion
				displayErr.Error(),
```

serprex (approved):

Issue #2019, opened 2024-08-29
[clickhouse] Add validation during mirror creation to check if destination tables exist and if they do, are valid to sync to


heavycrystal:
Cases:

- [ ] Check if table doesn't exist, if so proceed.
Table exists:
- [ ] Columns should match what is in Postgres incl. column exclusion.
- [ ] Presence of metadata columns.
- [ ] Types should match for columns.
- [ ] Table should be empty.
- [ ] Table should be `ReplacingMergeTree`

PR #2026, opened 2024-08-30
[clickhouse] validate certain aspects of existing destination tables

heavycrystal:
Partially addresses #2019 

Tests column name equality taking exclusion into account, PeerDB columns, engine and emptiness

### DIFF
--- a/flow/connectors/clickhouse/clickhouse.go
+++ b/flow/connectors/clickhouse/clickhouse.go
@@ -344,13 +344,17 @@ func (c *ClickhouseConnector) processTableComparison(dstTableName string, srcSch
 ) error {
 	for _, srcField := range srcSchema.Columns {
 		colName := srcField.Name
+		// if the column is mapped to a different name, find and use that name instead
 		for _, col := range tableMapping.Columns {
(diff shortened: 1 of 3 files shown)


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -273,3 +273,136 @@ func (c *ClickhouseConnector) execWithLogging(ctx context.Context, query string)
...
+}
+
+func (c *ClickhouseConnector) processTableComparison(dstTableName string, srcSchema *protos.TableSchema,
+	dstSchema []*protos.FieldDescription, peerDBColumns []string,
+) error {
+	for _, srcField := range srcSchema.Columns {
+		found := false
+		for _, dstField := range dstSchema {
+			// not doing type checks for now
+			if srcField.Name == dstField.Name {
```
serprex, line 349: #2010 makes it possible to override destination column name

heavycrystal: will hold off if it lands soon, else follow up

serprex (approved):

PR #2028, opened 2024-09-02
raise NonRetryable errors with some replication issues

heavycrystal:


serprex (approved):
make sure that these errors aren't getting wrapped in by fmt.Errorf by activity/callers, otherwise temporal won't pick up that it's a non retryable error

heavycrystal:
> make sure that these errors aren't getting wrapped in by fmt.Errorf by activity/callers, otherwise temporal won't pick up that it's a non retryable error

I think Temporal matches using `errors.As/Is` internally, tested one of the cases and it failed the workflow as expected

iskakaushik (approved):

PR #2029, opened 2024-09-02
Drop mirror : Drop stats option

Amogh-Bharadwaj:
Screenshot 2024-09-02 at 6 36 57 PM

This PR:
- Adds drop mirror stats (present in `peerdb_stats` schema) functionality to drop mirror. The failure of this step is not fatal to `drop_flow`. Exposed in UI as seen above.
- Resync now drops mirror stats
- In query layer, drop mirror does not drop stats. Extending the drop mirror command syntax to support this will be done if there's demand

Also:
- Adds a column - `parent_mirror_name (TEXT)` to `qrep_runs` and `qrep_partitions` tables. 
- Adds a field - `parentMirrorName (string)` to QRep Config proto
-  Adaps `monitoring.go` and initial load stats endpoint to use parent mirror name

Parent mirror name is the name of the parent peer flow in case of initial load clone. It is the same as flow name for QRep and XMin mirrors

- [ ] Functionally tested


iskakaushik (approved):

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -317,3 +317,32 @@ func UpdateRowsSyncedForPartition(ctx context.Context, pool *pgxpool.Pool, rowsS
 	}
 	return nil
 }
+
+func DeleteMirrorStats(ctx context.Context, pool *pgxpool.Pool, flowJobName string) error {
+	_, err := pool.Exec(ctx, `DELETE FROM peerdb_stats.qrep_partitions WHERE flow_name ILIKE 'clone\_' || replace(replace($1, '_', '\_'), '%', '\%') || '\_%'`, flowJobName)
```
Amogh-Bharadwaj, line 322: Runs into the case where two mirrors are substrings of each other
Need a new column - `mirror_name` or `parent_mirror` - in the qrep tables

Amogh-Bharadwaj: Fixed and updated description

PR #2040, opened 2024-09-04
Existing tables check in CH: lower case synced_at column

Amogh-Bharadwaj:
Validation to existing table which peerdb created fails because in SetupNormalize we lower case synced_at column. So that needs to be done in validation as well because by default it is upper case PEERDB_SYNCED_AT


Amogh-Bharadwaj:
Functionally tested

heavycrystal (approved):

PR #2056, opened 2024-09-08
[clickhouse] make table engines a warning instead of an error

heavycrystal:
There's engines like Distributed and other MergeTree family which could also be used in special workloads. Until we have more understanding on how these work and how common they are, don't block creation of mirrors.


woodhull:
👍 

I'm a PeerDB cloud user and the product is currently broken for me as a result.

iskakaushik (approved):

PR #2068, opened 2024-09-12
Advance LSN when PKM requested before any records received

iskakaushik:
When messages are received while cdc storage is empty we do not need to worry about confirming lsn too early before destination has processed previous records, so go ahead & progress lsn

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -16,6 +16,7 @@ import (
 	"github.com/lib/pq/oid"
 	"go.temporal.io/sdk/activity"
 
+	"github.com/PeerDB-io/peer-flow/connectors/external_metadata"
 	"github.com/PeerDB-io/peer-flow/connectors/utils"
 	geo "github.com/PeerDB-io/peer-flow/datatypes"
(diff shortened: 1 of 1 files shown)


serprex (approved):

flow/connectors/postgres/cdc.go
```diff
@@ -362,6 +362,10 @@ func PullCdcRecords[Items model.Items](
 
 	for {
 		if pkmRequiresResponse {
+			if cdcRecordsStorage.IsEmpty() {
+				req.ConsumedOffset.Store(int64(clientXLogPos))
```
serprex, line 366: ConsumedOffset must only be updated after catalog updated, otherwise pg may remove wal & restart will request old value in catalog

serprex (changes requested):
changes needed

flow/connectors/postgres/cdc.go
iskakaushik: makes sense

heavycrystal:
seems like a risky optimization

serprex (approved):

PR #2084, opened 2024-09-19
Numeric type: improved interface, fix Clickhouse support

Amogh-Bharadwaj:
This PR:
- Fixes casting of numeric values in Normalize insert-into-select statements in the Clickhouse connector
- Changes the behaviour of our default precision and scale setting. We now first check if we can reduce the precision to the maximum supported value, and only if not possible do we set it to our default. So for example for a numeric column on Postgres of type `numeric(78,70)`:

Previous behaviour for Clickhouse:
Since 78 is higher than the maximum supported precision on Clickhouse, we would set it to PeerDB's chosen default precision (76) and default scale (38)

New behaviour:
We first see that the precision is above max so we just change precision to the max of 76. 
Then we check if the new (precision, scale) tuple is valid. If not, then fallback to the behaviour from before and set it to defaults chosen by PeerDB. In this case it will now be valid, so the numeric on target will be `Decimal(76, 70)`. Notice how we do not change the scale unlike above

Motivation:
This allows us to now instead of setting a numeric like (78,0) to (76,36), to now just set it to (76,0) - which is more supportive than (76,36) when the incoming numerics are not going to have any fractional parts

- [x] Functionally tested

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -300,14 +300,15 @@ func (c *ClickhouseConnector) NormalizeRecords(
 
 			colSelector.WriteString(fmt.Sprintf("`%s`,", dstColName))
 			if clickhouseType == "" {
-				var err error
-				clickhouseType, err = colType.ToDWHColumnType(protos.DBType_CLICKHOUSE)
-				if err != nil {
-					return nil, fmt.Errorf("error while converting column type to clickhouse type: %w", err)
-				}
 				if colType == qvalue.QValueKindNumeric {
 					precision, scale := datatypes.GetNumericTypeForWarehouse(column.TypeModifier, datatypes.ClickHouseNumericCompatibility{})
 					clickhouseType = fmt.Sprintf("Decimal(%d, %d)", precision, scale)
+				} else {
+					var err error
+					clickhouseType, err = colType.ToDWHColumnType(protos.DBType_CLICKHOUSE)
+					if err != nil {
+						return nil, fmt.Errorf("error while converting column type to clickhouse type: %w", err)
+					}
 				}
 			}
 
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -305,6 +305,10 @@ func (c *ClickhouseConnector) NormalizeRecords(
 				if err != nil {
 					return nil, fmt.Errorf("error while converting column type to clickhouse type: %w", err)
 				}
+				if colType == qvalue.QValueKindNumeric {
+					precision, scale := datatypes.GetNumericTypeForWarehouse(column.TypeModifier, datatypes.ClickHouseNumericCompatibility{})
+					clickhouseType = fmt.Sprintf("Decimal(%d, %d)", precision, scale)
+				}
 			}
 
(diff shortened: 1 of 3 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -305,6 +305,10 @@ func (c *ClickhouseConnector) NormalizeRecords(
 				if err != nil {
 					return nil, fmt.Errorf("error while converting column type to clickhouse type: %w", err)
 				}
+				if colType == qvalue.QValueKindNumeric {
```
serprex, line 308: this can be if/else, since it disregards the above assignment to clickhouseType. Ideally `ToDWHColumnType` would take `column` & handle this

flow/datatypes/numeric.go
```diff
@@ -28,12 +27,20 @@ func (ClickHouseNumericCompatibility) MaxScale() int16 {
...
+func (c ClickHouseNumericCompatibility) IsValidPrecisionAndScale(precision, scale int16) bool {
+	return c.IsValidPrecision(precision) && c.IsValidScale(scale) && scale < precision
+}
+
+func (c ClickHouseNumericCompatibility) IsValidPrecision(precision int16) bool {
+	return precision > 0 && precision <= c.MaxPrecision()
+}
+
+func (ClickHouseNumericCompatibility) IsValidScale(scale int16) bool {
+	return scale >= 0
```
serprex, line 43: this isn't accurate, & if this is what we want this method seems unnecessary

flow/datatypes/numeric.go
```diff
@@ -113,8 +143,17 @@ func GetNumericTypeForWarehouse(typmod int32, warehouseNumeric WarehouseNumericC
 	}
 
 	precision, scale := ParseNumericTypmod(typmod)
+	if !warehouseNumeric.IsValidPrecision(precision) {
+		precision = warehouseNumeric.MaxPrecision()
+	}
+
+	if !warehouseNumeric.IsValidScale(scale) {
+		_, defaultScale := warehouseNumeric.DefaultPrecisionAndScale()
+		scale = defaultScale
```
serprex, line 152:
```suggestion
		_, scale = warehouseNumeric.DefaultPrecisionAndScale()
```
`_` is not just a throwaway identifier in Go

flow/datatypes/numeric.go
```diff
@@ -46,12 +53,20 @@ func (SnowflakeNumericCompatibility) MaxScale() int16 {
...
+	return s.MaxPrecision(), PeerDBSnowflakeScale
+}
+
+func (s SnowflakeNumericCompatibility) IsValidPrecisionAndScale(precision, scale int16) bool {
+	return s.IsValidPrecision(precision) && s.IsValidScale(scale) && scale < precision
 }
 
-func (SnowflakeNumericCompatibility) IsValidPrecisionAndScale(precision, scale int16) bool {
-	return precision > 0 && precision <= 38 && scale < precision
+func (s SnowflakeNumericCompatibility) IsValidPrecision(precision int16) bool {
```
serprex, line 59: when these methods all have same result they can instead take be free functions taking `WarehouseNumericCompatibility`

serprex (approved):

PR #2088, opened 2024-09-20
kafka: use activity logger instead of slog.Default for kafka client logger

serprex:


Amogh-Bharadwaj (approved):

flow/connectors/kafka/kafka.go
```diff
@@ -34,16 +34,46 @@ type KafkaConnector struct {
...
+func kgoLogger(logger log.Logger) kgo.Logger {
+	if sl, ok := logger.(*slog.Logger); ok {
+		return kslog.New(sl)
+	} else {
+		return kgoTemporalLogger{Logger: logger}
+	}
+}
+
+func (logger kgoTemporalLogger) Level() kgo.LogLevel {
+	return kgo.LogLevelInfo
```
serprex, line 50:
```suggestion
	return kslog.New(slog.Default()).Level()
```
Might be more accurate, would probably want to store result of this hack somewhere. But this should be fine & the above would not be necessarily correct

PR #2091, opened 2024-09-21
refactor numeric code and handle typmod better

heavycrystal:
started with numeric refactor, but discovered typmod is being handled improperly in some places
bumped BQ to BIGNUMERIC limits
added tests for typmod parsing


flow/datatypes/numeric.go
```diff
@@ -1,111 +1,163 @@
...
-	return 37
-}
-
-func (DefaultNumericCompatibility) DefaultPrecisionAndScale() (int16, int16) {
-	return 38, 20
-}
+// This is to reverse what make_numeric_typmod of Postgres does:
+// logic copied from: https://github.com/postgres/postgres/blob/c4d5cb71d229095a39fda1121a75ee40e6069a2a/src/backend/utils/adt/numeric.c#L929
+// Maps most "invalid" typmods to be unconstrained (same as -1)
+func NewParsedNumericTypmod(typmod int32) *NumericTypmod {
```
serprex, line 104:
```suggestion
func NewParsedNumericTypmod(typmod int32) NumericTypmod {
```
Seems like a type best passed by value & treated as immutable

flow/datatypes/numeric.go
```diff
@@ -1,111 +1,163 @@
...
 }
 
-// This is to reverse what make_numeric_typmod of Postgres does:
-// https://github.com/postgres/postgres/blob/21912e3c0262e2cfe64856e028799d6927862563/src/backend/utils/adt/numeric.c#L897
-func ParseNumericTypmod(typmod int32) (int16, int16) {
-	offsetMod := typmod - VARHDRSZ
-	precision := int16((offsetMod >> 16) & 0x7FFF)
-	scale := int16(offsetMod & 0x7FFF)
-	return precision, scale
+func (t *NumericTypmod) PrecisionAndScale() (int16, int16) {
```
serprex, line 143: Can make Precision/Scale public & remove this. No need for nil checks if you usually work with these by value

flow/datatypes/numeric.go
serprex: That'd be handled by a Valid bool

PR #2092, opened 2024-09-22
[cdc] update not present columns from oldtuple as well

heavycrystal:
rationale in comment

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -16,7 +16,7 @@ import (
 	"github.com/lib/pq/oid"
 	"go.temporal.io/sdk/activity"
 
-	"github.com/PeerDB-io/peer-flow/connectors/external_metadata"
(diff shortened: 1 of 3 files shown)


flow/connectors/postgres/cdc.go
```diff
@@ -501,6 +501,7 @@ func PullCdcRecords[Items model.Items](
 					// tableName here is destination tableName.
 					// should be ideally sourceTableName as we are in PullRecords.
 					// will change in future
+					// TODO: replident is cached here, should not cache since it can change
```
serprex, line 504: what do you mean by cached? is there no relation message for this?

serprex (approved):

flow/connectors/postgres/cdc.go
heavycrystal: it's populated from SetupFlow, and I'm not sure if we get this as a relation message

PR #2100, opened 2024-09-25
fix more !BADKEY logs

serprex:
after #2099 I ran a grepped a few more instances of this issue


flow/e2e/postgres/qrep_flow_pg_test.go
```diff
@@ -166,7 +166,7 @@ func (s PeerFlowE2ETestSuitePG) TestSimpleSlotCreation() {
 		setupError <- s.conn.SetupReplication(context.Background(), signal, setupReplicationInput)
 	}()
 
-	s.t.Log("waiting for slot creation to complete: ", flowJobName)
+	s.t.Log("waiting for slot creation to complete: " + flowJobName)
```
heavycrystal, line 169: can use `slog.String`?

flow/workflows/cdc_flow.go
```diff
@@ -289,7 +289,7 @@ func addCdcPropertiesSignalListener(
 	cdcPropertiesSignalChan.AddToSelector(selector, func(cdcConfigUpdate *protos.CDCFlowConfigUpdate, more bool) {
 		// do this irrespective of additional tables being present, for auto unpausing
 		state.FlowConfigUpdate = cdcConfigUpdate
-		logger.Info("CDC Signal received. Parameters on signal reception:",
+		logger.Info("CDC Signal received",
 			slog.Int("BatchSize", int(state.SyncFlowOptions.BatchSize)),
 			slog.Int("IdleTimeout", int(state.SyncFlowOptions.IdleTimeoutSeconds)),
 			slog.Any("AdditionalTables", cdcConfigUpdate.AdditionalTables),
```
heavycrystal, line 295: unrelated but can log `UpdatedEnv` as well, I missed this

heavycrystal (approved):

flow/e2e/postgres/qrep_flow_pg_test.go
serprex: s.t.Log isn't slog, this is testing. Moved to concat since it was a false positive on my scan

PR #2101, opened 2024-09-25
[solar] make sync flow restart less often

heavycrystal:


flow/workflows/sync_flow.go
```diff
@@ -18,7 +18,7 @@ import (
 // set this value high enough to never be met, relying on cdc restarts.
 // In the future cdc flow restarts could be decoupled from sync flow restarts.
 const (
-	maxSyncsPerSyncFlow = 64
+	maxSyncsPerSyncFlow = 514
```
serprex, line 21:
```suggestion
	maxSyncsPerSyncFlow = 256
```

serprex: 514 is intentional

PR #2102, opened 2024-09-25
Clickhouse Raw Table: Change order by to batch ID

Amogh-Bharadwaj:
This PR changes the table creation command of the raw table for Clickhouse. Instead of ordering by `_peerdb_uid`, we now order by `_peerdb_batch_id` 
The goal is to improve the performance of SELECTs from the raw table during normalize

### DIFF
--- a/flow/alerting/alerting.go
+++ b/flow/alerting/alerting.go
@@ -276,7 +276,7 @@ func (a *Alerter) checkAndAddAlertToCatalog(ctx context.Context, alertConfigId i
 	var createdTimestamp time.Time
 	err = row.Scan(&createdTimestamp)
 	if err != nil && err != pgx.ErrNoRows {
-		logger.LoggerFromCtx(ctx).Warn("failed to send alert: ", slog.String("err", err.Error()))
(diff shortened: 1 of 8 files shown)


flow/connectors/clickhouse/cdc.go
```diff
@@ -54,7 +54,7 @@ func (c *ClickHouseConnector) CreateRawTable(ctx context.Context, req *protos.Cr
 		_peerdb_match_data String,
 		_peerdb_batch_id Int,
 		_peerdb_unchanged_toast_columns String
-	) ENGINE = ReplacingMergeTree ORDER BY _peerdb_uid;`
+	) ENGINE = ReplacingMergeTree ORDER BY _peerdb_batch_id;`
```
iskakaushik, line 57: should we also include destination table name? cc: @heavycrystal

heavycrystal: will test and add

Amogh-Bharadwaj:
following here now: https://github.com/PeerDB-io/peerdb/pull/2104

PR #2107, opened 2024-09-26
RenameTableOptions: remove TableSchema

serprex:
In #2090 I got so caught up fixing bug @heavycrystal pointed out that I forgot I wanted to clean the code out of existence

This has no change in behavior, just cleans up protobuf so there isn't this weird 'workflow send nils into activity, activity fills in at start' protocol

### DIFF
--- a/flow/connectors/core.go
+++ b/flow/connectors/core.go
@@ -265,7 +265,7 @@ type RawTableConnector interface {
 type RenameTablesConnector interface {
 	Connector
 
-	RenameTables(context.Context, *protos.RenameTablesInput) (*protos.RenameTablesOutput, error)
+	RenameTables(context.Context, *protos.RenameTablesInput, map[string]*protos.TableSchema) (*protos.RenameTablesOutput, error)
 }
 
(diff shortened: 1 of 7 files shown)


heavycrystal (approved):

flow/connectors/core.go
```diff
@@ -265,7 +265,7 @@ type RawTableConnector interface {
 type RenameTablesConnector interface {
 	Connector
 
-	RenameTables(context.Context, *protos.RenameTablesInput) (*protos.RenameTablesOutput, error)
+	RenameTables(context.Context, *protos.RenameTablesInput, map[string]*protos.TableSchema) (*protos.RenameTablesOutput, error)
```
serprex, line 268: Another option, that'd avoid these map lookups, would be to have this method take a `[]model.RenameTableConfig` which we convert to from `*protos.RenameTablesInput`. It'd be `struct {CurrentName string; NewName string; TableSchema *protos.TableSchema}`

heavycrystal: I'd prefer this

PR #2111, opened 2024-09-30
chore: update dependencies

serprex:


iskakaushik:
@serprex  should we update golang version as well?

iskakaushik (approved):

PR #2112, opened 2024-09-30
_peerdb_uid: be uuid

serprex:


heavycrystal:
is there a reason we're doing this or is this just to look cleaner?

flow/connectors/postgres/postgres.go
```diff
@@ -451,7 +451,7 @@ func syncRecordsCore[Items model.Items](
 				}
 
 				row = []any{
-					uuid.New().String(),
+					uuid.New(),
```
heavycrystal, line 454: should test if switching this with the old raw table syntax will break syncs
ideally we'd have a fallback or a migration

flow/connectors/utils/stream.go
```diff
@@ -121,7 +121,7 @@ func recordToQRecordOrError[Items model.Items](batchID int64, record model.Recor
 		return nil, fmt.Errorf("unknown record type: %T", typedRecord)
 	}
 
-	entries[0] = qvalue.QValueString{Val: uuid.New().String()}
+	entries[0] = qvalue.QValueUUID{Val: uuid.New()}
```
heavycrystal, line 124: same here

serprex:
> is there a reason we're doing this or is this just to look cleaner?

it also fits in storage cleaner

flow/connectors/postgres/postgres.go
serprex: tested with old schema, postgres can handle inserting uuid into text column

iskakaushik:
@serprex existing mirrors in CH will continue to work after this and doesn't require any special migration i'm assuming?

heavycrystal (approved):

flow/connectors/utils/stream.go
heavycrystal: SF and BQ as well, unless we're not upgrading customers using them

serprex:
> @serprex existing mirrors in CH will continue to work after this and doesn't require any special migration i'm assuming?

correct

PR #2113, opened 2024-09-30
clickhouse: fix normalizing updates to primary key

serprex:
when primary key updated row becomes duplicate because old version isn't marked outdated,
generate deletion records behind updates

on an update heavy workload where primary keys aren't updated this can be disabled with `PEERDB_CLICKHOUSE_DISABLE_PRIMARY_UPDATE` to omit `UNION ALL`

if custom ordering is on a non replica identity column then row data may be missing,
causing this to fail to prevent duplication

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -15,6 +15,7 @@ import (
 	"github.com/PeerDB-io/peer-flow/generated/protos"
 	"github.com/PeerDB-io/peer-flow/model"
 	"github.com/PeerDB-io/peer-flow/model/qvalue"
+	"github.com/PeerDB-io/peer-flow/peerdbenv"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


serprex:
Given updating primary keys is quite rare (difficult to do with pg if you use foreign keys) it may be better to have this disabled by default

flow/connectors/clickhouse/normalize.go
```diff
@@ -321,23 +326,43 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
 		// add _peerdb_sign as _peerdb_record_type / 2
 		projection.WriteString(fmt.Sprintf("intDiv(_peerdb_record_type, 2) AS `%s`,", signColName))
+		// projectionUpdate generates delete on previous record, so _peerdb_record_type is filled in as 2
+		projectionUpdate.WriteString(fmt.Sprintf("1 AS `%s`,", signColName))
 		colSelector.WriteString(fmt.Sprintf("`%s`,", signColName))
 
 		// add _peerdb_timestamp as _peerdb_version
 		projection.WriteString(fmt.Sprintf("_peerdb_timestamp AS `%s`", versionColName))
+		// decrement timestamp by 1 so delete is ordered before update's new data
+		projectionUpdate.WriteString(fmt.Sprintf("_peerdb_timestamp - 1 AS `%s`", versionColName))
```
serprex, line 365: in theory this could be weird if two updates happened after each other by 1 tick with the second update on the primary key

heavycrystal (approved):
the perf impact of this seems noticeable
should be opt-in imo

PR #2114, opened 2024-09-30
Clickhouse: add test for schema changes

Amogh-Bharadwaj:
Quote target table name and add an E2E test for schema changes


flow/e2e/clickhouse/peer_flow_ch_test.go
```diff
@@ -263,3 +263,81 @@ func (s ClickHouseSuite) Test_Date32() {
 	env.Cancel()
 	e2e.RequireEnvCanceled(s.t, env)
 }
+
+func (s ClickHouseSuite) Test_Schema_Changes_CH() {
```
serprex, line 267: If CH implements `connectors.GetTableSchemaConnector` then generic schema changes test will be activated for CH

serprex (approved):

PR #2116, opened 2024-09-30
[clickhouse] defer dropping table during ValidateMirror

heavycrystal:
We're observing that this function (called while creating or updating a peer to validate it) can sometimes fail midway and leave this table around, trying to reduce the possibility of that

### DIFF
--- a/flow/connectors/clickhouse/clickhouse.go
+++ b/flow/connectors/clickhouse/clickhouse.go
@@ -114,7 +114,7 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
 	); err != nil {
 		return fmt.Errorf("failed to rename validation table %s: %w", validateDummyTableName, err)
 	}
-	validateDummyTableName = validateDummyTableName + "_renamed"
+	validateDummyTableName += "_renamed"
 
 	// insert a row
 	if err := c.exec(ctx, fmt.Sprintf("INSERT INTO %s VALUES (1, now64())", validateDummyTableName)); err != nil {
--- a/flow/connectors/clickhouse/clickhouse.go
+++ b/flow/connectors/clickhouse/clickhouse.go
@@ -83,22 +83,11 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
 		return fmt.Errorf("failed to create validation table %s: %w", validateDummyTableName, err)
 	}
 	defer func() {
-		// goroutine so API call doesn't hang
-		go func() {
-			// drop the table at any cost
-			for i := 1; i <= 5; i++ {
-				err = c.exec(ctx, "DROP TABLE IF EXISTS "+validateDummyTableName)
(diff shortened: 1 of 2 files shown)


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -77,24 +77,43 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
...
+		validateDummyTableName))
 	if err != nil {
 		return fmt.Errorf("failed to create validation table %s: %w", validateDummyTableName, err)
 	}
+	defer func() {
+		// goroutine so API call doesn't hang
+		go func() {
+			// drop the table at any cost
+			for i := 1; i <= 5; i++ {
+				err = c.database.Exec(ctx, "DROP TABLE IF EXISTS "+validateDummyTableName)
```
serprex, line 89: this will race on using connection if `c.database` is used after this method is called. Better to not use goroutine

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -77,24 +77,43 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
 	err := c.database.Exec(ctx, fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s (
 		id UInt64
 	) ENGINE = ReplacingMergeTree ORDER BY id;`,
```
serprex, line 79:
```suggestion
	) ENGINE = Memory ORDER BY id;`,
```
Are permissions engine specific?

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -77,24 +77,43 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
 	err := c.database.Exec(ctx, fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s (
```
serprex, line 77:
```suggestion
	err := c.database.Exec(ctx, fmt.Sprintf(`CREATE TEMPORARY TABLE IF NOT EXISTS %s (
```
This would have table cleaned up when connector disposed

flow/connectors/clickhouse/clickhouse.go
heavycrystal: unsure on this, will leave as is for now

flow/connectors/clickhouse/clickhouse.go
heavycrystal: can be worked around since the connector only exists for the duration of the API call
but `TEMPORARY` would work better

serprex: the better way would be to have caller responsible for returning status to user & on it's own it'd handle drop table before dropping connector

serprex:
Possibly replaced by #2119

serprex (approved):

PR #2120, opened 2024-10-01
clickhouse validation: use temporary table

serprex:
avoids leaking table due to errors

### DIFF
--- a/flow/connectors/clickhouse/clickhouse.go
+++ b/flow/connectors/clickhouse/clickhouse.go
@@ -75,44 +75,34 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
 	}
 	validateDummyTableName := "peerdb_validation_" + shared.RandomString(4)
 	// create a table
-	err := c.exec(ctx, fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s (
-		id UInt64
-	) ENGINE = ReplacingMergeTree ORDER BY id;`,
-		validateDummyTableName+"_temp"))
-	if err != nil {
+	if err := c.exec(ctx,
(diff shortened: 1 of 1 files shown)


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -75,44 +75,34 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
 	}
 	validateDummyTableName := "peerdb_validation_" + shared.RandomString(4)
 	// create a table
-	err := c.exec(ctx, fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s (
-		id UInt64
-	) ENGINE = ReplacingMergeTree ORDER BY id;`,
-		validateDummyTableName+"_temp"))
-	if err != nil {
+	if err := c.exec(ctx,
+		fmt.Sprintf(`CREATE TEMPORARY TABLE IF NOT EXISTS %s (id UInt64) ENGINE = ReplacingMergeTree ORDER BY id;`, validateDummyTableName),
```
iskakaushik, line 79: given that the intent is to validate the grants, can we confirm that create temp table and create table have the same grants?

serprex: Looking over, `CREATE TEMPORARY TABLE` & `CREATE TABLE` are separate

https://clickhouse.com/docs/en/sql-reference/statements/show#show-grants

Probably best to use `SHOW GRANTS` to minimize pollution. But have to recursively run on roles granted to user

heavycrystal: should we combine a normal table with a goroutine based drop
after addressing the race condition

serprex: We don't need a goroutine, validation is not slow

PR #2128, opened 2024-10-07
routes: remove ok/error_message

serprex:
We have a mix of ok: false & returning errors,
now that customers are using API this is negatively impacting their UX
(not to mention our own struggles)

Consolidate on consistently returning HTTP errors, rather than 200 OK with ok: false

### DIFF
--- a/ui/app/mirrors/[mirrorId]/page.tsx
+++ b/ui/app/mirrors/[mirrorId]/page.tsx
@@ -23,12 +23,17 @@ type EditMirrorProps = {
 
 export default function ViewMirror({ params: { mirrorId } }: EditMirrorProps) {
   const [mirrorState, setMirrorState] = useState<MirrorStatusResponse>();
+  const [errorMessage, setErrorMessage] = useState('');
   const [mounted, setMounted] = useState(false);
 
   const fetchState = useCallback(async () => {
-    const res = await getMirrorState(mirrorId);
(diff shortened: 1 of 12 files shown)


ui/app/mirrors/[mirrorId]/page.tsx
```diff
@@ -23,12 +23,17 @@ type EditMirrorProps = {
...
   const [mirrorState, setMirrorState] = useState<MirrorStatusResponse>();
+  const [errorMessage, setErrorMessage] = useState('');
   const [mounted, setMounted] = useState(false);
 
   const fetchState = useCallback(async () => {
-    const res = await getMirrorState(mirrorId);
-    setMirrorState(res);
     setMounted(true);
+    try {
+      const res = await getMirrorState(mirrorId);
```
Comment (serprex) [inline ui/app/mirrors/[mirrorId]/page.tsx:32]: couple other places where this is called, but no error handling

woodhull:
Love this!

As a follow up, it would be good if the front-end that consumes this API could also be updated so it handles errors more gracefully and surfaces things like network or server errors back to the user. I'm finding myself running into things frequently where the only way to see that a request has failed is to look in the Developer Console.

pjhampton (approved):

serprex:
@woodhull yes, it'd be nice to have a better abstraction over `fetch`, maybe with some of the ActionItem changes in React 19 there'll be a clearer path

PR #2134, opened 2024-10-11
replace clickhouse-go with ch-go

serprex:
not really used on any hot path so don't expect perf gains yet, but opens up better batch inserting based on discussion with @gingerwizard


serprex:
for io.EOF retries we'll need to reconnect by recreating client

serprex:
Looks like ColAuto is quite incomplete, started work on making it more complete upstream: https://github.com/ClickHouse/ch-go/pull/433

Also need to fix up decimal support, see https://github.com/ClickHouse/ch-go/issues/128

serprex:
using branch for better inference gets further in tests, but then needs https://github.com/ClickHouse/ch-go/pull/435 will update again once one of my 2 open PRs to ch-go merge

PR #2136, opened 2024-10-14
Refactor ClickHouse table creation SQL generation

iskakaushik:
- Separate order by column logic into new function getOrderedOrderByColumns
- Simplify primary key and order by clause generation
- Ensure primary key columns are always at the end of ORDER BY clause
- Remove redundant code and improve readability

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -167,56 +167,53 @@ func generateCreateTableSQLForNormalizedTable(
 		"`%s` %s, `%s` %s) ENGINE = %s",
 		signColName, signColType, versionColName, versionColType, engine))
 
-	var pkeyStr string
-	pkeys := tableSchema.PrimaryKeyColumns
-	if len(pkeys) > 0 {
-		if len(colNameMap) > 0 {
-			pkeys = slices.Clone(pkeys)
-			for idx, pk := range pkeys {
-				pkeys[idx] = getColName(colNameMap, pk)
-			}
-		}
-		pkeyStr = strings.Join(pkeys, ",")
+	orderByColumns := getOrderedOrderByColumns(tableMapping, tableSchema.PrimaryKeyColumns)
+
+	if len(orderByColumns) > 0 {
+		orderByStr := strings.Join(orderByColumns, ",")
 
 		stmtBuilder.WriteString("PRIMARY KEY (")
-		stmtBuilder.WriteString(pkeyStr)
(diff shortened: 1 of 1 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -167,56 +167,53 @@ func generateCreateTableSQLForNormalizedTable(
...
+		stmtBuilder.WriteString("ORDER BY (")
+		stmtBuilder.WriteString(orderByStr)
 		stmtBuilder.WriteString(") ")
 	}
 
-	var orderby []*protos.ColumnSetting
+	return stmtBuilder.String(), nil
+}
+
+// Returns a list of order by columns ordered by their ordering, and puts the pkeys at the end.
```
heavycrystal, line 187: shouldn't this be the opposite?

``` DB::Exception: Primary key must be a prefix of the sorting key```

**Review by heavycrystal — CHANGES_REQUESTED** [review]

File: flow/connectors/clickhouse/normalize.go (hunk 1 above)
Comment (iskakaushik) [reply flow/connectors/clickhouse/normalize.go:187]: we are making pkey and order key same.

Comment (heavycrystal) [reply flow/connectors/clickhouse/normalize.go:187]: Oh ok, is there any reason for this change?

Comment (heavycrystal) [reply flow/connectors/clickhouse/normalize.go:187]: Also,

```If we would have specified only the sorting key, then the primary key would be implicitly defined to be equal to the sorting key.`

can drop the clause entirely then?

**Comment by Amogh-Bharadwaj** [comment]
My only concern is :
```
A long primary key will negatively affect the insert performance and memory consumption, 
but extra columns in the primary key do not affect ClickHouse performance during SELECT queries.
```

as documented [here](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree#selecting-a-primary-key)

heavycrystal (approved):

saisrirampur:
> My only concern is :
> 
> ```
> A long primary key will negatively affect the insert performance and memory consumption, 
> but extra columns in the primary key do not affect ClickHouse performance during SELECT queries.
> ```
> 
> as documented [here](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree#selecting-a-primary-key)

The PRIMARY KEY in ClickHouse dictates the skip index. If the user defines a custom ordering key, it is implicit that they want the skip index to be that. Else queries might not be as performant.

Amogh-Bharadwaj (approved):

PR #2140, opened 2024-10-15
[cdc] don't cache replident during cdc

heavycrystal:
replident can be changed anytime, caching during CDC shouldn't be done
since we store TableSchema in catalog not removing `is_replica_identity_full` for now

### DIFF
--- a/protos/flow.proto
+++ b/protos/flow.proto
@@ -167,6 +167,8 @@ message CreateRawTableOutput { string table_identifier = 1; }
 message TableSchema {
   string table_identifier = 1;
   repeated string primary_key_columns = 2;
+  // only used during SetupFlow, CDC fetches it everytime
+  // TODO: rename at some point
   bool is_replica_identity_full = 3;
   TypeSystem system = 4;
   bool nullable_enabled = 5;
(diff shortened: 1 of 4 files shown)


protos/flow.proto
```diff
@@ -167,6 +167,8 @@ message CreateRawTableOutput { string table_identifier = 1; }
 message TableSchema {
   string table_identifier = 1;
   repeated string primary_key_columns = 2;
+  // only used during SetupFlow, CDC fetches it everytime
+  // TODO: rename at some point
```
serprex, line 171: can this be cleared out as migration so in future it can be removed?

heavycrystal: prefer not to remove it because I want entirety of SetupFlow to use the same value
want to just rename it to make sure it's not used elsewhere

serprex: Can create model
```
type SetupTableSchema {
  *protos.TableSchema
  ...rest
}
```
& pass that around instead of the proto

heavycrystal:
this may not be something we want to do; will revisit behaviour later

PR #2141, opened 2024-10-15
Optimize setup flow for table addition

Amogh-Bharadwaj:
In the CreateNormalizedTable activity, we fetch the tablenameschemamapping from catalog and pass it to the setupNormalize function of the connector.
This does not make sense for table addition however, where setup flow needs to be done for a handful of tables and not the entire set

This PR refactors the activity to pass in the tables provided as input to the activity and uses the catalog map to get the source schema of the tables

Also rename a few symbols to be more readable

- [ ] Automated test


flow/activities/flowable.go
```diff
@@ -247,7 +247,9 @@ func (a *FlowableActivity) CreateNormalizedTable(
 	defer shutdown()
 
 	tableExistsMapping := make(map[string]bool, len(tableNameSchemaMapping))
-	for tableIdentifier, tableSchema := range tableNameSchemaMapping {
+	for _, tableMapping := range config.TableMappings {
+		tableIdentifier := tableMapping.DestinationTableIdentifier
+		tableSchema := tableNameSchemaMapping[tableMapping.DestinationTableIdentifier]
```
```suggestion
		tableSchema := tableNameSchemaMapping[tableIdentifier]
```

PR #2142, opened 2024-10-15
Create ClickHouse mirror: New sorting key UI and logic

Amogh-Bharadwaj:
Screenshot 2024-10-22 at 4 04 20 AM

User can now specify order key list by selecting from a dropdown. If they do this, then a warning banner also pops up. This links to a document explaining the various caveats of custom order keys

Instead of suffixing the primary keys at the end of the ordering key, the behaviour is now:
If user doesn't define order key - PKEY and ORDER BY in CH is PKEY in Postgres.
If user defines order key - PKEY and ORDER BY in CH is defined Order Key in Postgres.


Amogh-Bharadwaj:
TODO: Link https://github.com/PeerDB-io/docs/pull/195

iskakaushik (approved):

PR #2143, opened 2024-10-15
Add alerting for bad WAL status

iskakaushik:
If the status of the slot is "lost" or "unreserved", then alert it via slack


iskakaushik:
i still need to functionally test this.

PR #2144, opened 2024-10-16
SetupReplication: clean up replicationErr handling

serprex:
defer close(replicationErr) would panic if SetupReplication returns error after defer

replicationErr channel write also hangs indefinitely were SlotCreated written to previously

reuse SlotCreated error mechanism so only one channel is being waited on

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -402,9 +402,7 @@ func (a *FlowableActivity) StartNormalize(
 		conn.DestinationName,
 	)
 	if errors.Is(err, errors.ErrUnsupported) {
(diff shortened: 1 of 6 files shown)


flow/activities/flowable.go
```diff
@@ -460,8 +458,7 @@ func (a *FlowableActivity) SetupQRepMetadataTables(ctx context.Context, config *
 	}
 	defer connectors.CloseConnector(ctx, conn)
 
-	err = conn.SetupQRepMetadataTables(ctx, config)
-	if err != nil {
+	if err := conn.SetupQRepMetadataTables(ctx, config); err != nil {
```
heavycrystal, line 461:
```suggestion
	if err = conn.SetupQRepMetadataTables(ctx, config); err != nil {
```
don't think we need this to be scoped, wherever applicable

heavycrystal (approved):

PR #2148, opened 2024-10-16
[clickhouse] apply TableSchemaDeltas in normalize

heavycrystal:
TableSchemaDeltas are already stored in catalog for audit
instead of applying in SyncFlow like other connectors, apply them during normalize
as part of the philosophy of not connecting to CH during SyncFlow

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -19,6 +19,7 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"github.com/PeerDB-io/peer-flow/connectors"
+	connmetadata "github.com/PeerDB-io/peer-flow/connectors/external_metadata"
 	connpostgres "github.com/PeerDB-io/peer-flow/connectors/postgres"
 	"github.com/PeerDB-io/peer-flow/connectors/utils/monitoring"
 	"github.com/PeerDB-io/peer-flow/generated/protos"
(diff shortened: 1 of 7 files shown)


Amogh-Bharadwaj:
Would be great if we could get in https://github.com/PeerDB-io/peerdb/pull/2114 as part of landing this
I can fix that PR up in the next couple of days

flow/connectors/clickhouse/normalize.go
```diff
@@ -235,6 +235,17 @@ func (c *ClickHouseConnector) NormalizeRecords(
 		return nil, err
 	}
 
+	schemaDeltas, err := c.GetSchemaDeltasForBatches(ctx, req.FlowJobName, req.SyncBatchID, normBatchID)
+	if err != nil {
+		c.logger.Error("[clickhouse] error while getting schema deltas for batches", "error", err)
+		return nil, err
+	}
+
+	if err = c.replayTableSchemaDeltas(ctx, schemaDeltas); err != nil {
```
serprex, line 244:
```suggestion
	if err := c.replayTableSchemaDeltas(ctx, schemaDeltas); err != nil {
```

flow/connectors/clickhouse/cdc.go
```diff
@@ -122,8 +119,16 @@ func (c *ClickHouseConnector) SyncRecords(ctx context.Context, req *model.SyncRe
 	return res, nil
 }
 
+// For CH specifically, we don't want to replay during sync, but rather during the normalization phase.
+// This a no-op for CH, private variant to be called during normalize.
 func (c *ClickHouseConnector) ReplayTableSchemaDeltas(ctx context.Context, flowJobName string,
```
serprex, line 124: should move this method to a separate connector interface, there's a few connectors that skip this already

flow/activities/flowable_core.go
```diff
@@ -172,6 +173,15 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	startTime := time.Now()
 	errGroup, errCtx := errgroup.WithContext(ctx)
 	errGroup.Go(func() error {
+		pgMetadata := connmetadata.NewPostgresMetadataFromCatalog(logger, a.CatalogPool)
```
serprex, line 176: Further down we have
```
syncBatchID, err := dstConn.GetLastSyncBatchID(errCtx, flowName)
```
Could lift that up to where we get lastOffset. Then use same value between two goroutines. Would make this logic work for connectors (pg:pg) which don't use catalog for metadata too

PR #2149, opened 2024-10-16
Validate mirror: fix get table schema

Amogh-Bharadwaj:
We were passing the type system value of the mirror as a hardcoded value of typesystem PG in the call to `GetTableSchema`, when we should be using the value in the config


iamKunalGupta (approved):
Assuming nothing breaks from nexus/ui requests

PR #2150, opened 2024-10-16
[postgres] application name has mirror suffix

heavycrystal:


flow/shared/postgres.go
```diff
@@ -28,16 +28,21 @@ const (
 	POSTGRES_15 PGVersion = 150000
 )
 
-func GetPGConnectionString(pgConfig *protos.PostgresConfig) string {
+func GetPGConnectionString(pgConfig *protos.PostgresConfig, flowName string) string {
 	passwordEscaped := url.QueryEscape(pgConfig.Password)
+	applicationName := "peerdb"
+	if flowName != "" {
+		applicationName = "peerdb_" + flowName
```
iamKunalGupta, line 35: do we need to crop the application name to <64 chars?

iamKunalGupta: Also should we make this an opt-in? Like `PEERDB_APPLICATION_NAME_USES_MIRRORNAME` or smth?

heavycrystal: testing, looks like PG will truncate it if needed
I haven't seen many users care about `application_name` as long as it's identified as PeerDB so not in favour to add another setting

iskakaushik: this should definitely be opt-in

serprex (approved):

flow/shared/postgres.go
iamKunalGupta: @heavycrystal can we make this opt in and finish up?

iamKunalGupta (approved):

PR #2163, opened 2024-10-19
[clickhouse] fix checkTablesEmptyAndEngine sql query

udev-21:
I was trying to create mirror from postgresql peer to clickhouse peer and got this error:
```bash
udev21=> create mirror my_test_mirror from mydb1 to clickhouse with table mapping ( public.products: products) WITH (do_initial_copy= true);

ERROR:  unable to submit job: "status: Unknown, message: \"invalid mirror: failed to get information for destination tables: code: 47, message: Missing columns: 'table' while processing query: 'SELECT name, engine, total_rows FROM system.tables WHERE (database = 'default') AND (table IN ('products'))', required columns: 'name' 'engine' 'total_rows' 'database' 'table', maybe you meant: ['name','engine','total_rows','database']\", details: [], metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\"} }"
```
after some research, found out that there's indeed no column named <b>table</b> in system.tables, I guess by table there's meant <b>name</b> column

### DIFF
--- a/flow/connectors/clickhouse/clickhouse.go
+++ b/flow/connectors/clickhouse/clickhouse.go
@@ -372,7 +372,7 @@ func (c *ClickHouseConnector) checkTablesEmptyAndEngine(ctx context.Context, tab
 		queryInput = append(queryInput, table)
 	}
 	rows, err := c.query(ctx,
-		fmt.Sprintf("SELECT name,engine,total_rows FROM system.tables WHERE database=? AND table IN (%s)",
+		fmt.Sprintf("SELECT name,engine,total_rows FROM system.tables WHERE database=? AND name IN (%s)",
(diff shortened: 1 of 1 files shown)


udev-21:
@serprex @iskakaushik @arajkumar

udev-21:
and for the record: it's my first time contributing to open-source project, feel free if I'm doing it wrong :)

heavycrystal:
The `table` alias for `name` in `system.tables` was added in 2021: https://github.com/ClickHouse/ClickHouse/pull/29882 but the docs may not be up to date on this.
All supported ClickHouse versions should have this column so this is not a bug.

Can still get this in though, thanks for your contribution.

serprex (approved):

PR #2168, opened 2024-10-22
Fix initial snapshot replay

serprex:
During replay connection is reset, invalidating snapshot name. In replays we want to drop slot from failed run & use newly exported snapshot name. That means slot/snapshot name cannot end up in temporal history

Another challenge is child qrep activities are running on peerflow workers, whereas connection is maintained by session on snapshot worker. So instead of relying on maps in session memory, use catalog to share names between workers

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -424,13 +424,12 @@ func replicateQRepPartition[TRead any, TWrite any, TSync connectors.QRepSyncConn
 	var rowsSynced int
 	errGroup, errCtx := errgroup.WithContext(ctx)
 	errGroup.Go(func() error {
-		snapshotFlowName := config.ParentMirrorName
-		if snapshotFlowName == "" {
-			snapshotFlowName = config.FlowJobName
-		}
-		_, snapshotName, _, err := shared.LoadSnapshotNameFromCatalog(ctx, a.CatalogPool, snapshotFlowName)
-		if err != nil {
-			return err
+		snapshotName := ""
+		if config.ParentMirrorName != "" {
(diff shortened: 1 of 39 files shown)


flow/activities/flowable_core.go
```diff
@@ -422,14 +424,22 @@ func replicateQRepPartition[TRead any, TWrite any, TSync connectors.QRepSyncConn
 	var rowsSynced int
 	errGroup, errCtx := errgroup.WithContext(ctx)
 	errGroup.Go(func() error {
+		snapshotName := ""
+		if config.ParentMirrorName != "" {
+			_, snapshotName, _, err = shared.LoadSnapshotNameFromCatalog(ctx, a.CatalogPool, config.ParentMirrorName)
+			if err != nil {
```
iskakaushik, line 430: a.Alerter.LogFlowError(ctx, config.FlowJobName, err)

flow/activities/flowable_core.go
```diff
@@ -501,6 +513,15 @@ func replicateXminPartition[TRead any, TWrite any, TSync connectors.QRepSyncConn
 	var currentSnapshotXmin int64
 	var rowsSynced int
 	errGroup.Go(func() error {
+		snapshotName := ""
+		if config.ParentMirrorName != "" {
+			var err error
+			_, snapshotName, _, err = shared.LoadSnapshotNameFromCatalog(ctx, a.CatalogPool, config.ParentMirrorName)
```
iskakaushik, line 519: a.Alerter.LogFlowError(ctx, config.FlowJobName, err)

serprex:
Somewhat addressed with #2703

PR #2172, opened 2024-10-23
Alerts for drop mirror

Amogh-Bharadwaj:
Add alerting for drop mirror's 2 activities - `DropFlowSource` and `DropFlowDestination`

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -621,22 +621,40 @@ func (a *FlowableActivity) DropFlowSource(ctx context.Context, req *protos.DropF
 	ctx = context.WithValue(ctx, shared.FlowNameKey, req.FlowJobName)
 	srcConn, err := connectors.GetByNameAs[connectors.CDCPullConnector](ctx, nil, a.CatalogPool, req.PeerName)
 	if err != nil {
-		return fmt.Errorf("failed to get source connector: %w", err)
+		srcConnErr := fmt.Errorf("[DropFlowSource] failed to get source connector: %w", err)
+		a.Alerter.LogFlowError(ctx, req.FlowJobName, srcConnErr)
+		return srcConnErr
 	}
 	defer connectors.CloseConnector(ctx, srcConn)
 
-	return srcConn.PullFlowCleanup(ctx, req.FlowJobName)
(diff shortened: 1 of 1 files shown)


flow/activities/flowable.go
```diff
@@ -621,22 +621,40 @@ func (a *FlowableActivity) DropFlowSource(ctx context.Context, req *protos.DropF
...
 
-	return dstConn.SyncFlowCleanup(ctx, req.FlowJobName)
+	err = dstConn.SyncFlowCleanup(ctx, req.FlowJobName)
+	if err != nil {
+		syncFlowCleanupErr := fmt.Errorf("[DropFlowDestination] failed to clean up destination: %w", err)
+		a.Alerter.LogFlowError(ctx, req.FlowJobName, syncFlowCleanupErr)
+		return syncFlowCleanupErr
+	}
+
+	return nil
```
Amogh-Bharadwaj, line 657: not sure if this should be `return err` instead

heavycrystal: the error is tagged in L651-L655 and returned if not nil, so this should be fine?

heavycrystal (approved):

flow/activities/flowable.go
```diff
@@ -621,22 +621,40 @@ func (a *FlowableActivity) DropFlowSource(ctx context.Context, req *protos.DropF
...
+		return srcConnErr
 	}
 	defer connectors.CloseConnector(ctx, srcConn)
 
-	return srcConn.PullFlowCleanup(ctx, req.FlowJobName)
+	err = srcConn.PullFlowCleanup(ctx, req.FlowJobName)
+	if err != nil {
+		pullCleanupErr := fmt.Errorf("[DropFlowSource] failed to clean up source: %w", err)
+		a.Alerter.LogFlowError(ctx, req.FlowJobName, pullCleanupErr)
+		return fmt.Errorf("[DropFlowSource] failed to clean up source: %w", pullCleanupErr)
```
serprex, line 634:
```suggestion
		return pullCleanupErr
```

flow/activities/flowable.go
serprex: nil is good, makes it obvious that at this point we've succeeded

serprex (approved):

PR #2178, opened 2024-10-23
crypt: support special key id `kms` to use KMS for encrypting catalog

serprex:
value of the `kms` key must be an empty string


serprex:
talking with @iamKunalGupta currently our KMS permissions only allow decryption. That'll need to change for this

PR #2179, opened 2024-10-24
KMS encrypt PEERDB_ENC_KEYS

serprex:
See #2178 for why KMS shouldn't itself be used for encryption/decryption


serprex:
As is this PR requires care upgrading so that PEERDB_ENC_KEYS is updated if KMS already used. Could introduce a bool or have a separate env key `PEERDB_KMS_ENC_KEYS`  with fallback logic

iskakaushik (approved):

PR #2180, opened 2024-10-24
kms: cache results

serprex:
kms should not be polled repeatedly, keep an in-memory cache of decrypted kms env variables

### DIFF
--- a/flow/peerdbenv/env.go
+++ b/flow/peerdbenv/env.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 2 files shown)


flow/peerdbenv/env.go
```diff
@@ -86,7 +87,15 @@ func decryptWithKMS(ctx context.Context, data []byte) ([]byte, error) {
 	return decrypted.Plaintext, nil
 }
 
+var KMSCache sync.Map
+
 func GetEnvBase64EncodedBytes(ctx context.Context, name string, defaultValue []byte) ([]byte, error) {
+	if cacheVal, ok := KMSCache.Load(name); ok {
```
serprex, line 93: There is a slight race here where two concurrent accesses will run in parallel. This is fine

flow/peerdbenv/env.go
```diff
@@ -86,7 +87,15 @@ func decryptWithKMS(ctx context.Context, data []byte) ([]byte, error) {
 	return decrypted.Plaintext, nil
 }
 
+var KMSCache sync.Map
+
 func GetEnvBase64EncodedBytes(ctx context.Context, name string, defaultValue []byte) ([]byte, error) {
```
heavycrystal, line 92: this function is used for fetching Temporal cert and key too
are they also going to be KMS encrypted

also function name should indicate KMS dependency

flow/peerdbenv/env.go
```diff
@@ -86,7 +87,15 @@ func decryptWithKMS(ctx context.Context, data []byte) ([]byte, error) {
 	return decrypted.Plaintext, nil
 }
 
+var KMSCache sync.Map
```
heavycrystal, line 90: don't see why this has to be public?

flow/peerdbenv/env.go
serprex: yes

iskakaushik (approved):

Issue #2182, opened 2024-10-24
Validate peer: set a timeout


Amogh-Bharadwaj:
Peer validation can take a long time especially for example when you enter the wrong port as it runs into the connection timeout waiting
Let's set a context deadline of 10 - 15 seconds for the validate peer handler

heavycrystal:
closed by #2183

PR #2183, opened 2024-10-24
validate: 1 minute timeout

serprex:
#2182

### DIFF
--- a/flow/cmd/validate_mirror.go
+++ b/flow/cmd/validate_mirror.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"log/slog"
 	"regexp"
-	"time"
 
 	"github.com/jackc/pgx/v5/pgtype"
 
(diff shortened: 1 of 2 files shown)


flow/cmd/validate_mirror.go
```diff
@@ -25,6 +26,8 @@ var (
 func (h *FlowRequestHandler) ValidateCDCMirror(
 	ctx context.Context, req *protos.CreateCDCFlowRequest,
 ) (*protos.ValidateCDCMirrorResponse, error) {
+	ctx, cancelCtx := context.WithTimeout(ctx, time.Minute)
```
iskakaushik, line 29: this can take longer make it 5 mins?

serprex: if validation takes that long it should have a status that gives a progress bar checking off which steps have passed

but for now we'll remove timeout on mirror validation

Amogh-Bharadwaj (approved):
functionally tested and timed with wrong port on rds instance

PR #2196, opened 2024-10-28
Alerts: tag context.Canceled / io.EOF / net.ErrClosed

serprex:
context.Canceled happens when mirrors are paused

connection drops happen & there's nothing actionable about them in isolation. They're better off being tracked as a frequency metric


flow/alerting/alerting.go
```diff
@@ -412,36 +410,46 @@ func (a *Alerter) LogNonFlowCritical(ctx context.Context, eventType telemetry.Ev
...
 		logger.Warn("failed to insert flow error", slog.Any("error", err))
 		return
 	}
-	a.sendTelemetryMessage(ctx, flowName, errorWithStack, telemetry.ERROR)
+	var tags []string
+	if errors.Is(err, context.Canceled) {
+		tags = append(tags, "errCanceled")
+	}
+	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
+		tags = append(tags, "errEOF")
```
iamKunalGupta, line 433: should we do "errType:errEOF" instead? might be more robust to parse for all the cases

serprex: or at least `err:EOF`

iamKunalGupta (approved):

PR #2200, opened 2024-10-29
feat(temporal): enable sdk metrics

iamKunalGupta:
- Adds a new env variable to enable specific metric from Temporal SDK:
   - `PEERDB_TEMPORAL_OTEL_METRICS_EXPORT_LIST` can be a comma separated list of metrics to expose. If the list is empty or no metric matches the elements in the list, then they are not exported
   - If the value is set to `__ALL__`, then all Temporal metrics are exposed
 - Additionally adds an initial interface of how temporal interceptors would look like
 - The added metrics include `temporal_workflow_task_execution_failed` metric having an attribute of `failure_reason` whose value can tell us about nondeterminism if the value is `nondeterminismerror`  (https://github.com/temporalio/sdk-go/pull/1295)
 - Additionally, the metrics are prefixed with `temporal.`

PeerDB metrics should not be affected as they use a separate exporter and meterprovider.

The rationale behind exporting a subset of metrics is noise and cases where metrics ingestion cannot be ignored

Full list of metrics can be viewed at https://docs.temporal.io/references/sdk-metrics


flow/otel_metrics/common.go
```diff
@@ -0,0 +1,7 @@
+package otel_metrics
+
+import "github.com/PeerDB-io/peer-flow/peerdbenv"
+
```
iskakaushik, line 4: shoud this be in `env.go`?

flow/cmd/logged_interceptor.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+
+	"go.temporal.io/sdk/interceptor"
+	"go.temporal.io/sdk/workflow"
+)
+
+type LoggedWorkflowInboundInterceptor struct {
+	interceptor.WorkflowInboundInterceptorBase
+	Next interceptor.WorkflowInboundInterceptor
```
serprex, line 12:
```suggestion
```
Why not get it from Base?

flow/otel_metrics/otel_manager.go
```diff
@@ -42,7 +48,52 @@ func setupGrpcOtelMetricsExporter() (sdkmetric.Exporter, error) {
 	return otlpmetricgrpc.New(context.Background())
 }
 
-func SetupOtelMetricsExporter(otelServiceName string) (*sdkmetric.MeterProvider, error) {
+func temporalMetricsFilteringView() sdkmetric.View {
+	exportListString := peerdbenv.GetEnvString("PEERDB_TEMPORAL_OTEL_METRICS_EXPORT_LIST", "")
```
iskakaushik, line 52: consolidate all env vars

flow/cmd/logged_interceptor.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (w *LoggedWorkflowInboundInterceptor) ExecuteWorkflow(
+	ctx workflow.Context,
+	in *interceptor.ExecuteWorkflowInput,
+) (interface{}, error) {
+	// Workflow starts here
+	result, err := w.Next.ExecuteWorkflow(ctx, in)
+	// Workflow ends here
+	return result, err
```
serprex, line 29:
```suggestion
	return w.Next.ExecuteWorkflow(ctx, in)
```

flow/otel_metrics/otel_manager.go
```diff
@@ -42,7 +48,52 @@ func setupGrpcOtelMetricsExporter() (sdkmetric.Exporter, error) {
...
+	// Special case for exporting all metrics
+	if exportListString == "__ALL__" {
```
iskakaushik, line 54: can we log to say something like we are exposing XYZ metrics

flow/cmd/logged_interceptor.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (c *LoggedActivityInboundInterceptor) ExecuteActivity(
+	ctx context.Context,
+	in *interceptor.ExecuteActivityInput,
+) (interface{}, error) {
+	// Activity starts here
+	out, err := c.Next.ExecuteActivity(ctx, in)
+	// Activity ends here
+	return out, err
```
serprex, line 51:
```suggestion
	return c.Next.ExecuteActivity(ctx, in)
```

flow/cmd/logged_interceptor.go
```diff
@@ -0,0 +1,79 @@
...
+) (interface{}, error) {
+	// Workflow starts here
+	result, err := w.Next.ExecuteWorkflow(ctx, in)
+	// Workflow ends here
+	return result, err
+}
+
+type LoggedActivityInboundInterceptor struct {
+	interceptor.ActivityInboundInterceptorBase
+	Next interceptor.ActivityInboundInterceptor
```
serprex, line 34:
```suggestion
```

flow/otel_metrics/otel_manager.go
```diff
@@ -42,7 +48,52 @@ func setupGrpcOtelMetricsExporter() (sdkmetric.Exporter, error) {
...
+				Name:        GetPeerDBOtelMetricsNamespace() + "temporal." + instrument.Name,
+				Description: instrument.Description,
+				Unit:        instrument.Unit,
+				Aggregation: sdkmetric.AggregationDrop{},
+			}, true
+		}
+	}
+
+	// Export only the metrics in the list
+	enabledMetrics := make(map[string]struct{})
```
serprex, line 78:
```suggestion
	enabledMetrics := make(map[string]struct{}, len(exportList))
```

flow/otel_metrics/otel_manager.go
```diff
@@ -42,7 +48,52 @@ func setupGrpcOtelMetricsExporter() (sdkmetric.Exporter, error) {
...
+		stream := sdkmetric.Stream{
+			Name:        GetPeerDBOtelMetricsNamespace() + "temporal." + instrument.Name,
+			Description: instrument.Description,
+			Unit:        instrument.Unit,
+		}
+		if _, ok := enabledMetrics[instrument.Name]; ok {
+			return stream, true
+		}
+		stream.Aggregation = sdkmetric.AggregationDrop{}
+		return stream, true
```
serprex, line 92:
```suggestion
		if _, ok := enabledMetrics[instrument.Name]; !ok {
			stream.Aggregation = sdkmetric.AggregationDrop{}
		}
		return stream, true
```

flow/otel_metrics/otel_manager.go
```diff
@@ -42,7 +48,52 @@ func setupGrpcOtelMetricsExporter() (sdkmetric.Exporter, error) {
...
+				Unit:        instrument.Unit,
+				Aggregation: sdkmetric.AggregationDrop{},
+			}, true
+		}
+	}
+
+	// Export only the metrics in the list
+	enabledMetrics := make(map[string]struct{})
+	for _, metricName := range exportList {
+		enabledMetrics[metricName] = struct{}{}
```
serprex, line 80: Maybe strings.Trim metricName since this comes from `,` split

flow/cmd/logged_interceptor.go
iamKunalGupta: This was just experimental and I was testing

flow/cmd/logged_interceptor.go
iamKunalGupta: Just experimental, we are not actually using it right now

flow/otel_metrics/otel_manager.go
iamKunalGupta: I think I can just log the env var, that should be good enough

iamKunalGupta: Added log above

PR #2202, opened 2024-10-29
improve logs pagination

serprex:
using LIMIT/OFFSET is inaccurate when new logs come it while logs being viewed

this makes a minor improvement by instead being based on records before/after an id from previous page,
which is how I'd like to also add pagination to sync batches in mirror status

### DIFF
--- a/protos/route.proto
+++ b/protos/route.proto
@@ -354,16 +354,19 @@ message MirrorLog {
   string error_message = 2;
   string error_type = 3;
   double error_timestamp = 4;
+  int32 id = 5;
 }
 message ListMirrorLogsRequest {
   string flow_job_name = 1;
   string level = 2;
-  int32 page = 3;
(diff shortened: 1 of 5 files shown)


protos/route.proto
```diff
@@ -354,16 +354,19 @@ message MirrorLog {
...
   string error_type = 3;
   double error_timestamp = 4;
+  int32 id = 5;
 }
 message ListMirrorLogsRequest {
   string flow_job_name = 1;
   string level = 2;
-  int32 page = 3;
-  int32 num_per_page = 4;
+  int32 num_per_page = 3;
```
iskakaushik, line 362: this will break other clients (internal usage), lets try to maintain proto compatibility.

serprex: done, can wait for 0.19 to remove ugly support for page param

pjhampton (approved):

PR #2203, opened 2024-10-30
feat: add request logging interceptor for API

iamKunalGupta:
Set `PEERDB_API_REQUEST_LOGGING_ENABLED` = true to enable request logging. Separating it from the oauth interceptor incase oauth is disabled and we wish to enable request logging

### DIFF
--- a/flow/peerdbenv/config.go
+++ b/flow/peerdbenv/config.go
@@ -161,6 +161,7 @@ func PeerDBGetIncidentIoToken() string {
 func PeerDBRAPIRequestLoggingEnabled() bool {
 	requestLoggingEnabled, err := strconv.ParseBool(GetEnvString("PEERDB_API_REQUEST_LOGGING_ENABLED", "false"))
 	if err != nil {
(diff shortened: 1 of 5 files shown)


flow/peerdbenv/config.go
```diff
@@ -156,3 +157,11 @@ func PeerDBGetIncidentIoUrl() string {
 func PeerDBGetIncidentIoToken() string {
 	return GetEnvString("PEERDB_INCIDENTIO_TOKEN", "")
 }
+
+func PeerDBRAPIRequestLoggingEnabled() bool {
+	requestLoggingEnabled, err := strconv.ParseBool(GetEnvString("PEERDB_API_REQUEST_LOGGING_ENABLED", "false"))
+	if err != nil {
```
serprex, line 163: should log error

iamKunalGupta: I think then I should do it for oauth as well, will do it

serprex (approved):

PR #2207, opened 2024-10-30
Sync batch pagination

serprex:
Some customers have enough batches now that response size was exceeding grpc limit

Changing this meant loading pages & aggregating graph data on server

Removed searching specific batch id & changing sort column from ui

### DIFF
--- a/flow/cmd/mirror_status.go
+++ b/flow/cmd/mirror_status.go
@@ -216,13 +216,15 @@ func (h *FlowRequestHandler) CDCGraph(ctx context.Context, req *protos.GraphRequ
 	if err != nil {
 		return nil, err
 	}
-	var data []*protos.GraphResponseItem
-	var t time.Time
-	var r int64
-	if _, err := pgx.ForEachRow(rows, []any{&t, &r}, func() error {
-		data = append(data, &protos.GraphResponseItem{Time: float64(t.UnixMilli()), Rows: float64(r)})
-		return nil
-	}); err != nil {
+	data, err := pgx.CollectRows(rows, func(row pgx.CollectableRow) (*protos.GraphResponseItem, error) {
+		var t time.Time
+		var r int64
+		if err := row.Scan(&t, &r); err != nil {
+			return nil, err
+		}
+		return &protos.GraphResponseItem{Time: float64(t.UnixMilli()), Rows: float64(r)}, nil
+	})
+	if err != nil {
 		return nil, err
 	}
 
--- a/flow/cmd/mirror_status.go
+++ b/flow/cmd/mirror_status.go
@@ -184,6 +184,11 @@ func (h *FlowRequestHandler) cdcFlowStatus(
 		cdcBatches = cdcBatchesResponse.CdcBatches
 	}
 
+	var rowsSynced int64
+	if err := h.pool.QueryRow(ctx, "select coalesce(sum(rows_in_batch), 0) from peerdb_stats.cdc_batches where flow_name=$1", req.FlowJobName).Scan(&rowsSynced); err != nil {
+		return nil, err
+	}
+
 	return &protos.CDCMirrorStatus{
 		Config:          config,
 		SourceType:      srcType,
@@ -192,6 +197,7 @@ func (h *FlowRequestHandler) cdcFlowStatus(
 			Clones: initialLoadResponse.TableSummaries,
 		},
 		CdcBatches: cdcBatches,
+		RowsSynced: rowsSynced,
 	}, nil
 }
 
--- a/flow/cmd/mirror_status.go
+++ b/flow/cmd/mirror_status.go
@@ -205,8 +205,13 @@ func (h *FlowRequestHandler) cdcFlowStatus(
 
 func (h *FlowRequestHandler) CDCGraph(ctx context.Context, req *protos.GraphRequest) (*protos.GraphResponse, error) {
 	truncField := "minute"
-	if slices.Contains([]string{"hour", "day", "month"}, req.AggregateType) {
-		truncField = req.AggregateType
+	switch req.AggregateType {
+	case "1hour":
+		truncField = "hour"
+	case "1day":
+		truncField = "day"
+	case "1month":
+		truncField = "month"
 	}
 	rows, err := h.pool.Query(ctx, `select tm, coalesce(sum(rows_in_batch), 0)
 	from generate_series(date_trunc($2, now() - $1::INTERVAL * 30), now(), $1::INTERVAL) tm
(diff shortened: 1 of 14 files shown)


serprex:
@pjhampton this PR removes being able to look up a specific ID & changing sort field

Specific ID can be implemented on a ui level by doing `AfterId: id-1`, but isn't generally useful. Is sorting by other fields something clickpipes would want?

flow/cmd/mirror_status.go
```diff
@@ -190,10 +198,39 @@ func (h *FlowRequestHandler) cdcFlowStatus(
...
 		},
-		CdcBatches: cdcBatchesResponse.CdcBatches,
+		CdcBatches: cdcBatches,
+		RowsSynced: rowsSynced,
 	}, nil
 }
 
+func (h *FlowRequestHandler) CDCGraph(ctx context.Context, req *protos.GraphRequest) (*protos.GraphResponse, error) {
+	truncField := "minute"
+	switch req.AggregateType {
```
heavycrystal, line 208: nit: could be a `slices.Contains`

flow/cmd/mirror_status.go
```diff
@@ -455,18 +492,39 @@ func (h *FlowRequestHandler) getMirrorCreatedAt(ctx context.Context, flowJobName
...
+		limitClause = fmt.Sprintf(" LIMIT %d", req.Limit)
 	}
-	q := `SELECT DISTINCT ON(batch_id) batch_id,start_time,end_time,rows_in_batch,batch_start_lsn,batch_end_lsn FROM peerdb_stats.cdc_batches
-	  WHERE flow_name=$1 AND start_time IS NOT NULL ORDER BY batch_id DESC, start_time DESC` + limitClause
-	rows, err := h.pool.Query(ctx, q, mirrorName)
+
+	whereExpr := ""
+	queryArgs := append(make([]any, 0, 2), req.FlowJobName)
+
+	sortOrderBy := "desc"
```
heavycrystal, line 507: nit: `asc` for symmetry

heavycrystal:
also seems like a candidate for API tests, can be done as a follow-up

heavycrystal:
image
image

not sure if this is easily addressable, but day and month aggregations are putting most of the batches in previous day/month for a mirror started a few minutes ago

flow/cmd/mirror_status.go
```diff
@@ -455,18 +492,39 @@ func (h *FlowRequestHandler) getMirrorCreatedAt(ctx context.Context, flowJobName
...
+	if req.BeforeId != 0 || req.AfterId != 0 {
+		if req.BeforeId != -1 {
```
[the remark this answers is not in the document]
serprex: I don't follow, what would be sensible behavior when both supplied?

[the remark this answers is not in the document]
serprex: Would end up being dead code, expecting this code will end up being touched with date filtering etc, so don't want to make things more complex & have unused code be broken in future. With API tests I'd be less reserved

serprex:
graph fixed, didn't realize interval was `1month` instead of `month` thus causing everything to be truncated by `minute`

flow/cmd/mirror_status.go
```diff
@@ -190,10 +198,43 @@ func (h *FlowRequestHandler) cdcFlowStatus(
...
-		CdcBatches: cdcBatchesResponse.CdcBatches,
+		CdcBatches: cdcBatches,
+		RowsSynced: rowsSynced,
 	}, nil
 }
 
+func (h *FlowRequestHandler) CDCGraph(ctx context.Context, req *protos.GraphRequest) (*protos.GraphResponse, error) {
+	truncField := "minute"
+	switch req.AggregateType {
+	case "1hour":
```
heavycrystal, line 209: could this be made a proto enum?

protos/route.proto
```diff
@@ -343,10 +345,29 @@ message InitialLoadSummaryResponse {
...
+}
+
+message GraphRequest {
+  string flow_job_name = 1;
+  string aggregate_type = 2; // TODO name?
+}
+
+message GraphResponseItem {
+  double time = 1;
+  double rows = 2;
```
heavycrystal, line 366: `double` for rows?

ui/app/mirrors/[mirrorId]/cdcDetails.tsx
```diff
@@ -151,8 +148,7 @@ const SyncIntervalLabel: React.FC<{ syncInterval?: number }> = ({
 
   if (!syncInterval) {
     return <ProgressCircle variant='determinate_progress_circle' />;
-  }
-  if (syncInterval >= 3600) {
+  } else if (syncInterval >= 3600) {
```
Comment (heavycrystal) [inline ui/app/mirrors/[mirrorId]/cdcDetails.tsx:151]: unrelated but ideally this should be hours + minutes in case it's not exactly an hour

ui/app/mirrors/[mirrorId]/qrepGraph.tsx
```diff
@@ -17,22 +17,20 @@ type QRepGraphProps = {
...
       timestamp: sync.startTime!,
       count: Number(sync.rowsInPartition) ?? 0,
     }));
 
-    let counts = aggregateCountsByInterval(rows, aggregateType);
-    counts = counts.slice(0, 29);
-    counts = counts.reverse();
-    setCounts(counts);
+    const counts = aggregateCountsByInterval(rows, aggregateType);
+    setCounts(counts.slice(0, 29).reverse());
```
Comment (heavycrystal) [inline ui/app/mirrors/[mirrorId]/qrepGraph.tsx:33]: nit: move 29 to a constant

ui/app/mirrors/[mirrorId]/cdcGraph.tsx
```diff
@@ -5,35 +5,47 @@ import {
...
+  useEffect(() => {
+    const fetchData = async () => {
+      const req: any = {
+        flowJobName: mirrorName,
+        aggregateType,
+      };
+
+      const res = await fetch('/api/v1/mirrors/cdc/graph', {
+        method: 'POST',
+        headers: {
```
Comment (heavycrystal) [inline ui/app/mirrors/[mirrorId]/cdcGraph.tsx:32]: nit: I'm not sure if `Content-Type` is necessary, looking at other routes

protos/route.proto
```diff
@@ -145,6 +145,7 @@ message CreatePeerResponse {
 message MirrorStatusRequest {
   string flow_job_name = 1;
   bool include_flow_info = 2;
+  bool exclude_batches = 3;
```
heavycrystal, line 148: nit: `exclude_cdc_batches`

flow/cmd/mirror_status.go
serprex: I'd rather not, it's postgres interval syntax, don't trust protobuf to preserve strings easily

protos/route.proto
serprex: int64 in protos gets put into typescript as number, but then sent as string. Then we have to convert down to number anyways for graph component

serprex:
I've removed sorting various columns because that'd complicate the code quite a bit & I'd like to wait & see what behavior clickpipes wants from this api

protos/route.proto
serprex: this'll also be excluding qrep batches in near future

heavycrystal (approved):

PR #2211, opened 2024-11-01
feat: add maintenance mode for upgrades

iamKunalGupta:
- Introduces Maintenance mode (status is available via dynamic config: `PEERDB_MAINTENANCE_MODE_ENABLED`) 
 - Maintenance mode consists of 2 Workflows:
   - `StartMaintenance` - for pre-upgrade, responsible for
     - Waiting for running snapshots
     - Updating dynamic config to true
     - Pausing and backing up currently running mirrors
   - `EndMaintenance` - for post-upgrade, responsible for
     - Resuming backed up mirrors
     - Updating dynamic config to false
 - During the upgrade (between `Start` and `End`), mirrors cannot be mutated/created in any way, 
 - There is also an instance info API which returns `Ready`/`Maintenance` which can be used for UI changes later.

There are 2 ways to trigger these 2 workflows:
1. API call to flow-api
2. Running the new `maintenance` entrypoint with the respective args

A new task queue is added so that the maintenance tasks can be spun up even during pre-upgrade hooks (from version earlier than ones containing this PR) and this also ensures that always the latest version of the maintenance flows run irrespective of the old version.


flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,298 @@
...
+	mirrorStateRestored = "restore"
+)
+
+type MaintenanceActivity struct {
+	CatalogPool    *pgxpool.Pool
+	Alerter        *alerting.Alerter
+	TemporalClient client.Client
+}
+
+type MaintenanceMirrorsInfo struct {
```
serprex, line 33: unnecessary type

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,298 @@
...
+
+func (a *MaintenanceActivity) getMirrorStatus(ctx context.Context, mirrorInfo MaintenanceMirrorInfoItem) (protos.FlowStatus, error) {
+	encodedState, err := a.TemporalClient.QueryWorkflow(ctx, mirrorInfo.WorkflowId, "", shared.FlowStatusQuery)
+	if err != nil {
+		slog.Error("Error querying mirror status for maintenance", "mirror", mirrorInfo.Name, "workflowId", mirrorInfo.WorkflowId, "error", err)
+		return protos.FlowStatus_STATUS_UNKNOWN, err
+	}
+	var state protos.FlowStatus
+	err = encodedState.Get(&state)
+	if err != nil {
```
serprex, line 74:
```suggestion
	if err := encodedState.Get(&state); err != nil {
```

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,298 @@
...
+
+func (a *MaintenanceActivity) BackgroundAlerter(ctx context.Context) error {
+	defer shared.Interval(ctx, 30*time.Second, func() {
+		activity.RecordHeartbeat(ctx, "Maintenance Workflow is still running")
+	})()
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-time.After(time.Duration(peerdbenv.PeerDBMaintenanceModeWaitAlertSeconds()) * time.Second):
```
serprex, line 293: use `time.Ticker`

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,298 @@
...
+	return nil
+}
+
+func (a *MaintenanceActivity) DisableMaintenanceMode(ctx context.Context) error {
+	slog.Info("Disabling maintenance mode")
+	return peerdbenv.UpdatePeerDBMaintenanceModeEnabled(ctx, a.CatalogPool, false)
+}
+
+func (a *MaintenanceActivity) BackgroundAlerter(ctx context.Context) error {
+	defer shared.Interval(ctx, 30*time.Second, func() {
```
serprex, line 286: this can be another ticker & be in the below `select`. See MaintainPull

flow/workflows/maintenance_flow.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	future := workflow.ExecuteActivity(ctx, maintenance.BackupAllPreviouslyRunningFlows, runningMirrors)
+	err = future.Get(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
+	logger.Info("StartMaintenance workflow completed")
+	return &protos.StartMaintenanceFlowOutput{
+		Version: peerdbenv.PeerDBVersionShaShort(),
```
serprex, line 130: This needs to be wrapped in a local activity. See `workflows/local_activities.go` for how other workflows access environment variables

flow/workflows/maintenance_flow.go
```diff
@@ -0,0 +1,275 @@
...
+
+	future := workflow.ExecuteActivity(ctx, maintenance.DisableMaintenanceMode)
+	err = future.Get(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	logger.Info("EndMaintenance workflow completed")
+	return &protos.EndMaintenanceFlowOutput{
+		Version: peerdbenv.PeerDBVersionShaShort(),
```
serprex, line 225: Again, local activity

flow/workflows/maintenance_flow.go
```diff
@@ -0,0 +1,275 @@
...
+		)
+
+		selector.AddFuture(f, func(f workflow.Future) {
+			err := f.Get(ctx, nil)
+			if err != nil {
+				logger.Error("Error resuming mirror", "mirror", mirror, "error", err)
+			} else {
+				logger.Info("Finished resuming mirror", "mirror", mirror)
+			}
+			ctx.Done()
```
serprex, line 253: this doesn't do anything. `ctx.Done()` returns channel which you can select on to detect ctx cancel

flow/workflows/maintenance_flow.go
```diff
@@ -0,0 +1,275 @@
...
+		})
+	}
+
+	for range mirrorsList.Mirrors {
```
serprex, line 257: Maybe better to use our BoundSelector tho I'm guessing we can handle plenty of these tasks running in parallel

flow/workflows/maintenance_flow.go
```diff
@@ -0,0 +1,275 @@
...
+		selector.Select(ctx)
```
serprex, line 258: Fun quirk: `selector.Select(ctx)` only uses ctx for metadata. If ctx completes while Select running it won't return. You have `selector.AddReceive(ctx.Done(), ...)` to not ignore workflow being cancelled while selecting here. See other workflows for examples of that

Tho in this case context cancel will complete the activies which should complete these futures

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	slog.Info("Checking if mirror is snapshot", "mirror", mirrorInfo.Name, "workflowId", mirrorInfo.WorkflowId, "status", mirrorStatus.String())
+	if mirrorStatus != protos.FlowStatus_STATUS_SNAPSHOT && mirrorStatus != protos.FlowStatus_STATUS_SETUP {
+		return mirrorStatus, nil
+	}
+	slog.Info("Waiting for mirror to finish snapshot",
+		"mirror", mirrorInfo.Name, "workflowId", mirrorInfo.WorkflowId, "status", mirrorStatus.String())
+	defer shared.Interval(
+		ctx, alertEvery, func() {
```
serprex, line 129:
```suggestion
	defer shared.Interval(ctx, alertEvery, func() {
```

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,298 @@
...
+			slog.Warn("[Maintenance] Still waiting for mirror to finish snapshot",
+				"mirror", mirrorInfo.Name, "workflowId", mirrorInfo.WorkflowId, "status", mirrorStatus.String())
+			a.Alerter.LogNonFlowWarning(ctx, telemetry.MaintenanceWait, mirrorInfo.Name, fmt.Sprintf(
+				"Maintenance mode is still waiting for mirror to finish snapshot, mirror=%s, workflowId=%s, status=%s",
+				mirrorInfo.Name, mirrorInfo.WorkflowId, mirrorStatus))
+		},
+	)()
+
+	defer shared.Interval(
+		ctx, logEvery, func() {
```
serprex, line 139:
```suggestion
	defer shared.Interval(ctx, logEvery, func() {
```

flow/peerdbenv/dynamicconf.go
```diff
@@ -261,6 +271,21 @@ func dynamicConfBool(ctx context.Context, env map[string]string, key string) (bo
 	return value, nil
 }
 
+func UpdateDynamicSetting(ctx context.Context, pool *pgxpool.Pool, name string, value *string) error {
+	catalogPool := pool
```
serprex, line 275:
```suggestion
```
just use pool, can reassign parameter

flow/peerdbenv/dynamicconf.go
```diff
@@ -354,3 +379,11 @@ func PeerDBIntervalSinceLastNormalizeThresholdMinutes(ctx context.Context, env m
 func PeerDBApplicationNamePerMirrorName(ctx context.Context, env map[string]string) (bool, error) {
 	return dynamicConfBool(ctx, env, "PEERDB_APPLICATION_NAME_PER_MIRROR_NAME")
 }
+
+func PeerDBMaintenanceModeEnabled(ctx context.Context, env map[string]string) (bool, error) {
+	return dynamicConfBool(ctx, env, "PEERDB_MAINTENANCE_MODE_ENABLED")
+}
+
+func UpdatePeerDBMaintenanceModeEnabled(ctx context.Context, pool *pgxpool.Pool, enabled bool) error {
+	return UpdateDynamicSetting(ctx, pool, "PEERDB_MAINTENANCE_MODE_ENABLED", ptr.String(strconv.FormatBool(enabled)))
```
serprex, line 388: If this is controlled by system I'm not sure we want to expose it to user where they can toggle it

flow/workflows/maintenance_flow.go
```diff
@@ -0,0 +1,275 @@
...
+
+func pauseAndGetRunningMirrors(
+	ctx workflow.Context,
+	mirrorsList activities.MaintenanceMirrorsInfo,
+	logger log.Logger,
+) (activities.MaintenanceMirrorsInfo, error) {
+	selector := workflow.NewSelector(ctx)
+	runningMirrors := make([]bool, len(mirrorsList.Mirrors))
+	for i, mirror := range mirrorsList.Mirrors {
+		activityInput := mirror
```
serprex, line 142:
```suggestion
```

flow/workflows/maintenance_flow.go
```diff
@@ -0,0 +1,275 @@
...
+			if err != nil {
+				logger.Error("Error checking and pausing mirror", "mirror", mirror, "error", err)
+			} else {
+				logger.Info("Finished check and pause for mirror", "mirror", mirror, "wasRunning", wasRunning)
+				runningMirrors[i] = wasRunning
+			}
+			ctx.Done()
+		})
+	}
+	onlyRunningMirrors := make([]activities.MaintenanceMirrorInfoItem, 0)
```
serprex, line 161:
```suggestion
	onlyRunningMirrors := make([]activities.MaintenanceMirrorInfoItem, 0, len(mirrorsList.Mirrors))
```
little extra cap is fine

flow/workflows/maintenance_flow.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+	}
+	return activities.MaintenanceMirrorsInfo{
+		Mirrors: onlyRunningMirrors,
+	}, nil
+}
+
+func getAllMirrors(ctx workflow.Context) (activities.MaintenanceMirrorsInfo, error) {
+	getMirrorsFuture := workflow.ExecuteActivity(ctx, maintenance.GetAllMirrors)
+	var mirrorsList activities.MaintenanceMirrorsInfo
```
serprex, line 180: we usually make activity results protobuf since temporal has special support for protobuf & helps avoid compatibility issues between versions

nexus/catalog/migrations/V40__maintenance_flows.sql
```diff
@@ -0,0 +1,15 @@
+CREATE SCHEMA IF NOT EXISTS maintenance;
+
+CREATE TABLE IF NOT EXISTS maintenance.maintenance_flows
+(
+    id              SERIAL PRIMARY KEY,
+    flow_id         serial    NOT NULL,
```
serprex, line 6: why is flow_id serial?

serprex, line 6:
```suggestion
    flow_id         bigint    NOT NULL,
```
why is flow_id serial? type in code is 64bit, not 32bit

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,298 @@
...
+
+func (a *MaintenanceActivity) WaitForRunningSnapshots(ctx context.Context) (MaintenanceMirrorsInfo, error) {
+	mirrors, err := a.GetAllMirrors(ctx)
+	if err != nil {
+		return MaintenanceMirrorsInfo{}, err
+	}
+
+	slog.Info("Found mirrors for snapshot check", "mirrors", mirrors, "len", len(mirrors.Mirrors))
+
+	waitBeforeAlertingDuration := 30 * time.Minute
```
serprex, line 90: a lot of this waiting logic could exist in workflow, using `workflow.Sleep` & `workflow.Go`

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,298 @@
...
+	slog.Info("Pausing mirror for maintenance", "mirror", mirrorInfo.Name, "workflowId", mirrorInfo.WorkflowId)
+
+	err = model.FlowSignal.SignalClientWorkflow(ctx, a.TemporalClient, mirrorInfo.WorkflowId, "", model.PauseSignal)
+	if err != nil {
+		slog.Error("Error signaling mirror running to pause for maintenance",
+			"mirror", mirrorInfo.Name, "workflowId", mirrorInfo.WorkflowId, "error", err)
+		return false, err
+	}
+	defer shared.Interval(
+		ctx, 30*time.Second, func() {
```
serprex, line 204:
```suggestion
	defer shared.Interval(ctx, 30*time.Second, func() {
```

flow/workflows/maintenance_flow.go
iamKunalGupta: Yup, I think it got left out from somewhere. Just pushed everything before sleeping once. Will fix up everything today

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,290 @@
...
+	return mirrors, nil
+}
+
+func (a *MaintenanceActivity) checkAndWaitIfSnapshot(
+	ctx context.Context,
+	mirrorInfo MaintenanceMirrorInfoItem,
+	logEvery time.Duration,
+	alertEvery time.Duration,
+) (protos.FlowStatus, error) {
+	// In case a mirror was just kicked off, it shows up in the running state, we wait for a bit before checking for snapshot
```
heavycrystal, line 108: this really shouldn't happen, seems like a bug

protos/route.proto
```diff
@@ -420,6 +420,37 @@ message ResyncMirrorRequest {
...
+
+message InstanceInfoRequest {
+}
+
+message InstanceInfoResponse {
+  InstanceStatus status = 1;
+}
+
+enum MaintenanceStatus {
+  MAINTENANCE_STATUS_START = 0;
```
heavycrystal, line 440: `MAINTENANCE_STATE_UNKNOWN` should be 0

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,290 @@
...
+		var info MaintenanceMirrorInfoItem
+		err := row.Scan(&info.Id, &info.Name, &info.WorkflowId, &info.CreatedAt, &info.IsCDC)
+		return info, err
+	})
+	return MaintenanceMirrorsInfo{
+		Mirrors: maintenanceMirrorItems,
+	}, err
+}
+
+func (a *MaintenanceActivity) getMirrorStatus(ctx context.Context, mirrorInfo MaintenanceMirrorInfoItem) (protos.FlowStatus, error) {
```
heavycrystal, line 65: this is a copy from `FlowRequestHandler`, can reuse instead
wherever applicable

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,290 @@
...
+	select distinct on(f.name)
+	  f.id, f.name, f.workflow_id,
+	  f.created_at, coalesce(f.query_string, '')='' is_cdc
+	from flows f
+	`)
+	if err != nil {
+		return MaintenanceMirrorsInfo{}, err
+	}
+
+	maintenanceMirrorItems, err := pgx.CollectRows(rows, func(row pgx.CollectableRow) (MaintenanceMirrorInfoItem, error) {
```
heavycrystal, line 55: nit: could use [RowToStructByName](https://pkg.go.dev/github.com/jackc/pgx/v5#RowToStructByName)

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,290 @@
...
+			mirrorInfo.Name, mirrorInfo.WorkflowId, mirrorStatus))
+	})()
+
+	defer shared.Interval(ctx, logEvery, func() {
+		slog.Info("[Maintenance] Waiting for mirror to finish snapshot",
+			"mirror", mirrorInfo.Name, "workflowId", mirrorInfo.WorkflowId, "status", mirrorStatus.String())
+	})()
+
+	snapshotWaitSleepInterval := 10 * time.Second
+	for mirrorStatus == protos.FlowStatus_STATUS_SNAPSHOT || mirrorStatus == protos.FlowStatus_STATUS_SETUP {
```
heavycrystal, line 140: waiting for a certain `mirrorStatus` can be made a helper

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,290 @@
...
+		mirrorStatus, err = a.getMirrorStatus(ctx, mirrorInfo)
+		if err != nil {
+			slog.Error("Error querying mirror status for maintenance of previously running mirror",
+				"mirror", mirrorInfo.Name, "workflowId", mirrorInfo.WorkflowId, "error", err)
+		}
+	}
+	return true, nil
+}
+
+func (a *MaintenanceActivity) CleanupBackupedFlows(ctx context.Context) error {
```
heavycrystal, line 215: nit: `CleanBackedUpFlows`

flow/cmd/maintenance.go
```diff
@@ -0,0 +1,212 @@
...
+			return err
+		}
+		var output *protos.StartMaintenanceFlowOutput
+		err = workflowRun.Get(ctx, &output)
+		if err != nil {
+			slog.Error("Error in start maintenance workflow", "error", err)
+			return err
+		}
+		slog.Info("Start maintenance workflow completed", "output", output)
+		return WriteMaintenanceOutput(args.OutputFile, StartMaintenanceResult{
```
heavycrystal, line 83: any reason why you're not preferring catalog for this?

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,290 @@
...
+
+	snapshotWaitSleepInterval := 10 * time.Second
+	for mirrorStatus == protos.FlowStatus_STATUS_SNAPSHOT || mirrorStatus == protos.FlowStatus_STATUS_SETUP {
+		time.Sleep(snapshotWaitSleepInterval)
+		activity.RecordHeartbeat(ctx, fmt.Sprintf("Waiting for mirror %s to finish snapshot", mirrorInfo.Name))
+		mirrorStatus, err = a.getMirrorStatus(ctx, mirrorInfo)
+		if err != nil {
+			return mirrorStatus, err
+		}
+	}
```
serprex, line 147: seems like loop body slightly unrolled, could all be one loop

docker-bake.hcl
```diff
@@ -65,6 +66,20 @@ target "flow-worker" {
   ]
 }
 
+target "flow-maintenance" {
```
heavycrystal, line 69: is this something that would go into a critical node pool?

flow/cmd/handler.go
```diff
@@ -327,6 +328,17 @@ func (h *FlowRequestHandler) FlowStateChange(
 ) (*protos.FlowStateChangeResponse, error) {
 	logs := slog.String("flowJobName", req.FlowJobName)
 	slog.Info("FlowStateChange called", logs, slog.Any("req", req))
+	underMaintenance, err := peerdbenv.PeerDBMaintenanceModeEnabled(ctx, nil)
```
heavycrystal, line 331: do we want to prohibit edit peer too?

docker-bake.hcl
iamKunalGupta: We can just yeet it on low cpu one

flow/cmd/handler.go
iamKunalGupta: I think edit peer should be fine, it does not lead to nondeterminism iirc

flow/activities/maintenance_activity.go
iamKunalGupta: It can happen that a mirror got paused by the time we were waiting. We just need to make sure that things are not in these 2 states, rest we will take care of pausing later

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,298 @@
...
+	IsCDC      bool
+}
+
+func (a *MaintenanceActivity) GetAllMirrors(ctx context.Context) (MaintenanceMirrorsInfo, error) {
+	rows, err := a.CatalogPool.Query(ctx,
+		`
+	select distinct on(f.name)
+	  f.id, f.name, f.workflow_id,
+	  f.created_at, coalesce(f.query_string, '')='' is_cdc
+	from flows f
```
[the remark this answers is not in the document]
iamKunalGupta: Was thinking whether we needed any joins when I was first writing this, will cleanup

flow/activities/maintenance_activity.go
iamKunalGupta: moving it to shared for re-use

flow/peerdbenv/dynamicconf.go
iamKunalGupta: This can be helpful for manual intervention or maintenance. We can hide it later from dynamic config if needed

flow/activities/maintenance_activity.go
iamKunalGupta: switched to proto

flow/activities/maintenance_activity.go
iamKunalGupta: will shift to shared.Interval and just wait on ctx.Done()

flow/activities/maintenance_activity.go
iamKunalGupta: `PeerDBMaintenanceModeWaitAlertSeconds` can be a bit large, that is why I had a separate timeout for the hearbeat

flow/workflows/maintenance_flow.go
iamKunalGupta: Yeah, I think we should be goo without it as well. I don't think we need child workflows just for this here

flow/activities/maintenance_activity.go
iamKunalGupta: I think I'll just remove it, `PeerDBMaintenanceModeWaitAlertSeconds` should be enough alerting for us, if anything is needed, we can just add it back in

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (a *MaintenanceActivity) BackgroundAlerter(ctx context.Context) error {
+	defer shared.Interval(ctx, 30*time.Second, func() {
+		activity.RecordHeartbeat(ctx, "Maintenance Workflow is still running")
+	})()
+	defer shared.Interval(ctx, time.Duration(peerdbenv.PeerDBMaintenanceModeWaitAlertSeconds())*time.Second, func() {
+		slog.Warn("Maintenance Workflow is still running")
+		a.Alerter.LogNonFlowWarning(ctx, telemetry.MaintenanceWait, "", "Maintenance mode is still running")
+	})()
+	<-ctx.Done()
```
serprex, line 259: these two can be combined into a single ticker & select in loop. See end of MaintainPull

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,261 @@
...
+	slog.Info("Pausing mirror for maintenance", "mirror", mirror.MirrorName, "workflowId", mirror.WorkflowId)
+
+	err = model.FlowSignal.SignalClientWorkflow(ctx, a.TemporalClient, mirror.WorkflowId, "", model.PauseSignal)
+	if err != nil {
+		slog.Error("Error signaling mirror running to pause for maintenance",
+			"mirror", mirror.MirrorName, "workflowId", mirror.WorkflowId, "error", err)
+		return false, err
+	}
+	defer shared.Interval(ctx, 30*time.Second, func() {
+		slog.Info("Waiting for mirror to pause",
```
serprex, line 172: can move this into loop alongside heartbeat

flow/activities/maintenance_activity.go
iamKunalGupta: I think it should be fine for now, if it was at a third place we can extract

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,261 @@
...
+		time.Sleep(30 * time.Second)
+	}
+
+	mirrorStatus, err := a.getMirrorStatus(ctx, mirror)
+	if err != nil {
+		return mirrorStatus, err
+	}
+	defer shared.Interval(ctx, alertEvery, func() {
+		slog.Warn("[Maintenance] Still waiting for mirror to finish snapshot",
+			"mirror", mirror.MirrorName, "workflowId", mirror.WorkflowId, "status", mirrorStatus.String())
```
serprex, line 99: mirrorStatus needs to either be atomic, have a mutex, or have this logic moved into loop below. You can setup a ticker for each frequency then have one select over all the tickers

iamKunalGupta: Yeah, I am doing that now

flow/activities/maintenance_activity.go
iamKunalGupta: wanted to have a different interval for this, don't need to spam logs for shorter runs

iamKunalGupta: just switching to a ticker here as well

flow/cmd/maintenance.go
iamKunalGupta: Reason at that time was possibility of using configmap and local debugging, replaced with catalog

flow/peerdbenv/dynamicconf.go
iamKunalGupta: Yes, maybe via oauth where different roles can map to different permissions like superadmin, admin or readonly

serprex (approved):

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+		if mirrorStatus == protos.FlowStatus_STATUS_SNAPSHOT || mirrorStatus == protos.FlowStatus_STATUS_SETUP {
+			return false, mirrorStatus, nil
+		}
+		return true, mirrorStatus, nil
+	}, 10*time.Second, fmt.Sprintf("Waiting for mirror %s to finish snapshot", mirror.MirrorName), logEvery)
+	return flowStatus, err
+}
+
+func (a *MaintenanceActivity) EnableMaintenanceMode(ctx context.Context) error {
```
heavycrystal, line 117: this and `DisableMaintenanceMode` could be one activity with different inputs

protos/route.proto
```diff
@@ -441,6 +441,38 @@ message ResyncMirrorRequest {
...
+
+enum MaintenanceStatus {
+  MAINTENANCE_STATUS_UNKOWN = 0;
+  MAINTENANCE_STATUS_START = 1;
+  MAINTENANCE_STATUS_END = 2;
+}
+
+message MaintenanceRequest {
+  MaintenanceStatus status = 1;
+  bool use_peerflow_task_queue = 2;
```
heavycrystal, line 468: why is this an option? can we stick to one task queue?

protos/route.proto
```diff
@@ -441,6 +441,38 @@ message ResyncMirrorRequest {
...
+
+message InstanceInfoRequest {
+}
+
+message InstanceInfoResponse {
+  InstanceStatus status = 1;
+}
+
+enum MaintenanceStatus {
+  MAINTENANCE_STATUS_UNKOWN = 0;
```
heavycrystal, line 461: nit: `UNKNOWN`

nexus/catalog/migrations/V40__maintenance_flows.sql
```diff
@@ -0,0 +1,27 @@
+CREATE SCHEMA IF NOT EXISTS maintenance;
+
+CREATE TABLE IF NOT EXISTS maintenance.maintenance_flows
+(
+    id              serial PRIMARY KEY,
```
heavycrystal, line 5: nit: inconsistent casing for types

flow/cmd/handler.go
```diff
@@ -527,3 +547,50 @@ func (h *FlowRequestHandler) ResyncMirror(
...
+		Status: protos.InstanceStatus_INSTANCE_STATUS_READY,
+	}, nil
+}
+
+func (h *FlowRequestHandler) Maintenance(ctx context.Context, in *protos.MaintenanceRequest) (*protos.MaintenanceResponse, error) {
+	taskQueueId := shared.MaintenanceFlowTaskQueue
+	if in.UsePeerflowTaskQueue {
+		taskQueueId = shared.PeerFlowTaskQueue
+	}
+	if in.Status == protos.MaintenanceStatus_MAINTENANCE_STATUS_START {
```
heavycrystal, line 574: nit: use switch case

flow/cmd/handler.go
```diff
@@ -527,3 +547,50 @@ func (h *FlowRequestHandler) ResyncMirror(
...
+		workflowRun, err := peerflow.RunEndMaintenanceWorkflow(ctx, h.temporalClient, &protos.EndMaintenanceFlowInput{}, taskQueueId)
+		if err != nil {
+			return nil, err
+		}
+		return &protos.MaintenanceResponse{
+			WorkflowId: workflowRun.GetID(),
+			RunId:      workflowRun.GetRunID(),
+		}, nil
+	}
+	return &protos.MaintenanceResponse{}, errors.New("invalid maintenance status")
```
heavycrystal, line 595: nit: return `nil`

flow/cmd/maintenance.go
```diff
@@ -0,0 +1,250 @@
...
+	SkipOnApiVersionMatch             bool
+	SkipOnNoMirrors                   bool
+	UseMaintenanceTaskQueue           bool
+	AssumeSkippedMaintenanceWorkflows bool
+}
+
+type StartMaintenanceResult struct {
+	SkippedReason *string `json:"skippedReason,omitempty"`
+	APIVersion    string  `json:"apiVersion,omitempty"`
+	CLIVersion    string  `json:"cliVersion,omitempty"`
```
heavycrystal, line 43: not sure if `CLIVersion` is the best name?

protos/route.proto
iamKunalGupta: a separate task queue is generally needed because we always want to run the latest version of the maintenance work flow. In some cases we may not have the maintenance task queue polling worker available, in that case flow worker + API will do it.

flow/cmd/maintenance.go
iamKunalGupta: yeah, this is the local version of the CLI which is running the maintenance mode

protos/route.proto
serprex: In which case is that?

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,292 @@
...
+	return &protos.MaintenanceMirrors{
+		Mirrors: maintenanceMirrorItems,
+	}, err
+}
+
+func (a *MaintenanceActivity) getMirrorStatus(ctx context.Context, mirror *protos.MaintenanceMirror) (protos.FlowStatus, error) {
+	return shared.GetWorkflowStatus(ctx, a.TemporalClient, mirror.WorkflowId)
+}
+
+func (a *MaintenanceActivity) WaitForRunningSnapshots(ctx context.Context) (*protos.MaintenanceMirrors, error) {
```
heavycrystal, line 61: I'm thinking that it's probably simpler and better to fail entering maintenance mode than wait for an unknown duration of time for running snapshots to finish, we can process upgrades for stuff that enters maintenance mode and try these instances later

iamKunalGupta: We have a background alerter for these uses cases where we wait, and it might not be a good idea to fail entering maintenance mode: the whole idea is for upgrades to be automatic.

iamKunalGupta: > and try these instances later
This is what maintenance mode will also aim to do

protos/route.proto
iamKunalGupta: Maintenance worker will only be available from the pre-upgrade (before instance gets upgraded) till end of post-upgrade (after an instance gets upgraded).

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,292 @@
...
+
+	slog.Info("Checking and waiting if mirror is snapshot", "mirror", mirror.MirrorName, "workflowId", mirror.WorkflowId, "status",
+		mirrorStatus.String())
+	if mirrorStatus != protos.FlowStatus_STATUS_SNAPSHOT && mirrorStatus != protos.FlowStatus_STATUS_SETUP {
+		return mirrorStatus, nil
+	}
+
+	flowStatus, err := RunEveryIntervalUntilFinish(ctx, func() (bool, protos.FlowStatus, error) {
+		activity.RecordHeartbeat(ctx, fmt.Sprintf("Waiting for mirror %s to finish snapshot", mirror.MirrorName))
+		mirrorStatus, err = a.getMirrorStatus(ctx, mirror)
```
serprex, line 105:
```suggestion
		mirrorStatus, err := a.getMirrorStatus(ctx, mirror)
```

flow/activities/maintenance_activity.go
```diff
@@ -0,0 +1,292 @@
...
+	mirrorStatus, err := a.getMirrorStatus(ctx, mirror)
+	if err != nil {
+		return mirrorStatus, err
+	}
+
+	slog.Info("Checking and waiting if mirror is snapshot", "mirror", mirror.MirrorName, "workflowId", mirror.WorkflowId, "status",
+		mirrorStatus.String())
+	if mirrorStatus != protos.FlowStatus_STATUS_SNAPSHOT && mirrorStatus != protos.FlowStatus_STATUS_SETUP {
+		return mirrorStatus, nil
+	}
```
serprex, line 101:
```suggestion
```

flow/cmd/maintenance.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	row := pool.QueryRow(ctx, `
+	select cli_version, api_version, skipped, skipped_reason
+	from maintenance.start_maintenance_outputs
+	order by created_at desc
+	limit 1
+	`)
+	var result StartMaintenanceResult
+	err = row.Scan(&result.CLIVersion, &result.APIVersion, &result.Skipped, &result.SkippedReason)
+	if err != nil {
```
serprex, line 227:
```suggestion
	var result StartMaintenanceResult
	if err := pool.QueryRow(ctx, `
	select cli_version, api_version, skipped, skipped_reason
	from maintenance.start_maintenance_outputs
	order by created_at desc
	limit 1
	`).Scan(&result.CLIVersion, &result.APIVersion, &result.Skipped, &result.SkippedReason); err != nil {
```

flow/shared/worklow.go
```diff
@@ -0,0 +1,28 @@
...
+	res, err := temporalClient.QueryWorkflow(ctx, workflowID, "", FlowStatusQuery)
+	if err != nil {
+		slog.Error(fmt.Sprintf("failed to get status in workflow with ID %s: %s", workflowID, err.Error()))
+		return protos.FlowStatus_STATUS_UNKNOWN,
+			fmt.Errorf("failed to get status in workflow with ID %s: %w", workflowID, err)
+	}
+	var state protos.FlowStatus
+	err = res.Get(&state)
+	if err != nil {
+		slog.Error(fmt.Sprintf("failed to get status in workflow with ID %s: %s", workflowID, err.Error()))
```
serprex, line 23:
```suggestion
		slog.Error("failed to get status in workflow with ID "+workflowID, slog.Any("error", err)))
```

flow/shared/worklow.go
```diff
@@ -0,0 +1,28 @@
...
+func GetWorkflowStatus(ctx context.Context, temporalClient client.Client, workflowID string) (protos.FlowStatus, error) {
+	res, err := temporalClient.QueryWorkflow(ctx, workflowID, "", FlowStatusQuery)
+	if err != nil {
+		slog.Error(fmt.Sprintf("failed to get status in workflow with ID %s: %s", workflowID, err.Error()))
+		return protos.FlowStatus_STATUS_UNKNOWN,
+			fmt.Errorf("failed to get status in workflow with ID %s: %w", workflowID, err)
+	}
+	var state protos.FlowStatus
+	err = res.Get(&state)
+	if err != nil {
```
serprex, line 22:
```suggestion
	if err := res.Get(&state); err != nil {
```

flow/shared/worklow.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/PeerDB-io/peer-flow/generated/protos"
+)
+
+func GetWorkflowStatus(ctx context.Context, temporalClient client.Client, workflowID string) (protos.FlowStatus, error) {
+	res, err := temporalClient.QueryWorkflow(ctx, workflowID, "", FlowStatusQuery)
+	if err != nil {
+		slog.Error(fmt.Sprintf("failed to get status in workflow with ID %s: %s", workflowID, err.Error()))
+		return protos.FlowStatus_STATUS_UNKNOWN,
+			fmt.Errorf("failed to get status in workflow with ID %s: %w", workflowID, err)
```
serprex, line 18:
```suggestion
		slog.Error(fmt.Sprintf("failed to query status in workflow with ID %s: %s", workflowID, err.Error()))
		return protos.FlowStatus_STATUS_UNKNOWN,
			fmt.Errorf("failed to query status in workflow with ID %s: %w", workflowID, err)
```
makes error messages distinct so that when looking up error messages in code we know which line was hit

flow/activities/maintenance_activity.go
iamKunalGupta: was planning to check before the loop and log, but I think it is fine anyways

heavycrystal:
was testing it
edit mirror during maintenance:
image

resume mirror doesn't error but doesn't work either

heavycrystal (approved):

PR #2214, opened 2024-11-04
[postgres] reduce unnecessary queries

### LINKED ISSUE #2210: Postgres connector: No need to get CustomTypeMap for API, non-engine purposes
We obtain the postgres connector in various API endpoints. As part of NewPostgresConnector, we load and store the custom types of the source postgres peer. However this custom type map will not be needed for many API endpoints

The SELECT query to get the custom types is an additional operation on source which is unnecessary in these cases. In some busy databases the query can also get locked for a while, resulting in long page load times in the UI

1. Identify places where we get a postgres connector
2. We can put that specific fetch under a flag maybe for the above

heavycrystal:
closes #2210

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -67,9 +67,9 @@ func (c *PostgresConnector) NewPostgresCDCSource(cdcConfig *PostgresCDCConfig) *
 		relationMessageMapping:       cdcConfig.RelationMessageMapping,
 		slot:                         cdcConfig.Slot,
 		publication:                  cdcConfig.Publication,
-		childToParentRelIDMapping:    cdcConfig.ChildToParentRelIDMap,
 		typeMap:                      pgtype.NewMap(),
 		commitLock:                   nil,
+		childToParentRelIDMapping:    cdcConfig.ChildToParentRelIDMap,
 		catalogPool:                  cdcConfig.CatalogPool,
 		flowJobName:                  cdcConfig.FlowJobName,
 		hushWarnUnhandledMessageType: make(map[pglogrepl.MessageType]struct{}),
@@ -111,6 +111,7 @@ type replProcessor[Items model.Items] interface {
 		p *PostgresCDCSource,
 		tuple *pglogrepl.TupleDataColumn,
 		col *pglogrepl.RelationMessageColumn,
+		customTypeMapping map[uint32]string,
 	) error
 }
 
@@ -125,6 +126,7 @@ func (pgProcessor) Process(
 	p *PostgresCDCSource,
 	tuple *pglogrepl.TupleDataColumn,
 	col *pglogrepl.RelationMessageColumn,
+	customTypeMapping map[uint32]string,
 ) error {
 	switch tuple.DataType {
 	case 'n': // null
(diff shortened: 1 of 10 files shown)


flow/connectors/postgres/cdc.go
```diff
@@ -85,21 +85,18 @@ func GetChildToParentRelIDMap(ctx context.Context, conn *pgx.Conn) (map[uint32]u
 		WHERE parent.relkind='p';
 	`
 
-	rows, err := conn.Query(ctx, query, pgx.QueryExecModeSimpleProtocol)
```
serprex, line 88: I remember putting this in for a reason

flow/shared/postgres.go
```diff
@@ -66,7 +66,7 @@ func GetCustomDataTypes(ctx context.Context, conn *pgx.Conn) (map[uint32]string,
 		var typeID pgtype.Uint32
```
serprex, line 66: Also change this to pgx.ForEachRow

flow/connectors/postgres/postgres.go
```diff
@@ -106,14 +100,25 @@ func NewPostgresConnector(ctx context.Context, env map[string]string, pgConfig *
...
-		customTypesMapping:     customTypeMap,
+		customTypesMapping:     nil,
 		metadataSchema:         metadataSchema,
 		hushWarnOID:            make(map[uint32]struct{}),
 		logger:                 logger,
 		relationMessageMapping: make(model.RelationMessageMapping),
 	}, nil
 }
 
+func (c *PostgresConnector) lazyInitCustomTypesMapping(ctx context.Context) error {
```
serprex, line 111: this is really error prone unless this method returns the custom types map & code doesn't directly access

flow/connectors/postgres/postgres.go
```diff
@@ -42,6 +42,7 @@ type PostgresConnector struct {
 	customTypesMapping     map[uint32]string
 	hushWarnOID            map[uint32]struct{}
 	relationMessageMapping model.RelationMessageMapping
+	pgVersion              *shared.PGVersion
```
serprex, line 45:
```suggestion
	pgVersion              shared.PGVersion
```
Can treat `0` as uninitialized

flow/connectors/postgres/postgres.go
heavycrystal: I'll switch to return map but I'm not sure about the threadsafe requirement, where is that possible? `SetupReplConn` is where I initialize it and that should be before any of the `ReplPing` stuff

serprex: Mostly pointing out that any call to this function needs to have exclusion on connection, but I don't think that's an issue

heavycrystal: can tie it to a mutex if needed in the future, same concern for caching version

heavycrystal: `customTypesMapping` is in some hot paths, keeping it as a field

serprex: it's okay, nil check is cheap

flow/connectors/postgres/qrep_query_executor_test.go
```diff
@@ -214,28 +181,20 @@ func TestAllDataTypes(t *testing.T) {
...
-		t.Fatalf("expected %v, got %v", expectedJSON, record[7].Value())
-	}
+	require.Equal(t, expectedJSON, record[7].Value(), "expected '{\"key\":\"value\"}'")
 
 	actualUUID := record[8].Value().([16]uint8)
-	if !bytes.Equal(actualUUID[:], savedUUID[:]) {
-		t.Fatalf("expected %v, got %v", savedUUID, actualUUID)
-	}
+	require.Equal(t, savedUUID[:], actualUUID[:], "expected savedUUID: %v", savedUUID)
+	// not sure why the values are unequal
```
serprex, line 194: might be caused by PG being microsecond precision. Try rounding `time.Now()` to seconds

serprex (approved):

PR #2215, opened 2024-11-04
CH Sorting Key: Introduce sorting key expression

Amogh-Bharadwaj:
This PR introduces the backend to support specifying sorting keys such as `function(column)` Ex: toDate(updated_at)

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -233,7 +233,11 @@ func getOrderedOrderByColumns(
 
 	orderbyColumns := make([]string, len(orderby))
 	for idx, col := range orderby {
-		orderbyColumns[idx] = getColName(colNameMap, col.SourceName)
+		sortingKeyName := col.SourceName
+		if col.SortingKeyExpression != nil && *col.SortingKeyExpression != "" {
+			sortingKeyName = *col.SortingKeyExpression
+		}
+		orderbyColumns[idx] = getColName(colNameMap, sortingKeyName)
 	}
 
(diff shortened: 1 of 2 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -233,7 +233,11 @@ func getOrderedOrderByColumns(
 
 	orderbyColumns := make([]string, len(orderby))
 	for idx, col := range orderby {
-		orderbyColumns[idx] = getColName(colNameMap, col.SourceName)
+		sortingKeyName := col.SourceName
+		if col.SortingKeyExpression != nil && *col.SortingKeyExpression != "" {
```
heavycrystal, line 237: checking for both defeats the purpose of an `optional string`
either switch to just `string` or ensure you don't send empty string from upstream

serprex: I prefer not using optional

flow/connectors/clickhouse/normalize.go
```diff
@@ -233,7 +233,11 @@ func getOrderedOrderByColumns(
...
+			sortingKeyName = *col.SortingKeyExpression
```
serprex, line 238: lack of validation means this opens up using `;` to run an arbitrary query. Hard to say what our threat model is, but we should explicitly get the okay that our threat model assumes anyone with access to ui has access to run arbitrary queries with any clickhouse peer

Amogh-Bharadwaj:
Closing for now as requires more consideration than initially estimated

PR #2217, opened 2024-11-04
 Enable wal heartbeats by default

serprex:
users predating PG14 should change the query or disable wal heartbeats


iamKunalGupta (approved):
let's test and merge

serprex:
Sending updated LSN & logging that updated LSN. PG is ignoring when we send unrequested standby status updates with updated LSN

heavycrystal (approved):

iamKunalGupta:
@serprex do we need to send `WalFlushPosition` instead for RDS?

iamKunalGupta:
> @serprex do we need to send `WalFlushPosition` instead for RDS?

nvm, the library does it for us

heavycrystal (approved):

PR #2220, opened 2024-11-05
Add more information logs for UI logs table

Amogh-Bharadwaj:
Currently the only information logs we store in catalog is "pushed x records"
It is important for us to show users more than just that

Note that this PR pertains to the logs table in UI. It has nothing to do with alerting

This PR logs info about some key moments during replication:
- Setting up of target tables
- Setup of replication slot and publication
- Replication of all partitions in QRep/Initial load
- Schema changes
- Resync completion
- Source cleanup of drop flow
- Destination cleanup of drop flow
- Removal of tables from publication
- Addition of tables from publication

- [x] Functional test

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -531,6 +531,8 @@ func (a *FlowableActivity) GetQRepPartitions(ctx context.Context,
 		}
 	}
 
+	a.Alerter.LogFlowInfo(ctx, config.FlowJobName, fmt.Sprintf("obtained partitions for table %s", config.WatermarkTable))
+
 	return &protos.QRepParitionResult{
 		Partitions: partitions,
 	}, nil
@@ -657,6 +659,8 @@ func (a *FlowableActivity) DropFlowSource(ctx context.Context, req *protos.DropF
 		return pullCleanupErr
 	}
 
+	a.Alerter.LogFlowInfo(ctx, req.FlowJobName, "Cleaned up source peer replication objects. Any replication slot or publication created by PeerDB has been removed.")
+
 	return nil
 }
 
@@ -675,6 +679,8 @@ func (a *FlowableActivity) DropFlowDestination(ctx context.Context, req *protos.
 			exceptions.NewDropFlowError(fmt.Errorf("[DropFlowDestination] failed to clean up destination: %w", err)),
 		)
 	}
+
+	a.Alerter.LogFlowInfo(ctx, req.FlowJobName, "Cleaned up destination peer replication objects. Any PeerDB metadata storage has been dropped.")
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


flow/activities/flowable.go
```diff
@@ -247,6 +248,7 @@ func (a *FlowableActivity) CreateNormalizedTable(
 	})
 	defer shutdown()
 
+	a.Alerter.LogFlowInfo(ctx, config.FlowName, "Setting up destination tables")
```
iamKunalGupta, line 251: why do we have this twice?

flow/activities/flowable.go
```diff
@@ -265,15 +267,19 @@ func (a *FlowableActivity) CreateNormalizedTable(
 		numTablesSetup.Add(1)
 		if !existing {
 			logger.Info("created table " + tableIdentifier)
+			a.Alerter.LogFlowInfo(ctx, config.FlowName, "created table "+tableIdentifier+" in destination")
```
iamKunalGupta, line 270: Might be too noisy for a large number of tables

flow/activities/flowable.go
```diff
@@ -509,6 +515,8 @@ func (a *FlowableActivity) GetQRepPartitions(ctx context.Context,
 		}
 	}
 
+	a.Alerter.LogFlowInfo(ctx, config.FlowJobName, fmt.Sprintf("obtained partitions for table %s", config.WatermarkTable))
```
iamKunalGupta, line 518: Maybe just log instead of sending an alert? (too noisy)

iamKunalGupta (commented):
overall, I think per table alerts should be removed and converted to logs as they can be noisy

flow/activities/flowable.go
Amogh-Bharadwaj: These are not alerts. This is for our logs table in UI
I think `LogFlowEvent` is for alerts ?

iamKunalGupta: My bad, but still do we want to do it for all tables?

flow/activities/flowable.go
serprex: think it's meant to say conn.StartupSetupNormalizedTables has created tables, beginning ingestion. Wording needs to be updated

flow/activities/snapshot_activity.go
```diff
@@ -58,7 +58,7 @@ func (a *SnapshotActivity) SetupReplication(
 ) (*protos.SetupReplicationOutput, error) {
 	ctx = context.WithValue(ctx, shared.FlowNameKey, config.FlowJobName)
 	logger := activity.GetLogger(ctx)
-
+	a.Alerter.LogFlowInfo(ctx, config.FlowJobName, "Setting up replication slot and publication")
 	a.Alerter.LogFlowEvent(ctx, config.FlowJobName, "Started Snapshot Flow Job")
```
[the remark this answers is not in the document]
Amogh-Bharadwaj: This is for alerting, we need this

flow/activities/flowable.go
Amogh-Bharadwaj: yeah because we have scarcity of user facing info during initial load - common feedback from customers cc @iskakaushik

iamKunalGupta (approved):

serprex (approved):

PR #2229, opened 2024-11-06
Implement table rename with atomic exchange

martin31821:
Currently, peerdb creates a _resync table and then tries to run 

```sql
DROP TABLE IF EXISTS target_table;
RENAME TABLE _resync_table TO target_table;
```

however, the problem is that this procedure breaks, once a depending dictionary is defined on these tables.

In order to enable such behavior, this PR modifies the ClickhouseConnector of Flow to 

- check whether the database engine supports the `EXCHANGE TABLES` command (https://clickhouse.com/docs/en/sql-reference/statements/exchange)
- If it does and the table already existed, runs `EXCHANGE TABLES` followed by a `DROP` which atomically swaps and therefore keeps the dictionary references
- If it doesnt, there's a fallback to the old method.

Currently, this will work with the Atomic() database engine.

### DIFF
--- a/flow/connectors/clickhouse/cdc.go
+++ b/flow/connectors/clickhouse/cdc.go
@@ -93,8 +93,7 @@ func (c *ClickHouseConnector) syncRecordsViaAvro(
 		return nil, err
 	}
 
-	err = c.ReplayTableSchemaDeltas(ctx, req.FlowJobName, req.Records.SchemaDeltas)
-	if err != nil {
+	if err := c.ReplayTableSchemaDeltas(ctx, req.FlowJobName, req.Records.SchemaDeltas); err != nil {
 		return nil, fmt.Errorf("failed to sync schema changes: %w", err)
 	}
 
@@ -113,8 +112,7 @@ func (c *ClickHouseConnector) SyncRecords(ctx context.Context, req *model.SyncRe
 		return nil, err
 	}
 
-	err = c.FinishBatch(ctx, req.FlowJobName, req.SyncBatchID, res.LastSyncedCheckpointID)
-	if err != nil {
+	if err := c.FinishBatch(ctx, req.FlowJobName, req.SyncBatchID, res.LastSyncedCheckpointID); err != nil {
 		c.logger.Error("failed to increment id", slog.Any("error", err))
 		return nil, err
 	}
@@ -137,15 +135,13 @@ func (c *ClickHouseConnector) ReplayTableSchemaDeltas(ctx context.Context, flowJ
 		for _, addedColumn := range schemaDelta.AddedColumns {
 			clickHouseColType, err := qvalue.QValueKind(addedColumn.Type).ToDWHColumnType(protos.DBType_CLICKHOUSE)
 			if err != nil {
-				return fmt.Errorf("failed to convert column type %s to ClickHouse type: %w",
-					addedColumn.Type, err)
+				return fmt.Errorf("failed to convert column type %s to ClickHouse type: %w", addedColumn.Type, err)
 			}
 			err = c.execWithLogging(ctx,
 				fmt.Sprintf("ALTER TABLE %s ADD COLUMN IF NOT EXISTS \"%s\" %s",
 					schemaDelta.DstTableName, addedColumn.Name, clickHouseColType))
 			if err != nil {
-				return fmt.Errorf("failed to add column %s for table %s: %w", addedColumn.Name,
-					schemaDelta.DstTableName, err)
+				return fmt.Errorf("failed to add column %s for table %s: %w", addedColumn.Name, schemaDelta.DstTableName, err)
 			}
 			c.logger.Info(fmt.Sprintf("[schema delta replay] added column %s with data type %s", addedColumn.Name,
 				addedColumn.Type),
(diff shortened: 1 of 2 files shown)


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -90,6 +94,31 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
...
+		if err := c.exec(ctx, "DROP TABLE IF EXISTS "+validateDummyTableName2); err != nil {
+			c.logger.Error("validation failed to drop table", slog.String("table", validateDummyTableName2), slog.Any("error", err))
+		}
+	}()
+
+	// exchange the two tables
+	if err := c.exec(ctx,
+		fmt.Sprintf("EXCHANGE TABLES %s AND %s", validateDummyTableName, validateDummyTableName2),
+	); err != nil {
+		c.logger.Warn("failed to run exchange tables on the database, will fall back to drop/rename", "err", err)
```
serprex, line 118:
```suggestion
		c.logger.Warn("failed to run exchange tables on the database, will fall back to drop/rename", "error", err)
```

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -29,11 +29,12 @@ import (
 
 type ClickHouseConnector struct {
 	*metadataStore.PostgresMetadata
-	database      clickhouse.Conn
-	logger        log.Logger
-	config        *protos.ClickhouseConfig
-	credsProvider *utils.ClickHouseS3Credentials
-	s3Stage       *ClickHouseS3Stage
+	database           clickhouse.Conn
+	dbSupportsExchange bool
```
serprex, line 33: this won't persist between validation & rename

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -180,12 +209,13 @@ func NewClickHouseConnector(
...
 	connector := &ClickHouseConnector{
-		database:         database,
-		PostgresMetadata: pgMetadata,
-		config:           config,
-		logger:           logger,
-		credsProvider:    &clickHouseS3CredentialsNew,
-		s3Stage:          NewClickHouseS3Stage(),
+		database:           database,
+		PostgresMetadata:   pgMetadata,
+		dbSupportsExchange: true,
```
serprex, line 214: can do a minversion check? we require a pretty recent version, so can probably assume `true` & drop variable

martin31821: true/false here depends on the used database engine - only atomic supports it, whereas lazy doesn't.

martin31821: however, I can just try to run an exchange and if there's an error just transparently fallback to the old drop/rename method.

flow/connectors/clickhouse/cdc.go
```diff
@@ -197,19 +197,34 @@ func (c *ClickHouseConnector) RenameTables(
...
-		// drop the dst table if exists
-		err = c.execWithLogging(ctx, "DROP TABLE IF EXISTS "+renameRequest.NewName)
-		if err != nil {
-			return nil, fmt.Errorf("unable to drop table %s: %w", renameRequest.NewName, err)
-		}
+		if c.dbSupportsExchange && originalTableExists {
+			c.logger.Info("attempting atomic exchange %s to %s", renameRequest.CurrentName, renameRequest.NewName)
+			// the clickhouse ATOMIC engine supports a special query to exchange two tables
+			// this means, we can have dependent (materialized) views and dictionaries on these tables.
+			err = c.execWithLogging(ctx, fmt.Sprintf("EXCHANGE TABLES %s and %s", renameRequest.NewName, renameRequest.CurrentName))
```
heavycrystal, line 204: there is a possibility of the resync table and the existing table having different schemas

would the EXCHANGE fail? would it fail if MVs depend on the schema of the existing table?

flow/connectors/clickhouse/clickhouse.go
heavycrystal: this needs to be populated by the `NewClickHouseConnector` method instead

can be populated by just running `SELECT engine_full FROM system.databases` unless EXCHANGE needs extra permissions

flow/connectors/clickhouse/cdc.go
```diff
@@ -197,19 +197,34 @@ func (c *ClickHouseConnector) RenameTables(
 			c.logger.Info(fmt.Sprintf("table '%s' does not exist, skipping soft-deletes transfer for it", renameRequest.NewName))
 		}
 
-		// drop the dst table if exists
-		err = c.execWithLogging(ctx, "DROP TABLE IF EXISTS "+renameRequest.NewName)
-		if err != nil {
-			return nil, fmt.Errorf("unable to drop table %s: %w", renameRequest.NewName, err)
-		}
+		if c.dbSupportsExchange && originalTableExists {
+			c.logger.Info("attempting atomic exchange %s to %s", renameRequest.CurrentName, renameRequest.NewName)
```
serprex, line 201: logger.Info is log/slog, not format strings

flow/connectors/clickhouse/clickhouse.go
serprex: yes, I think try & fallback on error makes sense. Even better if you can fallback on specific errors

serprex:
Thanks for contribution, when I have time I'll try it out & push any adjustments. Idea seems good

martin31821:
I've updated the PR with most of the comments - changed the implementation to try exchange first and when clickhouse returns ERR_NOT_IMPLEMENTED to fallback to the DROP+RENAME method + a warning in the validator:

image

I'd like to experiment with the possibility of schema differences mentioned by @heavycrystal. 
My use case actually is a dictionary which is (periodically re-)built out of the synced tables which I later need to add metadata to a downstream query.

flow/connectors/clickhouse/cdc.go
martin31821: So in my tests - the EXCHANGE itself works.

Case A: schema of _resync is compatible with the views (i.e. resync changes/drops/add columns not in View/Dictionary) everything works as expected.
Case B: schema of _resync impacts a dictionary: still works, however the dictionary won't update afterwards
Case C: schema of _resync impacts a view: EXCHANGE itself works, for materialized view the inserts of the data afterwards won't work because the MV references e.g. a nonexisting/incompatible column. Data insertion fails with Code 47.
image

PR #2239, opened 2024-11-11
QRep Overwrite Mode: introduce full refresh code path

Amogh-Bharadwaj:
Currently overwrite mode in query replication has the following behaviour:
Before new rows are synced, truncate the destination table

This PR introduces a dynconf-gated (default false) full refresh mode which if enabled, has the same behaviour as above but pulls the entire data from the source table at every wait time interval

### DIFF
--- a/flow/workflows/qrep_flow.go
+++ b/flow/workflows/qrep_flow.go
@@ -464,10 +464,7 @@ func QRepWaitForNewRowsWorkflow(ctx workflow.Context, config *protos.QRepConfig,
 	}
 
 	optedForOverwrite := config.WriteMode.WriteType == protos.QRepWriteType_QREP_WRITE_MODE_OVERWRITE
-	fullRefresh := false
-	if optedForOverwrite {
-		fullRefresh = getQRepOverwriteFullRefreshMode(ctx, logger, config.Env)
-	}
+	fullRefresh := optedForOverwrite && getQRepOverwriteFullRefreshMode(ctx, logger, config.Env)
 	// If no new rows are found, continue as new
 	if !hasNewRows || fullRefresh {
 		waitBetweenBatches := 5 * time.Second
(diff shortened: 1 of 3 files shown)


flow/workflows/qrep_flow.go
```diff
@@ -461,8 +463,13 @@ func QRepWaitForNewRowsWorkflow(ctx workflow.Context, config *protos.QRepConfig,
 		return fmt.Errorf("error checking for new rows: %w", err)
 	}
 
+	optedForOverwrite := config.WriteMode.WriteType == protos.QRepWriteType_QREP_WRITE_MODE_OVERWRITE
+	fullRefresh := false
+	if optedForOverwrite {
+		fullRefresh = getQRepOverwriteFullRefreshMode(ctx, logger, config.Env)
+	}
```
serprex, line 470:
```suggestion
	fullRefresh := optedForOverwrite && getQRepOverwriteFullRefreshMode(ctx, logger, config.Env)
```

flow/workflows/qrep_flow.go
```diff
@@ -473,6 +480,9 @@ func QRepWaitForNewRowsWorkflow(ctx workflow.Context, config *protos.QRepConfig,
 		}
 
 		logger.Info("QRepWaitForNewRowsWorkflow: continuing the loop")
```
serprex, line 482: this log should be moved below if-then-return

seems a little odd, why do we not wait for new rows when doing full refreshes?

flow/workflows/qrep_flow.go
```diff
@@ -545,8 +555,17 @@ func QRepFlowWorkflow(
 		return state, err
 	}
 
-	if !config.InitialCopyOnly && state.LastPartition != nil {
-		if err := q.waitForNewRows(ctx, signalChan, state.LastPartition); err != nil {
+	fullRefresh := false
+	optedForOverwrite := config.WriteMode.WriteType == protos.QRepWriteType_QREP_WRITE_MODE_OVERWRITE
+	lastPartition := state.LastPartition
+	if optedForOverwrite {
```
serprex, line 561:
```suggestion
	lastPartition := state.LastPartition
	if config.WriteMode.WriteType == protos.QRepWriteType_QREP_WRITE_MODE_OVERWRITE {
```

flow/workflows/qrep_flow.go
Amogh-Bharadwaj: There are always going to be new rows since it's full refresh so this is just to force it to wait for the requested wait time (`WaitBetweenBatchesSeconds`)

serprex (approved):

PR #2241, opened 2024-11-11
ci: replace s3 with minio

serprex:


docker-compose-dev.yml
```diff
@@ -209,7 +209,7 @@ services:
       - flow-api
 
   minio:
-    image: minio/minio:RELEASE.2024-07-16T23-46-41Z
+    image: minio/minio:RELEASE.2024-11-07T00-52-20Z
```
heavycrystal, line 212: can swap these to the bitnami image, or as a followup

.github/workflows/flow.yml
```diff
@@ -119,9 +141,20 @@ jobs:
...
         working-directory: ./flow
         env:
-          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          AWS_REGION: ${{ secrets.AWS_REGION }}
+          AWS_ENDPOINT_URL_S3: http://localhost:9999
+          AWS_ACCESS_KEY_ID: minio
+          AWS_SECRET_ACCESS_KEY: miniosecret
+          AWS_REGION: us-east-1
+          PEERDB_CLICKHOUSE_AWS_CREDENTIALS_AWS_ACCESS_KEY_ID: minio
```
heavycrystal, line 148: is there any way to not repeat the same values here?
user/password repeated 4 times

iskakaushik (approved):

flow/e2e/snowflake/qrep_flow_sf_test.go
```diff
@@ -123,6 +123,7 @@ func (s PeerFlowE2ETestSuiteSF) Test_Complete_QRep_Flow_Avro_SF_Upsert_Simple()
 }
 
 func (s PeerFlowE2ETestSuiteSF) Test_Complete_QRep_Flow_Avro_SF_S3() {
+	s.t.Skip("aws s3 broken in ci") // TODO fix
```
heavycrystal, line 126: TODO

flow/e2e/s3/s3_helper.go
```diff
@@ -24,28 +24,48 @@ type S3TestHelper struct {
...
+	case Aws:
+		credsPath = os.Getenv("TEST_S3_CREDS")
+		bucketName = "peerdb-test-bucket"
+	case Gcs:
 		credsPath = os.Getenv("TEST_GCS_CREDS")
 		bucketName = "peerdb_staging"
+		endpoint = "https://storage.googleapis.com"
+	case Minio:
+		bucketName = "peerdb"
+		endpoint = "http://localhost:9999" // TODO env variable
```
heavycrystal: TODO

flow/e2e/s3/s3_helper.go
```diff
@@ -24,28 +24,48 @@ type S3TestHelper struct {
...
+
+func NewS3TestHelper(s3environment S3Environment) (*S3TestHelper, error) {
+	var config utils.S3PeerCredentials
+	var endpoint string
+	var credsPath string
+	var bucketName string
+	switch s3environment {
+	case Aws:
+		credsPath = os.Getenv("TEST_S3_CREDS")
+		bucketName = "peerdb-test-bucket"
```
heavycrystal, line 43: not sure why this has to be in code?
if it does, make it a map?

docker-compose-dev.yml
serprex: would prefer as followup, these images work well enough as is, but would be nice to have docker-compose depend on same images as CI

.github/workflows/flow.yml
serprex: No

flow/e2e/s3/s3_helper.go
serprex: Hard to be a map when it has specific behavior (loading files for s3/gcs, using env for minio). This change is fleshing out two-state bool into tri-state enum since we already had this logic for aws/gcs, but now we support minio

PR #2253, opened 2024-11-14
fix lint

serprex:
should really not use auto-merge so much with linting not actually being required


heavycrystal (approved):
great for commit history

PR #2258, opened 2024-11-14
use disconnected ctx to ensure snapshot connection drops

heavycrystal:
Temporal has a thing for everything


serprex (approved):

flow/workflows/snapshot_flow.go
```diff
@@ -274,6 +274,13 @@ func (s *SnapshotFlowExecution) cloneTablesWithSlot(
 	if err != nil {
 		return fmt.Errorf("failed to setup replication: %w", err)
 	}
+	defer func() {
+		dCtx, cancel := workflow.NewDisconnectedContext(ctx)
```
serprex, line 278: why is there ctx & sessionCtx?

heavycrystal: good catch

PR #2262, opened 2024-11-17
refactor: reduce default batch and partition sizes for improved stability

iskakaushik:
Reduces default batch sizes from 1,000,000 to 250,000 rows across CDC and snapshot operations:

- Lower CDC batch size for incremental syncs
- Reduce snapshot partition size for initial loads
- Update UI defaults and tooltips to reflect new values

This change helps prevent memory pressure and timeout issues when processing large datasets.


serprex:
Seems like default could be dependent on instance size

serprex (approved):

PR #2269, opened 2024-11-19
chore: replace debezium in ci and test on multiple pg versions

iamKunalGupta:


iamKunalGupta:
once approved, I will update the merge rules to fix the status check requirement

serprex (approved):

.github/workflows/flow.yml
```diff
@@ -11,11 +11,12 @@ jobs:
     strategy:
       matrix:
         runner: [ubicloud-standard-16-ubuntu-2204-arm]
+        postgres-version: [15, 16, 17]
     runs-on: ${{ matrix.runner }}
     timeout-minutes: 30
     services:
       catalog:
-        image: imresamu/postgis:15-3.4-alpine
+        image: imresamu/postgis:${{ matrix.postgres-version }}-3.4-alpine
```
heavycrystal, line 19: I see PostGIS 3.5

iamKunalGupta: I did not update the image version as I was not sure

iamKunalGupta: Was waiting for tests to pass firt, will update this as well

.github/workflows/ci.yml
```diff
@@ -5,18 +5,19 @@ on:
...
     branches: [main, release/*]
-    paths: [nexus/**, protos/**]
+    paths: [nexus/**, protos/**, .github/workflows/ci.yml]
 
 jobs:
   build:
     strategy:
       matrix:
         runner: [ubicloud-standard-2-ubuntu-2204-arm]
+        postgres-version: [14, 15, 16, 17]
```
heavycrystal, line 15: can add PG 13 here as well; 12 will be EOL soon

iamKunalGupta: we have customers from pg 13?

.github/workflows/flow.yml
```diff
@@ -11,11 +11,12 @@ jobs:
     strategy:
       matrix:
         runner: [ubicloud-standard-16-ubuntu-2204-arm]
+        postgres-version: [15, 16, 17]
```
heavycrystal, line 14: we need only run a subset of tests ideally; PG-PG and some tests to ensure initial load/resync works; fine for now
13 and 14 should be here but they're pre-MERGE so will need work

.github/workflows/ci.yml
heavycrystal: we allow PG13 peers to be created

.github/workflows/flow.yml
iamKunalGupta: can you create an issue for now @heavycrystal ?

PR #2274, opened 2024-11-20
Introduce Completed flow status

Amogh-Bharadwaj:
Introduces a new flow state - `Completed` 
Currently this is the state for an initial load only mirror which has finished initial load

Functionally tested

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -485,6 +485,8 @@ func CDCFlowWorkflow(
 
 		// if initial_copy_only is opted for, we end the flow here.
 		if cfg.InitialSnapshotOnly {
+			logger.Info("initial snapshot only, ending flow")
+			state.CurrentFlowStatus = protos.FlowStatus_STATUS_COMPLETED
 			return state, nil
 		}
 	}
--- a/flow/e2e/test_utils.go
+++ b/flow/e2e/test_utils.go
@@ -221,7 +221,7 @@ func SetupCDCFlowStatusQuery(t *testing.T, env WorkflowRun, config *protos.FlowC
 			var status protos.FlowStatus
 			if err := response.Get(&status); err != nil {
 				t.Fatal(err)
-			} else if status == protos.FlowStatus_STATUS_RUNNING {
+			} else if status == protos.FlowStatus_STATUS_RUNNING || status == protos.FlowStatus_STATUS_COMPLETED {
 				return
 			} else if counter > 30 {
 				env.Cancel()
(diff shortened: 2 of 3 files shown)


Amogh-Bharadwaj:
> we already have a Terminated state that is mostly unused (currently an input to `FlowStateChange` and it drops the flow), can reuse that?

Good question, I would say no because:

1. These statuses are important for user interfaces, and `Completed` and `Terminated` are semantically different: terminated has negative semantics vs completed which is positive. Showing finished initial load mirrors as terminated is misleading to end users and gives the impression that no action be done on the mirror, but in fact:

2. Finished initial load only mirrors can be resynced or dropped - which involves sending the status as terminated so I believe drop mirror as part of the resync would fail, or even if not, it seems more correct to say "drop a completed mirror" versus "drop a dropped mirror"

cc @serprex @iskakaushik

heavycrystal:
> > we already have a Terminated state that is mostly unused (currently an input to `FlowStateChange` and it drops the flow), can reuse that?
> 
> Good question, I would say no because:
> 
>     1. These statuses are important for user interfaces, and `Completed` and `Terminated` are semantically different: terminated has negative semantics vs completed which is positive. Showing finished initial load mirrors as terminated is misleading to end users and gives the impression that no action be done on the mirror, but in fact:
> 
>     2. Finished initial load only mirrors can be resynced or dropped - which involves sending the status as terminated so I believe drop mirror as part of the resync would fail, or even if not, it seems more correct to say "drop a completed mirror" versus "drop a dropped mirror"
> 
> 
> cc @serprex @iskakaushik

1. status can be remapped by UI to show completed; don't think that would be a concern
2. This is probably a bigger question; but allowing an initial load only mirror to be resynced is interesting. Anyway, there's some confusion with statuses being used both as input to a function and as a display so I think we can land this new status

heavycrystal (approved):

flow/workflows/cdc_flow.go
```diff
@@ -485,6 +485,8 @@ func CDCFlowWorkflow(
 
 		// if initial_copy_only is opted for, we end the flow here.
 		if cfg.InitialSnapshotOnly {
+			logger.Info("initial snapshot only, ending flow")
+			state.CurrentFlowStatus = protos.FlowStatus_STATUS_COMPLETED
```
iamKunalGupta, line 489: @Amogh-Bharadwaj does the status ever go to running?

Amogh-Bharadwaj: Nope

iamKunalGupta: or does it go from snapshot to completed directly?

heavycrystal: it wouldn't, the workflow exits for good here I believe

Amogh-Bharadwaj: https://github.com/PeerDB-io/peerdb/pull/2275
just to be safe

PR #2276, opened 2024-11-21
otel: fetched bytes

serprex:
bit of code cleanup along with adding an otel counter of postgres cdc payload sizes


flow/connectors/postgres/cdc.go
```diff
@@ -463,6 +475,10 @@ func PullCdcRecords[Items model.Items](
 			continue
 		}
 
+		if fetchedBytesCounter != nil {
+			fetchedBytesCounter.Add(ctx, int64(len(msg.Data)))
```
serprex, line 479: Should add a bunch of attributes here

flow/connectors/postgres/cdc.go
```diff
@@ -463,6 +478,13 @@ func PullCdcRecords[Items model.Items](
 			continue
 		}
 
+		if fetchedBytesCounter != nil {
+			fetchedBytesCounter.Add(ctx, int64(len(msg.Data)), metric.WithAttributeSet(attribute.NewSet(
+				attribute.String(otel_metrics.FlowNameKey, req.FlowJobName),
+				attribute.String(otel_metrics.DeploymentUidKey, peerdbenv.PeerDBDeploymentUID()),
```
iamKunalGupta, line 484: @serprex I was thinking maybe we should move the deploymentUID key when we create the resource in otel_manager.go.

flow/connectors/postgres/cdc.go
```diff
@@ -361,6 +367,16 @@ func PullCdcRecords[Items model.Items](
...
 	}
 
+	var fetchedBytesCounter metric.Int64Counter
+	if p.otelManager != nil {
+		var err error
+		fetchedBytesCounter, err = p.otelManager.GetOrInitInt64Counter(otel_metrics.BuildMetricName(otel_metrics.FetchedBytesCounterName),
+			metric.WithUnit("By"), metric.WithDescription("Bytes received of CopyData over replication slot"))
+		if err != nil {
+			logger.Error("Could not get FetchedBytesCounter", slog.Any("error", err))
+		}
```
iamKunalGupta, line 377: we should also probably sent out an alert on this as it will be an important metric

iamKunalGupta (approved):

flow/connectors/postgres/cdc.go
serprex: will create followup PR

flow/connectors/postgres/cdc.go
serprex: made it return instead of ignore

PR #2280, opened 2024-11-21
Revamp settings page

Amogh-Bharadwaj:
This PR opts for a card grid display for the settings page rather than a table - as the content was being cramped in to fit in that tabular layout

Screenshot 2024-11-21 at 6 49 06 PM


serprex (approved):
how does this look when you reduce size of window?

Amogh-Bharadwaj:
> how does this look when you reduce size of window?

Good catch. Edited the grid setting to make it adapt to screen resize

Screenshot 2024-11-21 at 8 56 10 PM

iskakaushik (approved):

PR #2283, opened 2024-11-24
Fix slot guage

Amogh-Bharadwaj:
There seems to be an issue where RecordSlotSize panics in this line with nil pointer dereference
As far as I can see the only thing that can be nil is `slotMetricGauges.SlotLagGauge`


iskakaushik (approved):

flow/connectors/postgres/postgres.go
```diff
@@ -1215,6 +1215,10 @@ func (c *PostgresConnector) HandleSlotInfo(
 	logger.Info(fmt.Sprintf("Checking %s lag for %s", alertKeys.SlotName, alertKeys.PeerName),
 		slog.Float64("LagInMB", float64(slotInfo[0].LagInMb)))
 	alerter.AlertIfSlotLag(ctx, alertKeys, slotInfo[0])
+	if slotMetricGauges.SlotLagGauge == nil {
+		logger.Warn("warning: slotMetricGauges.SlotLagGauge is nil")
+		return nil
```
serprex, line 1220: we still want to append to monitoring tables, should just skip the gauge stuff

iskakaushik:
this is ok to land for now but @serprex as you rework otel lets fix the guage creation on startup or something

PR #2284, opened 2024-11-24
Wrap get current LSN in txn as pgx will retry

iskakaushik:
Per https://github.com/jackc/pgx/issues/74#issuecomment-139832623 moving reads into a transaction will initiate retries on network-level errors.

### DIFF
--- a/flow/connectors/postgres/client.go
+++ b/flow/connectors/postgres/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"strings"
+	"time"
 
 	"github.com/jackc/pgerrcode"
 	"github.com/jackc/pglogrepl"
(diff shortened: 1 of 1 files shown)


serprex (approved):

flow/connectors/postgres/client.go
```diff
@@ -643,35 +644,73 @@ func (c *PostgresConnector) getTableNametoUnchangedCols(
...
+	const maxRetries = 3
+	const retryDelay = 5 * time.Second
+
+	for range maxRetries {
+		if err := c.executeReadOnlyTxn(ctx, fn); err != nil {
+			lastErr = err
+			if ctx.Err() != nil {
+				return ctx.Err()
+			}
+			time.Sleep(retryDelay)
```
serprex, line 681: this will give an extra sleep on last try

flow/connectors/postgres/client.go
```diff
@@ -643,35 +644,73 @@ func (c *PostgresConnector) getTableNametoUnchangedCols(
...
-	query := `
-		SELECT CASE
-			WHEN pg_is_in_recovery() THEN pg_last_wal_receive_lsn()
-			ELSE pg_current_wal_lsn()
-		END`
-
-	var lsnString string
-	if err := tx.QueryRow(ctx, query).Scan(&lsnString); err != nil {
-		return 0, fmt.Errorf("query failed: %w", err)
+	if err := fn(tx); err != nil {
```
serprex, line 659: I don't think this accomplishes what you want still, you need to reconnect on net errors

PR #2292, opened 2024-11-26
DropFlow: remove context deadline in shutdown flow

Amogh-Bharadwaj:
Not sure why we are setting a context deadline of 5 minutes for drop flow because after 5 minutes we are anyways performing `handleCancelWorkflow` in the event of drop flow taking a long time (to drop the slot etc)

Currently the context is getting cancelled after 5 minutes, taking us to the first case:
```golang
	case err := <-errChan:
		if err != nil {
			slog.Error(fmt.Sprintf("unable to cancel PeerFlow workflow: %s. Attempting to terminate.", err.Error()))
			terminationReason := fmt.Sprintf("workflow %s did not cancel in time.", workflowID)
			if err := h.temporalClient.TerminateWorkflow(ctx, workflowID, runID, terminationReason); err != nil {
				return fmt.Errorf("unable to terminate PeerFlow workflow: %w", err)
			}
		}
```
and erroring out

### DIFF
--- a/flow/cmd/handler.go
+++ b/flow/cmd/handler.go
@@ -298,12 +298,9 @@ func (h *FlowRequestHandler) shutdownFlow(
 		return fmt.Errorf("unable to start DropFlow workflow: %w", err)
 	}
 
-	cancelCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
-	defer cancel()
-
 	errChan := make(chan error, 1)
 	go func() {
-		errChan <- dropFlowHandle.Get(cancelCtx, nil)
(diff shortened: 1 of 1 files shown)


serprex (approved):

heavycrystal:
nit: the workflow itself can run indefinitely, it's only the `.Get` call that gets `cancelCtx` so it waits 5 minutes for the result basically

I agree the select branches for DropFlow seem redundant (the cancel branch that we probably wanted will almost never be taken); but the core decision to take here is whether we want DropFlow to run indefinitely or not.

serprex:
Less of an issue now that drop api is async

PR #2297, opened 2024-11-26
Add tags for Flows along with routes to set and get

iskakaushik:
- add tags to telemetry to provide additional visibility
- later these can be added to otel metrics etc

### DIFF
--- a/flow/cmd/tags_handler.go
+++ b/flow/cmd/tags_handler.go
@@ -37,7 +37,7 @@ func (h *FlowRequestHandler) CreateOrReplaceFlowTags(
 		return nil, fmt.Errorf("flow %s does not exist", flowName)
 	}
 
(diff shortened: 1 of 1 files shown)


flow/cmd/tags_handler.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if !exists {
+		slog.Error("flow does not exist", slog.String("flow_name", flowName))
+		return nil, fmt.Errorf("flow %s does not exist", flowName)
+	}
+
+	tags := make(map[string]string)
```
serprex, line 40:
```suggestion
	tags := make(map[string]string,len(in.Tags))
```

serprex (approved):

PR #2305, opened 2024-11-29
ClickHouse: Put numeric-string mapping behind feature flag

Amogh-Bharadwaj:
### Why
Unbounded numerics in Postgres databases statistically are unlikely to have values with dimensions larger than 76,38.
Currently by default unbounded `numeric` in postgres is mapped to String in ClickHouse - an unideal mapping from data nativity point of view and only useful for the scarce cases mentioned above

### What
This PR puts this mapping choice behind a dynamic setting flag whose default is false - thereby restoring the default mapping to numeric to Decimal in all cases

Adapted the E2E test for this


iskakaushik (approved):

serprex (approved):

flow/connectors/clickhouse/normalize.go
```diff
@@ -81,14 +81,26 @@ func getColName(overrides map[string]string, name string) string {
...
 	rawPrecision, _ := datatypes.ParseNumericTypmod(column.TypeModifier)
-	if rawPrecision > datatypes.PeerDBClickHouseMaxPrecision {
-		return "String"
-	} else {
+	numericAsStringEnabled, err := peerdbenv.PeerDBEnableClickHouseNumericAsString(ctx, env)
+	if err != nil {
+		return "", err
+	}
+
+	if column.TypeModifier == -1 {
```
serprex, line 91:
```suggestion
	if column.TypeModifier == -1 {
		numericAsStringEnabled, err := peerdbenv.PeerDBEnableClickHouseNumericAsString(ctx, env)
		if err != nil {
			return "", err
		}
```

flow/connectors/clickhouse/normalize.go
```diff
@@ -81,14 +81,26 @@ func getColName(overrides map[string]string, name string) string {
 	return name
 }
 
-func getClickhouseTypeForNumericColumn(column *protos.FieldDescription) string {
+func getClickhouseTypeForNumericColumn(ctx context.Context, column *protos.FieldDescription, env map[string]string) (string, error) {
 	rawPrecision, _ := datatypes.ParseNumericTypmod(column.TypeModifier)
```
serprex, line 85: this line can come after the -1 check

PR #2308, opened 2024-11-30
feat(clickhouse): add default precision/scale for numeric columns

iskakaushik:
When TypeModifier is -1 (unset), use default precision and scale values from ClickHouseNumericCompatibility instead of raw values. This ensures numeric columns have appropriate precision/scale settings even when not explicitly specified.


Amogh-Bharadwaj:
I've pushed a change to https://github.com/PeerDB-io/peerdb/pull/2305 which I think takes care of this
Could you take a look ? @iskakaushik

serprex:
replaced by #2305

PR #2313, opened 2024-12-03
[clickhouse] optional enforcement of SMT tables

heavycrystal:


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -266,6 +266,21 @@ func Connect(ctx context.Context, env map[string]string, config *protos.Clickhou
 		return nil, fmt.Errorf("failed to ping to ClickHouse peer: %w", err)
 	}
 
+	if peerdbenv.PeerDBAllowedTargets() == strings.ToLower(protos.DBType_CLICKHOUSE.String()) {
```
iamKunalGupta, line 269: maybe we should put the condition in a function for easier usage

flow/peerdbenv/config.go
```diff
@@ -133,6 +134,10 @@ func PeerDBAllowedTargets() string {
 	return GetEnvString("PEERDB_ALLOWED_TARGETS", "")
 }
 
+func PeerDBOnlyClickHouseAllowed() bool {
+	return PeerDBAllowedTargets() == strings.ToLower(protos.DBType_CLICKHOUSE.String())
```
serprex, line 138:
```suggestion
	return strings.EqualFold(PeerDBAllowedTargets(), protos.DBType_CLICKHOUSE.String())
```

serprex (approved):

PR #2314, opened 2024-12-03
BQ: fix create/replace logic for resync

Amogh-Bharadwaj:
In resync, in SetupNormalize we need to drop the resync tables if it is in resync mode
This PR fixes a bug in this where we check column existence and return before we get the chance to delete the resync table


heavycrystal (approved):

flow/connectors/bigquery/bigquery.go
```diff
@@ -636,11 +636,25 @@ func (c *BigQueryConnector) SetupNormalizedTable(
...
+				slog.String("table", tableIdentifier),
+				slog.Any("existingMetadata", existingMetadata))
+			return true, nil
+		}
+	} else {
+		if !strings.Contains(err.Error(), "notFound") {
+			return false, fmt.Errorf("error while checking metadata for BigQuery table existence %s: %w",
+				tableIdentifier, err)
+		}
 	}
```
serprex, line 658:
```suggestion
	} else if !strings.Contains(err.Error(), "notFound") {
		return false, fmt.Errorf("error while checking metadata for BigQuery table existence %s: %w",
			tableIdentifier, err)
	}
```

serprex (approved):

PR #2316, opened 2024-12-03
nexus: support uuid arrays

### LINKED ISSUE #2315: ARRAY<UUID> not handled properly
[array.rs](https://github.com/PeerDB-io/peerdb/blob/main/nexus/value/src/array.rs) doesn't handle arrays of UUIDs

the [lib.rs](https://github.com/PeerDB-io/peerdb/blob/main/nexus/value/src/lib.rs) does handle UUIDs

Can we please extend array processing to include `ARRAY<UUID>`

serprex:
fixes #2315


lukejudd-lux:
nice! @serprex 

https://docs.peerdb.io/datatypes/datatype-matrix#primitive-data-types

will need updating as well

PR #2323, opened 2024-12-04
[cdc] better manual handling of partitioned tables

heavycrystal:
1. log when we're remapping child table OID to parent table OID
2. partition fetching query also allows non-partitioned tables, extensions use this
3. in case we see an unknown table OID after checking `childToParentRelIDMapping`, check again to ensure the table wasn't created in the middle of CDC
4. remove log since it could log PII

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -682,7 +682,7 @@ func processMessage[Items model.Items](
 		p.commitLock = nil
 	case *pglogrepl.RelationMessage:
 		// treat all relation messages as corresponding to parent if partitioned.
-		msg.RelationID, err = p.checkIfUnknownTableInherits(msg.RelationID)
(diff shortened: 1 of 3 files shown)


flow/connectors/postgres/cdc.go
```diff
@@ -964,8 +970,43 @@ func processRelationMessage[Items model.Items](
...
+
+// since we generate the child to parent mapping at the beginning of the CDC stream,
+// some tables could be created after the CDC stream starts,
+// and we need to check if they inherit from a known table
+func (p *PostgresCDCSource) checkIfUnknownTableInherits(relID uint32) (uint32, error) {
+	relID = p.getParentRelIDIfPartitioned(relID)
+
+	if _, ok := p.srcTableIDNameMapping[relID]; !ok {
+		var parentRelID uint32
+		err := p.conn.QueryRow(context.Background(),
```
serprex, line 994: should take ctx parameter

serprex (approved):

PR #2329, opened 2024-12-06
move normalize into sync activity

serprex:
can still run in parallel,
in fact now sync batches won't sync ahead if normalize is slow


flow/workflows/cdc_flow.go
```diff
@@ -492,10 +490,8 @@ func CDCFlowWorkflow(
 	}
 
 	syncFlowID := GetChildWorkflowID("sync-flow", cfg.FlowJobName, originalRunID)
-	normalizeFlowID := GetChildWorkflowID("normalize-flow", cfg.FlowJobName, originalRunID)
 
 	var restart, finished bool
-	syncCount := 0
```
serprex, line 498: cdc_flow no longer gets notified when sync flows end, need to somehow pass this down to sync flow

flow/activities/flowable.go
```diff
@@ -297,16 +303,39 @@ func (a *FlowableActivity) MaintainPull(
...
+				req.Done <- res
+			}
+			if req.Done != nil {
+				close(req.Done)
+			}
+		}
+		close(normDone)
+	}()
+	defer func() { <-normDone }()
+	defer close(normalize) // TODO race, this will cause sync to panic if it tries to send to normalize after maintainpull ends
```
serprex, line 337: TODO

serprex:
Looking over failures, I think this needs to also move schema changes into the activity

flow/activities/flowable.go
```diff
@@ -297,16 +303,41 @@ func (a *FlowableActivity) MaintainPull(
...
 
+	normDone := make(chan struct{})
+	go func() {
+		for req := range normalize {
+			res, err := a.StartNormalize(ctx, &protos.StartNormalizeInput{
+				FlowConnectionConfigs: config,
+				SyncBatchID:           req.BatchID,
+			})
+			if err != nil {
+				a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
```
serprex, line 326: currently we wait until next sync batch to retry, could instead do something like
```
select {
normalize <- req:
default:
}
```
to retry unless another normalize is already queued up

serprex: I'm an idiot, just drain channel & use goto, no need for brittle code risking self deadlock

flow/activities/flowable.go
```diff
@@ -38,9 +38,16 @@ type CheckConnectionResult struct {
...
+type NormalizeBatchRequest struct {
+	Done    chan struct{}
+	BatchID int64
+}
+
 type CdcCacheEntry struct {
-	connector connectors.CDCPullConnectorCore
-	done      chan struct{}
+	connector     connectors.CDCPullConnectorCore
+	done          chan struct{}
```
iskakaushik, line 48: call this `syncDone`

flow/activities/flowable.go
```diff
@@ -297,16 +304,56 @@ func (a *FlowableActivity) MaintainPull(
...
+		done:          done,
+		normalize:     normalize,
+		normalizeDone: normalizeDone,
 	}
 	a.CdcCacheRw.Unlock()
 
 	ticker := time.NewTicker(15 * time.Second)
 	defer ticker.Stop()
 
+	go func() {
```
iskakaushik, line 321: move all inner stuff to a new function.

flow/activities/flowable.go
```diff
@@ -297,16 +304,56 @@ func (a *FlowableActivity) MaintainPull(
...
+				} else if req.Done != nil {
+					close(req.Done)
+				}
+			case <-done:
+				break loop
+			case <-ctx.Done():
+				break loop
+			}
+		}
+		close(normalizeDone)
```
iskakaushik, line 354: defer and remove `loop` label in favor of `return`

flow/activities/flowable.go
```diff
@@ -297,16 +304,56 @@ func (a *FlowableActivity) MaintainPull(
...
+					break loop
+				}
+			retry:
+				if err := a.StartNormalize(ctx, config, req.BatchID); err != nil {
+					a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
+					for {
+						// update req to latest normalize request & retry
+						select {
+						case req = <-normalize:
+						case <-done:
```
iskakaushik, line 336: log why we return. Wherever we return in this function.

flow/activities/flowable.go
```diff
@@ -297,16 +304,56 @@ func (a *FlowableActivity) MaintainPull(
...
+							break loop
+						case <-ctx.Done():
+							break loop
+						default:
+							time.Sleep(time.Second)
```
iskakaushik, line 341: or change to 30 seconds for retry.

flow/activities/flowable.go
```diff
@@ -297,16 +304,56 @@ func (a *FlowableActivity) MaintainPull(
...
 
+	go func() {
+	loop:
+		for {
+			select {
+			case req, ok := <-normalize:
+				if !ok {
+					break loop
+				}
+			retry:
```
iskakaushik, line 329: just put in an inner loop.

flow/activities/flowable.go
serprex: at 30s I want to select on ctx.Done/syncDone to avoid delays reacting to pause, but leaving it since having this dance a third time would be a bit of a mess. Maybe something like shared.Sleep(context.Context) could work, but that won't watch over syncDone

iskakaushik (commented):
pending:
1. buffered channel and test (default unbuffered) -- make it mirror conf, env.
2. test schema changes

iskakaushik (approved):

PR #2333, opened 2024-12-07
e2e test CustomSyncs

serprex:


serprex:
goal is to make #2329 fail ci since it currently rips out this feature

Amogh-Bharadwaj (approved):

PR #2337, opened 2024-12-09
ch: implement GetTableSchemaConnector

serprex:
for generic schema changes e2e, alternative to #2114


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -542,3 +543,63 @@ func (c *ClickHouseConnector) GetVersion(ctx context.Context) (string, error) {
...
+	q, err := c.database.Query(ctx, fmt.Sprintf("select * from %s limit 0", tableName))
+	if err != nil {
+		return nil, err
+	}
+
+	columns := q.ColumnTypes()
+	colFields := make([]*protos.FieldDescription, 0, len(columns))
+	for _, column := range columns {
+		var qkind qvalue.QValueKind
+		switch column.DatabaseTypeName() {
```
heavycrystal, line 558: would prefer to lift this into a function so it can be reused here and in `.GetRows`

heavycrystal (approved):

PR #2353, opened 2024-12-12
[cdc] treat all pkms as reply_requested

heavycrystal:
Looking at walsender code, `PrimaryKeepaliveMessages` are _mostly_ only sent with `reply_requested` when `wal_sender_timeout` is non-zero. We recommend `wal_sender_timeout` be set to 0. We currently only send a `StandbyStatusMessage` when we see a pkm that has `reply_requested` so we almost never send them.

Treat all pkms as `reply_requested` for hopefully more aggressive slot flushes.


serprex (approved):

iamKunalGupta:
@heavycrystal can you do a large cdc update test on this once?

PR #2355, opened 2024-12-13
support tls_cert/tls_key in nexus

serprex:
Need to contribute upstream to pgwire:
1. tokio-rustls should be reexported since we must use same version of TlsAcceptor as them
2. feature should not add parameter to public facing api method, as adding features should never break code


serprex:
https://github.com/sunng87/pgwire/pull/228

iamKunalGupta (approved):

PR #2356, opened 2024-12-13
add throttling when there is a syncErr

iskakaushik:


serprex (approved):

flow/workflows/sync_flow.go
```diff
@@ -162,6 +162,15 @@ func SyncFlowWorkflow(
 		}
 
 		restart := syncErr || workflow.GetInfo(ctx).GetContinueAsNewSuggested()
+
+		if syncErr {
+			logger.Info("sync flow error, sleeping for 30 seconds...")
```
serprex, line 167: no need to log, already logged error when syncErr was set

PR #2369, opened 2024-12-17
[cdc] always initialize childToParentRelIDMapping

heavycrystal:
https://github.com/PeerDB-io/peerdb/pull/2323#discussion_r1888186520

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -404,6 +404,8 @@ func pullCore[Items model.Items](
 		if err != nil {
 			return fmt.Errorf("error getting child to parent relid map: %w", err)
 		}
+	} else {
+		childToParentRelIDMap = make(map[uint32]uint32)
 	}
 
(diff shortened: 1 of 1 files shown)


flow/connectors/postgres/postgres.go
```diff
@@ -404,6 +404,8 @@ func pullCore[Items model.Items](
 		if err != nil {
 			return fmt.Errorf("error getting child to parent relid map: %w", err)
 		}
+	} else {
+		childToParentRelIDMap = make(map[uint32]uint32)
```
serprex, line 408: why is empty map okay here? elsewhere we write to this hash, could those writes not happen when this was previously nil?

heavycrystal: yes, writes to a nil map panic

serprex: right, but you could do
```go
if map != nil {
  map[k] = v
}
```
my question is if that would be an equivalent fix

serprex: I prefer checking for nil to avoid code having half filled maps in some states, nil is a very simple edge case in comparison

PR #2373, opened 2024-12-18
[validate] run IDENTIFY_SYSTEM on replication conn

heavycrystal:


flow/connectors/postgres/validate.go
```diff
@@ -139,6 +139,11 @@ func (c *PostgresConnector) CheckReplicationConnectivity(ctx context.Context) er
 	if err != nil {
 		return fmt.Errorf("failed to create replication connection: %v", err)
 	}
+	defer conn.Close(ctx)
+
+	if _, err := conn.Exec(ctx, "IDENTIFY_SYSTEM"); err != nil {
+		return fmt.Errorf("failed to execute IDENTIFY_SYSTEM on replication connection: %v", err)
```
serprex, line 145:
```suggestion
		return fmt.Errorf("failed to execute IDENTIFY_SYSTEM on replication connection: %w", err)
```

flow/connectors/postgres/validate.go
```diff
@@ -139,6 +139,11 @@ func (c *PostgresConnector) CheckReplicationConnectivity(ctx context.Context) er
 	if err != nil {
 		return fmt.Errorf("failed to create replication connection: %v", err)
 	}
+	defer conn.Close(ctx)
```
serprex, line 142: this will double close, move into error branch instead of defer

heavycrystal: pgx is fine with calling `.Close()` repeatedly

```
Close closes a connection. It is safe to call Close on an already closed connection.
```

serprex (approved):

Issue #2341, opened 2024-12-09
Emit more metrics for broader visibility


iamKunalGupta:
1. `flow_status` metrics: for every mirror emit a value of `1` with `flowName` and `status` metric labels, include `unknown` status as well
2. `maintenance_running` emit `1` with `maintenance_type` = `start`/`end`
3. `flow_source_table_count` emit table count for a given `flowName`
4. Add `peerSourceType`, `peerSourceHost`, and `peerDestinationType`,`peerDestinationHost` to all metrics emitted.
5. Whenever we add alert to catalog, emit the alert count as a metric
6. Add rows synced metric

iskakaushik:
(
Orgid
Serviceid
Pipeid
Total bytes moved
Minute (TIMESTAMP)
Initial Load Bytes Moved
CDC Bytes Moved
Resync Bytes Moved
);

iamKunalGupta:
> ( Orgid Serviceid Pipeid Total bytes moved Minute (TIMESTAMP) Initial Load Bytes Moved CDC Bytes Moved Resync Bytes Moved );

ServiceID and PipeID are auto-added as long as flowName and deploymentUID is set, services can move between orgs, so I think we can do without org id.

PR #2378, opened 2024-12-19
gauges for current batch id & normalized batch id

serprex:
important once #2371 prevents seeing batch history in temporal


flow/activities/flowable_core.go
```diff
@@ -349,6 +349,16 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	activity.RecordHeartbeat(ctx, pushedRecordsWithCount)
 	a.Alerter.LogFlowInfo(ctx, flowName, pushedRecordsWithCount)
 
+	if a.OtelManager != nil {
+		currentBatchID, err := a.OtelManager.GetOrInitInt64Gauge(
```
serprex, line 353: ideally we'd have something like I have for typed temoral slots & then this could just be
```
otel_metrics.CurrentBatchId.Record(ctx, a.OtelManager, res.CurrentSyncBatchID)
```

heavycrystal:
need to validate against #2382

flow/otel_metrics/otel_manager.go
```diff
@@ -18,15 +18,19 @@ import (
 )
 
 const (
-	SlotLagGaugeName                    string = "cdc_slot_lag"
-	OpenConnectionsGaugeName            string = "open_connections"
-	OpenReplicationConnectionsGaugeName string = "open_replication_connections"
-	IntervalSinceLastNormalizeGaugeName string = "interval_since_last_normalize"
-	FetchedBytesCounterName             string = "fetched_bytes"
+	SlotLagGaugeName                    = "cdc_slot_lag"
+	CurrentBatchIdGaugeName             = "current_batch_id"
```
iamKunalGupta, line 22: Maybe we should call it sync_batch_id

Also wondering if we should emit this twice, one at start and one at end, can attach different labels for both the times.

iamKunalGupta (commented):
@serprex if you have bandwidth can you add a couple of metrics more from https://github.com/PeerDB-io/peerdb/issues/2341?

Also as @heavycrystal mentioned, would be great if you could check https://github.com/PeerDB-io/peerdb/issues/2382 while you're at it?

serprex:
> @serprex if you have bandwidth can you add a couple of metrics more from #2341?
> 
> Also as @heavycrystal mentioned, would be great if you could check #2382 while you're at it?

those would be separate PRs

heavycrystal (approved):

PR #2394, opened 2024-12-24
golangci-lint 1.62

serprex:
Enable new linters: iface & recvcheck


flow/connectors/sql/query_executor.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/PeerDB-io/peer-flow/model/qvalue"
 )
 
+//nolint:iface
 type SQLQueryExecutor interface {
```
serprex, line 23: Long term we should just get rid of this, only used by e2e, only implemented with snowflake/sqlserver

heavycrystal (approved):

PR #2396, opened 2024-12-27
LSN Fetch for stats: Make it non-fatal for null LSN

Amogh-Bharadwaj:
Follow-up to https://github.com/PeerDB-io/peerdb/pull/2372

Since this function is just for recording in a stat table, the error can just be logged.


flow/connectors/postgres/postgres.go
```diff
@@ -433,12 +433,8 @@ func pullCore[Items model.Items](
 	latestLSN, err := c.getCurrentLSN(ctx)
```
serprex, line 433: should comment indicating not returning error is intentional

heavycrystal (approved):

PR #2397, opened 2024-12-27
Stop sending non-actionable errors to incident.io

iskakaushik:


Amogh-Bharadwaj (approved):

serprex:
PR description should describe why this is preferable to handling tags on incident.io end

PR #2398, opened 2024-12-27
QRecordSchema: close schemaLatch in Close

serprex:
goroutines were leaking waiting on schemaLatch,
if provider hit error then schemaLatch could be left indefinitely open

### DIFF
--- a/flow/connectors/bigquery/qrep.go
+++ b/flow/connectors/bigquery/qrep.go
@@ -23,9 +23,9 @@ func (c *BigQueryConnector) SyncQRepRecords(
 ) (int, error) {
 	// Ensure the destination table is available.
 	destTable := config.DestinationTableIdentifier
-	srcSchema := stream.Schema()
-	if srcSchema.Fields == nil {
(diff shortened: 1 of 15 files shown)


serprex:
Some options: have Schema method take context.Context, have Schema return error

iskakaushik:
> Some options: have Schema method take context.Context, have Schema return error

This option is definitely better.

serprex:
So I went this route because I think context will make the error a bit flaky (between context error which is probably tied with stream error)

This way makes `flow/pua/stream_adapter.go` work unchanged, ie stream transforms don't have an error path

If stream is closed without error without schema it's a strange scenario. Seems best in that case to `return 0, nil`, whereas with error return we'd get back nil & go ahead with no schema. Guess we could make a separate `errors.New("NoSchema")`

So this seemed like the simplest path. But could have `Schema()` still return error, which would just be `stream.Err()`, so that it isn't so easy to call Schema without error handling

heavycrystal (approved):

PR #2401, opened 2024-12-29
SyncFlow: top level heartbeat

serprex:
was seeing heartbeat timeouts

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -302,7 +302,7 @@ func (a *FlowableActivity) SyncFlow(
 	group, groupCtx := errgroup.WithContext(ctx)
 	group.Go(func() error {
 		// returning error signals sync to stop, normalize can recover connections without interrupting sync, so never return error
-		a.normalizeLoop(groupCtx, config, syncDone, normRequests, &normalizingBatchID, &normalizeWaiting)
+		a.normalizeLoop(groupCtx, logger, config, syncDone, normRequests, &normalizingBatchID, &normalizeWaiting)
 		return nil
 	})
 	group.Go(func() error {
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -250,28 +250,26 @@ func (a *FlowableActivity) SyncFlow(
 	ctx = context.WithValue(ctx, shared.FlowNameKey, config.FlowJobName)
 	logger := activity.GetLogger(ctx)
 
-	currentSyncFlowNum := atomic.Int32{}
-	totalRecordsSynced := atomic.Int64{}
-	normalizingBatchID := atomic.Int64{}
-	normalizeWaiting := atomic.Bool{}
-	syncingBatchID := atomic.Int64{}
-	syncWaiting := atomic.Bool{}
+	var currentSyncFlowNum atomic.Int32
+	var totalRecordsSynced atomic.Int64
+	var normalizingBatchID atomic.Int64
+	var normalizeWaiting atomic.Bool
+	var syncingBatchID atomic.Int64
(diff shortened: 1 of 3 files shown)


heavycrystal:
we lose more visibility into which phase of a batch we are in

serprex:
yes, tho with parallelism heartbeats were already stepping over each other

iskakaushik:
@serprex  - can we have this heartbeat contain a message that shows the consolidated status of syncs and normalizes, something like:

```
{
   start_sync_batch_id: ...,
   start_normalize_batch_id: ...,
   current_sync_batch_id: ...,
   current_sync_batch_start_time_utc: ...,
   current_normalize_batch_id: ...,
   current_normalize_start_time_utc: ...,
   current_time_utc: ...
}
```

serprex:
updated with heartbeat state I think will be useful. Temporal already has timestamps on heartbeat, no need to put timestamps in message

flow/activities/flowable_core.go
```diff
@@ -202,24 +200,25 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
...
 			if temporal.IsApplicationError(err) {
-				return 0, err
+				return nil, err
 			} else {
-				return 0, fmt.Errorf("failed in pull records when: %w", err)
+				return nil, fmt.Errorf("failed in pull records when: %w", err)
 			}
 		}
 		logger.Info("no records to push")
+		syncWaiting.Store(true)
```
heavycrystal, line 209: why would this be true? also unsure how we hit case in 208

flow/activities/flowable_core.go
```diff
@@ -668,9 +672,11 @@ func (a *FlowableActivity) normalizeLoop(
 				break
 			}
 		case <-syncDone:
+			normalizeWaiting.Store(false)
 			logger.Info("[normalize-loop] syncDone closed")
 			return
 		case <-ctx.Done():
+			normalizeWaiting.Store(false)
```
heavycrystal, line 679: can we defer this instead?

flow/activities/flowable_core.go
serprex: `Waiting` should generally be true as long as a sync/normalize isn't happening. Should probably initialize them as true. For normalize I settled on having that be true while channel is being waited on, with syncWaiting it's less clear, maybe a string saying what state it's in would be better, between setup, sync, schema deltas, & waiting on normalize

flow/activities/flowable_core.go
serprex: Yes

flow/activities/flowable_core.go
heavycrystal: sync being a string makes sense; or you could add an `Idle` state

flow/activities/flowable.go
```diff
@@ -249,40 +242,67 @@ func (a *FlowableActivity) CreateNormalizedTable(
...
-) (CdcState, context.Context, error) {
+	options *protos.SyncFlowOptions,
+) error {
 	ctx = context.WithValue(ctx, shared.FlowNameKey, config.FlowJobName)
-	srcConn, err := connectors.GetByNameAs[connectors.CDCPullConnector](ctx, config.Env, a.CatalogPool, config.SourceName)
+	logger := activity.GetLogger(ctx)
+
+	var currentSyncFlowNum atomic.Int32
+	var totalRecordsSynced atomic.Int64
+	var normalizingBatchID atomic.Int64
```
heavycrystal, line 255: nit: could probably store some of these in substruct of `FlowableActivity` and keep calls clean

heavycrystal (approved):

PR #2405, opened 2024-12-30
Account for columns on Postgres without SELECT permission

Amogh-Bharadwaj:
We currently require all columns to be SELECT-able in a table for it to be synced. However, if a user is excluding columns from a table, it is a valid expectation that SELECT permissions would not be required for those excluded columns.

This PR takes a crack at supporting this use-case.

It involves:
- Passing column exclusion information to `getTableSchema`
- Passing column exclusion information to `CheckSourceTables` in `validate_mirror.go`

It also adds a new E2E test and a utility function for it which grants only selected columns SELECT privileges


flow/connectors/postgres/postgres.go
```diff
@@ -812,9 +850,17 @@ func (c *PostgresConnector) getTableSchemaForTable(
 		return nil, err
 	}
 
+	selectedColumns, err := c.GetSelectedColumns(ctx, schemaTable, excludedColumns)
+	if err != nil {
+		return nil, err
+	}
+	selectedColumnsStr := strings.Join(selectedColumns, ",")
```
serprex, line 857: Need to escape identifiers. Can skip this all when `len(selectedColumns) == 0`

flow/connectors/postgres/postgres.go
```diff
@@ -771,11 +772,48 @@ func (c *PostgresConnector) GetTableSchema(
...
+		if err := rows.Scan(&columnName, &dataType, &isNullable); err != nil {
+			return nil, fmt.Errorf("error scanning column while getting selected columns: %w", err)
+		}
+
+		for _, excludedColumn := range excludedColumns {
+			if columnName == excludedColumn {
+				continue
+			}
+			columns = append(columns, QuoteIdentifier(columnName))
+		}
```
serprex, line 805: this will append columnName a bunch of times for each exclusion. Could use slices.Contains, but instead better if excludedColumns is `map[string]struct{}` for fast membership test

flow/shared/postgres.go
```diff
@@ -150,3 +150,11 @@ func IsSQLStateError(err error, sqlStates ...string) bool {
 	var pgErr *pgconn.PgError
 	return errors.As(err, &pgErr) && slices.Contains(sqlStates, pgErr.Code)
 }
+
+func ConstructExcludedColumnsMap(tableMappings []*protos.TableMapping) map[string][]string {
```
serprex, line 154:
```suggestion
func ConstructExcludedColumnsMap(tableMappings []*protos.TableMapping) map[string]map[string]struct{} {
```

flow/connectors/core.go
```diff
@@ -50,6 +50,7 @@ type GetTableSchemaConnector interface {
 		env map[string]string,
 		system protos.TypeSystem,
 		tableIdentifiers []string,
+		excludedColumnsMap map[string][]string,
```
heavycrystal, line 53: pass `protos.TableMapping` instead if possible

flow/connectors/postgres/postgres.go
```diff
@@ -771,11 +772,48 @@ func (c *PostgresConnector) GetTableSchema(
...
 }
 
+func (c *PostgresConnector) GetSelectedColumns(
+	ctx context.Context,
+	sourceTable *utils.SchemaTable,
+	excludedColumns []string,
+) ([]string, error) {
+	// Given the list of excluded columns,
+	// we need to get the list of columns that are not excluded
+	getColumnsSQL := "SELECT attname AS column_name, format_type(atttypid, atttypmod) AS data_type, attnotnull AS is_nullable FROM pg_attribute WHERE attrelid = '%s.%s'::regclass AND attnum > 0 AND NOT attisdropped"
```
heavycrystal, line 782: can't you just pass the excluded columns into the query and filter on `attname`? handling empty array of course

flow/connectors/postgres/postgres.go
```diff
@@ -771,11 +772,48 @@ func (c *PostgresConnector) GetTableSchema(
...
+
+	return columns, nil
+}
+
 func (c *PostgresConnector) getTableSchemaForTable(
 	ctx context.Context,
 	env map[string]string,
 	tableName string,
 	system protos.TypeSystem,
+	excludedColumns []string,
```
heavycrystal, line 816: meta: column exclusion is now applied by the columns missing from the schema entirely instead of them being present and filtered at several places
need to test this, I think you'll have schema changes detecting excluded columns and trying to add them

flow/e2e/congen.go
```diff
@@ -53,6 +53,31 @@ func cleanPostgres(conn *pgx.Conn, suffix string) error {
...
+	}
+
+	// 1. Revoke all permissions on the table
+	if _, err := conn.Exec(context.Background(),
+		fmt.Sprintf("REVOKE ALL ON TABLE %s.%s FROM public", schemaTable.Schema, schemaTable.Table)); err != nil {
+		return fmt.Errorf("failed to revoke permissions on table %s: %w", schemaTable.String(), err)
+	}
+
+	// 2. Grant SELECT permission on the columns
+	for _, col := range selectedColumns {
```
heavycrystal, line 71: nit: coalesce to single `GRANT`

Amogh-Bharadwaj:
Functionally tested

flow/connectors/postgres/postgres.go
Amogh-Bharadwaj: Functionally tested and it works.  Looks like we take care of excluded columns in schema changes [here](https://github.com/PeerDB-io/peerdb/blob/6e5ed879ec653299729cc8257f76270840fb11db/flow/connectors/postgres/cdc.go#L934-L935)

flow/connectors/core.go
Amogh-Bharadwaj: `tableIdentifiers` then becomes redundant so removing that will be another refactor. Can do it as a follow-up where we remove both in place of tablemapping

flow/connectors/postgres/validate.go
```diff
@@ -25,10 +26,26 @@ func (c *PostgresConnector) CheckSourceTables(ctx context.Context,
 		var row pgx.Row
 		tableArr = append(tableArr, fmt.Sprintf(`(%s::text,%s::text)`,
 			QuoteLiteral(parsedTable.Schema), QuoteLiteral(parsedTable.Table)))
+
+		selectedColumnsStr := "*"
+		if _, ok := excludedColumnsMap[parsedTable.Schema+"."+parsedTable.Table]; ok {
```
serprex, line 31: This is always going to be a pain, since we parsed `req.ConnectionConfigs.TableMappings` for `tableNames` while `excludedColumnsMap` is built on the raw strings from `req.ConnectionConfigs.TableMappings`. So you can't really use `parsedTable.String()` but that's what you'd want if you were going to properly handle cases like `"sch.e""ma"."ta.b""le"` which parses to `sch.e"ma.ta.b"le`

flow/shared/postgres.go
```diff
@@ -150,3 +150,11 @@ func IsSQLStateError(err error, sqlStates ...string) bool {
 	var pgErr *pgconn.PgError
 	return errors.As(err, &pgErr) && slices.Contains(sqlStates, pgErr.Code)
 }
+
+func ConstructExcludedColumnsList(tableMappings []*protos.TableMapping) map[string][]string {
+	excludedColumns := make(map[string][]string)
+	for _, tm := range tableMappings {
+		excludedColumns[tm.SourceTableIdentifier] = slices.Clone(tm.Exclude)
```
serprex, line 157:
```suggestion
		excludedColumns[tm.SourceTableIdentifier] = tm.Exclude
```
`Exclude` should never be mutated (I mean should as in people should not do that, code today does not do that, because doing that would be bad)

flow/e2e/congen.go
```diff
@@ -53,6 +54,37 @@ func cleanPostgres(conn *pgx.Conn, suffix string) error {
...
+	selectedColumns []string,
+) error {
+	schemaTable, err := utils.ParseSchemaTable(tableIdentifier)
+	if err != nil {
+		return fmt.Errorf("failed to parse table identifier %s: %w", tableIdentifier, err)
+	}
+
+	// 1. Revoke all permissions on the table
+	if _, err := conn.Exec(context.Background(),
+		fmt.Sprintf("REVOKE ALL ON TABLE %s.%s FROM postgres", schemaTable.Schema, schemaTable.Table)); err != nil {
```
serprex, line 68:
```suggestion
		fmt.Sprintf("REVOKE ALL ON TABLE %s FROM postgres", schemaTable)); err != nil {
```

flow/e2e/clickhouse/peer_flow_ch_test.go
```diff
@@ -776,3 +776,53 @@ func (s ClickHouseSuite) Test_Types_CH() {
...
 }
+
+func (s ClickHouseSuite) Test_Unprivileged_Postgres_Columns() {
+	// This test is for checking that the flow can handle a table with columns that the user does not have SELECT privileges on
+	srcTableName := "test_unprivileged_columns"
+	srcFullName := s.attachSchemaSuffix(srcTableName)
+	dstTableName := "test_unprivileged_columns_dst"
+
+	_, err := s.Conn().Exec(context.Background(), fmt.Sprintf(`
+		CREATE TABLE IF NOT EXISTS %s (
```
serprex, line 787: should include some oddly named columns (apostrophes etc), including in the set of excluded columns

flow/activities/flowable_core.go
```diff
@@ -83,17 +83,11 @@ func (a *FlowableActivity) applySchemaDeltas(
...
 	tableSchemaDeltasCount := len(schemaDeltas)
 	if tableSchemaDeltasCount > 0 {
-		modifiedSrcTables := make([]string, 0, tableSchemaDeltasCount)
-		for _, tableSchemaDelta := range schemaDeltas {
-			modifiedSrcTables = append(modifiedSrcTables, tableSchemaDelta.SrcTableName)
-		}
-
 		if err := a.SetupTableSchema(ctx, &protos.SetupTableSchemaBatchInput{
-			PeerName:         config.SourceName,
-			TableIdentifiers: modifiedSrcTables,
```
serprex, line 93: @heavycrystal afaict `TableIdentifiers` wasn't used by SetupTableSchema, making this code look a bit silly. Makes me worry I'm missing something here

serprex (approved):

PR #2181, opened 2024-10-24
[clickhouse] misc type fixes

heavycrystal:
1. bytes was inserted raw during initial load and base64 encoded in CDC, synchronize to base64
2. timestamp types were inserted with precision 3 in CDC as opposed to 6 in initial load, synchronized to 6
3. Time and TimeTZ are inserted as `DateTime(6)`
4. Fix handling of `ArrayDate` `ArrayTimestamp` and `ArrayTimestampTZ`

PR #2407, opened 2025-01-02
PEERDB_CLICKHOUSE_BINARY_FORMAT

serprex:
In #2181 we consolidated on transmitting bytea columns as base64 strings

ClickHouse supports binary data in strings. They need not be valid utf8

This adds PEERDB_CLICKHOUSE_BINARY_FORMAT with 3 formats: raw, hex, & base64
Default to base64 to avoid breaking existing setups

Also add caching logic for non-immediate dynamic settings. Helps here where we'd hit catalog for every bytea field processed

Included `hex` to demonstrate how other formats would fit in, but expect this feature to be removed once we move everyone to `raw`


flow/connectors/clickhouse/normalize.go
```diff
@@ -390,14 +390,47 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
+						if enablePrimaryUpdate {
+							projectionUpdate.WriteString(fmt.Sprintf(
+								"base64Decode(JSONExtract(_peerdb_match_data, '%s', '%s')) AS `%s`,",
+								colName,
+								clickHouseType,
+								dstColName,
+							))
+						}
+					case peerdbenv.BinaryFormatHex:
+						projection.WriteString(fmt.Sprintf("hex(base64Decode(JSONExtract(_peerdb_data, '%s', '%s'))) AS `%s`,", colName, clickHouseType, dstColName))
```
serprex, line 411: might be ideal to send hex to raw table rather in this case, but:
1. that would mean passing down ctx/env somehow to json serialization
2. things would get weird with raw tables being invalidated when changing PEERDB_BINARY_FORMAT
3. base64 is actually a pretty good binary encoding when raw binary isn't available (relatively dense while being computationally straightforward compared to base85, generally available decoding functions on destination)

flow/e2e/clickhouse/peer_flow_ch_test.go
```diff
@@ -666,6 +665,66 @@ func (s ClickHouseSuite) Test_Unbounded_Numeric_Without_FF() {
 	s.testNumericFF(false)
 }
 
+const rawBinaryFormatTestcase = "\x00\x010123\x7f\xff"
```
heavycrystal, line 668: nit: `binaryFormatTestcase`

flow/peerdbenv/dynamicconf.go
```diff
@@ -107,6 +107,14 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_NEW_MIRROR,
 		TargetForSetting: protos.DynconfTarget_ALL,
 	},
+	{
+		Name:             "PEERDB_BINARY_FORMAT",
```
heavycrystal, line 111: nit: add `CLICKHOUSE` to setting

flow/peerdbenv/dynamicconf.go
```diff
@@ -398,6 +428,23 @@ func PeerDBNullable(ctx context.Context, env map[string]string) (bool, error) {
 	return dynamicConfBool(ctx, env, "PEERDB_NULLABLE")
 }
 
+func PeerDBBinaryFormat(ctx context.Context, env map[string]string) (BinaryFormat, error) {
+	format, err := dynLookup(ctx, env, "PEERDB_BINARY_FORMAT")
+	if err != nil {
+		return 0, err
+	}
+	switch format {
+	case "raw":
```
heavycrystal, line 437: nit: make it case-insensitive since UI wouldn't really have any validation for this

flow/peerdbenv/dynamicconf.go
```diff
@@ -107,6 +107,14 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_NEW_MIRROR,
 		TargetForSetting: protos.DynconfTarget_ALL,
 	},
+	{
+		Name:             "PEERDB_BINARY_FORMAT",
+		Description:      "Binary field encoding; either raw, hex, or base64",
+		DefaultValue:     "base64",
```
heavycrystal, line 113: change to raw as a TODO

flow/peerdbenv/dynamicconf.go
```diff
@@ -238,6 +246,14 @@ var DynamicIndex = func() map[string]int {
 	return defaults
 }()
 
+type BinaryFormat int
+
+const (
+	BinaryFormatRaw = iota
```
heavycrystal, line 252: nit: 0 as Unknown since we return 0 in error case

flow/model/qvalue/avro_converter.go
```diff
@@ -509,9 +520,17 @@ func (c *QValueAvroConverter) processNumeric(num decimal.Decimal) any {
...
+func (c *QValueAvroConverter) processBytes(byteData []byte, format peerdbenv.BinaryFormat) interface{} {
+	if c.TargetDWH == protos.DBType_CLICKHOUSE && format != peerdbenv.BinaryFormatRaw {
+		var encoded string
+		switch format {
+		case peerdbenv.BinaryFormatBase64:
+			encoded = base64.StdEncoding.EncodeToString(byteData)
+		case peerdbenv.BinaryFormatHex:
+			encoded = strings.ToUpper(hex.EncodeToString(byteData))
+		default:
+			panic(fmt.Sprintf("unhandled binary format: %d", format))
```
heavycrystal, line 532: a panic can cause issues with maintenance mode (relies on flow-worker being active to pause mirrors)

serprex: panic can never happen unless a logic error is introduced in code, & bad code can always panic anyways

flow/peerdbenv/dynamicconf.go
serprex: by time we change default to raw we can probably remove this option

heavycrystal (approved):

PR #2409, opened 2025-01-05
[snapshot] close stream when fail to get srcConn in errgroup

heavycrystal:
if pull connector fails to initialize due to some reason and sync is already stuck on `chan receive`, then partClone workflow will hang indefinitely. fixed by closing stream explicitly before pull side errgroup goroutine returns.

### DIFF
--- a/flow/model/qrecord_stream.go
+++ b/flow/model/qrecord_stream.go
@@ -4,6 +4,10 @@ import (
 	"github.com/PeerDB-io/peer-flow/model/qvalue"
 )
 
+type StreamCloser interface {
+	Close(error)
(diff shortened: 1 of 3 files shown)


flow/model/qrecord_stream.go
```diff
@@ -4,6 +4,10 @@ import (
 	"github.com/PeerDB-io/peer-flow/model/qvalue"
 )
 
+type StreamCloser interface {
```
serprex, line 7: to fix lint move this interface to `flow/activities/flowable.go`

serprex (approved):

PR #2419, opened 2025-01-09
revive custom sync test

serprex:
removed in #2418

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -506,8 +506,15 @@ func CDCFlowWorkflow(
 			} else {
 				logger.Error("error in sync flow", slog.Any("error", err))
 			}
-			logger.Info("sync flow errored, sleeping for 10 minutes before retrying")
-			_ = workflow.Sleep(ctx, 10*time.Minute)
+
+			// cannot use shared.IsSQLStateError because temporal serialize/deserialize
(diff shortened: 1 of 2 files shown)


flow/workflows/cdc_flow.go
```diff
@@ -506,8 +506,8 @@ func CDCFlowWorkflow(
 			} else {
 				logger.Error("error in sync flow", slog.Any("error", err))
 			}
-			logger.Info("sync flow errored, sleeping for 10 minutes before retrying")
-			_ = workflow.Sleep(ctx, 10*time.Minute)
+			logger.Info("sync flow errored, sleeping for 30 seconds before retrying")
+			_ = workflow.Sleep(ctx, 30*time.Second)
```
serprex, line 510: if this is issue then better fix can be worked out so that NumberOfSyncs isn't hitting error path

flow/activities/flowable.go
```diff
@@ -346,7 +346,8 @@ func (a *FlowableActivity) SyncFlow(
 				slog.Int64("totalRecordsSynced", totalRecordsSynced.Load()))
 
 			if options.NumberOfSyncs > 0 && currentSyncFlowNum.Load() >= options.NumberOfSyncs {
-				break
+				close(syncDone)
+				return nil
```
serprex, line 350: don't like this fix since returning early with parallel normalize can cause context cancellation, preventing normalization

PR #2420, opened 2025-01-09
cdc flow: continue as new after state updates

serprex:
this prevents rewinding state to start of workflow from undoing changes to settings (such as table additions/removals)


serprex:
ideally we'd move to fully immutable state, but that just isn't the golang way

iskakaushik:
is there a way to write a test for this?

serprex:
> is there a way to write a test for this?

Maybe by checking that RunID changes. Will look into

iskakaushik (approved):

Issue #2382, opened 2024-12-19
Fix stale metrics being reported by otel


iamKunalGupta:
When slots are dropped and if the worker is still alive, the metrics are still exported.

This might be a regression due to #2272 and related to https://github.com/open-telemetry/opentelemetry-specification/issues/3062 and https://github.com/open-telemetry/opentelemetry-go/discussions/4619

status:
1. Gauges are fixed via #2421
2. Counters are TODO

iamKunalGupta:
@serprex should we revert back to using Observables? https://github.com/PeerDB-io/peerdb/pull/2272

PR #2421, opened 2025-01-09
refactor: revert to using observables modified to sync gauges

iamKunalGupta:
#2382

### DIFF
--- a/flow/otel_metrics/observables.go
+++ b/flow/otel_metrics/observables.go
@@ -1,0 +1,199 @@
+package otel_metrics
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"sync"
+	"sync/atomic"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/embedded"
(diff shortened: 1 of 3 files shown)


flow/otel_metrics/observables.go
```diff
@@ -0,0 +1,199 @@
...
+
+func Float64Gauge(meter metric.Meter, name string, opts ...metric.Float64GaugeOption) (metric.Float64Gauge, error) {
+	gaugeConfig := metric.NewFloat64GaugeConfig(opts...)
+	return NewFloat64SyncGauge(meter, name,
+		metric.WithDescription(gaugeConfig.Description()),
+		metric.WithUnit(gaugeConfig.Unit()),
+	)
+}
+
+// TODO this implementation is still pending
```
serprex, line 192: should either drop this or switch to it before merging

iamKunalGupta: yup, will test and check

iskakaushik (approved):

flow/otel_metrics/observables.go
iamKunalGupta: will take up counters later

PR #2422, opened 2025-01-09
fix: space pkm response to max once per minute

iamKunalGupta:


flow/connectors/postgres/cdc.go
```diff
@@ -386,7 +386,8 @@ func PullCdcRecords[Items model.Items](
 
 	for {
 		if pkmRequiresResponse {
-			if cdcRecordsStorage.IsEmpty() && int64(clientXLogPos) > req.ConsumedOffset.Load() {
+			if cdcRecordsStorage.IsEmpty() && int64(clientXLogPos) > req.ConsumedOffset.Load() &&
+				time.Since(standByLastLogged) >= 1*time.Minute {
```
serprex, line 390: ideally we'd throttle this based on this update, not alongside existing logging code. Or move logging code in here & only log once a minute

As is we may be logging every 10s & never update offset

flow/connectors/postgres/cdc.go
```diff
@@ -383,7 +384,12 @@ func PullCdcRecords[Items model.Items](
 
 	pkmRequiresResponse := false
 	waitingForCommit := false
-
+	pkmEmptyBatchThrottleEnabled, err := peerdbenv.PeerDBPKMEmptyBatchThrottleEnabled(ctx, nil)
```
serprex, line 387:
```suggestion
	pkmEmptyBatchThrottleEnabled, err := peerdbenv.PeerDBPKMEmptyBatchThrottleEnabled(ctx, req.Env)
```

flow/connectors/postgres/cdc.go
```diff
@@ -501,7 +508,13 @@ func PullCdcRecords[Items model.Items](
...
 				clientXLogPos = pkm.ServerWALEnd
 			}
-			pkmRequiresResponse = true
+			if pkmEmptyBatchThrottleEnabled {
+				if pkm.ReplyRequested || time.Since(lastEmptyBatchPkmSentTime) >= 1*time.Minute {
+					pkmRequiresResponse = true
+				}
+			} else {
+				pkmRequiresResponse = true
+			}
```
serprex, line 517:
```suggestion
			if pkmEmptyBatchThrottleEnabled || pkm.ReplyRequested || time.Since(lastEmptyBatchPkmSentTime) >= 1*time.Minute {
				pkmRequiresResponse = true
			}
```

iamKunalGupta: `pkmEmptyBatchThrottleEnabled` = false by default -> we want to not throttle

serprex: right, I'm not thinking straight, long day

serprex (approved):

PR #2423, opened 2025-01-10
fix: use condition instead of coalesce for updates on cdc_batches

iamKunalGupta:
Existing query can cause problems on mirrors with very low sync interval as the batches table grows large

### DIFF
--- a/nexus/catalog/migrations/v42__add_null_end_time_index_cdc_batches.sql
+++ b/nexus/catalog/migrations/v42__add_null_end_time_index_cdc_batches.sql
@@ -1,1 +1,0 @@
-CREATE INDEX IF NOT EXISTS idx_cdc_batches_end_time_null ON peerdb_stats.cdc_batches(end_time) WHERE end_time IS NULL;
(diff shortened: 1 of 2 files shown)


serprex (approved):

iskakaushik (approved):

nexus/catalog/migrations/v42__add_null_end_time_index_cdc_batches.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX IF NOT EXISTS idx_cdc_batches_end_time_null ON peerdb_stats.cdc_batches(end_time) WHERE end_time IS NULL;
```
serprex, line 1:
```suggestion
CREATE INDEX IF NOT EXISTS idx_cdc_batches_end_time_null ON peerdb_stats.cdc_batches(flow_name) WHERE end_time IS NULL;
```

iamKunalGupta: I think we already have one for flow name, let me check

iamKunalGupta: Yes we do https://github.com/PeerDB-io/peerdb/blob/main/nexus/catalog/migrations/V9__mirror_stats_rels.sql#L23

PR #2427, opened 2025-01-12
changed filename from generate_protos.sh to generate-protos.sh for consistency with readme

nirnayroy:
Just the change from _ to - in the filename. 
I agree that separate PR for this is overkill, but just in case it helps the devs.


serprex (approved):

iamKunalGupta:
Thank you for your contribution!

PR #2436, opened 2025-01-13
remove usehooks-ts

serprex:
we only use one hook from package, & usehooks-ts is one of the packages blocking react 19


iskakaushik (approved):

iamKunalGupta:
@serprex should I re-enable react major updates https://github.com/PeerDB-io/peerdb/blob/main/renovate.json#L27-L32

eslint-config might have to be migrated, but if nothing is blocking, maybe we can do it

PR #2443, opened 2025-01-14
cdc flow: only sleep for 10 minutes for panics & application errors

serprex:
In development heartbeat timeout happens between runs, which set off 10 minute timeout

Also, long sleeps in workflows using signals blocks those signals, causing things like taking minutes to start pausing.
Wrap `workflow.Sleep` in `workflow.Go` which sets off a `workflow.Future` which can be waited on alongside signals

### DIFF
--- a/flow/model/signals.go
+++ b/flow/model/signals.go
@@ -84,11 +84,6 @@ func (self TypedReceiveChannel[T]) ReceiveAsyncWithMoreFlag() (T, bool, bool) {
 	return result, ok, more
 }
 
-func (self TypedReceiveChannel[T]) Drain() {
-	for self.Chan.ReceiveAsync(nil) {
-	}
-}
-
 func (self TypedReceiveChannel[T]) AddToSelector(selector workflow.Selector, f func(T, bool)) workflow.Selector {
 	return selector.AddReceive(self.Chan, func(c workflow.ReceiveChannel, more bool) {
 		var result T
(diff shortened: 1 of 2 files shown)


flow/workflows/cdc_flow.go
```diff
@@ -501,21 +501,22 @@ func CDCFlowWorkflow(
 			var panicErr *temporal.PanicError
 			if errors.As(err, &panicErr) {
 				logger.Error(
-					"panic in sync flow",
+					"panic in sync flow, sleeping for 10 minutes",
```
iamKunalGupta, line 504: does this include nondeterminism?

serprex: no, non determinism isn't an activity error, but a workflow error which prevents workflow resuming

iamKunalGupta: my bad wasn't thinking straight, nondeterminism can only be probably caught by an interceptor

flow/model/signals.go
```diff
@@ -84,11 +84,6 @@ func (self TypedReceiveChannel[T]) ReceiveAsyncWithMoreFlag() (T, bool, bool) {
 	return result, ok, more
 }
 
-func (self TypedReceiveChannel[T]) Drain() {
```
serprex, line 87: Removing this because:
1. after moving sync flow to activity it's not used
2. draining channels is generally a sign that you're doing something wrong; stuff like having to use this method is why I now avoid signals with temporal

iskakaushik (approved):

PR #2444, opened 2025-01-14
UI: Reflect column and engine values for existing tables

Amogh-Bharadwaj:
Currently, excluded columns and engine selection are not populated in the table box for an existing table in the mirror (where the destination field, engine, columns are disabled)

This PR wires those through. Tested

### DIFF
--- a/ui/app/mirrors/create/cdc/schemabox.tsx
+++ b/ui/app/mirrors/create/cdc/schemabox.tsx
@@ -49,7 +49,7 @@ interface SchemaBoxProps {
     SetStateAction<{ tableName: string; columns: string[] }[]>
   >;
   peerType?: DBType;
-  alreadySelectedTables: TableMapping[] | undefined;
+  alreadySelectedTables: TableMapping[];
   initialLoadOnly?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


ui/app/mirrors/create/cdc/schemabox.tsx
```diff
@@ -199,12 +199,17 @@ export default function SchemaBox({
               ?.map((tableMap) => tableMap.sourceTableIdentifier)
               .includes(row.source)
           ) {
+            const existingRow = alreadySelectedTables?.find(
```
serprex, line 202:
```suggestion
            const existingRow = alreadySelectedTables.find(
```
can remove the `?.map` above too. As is it'll hit an error about trying to call `undefined` as a function. You'd have to use the `find?.(...)` to null check the call too

ui/app/mirrors/create/cdc/schemabox.tsx
```diff
@@ -199,12 +199,17 @@ export default function SchemaBox({
...
             row.editingDisabled = true;
-            row.destination =
-              alreadySelectedTables?.find(
-                (tableMap) => tableMap.sourceTableIdentifier === row.source
-              )?.destinationTableIdentifier ?? '';
+            row.exclude = new Set(existingRow?.exclude ?? []);
+            row.destination = existingRow?.destinationTableIdentifier ?? '';
+            alreadySelectedTables?.find(
+              (tableMap) => tableMap.sourceTableIdentifier === row.source
+            )?.destinationTableIdentifier ?? '';
```
serprex, line 212:
```suggestion
```

serprex (approved):

Issue #2430, opened 2025-01-13
nexus: support querying peers with hyphens in their name


iamKunalGupta:

iamKunalGupta:
not reproducible as of now

PR #2445, opened 2025-01-15
nexus: CREATE PEER should not include quotes in name

serprex:
Only look at first identifier in ObjectName, ideally parser should only accept single identifier there


serprex:
noticed while reproducing #2430

PR #2457, opened 2025-01-18
consolidate drop and resync into a single workflow

heavycrystal:


flow/activities/flowable.go
```diff
@@ -1063,3 +1067,69 @@ func (a *FlowableActivity) RemoveFlowEntryFromCatalog(ctx context.Context, flowN
 
 	return nil
 }
+
+func (a *FlowableActivity) HandleCancelWorkFlow(ctx context.Context, workflowID string) error {
+	errChan := make(chan error, 1)
```
serprex, line 1072: Trying to get rid of this pattern, it's really a misuse of channels

Instead you should prefer a `chan struct{}` which closes after setting an `err` variable

flow/activities/flowable.go
```diff
@@ -1063,3 +1067,69 @@ func (a *FlowableActivity) RemoveFlowEntryFromCatalog(ctx context.Context, flowN
...
+		err := a.TemporalClient.CancelWorkflow(ctxWithTimeout, workflowID, "")
+		errChan <- err
+	}()
+
+	select {
+	case err := <-errChan:
+		if err != nil {
+			slog.Error(fmt.Sprintf("unable to cancel workflow: %s. Attempting to terminate.", err.Error()))
+			terminationReason := fmt.Sprintf("error encountered while trying to cancel workflow %s", workflowID)
+			if err := a.TemporalClient.TerminateWorkflow(ctx, workflowID, "", terminationReason); err != nil {
```
serprex, line 1088: Termination seems like the wrong solution. If cancel isn't working it's because the workflow is stuck, possibly waiting on activity to close. Termination won't change that, it'll just orphan these stuck goroutines while we should be fixing them

flow/activities/flowable.go
```diff
@@ -1063,3 +1067,69 @@ func (a *FlowableActivity) RemoveFlowEntryFromCatalog(ctx context.Context, flowN
...
+
+	go func() {
+		err := a.TemporalClient.CancelWorkflow(ctxWithTimeout, workflowID, "")
+		errChan <- err
+	}()
+
+	select {
+	case err := <-errChan:
+		if err != nil {
+			slog.Error(fmt.Sprintf("unable to cancel workflow: %s. Attempting to terminate.", err.Error()))
```
serprex, line 1086: activities should use activity logger

flow/cmd/handler.go
```diff
@@ -515,30 +408,37 @@ func (h *FlowRequestHandler) ResyncMirror(
...
+		ID:                    dropFlowWorkflowID,
+		TaskQueue:             h.peerflowTaskQueueID,
+		TypedSearchAttributes: shared.NewSearchAttributes(req.FlowJobName),
 	}
 
-	if _, err := h.CreateCDCFlow(ctx, &protos.CreateCDCFlowRequest{
-		ConnectionConfigs: config,
-	}); err != nil {
-		return nil, err
+	if _, err := h.temporalClient.ExecuteWorkflow(ctx, workflowOptions,
```
serprex, line 430: Response should be workflow id so ui can poll status of operation

serprex:
Not clear how much consolidation is happening here, would expect to see code removal from existing resync logic. It that a planned followup?

flow/activities/flowable.go
iamKunalGupta: +1, we should only cancel. In case of termination issue, just raise an error because termination = manual intervention.

Also, you don't need a channel I believe now, as the context.WithTimeout should take care of non-await of the call, you can just do a select on `ctx.Done`?

iamKunalGupta: Also, need to skip in case PeerFlow was already canceled, does temporal client complain then?

flow/workflows/drop_flow.go
```diff
@@ -82,6 +86,18 @@ func DropFlowWorkflow(ctx workflow.Context, input *protos.DropFlowInput) error {
 	workflow.GetLogger(ctx).Info("performing cleanup for flow",
 		slog.String(string(shared.FlowNameKey), input.FlowJobName))
 
+	cancelWorkFlowCtx := workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		StartToCloseTimeout: 2 * time.Minute,
+		RetryPolicy: &temporal.RetryPolicy{
+			MaximumAttempts: 2,
+		},
+	})
+	if err := workflow.ExecuteActivity(cancelWorkFlowCtx,
```
iamKunalGupta, line 95: We need to validate mirror before any destructive operations

iamKunalGupta (commented):
tbh feels a little weird to see DropFlow execute resync code. Might make it a bit hard to distinguish between request type unless reviewed in temporal. better idea might be to have a drop as child workflow and execute the rest of the pieces in their activities as long as they are repeatable.

Also, alert in cases we cannot proceed

flow/activities/flowable.go
heavycrystal: I'll double check the behaviour, but https://github.com/temporalio/temporal/issues/2860 suggests cancelling a cancelled workflow doesn't error

PR #2463, opened 2025-01-20
E2E: add test for drop mirror

Amogh-Bharadwaj:
Adds a test for drop mirror in ClickHouse E2E:
- New helper functions in `test_utils.go`: `EnvWaitForEmptyStatsInCatalog` and `checkIfMirrorEntriesRemoved`
- Checks if there are any rows pertaining to the mirror in these tables:

```golang
	metadataTable := "metadata_last_sync_state"

	catalogTables := []string{
		metadataTable,
		"peerdb_stats.cdc_batches",
		"peerdb_stats.cdc_batch_table",
		"peerdb_stats.qrep_partitions",
		"peerdb_stats.qrep_runs",
	}
```

PR #2464, opened 2025-01-20
call SyncFlowCleanup during SetupFlow

heavycrystal:


serprex (approved):

flow/activities/flowable.go
```diff
@@ -78,6 +78,13 @@ func (a *FlowableActivity) SetupMetadataTables(ctx context.Context, config *prot
 	}
 	defer connectors.CloseConnector(ctx, dstConn)
 
+	// this should have been done by DropFlowDestination
+	// but edge case due to late context cancellation
+	if err := dstConn.SyncFlowCleanup(ctx, config.FlowName); err != nil {
```
serprex, line 83: can e2e test this with a drop test that passes false for deleting stats

heavycrystal: can be done in #2463

PR #2475, opened 2025-01-22
checkpoint: add text alongside int64

serprex:
for pg we currently use destination instead of catalog. mysql->pg will use catalog. therefore metadata tables on destination pg can ignore this change, avoiding issues upgrading existing pgpg setups

also update connector interfaces a bit

split out from #2395

### DIFF
--- a/flow/connectors/postgres/client.go
+++ b/flow/connectors/postgres/client.go
@@ -25,7 +25,7 @@ import (
 const (
 	mirrorJobsTableIdentifier = "peerdb_mirror_jobs"
 	createMirrorJobsTableSQL  = `CREATE TABLE IF NOT EXISTS %s.%s(mirror_job_name TEXT PRIMARY KEY,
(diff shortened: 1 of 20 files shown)


flow/connectors/postgres/client.go
```diff
@@ -44,7 +44,7 @@ const (
 	upsertJobMetadataForSyncSQL = `INSERT INTO %s.%s AS j (mirror_job_name,lsn_offset,sync_batch_id,normalize_batch_id)
 		VALUES ($1,$2,$3,0) ON CONFLICT(mirror_job_name) DO UPDATE SET lsn_offset=GREATEST(j.lsn_offset, EXCLUDED.lsn_offset),
 		sync_batch_id=EXCLUDED.sync_batch_id`
-	checkIfJobMetadataExistsSQL          = "SELECT COUNT(1)::TEXT::BOOL FROM %s.%s WHERE mirror_job_name=$1"
+	checkIfJobMetadataExistsSQL          = "SELECT EXISTS(SELECT * FROM %s.%s WHERE mirror_job_name=$1)"
```
serprex, line 47: `::TEXT::BOOL` only works for 0 & 1

iskakaushik (approved):

PR #2477, opened 2025-01-22
mysql: GetVersionConnector

serprex:
split from #2395 this will reduce the only changed lines to specialized lastOffset in flowable_core.go


flow/connectors/mysql/mysql.go
```diff
@@ -1,15 +1,125 @@
...
+
+func (c *MySqlConnector) ConnectionActive(context.Context) error {
 	return nil
 }
+
+func (c *MySqlConnector) connect(ctx context.Context) (*client.Conn, error) {
+	conn := c.conn[ctx]
+	if conn == nil {
+		argF := []client.Option{func(conn *client.Conn) error {
+			conn.SetCapability(mysql.CLIENT_COMPRESS)
```
heavycrystal, line 60: https://github.com/go-mysql-org/go-mysql/blob/97aa9858c8dc7ec8c22d7f37e7dcc444fb5a3ed0/driver/driver_options.go#L36 implies ZSTD is also an option

might want to either make it configurable to use this; also I'm not sure if the server gracefully handles this

flow/connectors/mysql/mysql.go
```diff
@@ -1,15 +1,125 @@
...
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/shared"
+)
 
-func (MySqlConnector) Close() error {
-	return nil
+type MySqlConnector struct {
+	*metadataStore.PostgresMetadata
+	config *protos.MySqlConfig
+	conn   map[context.Context]*client.Conn
```
heavycrystal, line 24: add a comment since this is pretty different behaviour from other connectors

flow/connectors/mysql/mysql.go
```diff
@@ -1,15 +1,125 @@
...
+		if _, err := conn.Execute("SET sql_mode = ANSI"); err != nil {
+			return nil, fmt.Errorf("failed to set sql_mode to ANSI: %w", err)
+		}
+	}
+	return conn, nil
+}
+
+// withRetries return an iterable over connections,
+// consumer should break out of loop on success or error,
+// to retry for mysql.ErrBadConn
```
heavycrystal, line 81: does this error type encompass network errors also?

flow/connectors/mysql/mysql.go
serprex: that would fit as a followup adding compression mode to mysql config. Removed setting this capability in this PR, extending proto being kept to main mysql PR or will be split off alongside flavor

heavycrystal (approved):

PR #2478, opened 2025-01-22
PEERDB_CLICKHOUSE_NORMALIZATION_PARTS

serprex:


flow/connectors/clickhouse/normalize.go
```diff
@@ -301,7 +301,18 @@ func (c *ClickHouseConnector) NormalizeRecords(
 					slog.Int64("syncBatchId", req.SyncBatchID),
 					slog.Int64("normalizeBatchId", normBatchID),
 					slog.String("query", query))
-				if err := chConn.Exec(errCtx, query); err != nil {
+
+				numParts, err := peerdbenv.PeerDBClickHouseNormalizationChunkingParts(ctx, req.Env)
+				if err != nil {
+					c.logger.Warn("failed to get chunking parts, proceeding without chunking", slog.Any("error", err))
+				} else if numParts > 1 {
```
iskakaushik, line 308: can we add "info" logging to indicate which branch we are in?

iskakaushik (approved):

flow/connectors/clickhouse/normalize.go
serprex: the query will be logged & include the partition value at the end

iskakaushik:
@serprex  can we do this for initial load as well please? a separate flag.

serprex:
> @serprex can we do this for initial load as well please? a separate flag.

alright, as separate PR too

PR #2480, opened 2025-01-22
Fix chunking with enabled primary key updates

serprex:
Appending to end of query broke with UNION ALL introduced when using primary key updates

This method also works better with setting to run inserts in parallel, where chunks will be distributed to these inserts,
which can speed up single table mirrors


iskakaushik:
can we add a test?

serprex:
> can we add a test?

will try find time tomorrow

iskakaushik (approved):

PR #2485, opened 2025-01-24
PEERDB_CLICKHOUSE_INITIAL_LOAD_PARTS_PER_PARTITION

serprex:


flow/connectors/clickhouse/qrep_avro_sync.go
```diff
@@ -155,13 +157,36 @@ func (s *ClickHouseAvroSyncMethod) SyncQRepRecords(
...
+	}
+	numParts = max(numParts, 1)
+
+	for i := range numParts {
+		var whereClause string
+		if numParts > 1 {
+			whereClause = fmt.Sprintf(" WHERE cityHash64(`%s`) %% %d = %d", hashColName, numParts, i)
+		}
+		query := fmt.Sprintf(
+			"INSERT INTO `%s`(%s) SELECT %s FROM s3('%s','%s','%s'%s, 'Avro')%s SETTINGS throw_on_max_partitions_per_insert_block = 0",
```
serprex, line 175:
```suggestion
			"INSERT INTO `%s`(%s) SELECT %s FROM s3('%s','%s','%s'%s, 'Avro')%s",
```
don't think we want to add that setting here

iskakaushik (approved):

PR #2489, opened 2025-01-27
Raise HeartbeatTimeout to 5m

serprex:
Despite sending heartbeats every 15s, we're seeing heartbeats regularly take almost 1m to register

When that latency exceeds 1m it causes activity failure. Avoid that


iamKunalGupta:
> Despite sending heartbeats every 15s, we're seeing heartbeats regularly take almost 1m to register
> 
> When that latency exceeds 1m it causes activity failure. Avoid that

Maybe we need to check this a little, because I feel this can "feel" slow to the end user when we actually run into an issue

heavycrystal (approved):

iskakaushik (changes requested):
we need to think more carefully about this.

serprex:
Hoping #2490 addresses issue

PR #2493, opened 2025-01-28
feat: add context propagator and use it to enrich metrics with flow metadata

iamKunalGupta:
The context is propagated as temporal headers:
image

Metrics automatically have the desired attributes:
image

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -1114,7 +1114,7 @@ func (a *FlowableActivity) GetFlowMetadata(
 	flowName string,
 	sourceName string,
 	destinationName string,
-) (*shared.FlowMetadata, error) {
+) (*protos.FlowContextMetadata, error) {
 	logger := log.With(activity.GetLogger(ctx), slog.String(string(shared.FlowNameKey), flowName))
 	peerTypes, err := connectors.LoadPeerTypes(ctx, a.CatalogPool, []string{sourceName, destinationName})
 	if err != nil {
@@ -1124,13 +1124,13 @@ func (a *FlowableActivity) GetFlowMetadata(
 	logger.Info("loaded peer types for flow", slog.String("flowName", flowName),
 		slog.String("sourceName", sourceName), slog.String("destinationName", destinationName),
 		slog.Any("peerTypes", peerTypes))
-	return &shared.FlowMetadata{
+	return &protos.FlowContextMetadata{
 		FlowName: flowName,
-		Source: shared.PeerMetadata{
+		Source: &protos.PeerContextMetadata{
 			Name: sourceName,
 			Type: peerTypes[sourceName],
 		},
-		Destination: shared.PeerMetadata{
+		Destination: &protos.PeerContextMetadata{
 			Name: destinationName,
 			Type: peerTypes[destinationName],
 		},
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -378,9 +378,7 @@ func (a *FlowableActivity) SyncFlow(
 				slog.Int64("totalRecordsSynced", totalRecordsSynced.Load()))
 			if a.OtelManager != nil {
 				a.OtelManager.Metrics.RecordsSyncedGauge.Record(ctx, syncResponse.NumRecordsSynced, metric.WithAttributeSet(attribute.NewSet(
-					attribute.String(otel_metrics.FlowNameKey, config.FlowJobName),
 					attribute.String(otel_metrics.BatchIdKey, strconv.FormatInt(syncResponse.CurrentSyncBatchID, 10)),
-					attribute.String(otel_metrics.SourcePeerType, fmt.Sprintf("%T", srcConn)),
 				)))
 			}
 			if options.NumberOfSyncs > 0 && currentSyncFlowNum.Load() >= options.NumberOfSyncs {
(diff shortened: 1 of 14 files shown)


flow/shared/context.go
```diff
@@ -0,0 +1,83 @@
...
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	FlowMetadataKey = "x-peerdb-flow-metadata"
+)
+
+type FlowMetadata struct {
+	FlowName    string
```
serprex, line 16: will modifications to this struct cause non determinism errors? would using protobuf for encoding/decoding payload help versioning issues?

iamKunalGupta: I will test this, but I assume that it should not cause issues as headers should not be considered inputs

flow/activities/flowable.go
```diff
@@ -1110,3 +1108,27 @@ func (a *FlowableActivity) RemoveFlowEntryFromCatalog(ctx context.Context, flowN
 
 	return nil
 }
+
+func (a *FlowableActivity) GetFlowMetadata(ctx context.Context, flowName string, sourceName string,
+	destinationName string) (*shared.FlowMetadata, error) {
+	logger := log.With(activity.GetLogger(ctx), slog.String(string(shared.FlowNameKey), flowName))
+	peerTypes, err := connectors.LoadPeerTypes(ctx, a.CatalogPool, []string{sourceName, destinationName})
```
heavycrystal, line 1115: do we want to cache this somewhere? not sure how often this will hit catalog with the metrics we will add

serprex: This function is meant to be run by workflow & then passed down to activities

heavycrystal: missed the shared function my bad
could be a `LocalActivity`?

iamKunalGupta: Yep, was thinking the same

flow/otel_metrics/observables.go
```diff
@@ -99,6 +102,52 @@ func NewFloat64SyncGauge(meter metric.Meter, gaugeName string, opts ...metric.Fl
...
+func (a *ContextAwareInt64SyncGauge) Record(ctx context.Context, value int64, options ...metric.RecordOption) {
+	flowMetadata := shared.GetFlowMetadata(ctx)
+	if flowMetadata != nil {
+		options = append(options, buildFlowMetadataAttributes(flowMetadata))
+	}
+	a.Int64SyncGauge.Record(ctx, value, options...)
+}
+
+type ContextAwareFloat64SyncGauge struct {
+	Float64SyncGauge
```
serprex, line 128:
```suggestion
	metric.Float64Gauge
```
can we keep this wrapping interface instead of returning interface & then downcasting?

if not methods like Float64Gauge should return Float64SyncGauge. Probably should either way

flow/workflows/drop_flow.go
```diff
@@ -85,7 +86,12 @@ func DropFlowWorkflow(ctx workflow.Context, input *protos.DropFlowInput) error {
 	ctx = workflow.WithValue(ctx, shared.FlowNameKey, input.FlowJobName)
 	workflow.GetLogger(ctx).Info("performing cleanup for flow",
 		slog.String(string(shared.FlowNameKey), input.FlowJobName))
-
+	newCtx, err := GetFlowMetadataContext(ctx,
+		input.FlowJobName, input.FlowConnectionConfigs.SourceName, input.FlowConnectionConfigs.DestinationName)
+	if err != nil {
+		return fmt.Errorf("failed to get flow metadata context: %w", err)
+	}
+	ctx = newCtx
```
serprex, line 94:
```suggestion
	var err error
	ctx, err = GetFlowMetadataContext(ctx,
		input.FlowJobName, input.FlowConnectionConfigs.SourceName, input.FlowConnectionConfigs.DestinationName)
	if err != nil {
		return fmt.Errorf("failed to get flow metadata context: %w", err)
	}
```

flow/workflows/cdc_flow.go
```diff
@@ -568,3 +574,15 @@ func CDCFlowWorkflow(
 		}
 	}
 }
+
+func GetFlowMetadataContext(ctx workflow.Context, flowJobName string, sourceName string, destinationName string) (workflow.Context, error) {
```
serprex, line 578: there's another utility method around in these workflow files (GetSideEffect?), these could be in a util.go

flow/workflows/cdc_flow.go
```diff
@@ -568,3 +574,15 @@ func CDCFlowWorkflow(
...
+	metadataCtx := workflow.WithLocalActivityOptions(ctx, workflow.LocalActivityOptions{
+		StartToCloseTimeout: 10 * time.Second,
```
serprex, line 580: give more time here, local activity has no heartbeat

serprex (approved):

flow/workflows/cdc_flow.go
iamKunalGupta: I think I can put in 30 secs

flow/otel_metrics/observables.go
iamKunalGupta: my bad, forgot to refactor

flow/workflows/drop_flow.go
iamKunalGupta: makes sense, ctx is not needed once we have an error

PR #2494, opened 2025-01-28
feat: add alert classifier

iamKunalGupta:
Will be good to have #2493 out first if possible

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -49,7 +49,7 @@ var (
 		Class: "NOTIFY_TERMINATE", action: NotifyUser,
 	}
 	ErrorNotifyConnectTimeout = ErrorClass{
-		// TODO
+		// TODO(TBD)
 		Class: "NOTIFY_CONNECT_TIMEOUT", action: NotifyUser,
 	}
 	ErrorEventInternal = ErrorClass{
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -126,6 +126,8 @@ func GetErrorClass(ctx context.Context, err error) ErrorClass {
 		switch pgErr.Code {
 		case "28P01": // invalid_password
 			return ErrorNotifyConnectivity
+		case "42P01": // undefined_table
+			return ErrorNotifyConnectivity
 		case "57P01": // admin_shutdown
 			return ErrorNotifyTerminate
 		case "57P03": // cannot_connect_now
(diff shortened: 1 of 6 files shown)


flow/alerting/classifier.go
```diff
@@ -0,0 +1,134 @@
...
+	return e.Class
+}
+func (e ErrorClass) ErrorAction() ErrorAction {
+	if e.action != "" {
+		return e.action
+	}
+	return NotifyTelemetry
+}
+
+func GetErrorClass(ctx context.Context, err error) ErrorClass {
```
serprex, line 87:
```suggestion
func GetErrorClass(ctx context.Context, err error) (ErrorClass, []string) {
```
this might as well compute tags, lots of overlap

iamKunalGupta: > Appending logic currently will cause duplicate tagging

Duplicate tags are fine, will do the same in phases, will have a different prefix for classification and action

> Logic a bit annoying with early returns here

Did it to prevent going through all if cases, can remove the return from switch and move after the switches, IIRC the switches use binary search

flow/shared/exceptions/postgres.go
```diff
@@ -0,0 +1,14 @@
+package exceptions
+
+// PostgresSetupError represents errors during setup of Postgres peers, maybe we can later replace with a more generic error type
+type PostgresSetupError struct {
```
serprex, line 4: should implement `Unwrap() error` to work better with various functions from `errors`

iamKunalGupta: > I don't see what this adds over fmt.Errorf

This is just for easier check, so we don't have to do string matching, have used it in 1 place and can gradually increase usage

serprex: ideal would be for there to be interface that has `ErrorClass() ErrorClass`

iamKunalGupta: agreed, I think we can do that once we have more error types

flow/alerting/classifier.go
```diff
@@ -0,0 +1,165 @@
...
+	// SSH related errors
+	var sshErr *ssh.OpenChannelError
+	if errors.As(err, &sshErr) {
+		return ErrorNotifyConnectivity
+	}
+	return ErrorOther
+}
+
+func isClickHouseMvError(exception *clickhouse.Exception) bool {
+	return strings.Contains(exception.Message, "while pushing to view")
```
serprex, line 164:
```suggestion
	return exception.Code == 241 && strings.Contains(exception.Message, "while pushing to view")
```

iskakaushik (approved):

serprex (approved):

PR #2502, opened 2025-01-29
mysql: transactional consistency

serprex:
prevent splitting transaction across multiple batches

### DIFF
--- a/flow/connectors/mysql/cdc.go
+++ b/flow/connectors/mysql/cdc.go
@@ -230,7 +230,6 @@ func (c *MySqlConnector) startCdcStreamingFilePos(
 func (c *MySqlConnector) startCdcStreamingGtid(
 	gset mysql.GTIDSet,
 ) (*replication.BinlogSyncer, *replication.BinlogStreamer, mysql.GTIDSet, mysql.Position, error) {
-	// https://hevodata.com/learn/mysql-gtids-and-replication-set-up
 	syncer := c.startSyncer()
 	stream, err := syncer.StartSyncGTID(gset)
 	if err != nil {
@@ -322,7 +321,7 @@ func (c *MySqlConnector) PullRecords(
 		return nil
 	}
 
-	for recordCount < req.MaxBatchSize {
+	for inTx || recordCount < req.MaxBatchSize {
 		getCtx := ctx
 		if !inTx {
 			getCtx = timeoutCtx
--- a/flow/connectors/mysql/cdc.go
+++ b/flow/connectors/mysql/cdc.go
@@ -140,9 +140,15 @@ func (c *MySqlConnector) SetupReplication(
 	ctx context.Context,
 	req *protos.SetupReplicationInput,
 ) (model.SetupReplicationResult, error) {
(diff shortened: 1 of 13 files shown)


flow/connectors/mysql/cdc.go
```diff
@@ -333,6 +344,7 @@ func (c *MySqlConnector) PullRecords(
 		case *replication.PreviousGTIDsEvent:
 			// TODO look into this, maybe we just do gset.Update(ev.GTIDSets)
 		case *replication.RowsEvent:
+			inTx = true
```
serprex, line 347: Maybe this should be on QueryEvent `"BEGIN"` instead. Current logic wait for transaction when dealing with non transactional updates

flow/connectors/mysql/cdc.go
```diff
@@ -298,6 +303,12 @@ func (c *MySqlConnector) PullRecords(
 		}
 
 		switch ev := event.Event.(type) {
+		case *replication.QueryEvent:
+			if string(ev.Query) == "COMMIT" {
+				inTx = false
+			}
+		case *replication.XIDEvent:
+			inTx = false
```
serprex, line 311: handle both based on https://github.com/krowinski/php-mysql-replication/issues/4

iskakaushik:
can you write a test for this?

flow/connectors/mysql/cdc.go
```diff
@@ -298,6 +303,12 @@ func (c *MySqlConnector) PullRecords(
 		}
 
 		switch ev := event.Event.(type) {
+		case *replication.QueryEvent:
+			if string(ev.Query) == "COMMIT" {
```
heavycrystal, line 307: what about `ROLLBACK`?
also `COMMIT WORK` and other MySQLisms: https://mariadb.com/kb/en/start-transaction/

serprex: rolled back transactions should not appear in binlog (otherwise we'd need to support rollback like in pg v2)

serprex:
> can you write a test for this?

I'm not sure how. It's easy enough to write a test which hangs a transaction open & shows no batch finishing, but otherwise what I'm more worried about is this getting stuck when it shouldn't, & that's not something I can write a test for

flow/connectors/mysql/cdc.go
serprex: not even going to punt on this for now, removed

heavycrystal (approved):

flow/e2e/mysql.go
```diff
@@ -81,13 +69,41 @@ func setupMyCore(t *testing.T, suffix string, isMaria bool) (*MySqlSource, error
...
+		}
+		gtidMode, err := rs.GetString(0, 0)
+		if err != nil {
+			connector.Close()
+			return nil, err
+		}
+		if !strings.EqualFold(gtidMode, "on") {
+			for _, sql := range []string{
+				"set global binlog_row_metadata=full",
+				"set global enforce_gtid_consistency=on",
```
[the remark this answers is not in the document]
serprex: I added it there before merging

PR #2504, opened 2025-01-30
QRepConsolidate: combine activities

serprex:


heavycrystal (commented):
rename `ConsolidateQRepPartitions` to indicate new things it does

PR #2506, opened 2025-01-30
mysql: cdcRecordsStorage

serprex:
port over code from postgres cdc to hook into cdcRecordsStorage, needed for when without `binlog_row_image = full`

but in pg we have an escape with tracking replica identity full. so need to weigh if this works here the same


serprex:
in pg with replica identity full we don't do anything with cdc storage

I think mysql can pass on this

PR #2509, opened 2025-01-30
mysql UI

serprex:


flow/connectors/mysql/schema.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+
+	schemas := make([]string, 0, rs.RowNumber())
+	for idx := range rs.RowNumber() {
+		val, err := rs.GetString(idx, 0)
+		if err != nil {
+			return nil, err
+		}
+		if val != "information_schema" && val != "performance_schema" && val != "sys" {
```
heavycrystal, line 40: nit: slices.Contains

flow/connectors/mysql/schema.go
```diff
@@ -0,0 +1,101 @@
...
+		val, err := rs.GetString(idx, 0)
+		if err != nil {
+			return nil, err
+		}
+		tables = append(tables, val)
+	}
+	return &protos.AllTablesResponse{Tables: tables}, nil
+}
+
+func (c *MySqlConnector) GetSchemas(ctx context.Context) (*protos.PeerSchemasResponse, error) {
```
heavycrystal, line 28: I'm not sure if this should even be a thing, because MySQL doesn't have schemas and we're essentially showing tables across all databases with this
```
In MySQL, physically, a schema is synonymous with a database. You can substitute the keyword SCHEMA instead of DATABASE in MySQL SQL syntax, for example using CREATE SCHEMA instead of CREATE DATABASE.
```

heavycrystal (commented):
need clarity on schema/database handling

flow/cmd/peer_data.go
```diff
@@ -138,8 +135,8 @@ func (h *FlowRequestHandler) ListPeers(
 ) (*protos.ListPeersResponse, error) {
 	query := "SELECT name, type FROM peers"
 	if peerdbenv.PeerDBOnlyClickHouseAllowed() {
-		// only postgres and clickhouse peers
-		query += " WHERE type IN (3, 8)"
+		// only postgres, mysql, and clickhouse
```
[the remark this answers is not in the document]
serprex: https://en.wikipedia.org/wiki/Serial_comma

https://www.cnn.com/2018/02/09/us/dairy-drivers-oxford-comma-case-settlement-trnd/index.html

flow/connectors/mysql/schema.go
serprex: Yes, I think it's better here to treat mysql databases like schemas rather than postgres databases

heavycrystal: then it makes sense to remove database from the config of mysql peers

iskakaushik (approved):

PR #2529, opened 2025-02-05
Remove consistency sleep and account for nil env

Amogh-Bharadwaj:
This PR fixes two issues:
1) Removes the 3 second sleep in NormalizeRecords in ClickHouse. This was a shot-in-the-dark measure to mitigate a data loss issue which we now know is fixed by specifying select_sequential_consistency in the case of non SMT tables.
2) In the case where `env` is not provided when creating a mirror, `cfg.Env` can be nil. When the mirror is edited and `updatedEnv` is passed, we get a `assignment to entry in nil map` error

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -10,7 +10,6 @@ import (
 	"slices"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
 	"golang.org/x/sync/errgroup"
@@ -230,9 +229,6 @@ func (c *ClickHouseConnector) NormalizeRecords(
 	ctx context.Context,
 	req *model.NormalizeRecordsRequest,
 ) (model.NormalizeResponse, error) {
-	// fix for potential consistency issues
-	time.Sleep(3 * time.Second)
-
 	normBatchID, err := c.GetLastNormalizeBatchID(ctx, req.FlowJobName)
 	if err != nil {
 		c.logger.Error("[clickhouse] error while getting last sync and normalize batch id", "error", err)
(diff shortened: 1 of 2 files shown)


iamKunalGupta (approved):

flow/workflows/cdc_flow.go
```diff
@@ -91,6 +91,9 @@ func processCDCFlowConfigUpdate(
 		state.SyncFlowOptions.NumberOfSyncs = 0
 	}
 	if flowConfigUpdate.UpdatedEnv != nil {
+		if cfg.Env == nil {
```
heavycrystal, line 94: I'm not sure if we should skip adding the env entirely in this case, need to think it through

flow/connectors/clickhouse/normalize.go
```diff
@@ -230,9 +229,6 @@ func (c *ClickHouseConnector) NormalizeRecords(
 	ctx context.Context,
 	req *model.NormalizeRecordsRequest,
 ) (model.NormalizeResponse, error) {
-	// fix for potential consistency issues
-	time.Sleep(3 * time.Second)
```
heavycrystal, line 234: unrelated, why are we removing this?

Amogh-Bharadwaj: It doesn't serve any purpose

flow/workflows/cdc_flow.go
```diff
@@ -91,6 +91,9 @@ func processCDCFlowConfigUpdate(
...
+			cfg.Env = make(map[string]string)
```
serprex, line 95:
```suggestion
			cfg.Env = make(map[string]string, len(flowConfigUpdate.UpdatedEnv))
```

serprex (approved):

PR #2531, opened 2025-02-05
Fix LoggerFromCtx to include flowName

iskakaushik:


flow/shared/logger.go
```diff
@@ -10,18 +10,20 @@ import (
...
-		}
-	} else if hasName {
-		return slog.With(string(FlowNameKey), flowName)
+		logger = activity.GetLogger(ctx)
 	} else {
-		return slog.Default()
+		logger = log.NewStructuredLogger(slog.Default())
 	}
+
+	flowName, hasName := ctx.Value(FlowNameKey).(string)
```
iamKunalGupta, line 21: @iskakaushik there are more context keys now if we want, check https://github.com/PeerDB-io/peerdb/blob/main/flow/shared/context.go#L27-L32

serprex, line 21: can remove a bunch of activity code that does this redundantly

serprex (approved):

flow/shared/logger.go
serprex: yes, we do want

flow/activities/flowable_core.go
```diff
@@ -628,7 +628,7 @@ func (a *FlowableActivity) startNormalize(
 	batchID int64,
 ) error {
 	ctx = context.WithValue(ctx, shared.FlowNameKey, config.FlowJobName)
-	logger := activity.GetLogger(ctx)
+	logger := shared.LoggerFromCtx(ctx)
```
serprex, line 631: this should be removed since this method is no longer a top level activity. Caller should pass logger down

PR #2536, opened 2025-02-07
feat(classifier): add pg insufficient_privilege

iamKunalGupta:


flow/alerting/classifier.go
```diff
@@ -137,6 +137,8 @@ func GetErrorClass(ctx context.Context, err error) ErrorClass {
 			return ErrorNotifyConnectivity
 		case "42P01": // undefined_table
 			return ErrorNotifyConnectivity
+		case "42501": // insufficient_privilege
```
heavycrystal, line 140: nit: you can coalesce these

heavycrystal (approved):

flow/alerting/classifier.go
iamKunalGupta: yeah, if the list grows we can do that. wanted to leave comments on the error code.
I think case is anyways condensed to a binary search

PR #2553, opened 2025-02-11
pg: client_encoding=UTF8

serprex:
Default is to use the database encoding. Our code was assuming UTF8. This hit issues on Neon


serprex:
@heavycrystal pointed out we already have this

PR #2564, opened 2025-02-12
[mysql] container with config for testing

heavycrystal:


stacks/mysql/my.cnf.sample
```diff
@@ -0,0 +1,21 @@
...
+binlog_format=ROW
+
+# Maximum size of a binary log file before a new one is created
+max_binlog_size=100M
+
+# Ensure the binary log is flushed to disk after each transaction
+sync_binlog=1
+
+# Set the binary log row image format (FULL, MINIMAL, or NOBLOB)
+binlog_row_image=full
```
serprex, line 18:
```suggestion
binlog_row_image=full

# Not required, but needed for excluded columns, & makes system more resilient
binlog_row_metadata=full
```

serprex (approved):

PR #2557, opened 2025-02-11
[mysql] ALTER TABLE ADD COLUMN

heavycrystal:
TODO e2e
TODO types
TODO swap out replace or as a follow-up

PR #2566, opened 2025-02-12
mysql cdc: bounds check pulling schema info

serprex:
This is an incomplete fix which avoids immediate panic when incoming row has more columns than we expect

Unfortunately without precise schema information we struggle to map cdc columns to type information which isn't available in mysql's cdc payloads

#2557 should avoid this scenario to avoid data loss


flow/connectors/mysql/cdc.go
```diff
@@ -370,7 +370,10 @@ func (c *MySqlConnector) PullRecords(
 						}
 						return nil
 					}
-					return schema.Columns[idx]
+					if idx < len(schema.Columns) {
+						return schema.Columns[idx]
+					}
+					return nil
```
heavycrystal, line 376: should add a warning here since data read would be incomplete, but could spam so try to hushWarn

heavycrystal (approved):

PR #2575, opened 2025-02-14
[mysql] attempt to autodetermine flavour

heavycrystal:
MariaDB frequently cosplays as MySQL and expecting all users to know what they have can lead to issues as we scale

### DIFF
--- a/flow/connectors/mysql/mysql.go
+++ b/flow/connectors/mysql/mysql.go
@@ -47,15 +47,15 @@ func NewMySqlConnector(ctx context.Context, config *protos.MySqlConfig) (*MySqlC
 	}, nil
 }
 
-func (c *MySqlConnector) Flavor() string {
-	switch c.config.Flavor {
-	case protos.MySqlFlavor_MYSQL_MYSQL:
-		return mysql.MySQLFlavor
-	case protos.MySqlFlavor_MYSQL_MARIA:
-		return mysql.MariaDBFlavor
-	default:
+func (c *MySqlConnector) Flavor(ctx context.Context) string {
+	// it looks like gtid_mode was never supported in MariaDB, so hoping this clears the flavor up
+	if _, err := c.Execute(ctx, "select @@gtid_mode"); err != nil {
+		if mysql.ErrorCode(err.Error()) == mysql.ER_UNKNOWN_SYSTEM_VARIABLE {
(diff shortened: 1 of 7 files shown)


flow/connectors/mysql/mysql.go
```diff
@@ -256,7 +256,8 @@ func (c *MySqlConnector) GetMasterGTIDSet(ctx context.Context) (mysql.GTIDSet, e
 	if err != nil {
 		return nil, fmt.Errorf("failed to GetString for %s: %w", query, err)
 	}
-	gset, err := mysql.ParseGTIDSet(c.Flavor(), gx)
+	c.logger.Info("QQQQ", slog.String("gset", gx))
+	gset, err := mysql.ParseGTIDSet(c.Flavor(ctx), gx)
```
serprex, line 260:
```suggestion
	gset, err := mysql.ParseGTIDSet(c.Flavor(ctx), gx)
```

serprex:
For ui there's plans to have separate tiles for mysql vs mariadb, which is what other ETL tools do. People generally know if they're on mariadb in recent years since mysql/maria diverged after v5

flow/connectors/mysql/mysql.go
```diff
@@ -47,15 +47,15 @@ func NewMySqlConnector(ctx context.Context, config *protos.MySqlConfig) (*MySqlC
...
-	switch c.config.Flavor {
-	case protos.MySqlFlavor_MYSQL_MYSQL:
-		return mysql.MySQLFlavor
-	case protos.MySqlFlavor_MYSQL_MARIA:
-		return mysql.MariaDBFlavor
-	default:
+func (c *MySqlConnector) Flavor(ctx context.Context) string {
+	// it looks like gtid_mode was never supported in MariaDB, so hoping this clears the flavor up
+	if _, err := c.Execute(ctx, "select @@gtid_mode"); err != nil {
+		if mysql.ErrorCode(err.Error()) == mysql.ER_UNKNOWN_SYSTEM_VARIABLE {
```
serprex, line 53: better off using `errors.As` to get `mysql.MyError` so you can pull Code field instead of using this function which parses error message

heavycrystal:
talking with @serprex, we'll instead validate flavour during creation and error out (retain the variable)

PR #2580, opened 2025-02-14
split out catalog errors as a separate class

serprex:


iamKunalGupta (approved):
lgtm, pushed a small change

PR #2594, opened 2025-02-18
add mysql validations

panbhag-ch:


flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
+}
+
+func (c *MySQLConnector) CheckBinlogSettings(ctx context.Context) error {
+	if c.conn == nil {
+		return errors.New("check binlog settings: conn is nil")
+	}
+
+	// Check binlog_expire_logs_seconds
+	var expireSeconds int
+	err := c.conn.QueryRowContext(ctx, "SELECT @@binlog_expire_logs_seconds").Scan(&expireSeconds)
```
serprex, line 89: Settings are flavor specific. Fine to return nil when flavor is not MySQL for now

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
 package connmysql
 
+
 import (
 	"context"
+	"database/sql"
```
serprex, line 6: have been avoiding `database/sql`, you can use existing Execute method on MySqlConnector

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
 )
 
+type MySQLConnector struct {
+	conn *sql.DB
+	config *MySQLConfig
+}
+
+func (c *MySQLConnector) CheckSourceTables(ctx context.Context, tableNames []*utils.SchemaTable) error {
+	if c.conn == nil {
+		return errors.New("check tables: conn is nil")
```
serprex, line 26: MySql connector doesn't maintain a single long lived connection, let Execute manage this

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
+
 import (
 	"context"
+	"database/sql"
 	"errors"
+	"fmt"
+	"strings"
 
 	"github.com/go-mysql-org/go-mysql/mysql"
+	_ "github.com/go-sql-driver/mysql"
```
serprex, line 12: we're using go-mysql-org/go-mysql, no need to mix in go-sql-driver/mysql

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
+	if c.conn == nil {
+		return errors.New("check tables: conn is nil")
+	}
+
+	for _, parsedTable := range tableNames {
+		query := fmt.Sprintf("SELECT 1 FROM `%s`.`%s` LIMIT 1", parsedTable.Schema, parsedTable.Table)
+		_, err := c.conn.QueryContext(ctx, query)
+		if err != nil {
+			return fmt.Errorf("error checking table %s.%s: %v", parsedTable.Schema, parsedTable.Table, err)
+		}
```
serprex, line 34:
```suggestion
		if _, err := c.conn.QueryContext(ctx, fmt.Sprintf("SELECT 1 FROM %s LIMIT 1", parsedTable.MySQL())); err != nil {
			return fmt.Errorf("error checking table %s: %w", parsedTable.MySQL(), err)
		}
```

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
+
+func (c *MySQLConnector) CheckReplicationConnectivity(ctx context.Context) error {
+	if c.conn == nil {
+		return errors.New("check replication connectivity: conn is nil")
+	}
+
+	var masterLogFile string
+	var masterLogPos int
+
+	err := c.conn.QueryRowContext(ctx, "SHOW MASTER STATUS").Scan(&masterLogFile, &masterLogPos)
```
serprex, line 65: this has flavor/versioning considerations, see `GetMasterPos` & `GetMasterGTIDSet`. Can probably use those methods instead of this

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
+	}
+
+	// Check binlog_row_metadata
+	var binlogRowMetadata string
+	err = c.conn.QueryRowContext(ctx, "SELECT @@binlog_row_metadata").Scan(&binlogRowMetadata)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve binlog_row_metadata: %v", err)
+	}
+	if binlogRowMetadata != "FULL" {
+		return errors.New("binlog_row_metadata must be set to 'FULL' for column exclusion support")
```
serprex, line 114: this should only be warning, not error. might be best to defer until mirror validation

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
+	}
+	if expireSeconds <= 86400 {
+		return errors.New("binlog_expire_logs_seconds is too low. Must be greater than 1 day")
+	}
+
+	// Check binlog_format
+	var binlogFormat string
+	err = c.conn.QueryRowContext(ctx, "SELECT @@binlog_format").Scan(&binlogFormat)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve binlog_format: %v", err)
```
serprex, line 101:
```suggestion
	if err := c.conn.QueryRowContext(ctx, "SELECT @@binlog_format").Scan(&binlogFormat); err != nil {
		return fmt.Errorf("failed to retrieve binlog_format: %w", err)
```
been trying to have consistent error scoping, this decision has avoided a couple issues

%w over %v in `fmt.Errorf` allows `errors.As` / `errors.Is` to pick up the underlying error, which is important for us to get mysql error codes

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
+
+	return nil
+}
+
+func (c *MySQLConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.FlowConnectionConfigs) error {
+	sourceTables := make([]*utils.SchemaTable, 0, len(cfg.TableMappings))
+	for _, tableMapping := range cfg.TableMappings {
+		parsedTable, parseErr := utils.ParseSchemaTable(tableMapping.SourceTableIdentifier)
+		if parseErr != nil {
+			return fmt.Errorf("invalid source table identifier: %s", parseErr)
```
serprex, line 145:
```suggestion
			return fmt.Errorf("invalid source table identifier: %w", parseErr)
```

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
 package connmysql
 
+
```
serprex, line 3:
```suggestion
```

serprex (changes requested):

flow/connectors/mysql/validate.go
serprex: to check binlog is enabled should only need to check `select @@log_bin`

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
+	}
+
+	// Check binlog_row_image
+	var binlogRowImage string
+	err = c.conn.QueryRowContext(ctx, "SELECT @@binlog_row_image").Scan(&binlogRowImage)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve binlog_row_image: %v", err)
+	}
+	if binlogRowImage != "FULL" {
+		return errors.New("binlog_row_image must be set to 'FULL' (equivalent to PostgreSQL's REPLICA IDENTITY FULL)")
```
heavycrystal, line 124: would prefer to avoid `REPLICA IDENTITY FULL` comparisons

flow/connectors/mysql/validate.go
```diff
@@ -1,14 +1,172 @@
...
+	return nil
+}
+
+func (c *MySQLConnector) CheckReplicationPermissions(ctx context.Context) error {
+	if c.conn == nil {
+		return errors.New("check replication permissions: conn is nil")
+	}
+
+	var replicationPrivilege string
+	err := c.conn.QueryRowContext(ctx, "SHOW GRANTS FOR CURRENT_USER()").Scan(&replicationPrivilege)
```
heavycrystal, line 45: it's more robust to parse this into a string array and then check to avoid matching partial permission names later

flow/connectors/mysql/validate.go
```diff
@@ -41,7 +40,7 @@ func (c *MySqlConnector) CheckReplicationPermissions(ctx context.Context) error
 }
 
 func (c *MySqlConnector) CheckReplicationConnectivity(ctx context.Context) error {
-	rs, err := c.Execute(ctx, "SHOW MASTER STATUS")
+	rs, err := c.Execute(ctx, "SHOW BINARY LOG STATUS")
```
serprex, line 43: earlier versions don't support `BINARY LOG`, see https://github.com/PeerDB-io/peerdb/blob/eb37de39360e86d724283942d69b9b4c3d09a787/flow/connectors/mysql/mysql.go#L215

flow/connectors/mysql/validate.go
panbhag-ch: at this point, not calling this when the flavour is Maria

flow/connectors/mysql/validate.go
panbhag-ch: its a warning now

flow/connectors/mysql/validate.go
```diff
@@ -3,12 +3,187 @@ package connmysql
...
 	"github.com/go-mysql-org/go-mysql/mysql"
 
+	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/shared"
 )
 
+func (c *MySqlConnector) CheckSourceTables(ctx context.Context, tableNames []*utils.SchemaTable) error {
+	for _, parsedTable := range tableNames {
+		if _, err := c.Execute(ctx, fmt.Sprintf("SELECT 1 FROM %s LIMIT 1", parsedTable.MySQL())); err != nil {
```
serprex, line 18:
```suggestion
		if _, err := c.Execute(ctx, fmt.Sprintf("SELECT * FROM %s LIMIT 0", parsedTable.MySQL())); err != nil {
```

flow/connectors/mysql/validate.go
```diff
@@ -3,12 +3,187 @@ package connmysql
...
+			grantSet[part] = struct{}{}
+		}
+
+		// Check for exact match of required privileges
+		if _, ok := grantSet["REPLICATION SLAVE"]; ok {
+			return nil
+		}
+		if _, ok := grantSet["REPLICATION CLIENT"]; ok {
+			return nil
+		}
```
serprex, line 48: Can compare parts in loop, no need to build map, it's O(N) time either way

flow/connectors/mysql/validate.go
```diff
@@ -3,12 +3,187 @@ package connmysql
...
+func (c *MySqlConnector) CheckReplicationPermissions(ctx context.Context) error {
+	rs, err := c.Execute(ctx, "SHOW GRANTS FOR CURRENT_USER()")
+	if err != nil {
+		return fmt.Errorf("failed to check replication privileges: %w", err)
+	}
+
+	for _, row := range rs.Values {
+		grant := shared.UnsafeFastReadOnlyBytesToString(row[0].AsString())
+		// Normalize and split grant string into individual permissions
+		grantParts := strings.FieldsFunc(grant, func(r rune) bool { return r == ' ' || r == ',' })
```
serprex, line 34: can use `strings.FieldsFuncSeq` instead

flow/connectors/mysql/validate.go
```diff
@@ -3,12 +3,187 @@ package connmysql
...
+
+	if len(rs.Values) == 0 {
+		c.logger.Warn("failed to retrieve binlog_row_metadata: no rows returned")
+		return nil
+	}
+
+	binlogRowMetadata := shared.UnsafeFastReadOnlyBytesToString(rs.Values[0][0].AsString()) // Convert FieldValue to string
+
+	if binlogRowMetadata != "FULL" {
+		c.logger.Warn("binlog_row_metadata must be set to 'FULL' for column exclusion support")
```
serprex, line 132: should be error for mirrors with column exclusion

flow/connectors/mysql/validate.go
```diff
@@ -3,12 +3,187 @@ package connmysql
...
+		return nil
```
serprex, line 133: returning nil here will overlook later errors when warning about this

flow/connectors/mysql/validate.go
```diff
@@ -3,30 +3,191 @@ package connmysql
...
+
 func (c *MySqlConnector) ValidateCheck(ctx context.Context) error {
 	if _, err := c.Execute(ctx, "select @@gtid_mode"); err != nil {
 		var mErr *mysql.MyError
 		// seems to be MariaDB
 		if errors.As(err, &mErr) && mErr.Code == mysql.ER_UNKNOWN_SYSTEM_VARIABLE && c.config.Flavor != protos.MySqlFlavor_MYSQL_MARIA {
 			return errors.New("server appears to be MariaDB but flavor is not set to MariaDB")
 		} else {
-			return err
+			return nil
```
serprex, line 171: @heavycrystal I think your validation was failing for mariadb

serprex (approved):

Issue #2259, opened 2024-11-15
maintenance mode enhancements


iamKunalGupta:
- [ ] Fix log for mirrorStatus to do toString before
- [ ] RunIntervalUntil should execute func before the ticker starts as well to optimize wait time
- [ ] Fix from_version when inserting start maintenance output
- [ ] #2254

iamKunalGupta:
The rest are out of scope as pausing is expected to ALWAYS work for running mirrors

Issue #2254, opened 2024-11-14
add UI components for maintenance mode


iamKunalGupta:
The APIs do not allow mirror mutations during maintenance, but the UI also should reflect that maintenance mode is enabled:

1.  a banner at the top (uses the instance info API) - "Maintenance is undergoing, any mirror operations will be disabled ⚠️ "
2. Disable the APIs from the UI? (or the banner in (1)) is enough?

PR #2596, opened 2025-02-18
feat(maintenance): improvements part 1

iamKunalGupta:
- [x] > RunIntervalUntil should execute func before the ticker starts as well to optimize wait time
    Faster StartMaintenance, no waiting 30 secs for each snapshot
- [x] > Fix from_version when inserting start maintenance output
- [x] > add UI components for maintenance mode

   Currently achieving this via sidebar, should be good enough I guess

   image

   image

   image

   image

Ref #2259 #2254


flow/activities/maintenance_activity.go
```diff
@@ -90,7 +90,7 @@ func (a *MaintenanceActivity) checkAndWaitIfSnapshot(
 	if mirror.MirrorCreatedAt.AsTime().After(time.Now().Add(-30 * time.Second)) {
 		slog.Info("Mirror was created less than 30 seconds ago, waiting for it to be ready before checking for snapshot",
 			"mirror", mirror.MirrorName, "workflowId", mirror.WorkflowId)
-		time.Sleep(30 * time.Second)
+		time.Sleep(time.Now().Add(35 * time.Second).Sub(mirror.MirrorCreatedAt.AsTime()))
```
serprex, line 93:
```suggestion
		time.Sleep(time.Now().Add(35 * time.Second).Sub(mirror.MirrorCreatedAt.AsTime()))
```
Shouldn't we be waiting on `mirror created at` + `30s` - `now` instead?

Should adjust condition too, set a targetTime which is createdAt+30, then check that time is before that, then sleep for difference of those values. Only call `time.Now()` once because logic will be cleaner (for stupid scenarios like slog.Info hangs for a minute)

flow/activities/maintenance_activity.go
```diff
@@ -264,7 +264,18 @@ func RunEveryIntervalUntilFinish[T any](
...
+	runBeforeFirstTick bool,
 ) (T, error) {
+	if runBeforeFirstTick {
+		finished, result, err := runFunc()
+		if err != nil {
+			return result, err
+		}
+		if finished {
+			return result, err
+		}
```
serprex, line 276:
```suggestion
		if err != nil || finished {
			return result, err
		}
```

flow/activities/maintenance_activity.go
iamKunalGupta: makes sense

ui/app/api/instance/info/route.ts
```diff
@@ -0,0 +1,10 @@
+import { GetFlowHttpAddressFromEnv } from '@/rpc/http';
+
+export const dynamic = 'force-dynamic';
+
+export async function GET() {
+  const flowServiceAddr = GetFlowHttpAddressFromEnv();
+  return fetch(`${flowServiceAddr}/v1/instance/info`, {
```
serprex, line 7: This file isn't needed, send request to `/api/v1/instance/info`

ui/components/SidebarComponent.tsx
```diff
@@ -106,17 +148,28 @@ export default function SidebarComponent(props: { showLogout: boolean }) {
...
+                <Label as='label' style={{ textAlign: 'center', fontSize: 15 }}>
+                  {' '}
+                  <b>Version: </b>
+                  {isLoading ? 'Loading...' : version?.version}
+                </Label>
+              </div>
+            </div>
+          ) : (
+            <></>
+          )}
```
serprex, line 164:
```suggestion
          {sidebarState === 'open' && (
            <div>
              <div style={centerFlexStyle}>
                <Label as='label' style={{ textAlign: 'center', fontSize: 15 }}>
                  {' '}
                  <b>Version: </b>
                  {isLoading ? 'Loading...' : version?.version}
                </Label>
              </div>
            </div>
          )}
```

ui/components/SidebarComponent.tsx
```diff
@@ -34,10 +45,41 @@ export default function SidebarComponent(props: { showLogout: boolean }) {
...
 
+  const {
+    data: instanceInfo,
+    error: instanceInfoError,
+    isLoading: isInstanceInfoLoading,
+  }: {
+    data: InstanceInfoResponse;
+    error: any;
+    isLoading: boolean;
+  } = useSWR('/api/instance/info', fetcher);
```
serprex, line 56:
```suggestion
  } = useSWR('/api/v1/instance/info', fetcher);
```

ui/app/api/instance/info/route.ts
iamKunalGupta: sidebar component is client side

serprex: left suggestion on SidebarComponent.tsx, see `ui/app/api/v1` for how requests get transparently proxied

iamKunalGupta: my bad, I see we have a proxy

serprex (approved):

PR #2606, opened 2025-02-19
feat(flow-api): add grpc server metrics

iamKunalGupta:
Currently we return nil instead of error from grpc in many cases -> this leads to incorrect grpc status code

### DIFF
--- a/flow/cmd/api.go
+++ b/flow/cmd/api.go
@@ -222,7 +222,6 @@ func APIMain(ctx context.Context, args *APIServerParams) error {
 		if err != nil {
 			return fmt.Errorf("unable to create otel manager: %w", err)
 		}
-		_ = otelManager.Meter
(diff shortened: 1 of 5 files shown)


flow/cmd/api.go
```diff
@@ -205,13 +208,27 @@ func APIMain(ctx context.Context, args *APIServerParams) error {
...
+		),
+	}
+
+	var otelManager *otel_metrics.OtelManager
+	if args.EnableOtelMetrics {
+		otelManager, err = otel_metrics.NewOtelManager(otel_metrics.FlowApiServiceName)
+		if err != nil {
+			return fmt.Errorf("unable to create otel manager: %w", err)
+		}
+		_ = otelManager.Meter
```
serprex, line 225: ?

iamKunalGupta: was testing smth locally, this chunk got added

serprex (approved):

PR #2608, opened 2025-02-19
cdc flow: increase timeout on repeated errors

serprex:


flow/workflows/cdc_flow.go
```diff
@@ -570,6 +571,11 @@ func CDCFlowWorkflow(
 				return nil, err
 			}
 
+			if finishedError {
```
serprex, line 574: this seems weak, since the only way to hit this with success is pause/resume, editing mirror, or waiting an absurdly long time

flow/workflows/cdc_flow.go
```diff
@@ -502,27 +503,27 @@ func CDCFlowWorkflow(
...
-					sleepFor = 10 * time.Minute
+					sleepFor = time.Duration(5+min(state.ErrorCount, 5)*15) * time.Minute
 				}
+
+				logger.Error("error in sync flow", slog.Any("error", err), slog.Any("sleepFor", sleepFor))
 			}
 			mainLoopSelector.AddFuture(model.SleepFuture(ctx, sleepFor), func(_ workflow.Future) {
 				logger.Info("sync finished after waiting after error")
 				finished = true
+				finishedError = true
```
serprex, line 526: maybe set this sooner so pause doesn't reset counter, but having counter reset on pause/resume seems like a feature

iamKunalGupta: +1, pause + resume should reset the counter.
In cases user is trying to fix the error, not having to wait to see if the fix worked would by resetting this externally would be valuable

iamKunalGupta (approved):

PR #2612, opened 2025-02-20
Classify normalization errors specially

serprex:
If no other classification applies by the end of an error from normalization, it is most likely user error

For example, this can be non-MV related in the case of certain schema changes


flow/alerting/classifier.go
```diff
@@ -57,6 +57,9 @@ var (
 		// TODO(this is mostly done via NOTIFY_CONNECTIVITY, will remove later if not needed)
 		Class: "NOTIFY_CONNECT_TIMEOUT", action: NotifyUser,
 	}
+	ErrorNotifyGeneral = ErrorClass{
+		Class: "NOTIFY_GENERAL", action: NotifyTelemetry,
+	}
```
iamKunalGupta, line 62: Let's call it ErrorNormalize for now so that we can get more info during alerting
```suggestion
	ErrorNormalize = ErrorClass{
		Class: "NORMALIZE", action: NotifyTelemetry,
	}
```

iamKunalGupta (approved):

iamKunalGupta (approved):

PR #2619, opened 2025-02-21
mysql: context

serprex:
based on https://github.com/go-mysql-org/go-mysql/pull/997 without having to switch to `database/sql`

### DIFF
--- a/flow/connectors/mysql/mysql.go
+++ b/flow/connectors/mysql/mysql.go
@@ -25,7 +25,7 @@ type MySqlConnector struct {
 	*metadataStore.PostgresMetadata
 	config   *protos.MySqlConfig
 	ssh      utils.SSHTunnel
-	conn     atomic.Pointer[client.Conn]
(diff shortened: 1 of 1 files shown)


flow/connectors/mysql/mysql.go
```diff
@@ -22,10 +23,11 @@ import (
 
 type MySqlConnector struct {
 	*metadataStore.PostgresMetadata
-	config *protos.MySqlConfig
-	ssh    utils.SSHTunnel
-	conn   *client.Conn
-	logger log.Logger
+	config   *protos.MySqlConfig
+	ssh      utils.SSHTunnel
+	conn     atomic.Pointer[client.Conn]
```
serprex, line 28: note that the use of atomics here does not expose a threadsafe interface, this is only to avoid race conditions on `c.conn` being replaced during retries & context watcher reading `c.conn`

heavycrystal: can be a comment just in case

heavycrystal (approved):

PR #2634, opened 2025-02-26
feat: persist flow status in catalog

### LINKED ISSUE #2623: Persist Mirror Status in Catalog
This is for 2 reasons:
1. Completed mirrors
   - WF which are beyond temporal retention period
   - Non Determinism of completed workflows
2. Costs associated with querying workflows

iamKunalGupta:
Resolves #2623


flow/internal/workflow.go
```diff
@@ -11,18 +11,58 @@ import (
...
+			slog.Info("flow status from catalog is unknown, will fall back to temporal",
+				slog.String("flowID", workflowID))
+		}
+
+		// this should only trigger for existing mirrors once
+		// except QRep/XMin mirrors, they will always hit this codepath
+		status, tctlErr := getWorkflowStatusFromTemporal(ctx, temporalClient, workflowID)
+		if tctlErr != nil {
+			return status, tctlErr
+		}
```
iamKunalGupta, line 48: @heavycrystal I don't his this is true, once we update the status in catalog, the status will ALWAYS be stale, which would lead to weird statuses

Maybe we need to add the same state saving logic in qrep and xmin

iamKunalGupta: But we also need to handle the case where QRep is actually a child of CDCFlowWorkflow, then we need to make sure to not update the status

heavycrystal: we could have skipped saving if the workflow type was not CDC but this is also good

flow/e2e/test_utils.go
```diff
@@ -671,7 +671,10 @@ func (env WorkflowRun) Query(ctx context.Context, queryType string, args ...any)
 
 func (env WorkflowRun) GetFlowStatus(t *testing.T) protos.FlowStatus {
 	t.Helper()
-	flowStatus, err := internal.GetWorkflowStatus(t.Context(), env.c, env.GetID())
```
serprex, line 674: why does this change? bypassing cache?

heavycrystal: e2e tests don't end up storing in catalog at all, so we get UNKNOWN status
bypassing that logic

serprex (approved):

flow/e2e/test_utils.go
iamKunalGupta: I'm not sure, but I think we should not bypass cache during tests

@heavycrystal changed this I think

flow/internal/workflow.go
```diff
@@ -11,18 +11,57 @@ import (
...
+
+		// this should only trigger for existing mirrors once
+		status, tctlErr := getWorkflowStatusFromTemporal(ctx, temporalClient, workflowID)
+		if tctlErr != nil {
+			return status, tctlErr
+		}
+		// only update the catalog if the status is different
+		if err == nil && status != flowStatus {
+			return UpdateFlowStatusInCatalog(ctx, pool, workflowID, status)
+		}
```
iamKunalGupta, line 51: There is an issue here actually wrt maintenance

Maintenance tries to get mirror status, then issues pause and then waits for the mirror to pause.

In this case, the first get status will write through the cache as `RUNNING`, then the check for `PAUSE` will always return the `RUNNING` status because workflow is on an old version.

I will comment out this piece and we can re-add it when we know that the worker is on a high enough version

flow/internal/workflow.go
```diff
@@ -11,18 +11,57 @@ import (
...
+
+		// this should only trigger for existing mirrors once
+		status, tctlErr := getWorkflowStatusFromTemporal(ctx, temporalClient, workflowID)
+		if tctlErr != nil {
+			return status, tctlErr
+		}
+		// only update the catalog if the status is different, this can enabled in later upgrades
+		//if err == nil && status != flowStatus {
+		//	return UpdateFlowStatusInCatalog(ctx, pool, workflowID, status)
+		//}
```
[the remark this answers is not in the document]
iamKunalGupta: For `COMPLETED` workflows we need to update the status on read as they are no longer running, I have a commit locally to enable this just for completed workflows, I'll push it

flow/internal/workflow.go
```diff
@@ -11,18 +11,58 @@ import (
...
+		// this should only trigger for existing mirrors once
+		status, tctlErr := getWorkflowStatusFromTemporal(ctx, temporalClient, workflowID)
+		if tctlErr != nil {
+			return status, tctlErr
+		}
+		// only update the catalog if the status is COMPLETED, this a long state, maintenance wfs do not alter these and we can assume
+		// that any changes post read-through cache happen only on new workers
+		if err == nil && status == protos.FlowStatus_STATUS_COMPLETED {
+			return UpdateFlowStatusInCatalog(ctx, pool, workflowID, status)
+		}
```
iamKunalGupta, line 52: @serprex we can assume here that resyncs for Completed flows only happen on upgraded workers

serprex: TERMINATING we'll get stuck on since we won't pick up when it gets to TERMINATED, it's okay to query

heavycrystal (approved):

PR #2638, opened 2025-02-27
feat(metrics): add operation and temporal context to metrics

iamKunalGupta:


flow/activities/flowable.go
```diff
@@ -336,9 +338,10 @@ func (a *FlowableActivity) SyncFlow(
 	normRequests := make(chan NormalizeBatchRequest, normalizeBufferSize)
 
 	group, groupCtx := errgroup.WithContext(ctx)
+	normalizeCtx := internal.WithOperationContext(groupCtx, protos.FlowOperation_FLOW_OPERATION_NORMALIZE)
```
serprex, line 341: can move this into the group.Go scope

serprex (approved):

flow/activities/flowable.go
iamKunalGupta: groupCtx also involves `syncRecords` and `maintainReplConn` which I think needs to be in the `sync` operation?

PR #2641, opened 2025-02-27
refactor empty signal

serprex:


flow/model/cdc_stream.go
```diff
@@ -81,23 +85,27 @@ func (r *CDCStream[T]) AddRecord(ctx context.Context, record Record[T]) error {
...
 func (r *CDCStream[T]) SignalAsEmpty() {
-	r.emptySignal <- true
+	r.emptySet = true
+	close(r.emptySignal)
 }
 
 func (r *CDCStream[T]) SignalAsNotEmpty() {
-	r.emptySignal <- false
+	r.empty = false
+	r.emptySet = true
```
heavycrystal, line 94: can we make `empty` a `*bool` and have nil, false and true be the values instead? or some other form of tristate variable
having 2 of them is unnecessary

flow/model/cdc_stream.go
```diff
@@ -32,11 +34,13 @@ func NewCDCStream[T Items](channelBuffer int) *CDCStream[T] {
 	return &CDCStream[T]{
 		records:            make(chan Record[T], channelBuffer),
 		SchemaDeltas:       make([]*protos.TableSchemaDelta, 0),
-		emptySignal:        make(chan bool, 1),
+		emptySignal:        make(chan struct{}),
```
heavycrystal, line 37: is removing buffered intentional?

flow/model/cdc_stream.go
serprex: I'd rather keep 2 bools without pointer indirection. Could make it a single byte which we pull bits out of, but that muddies things, & with alignment there's no impact on struct size

flow/model/cdc_stream.go
serprex: yes, since with this change nothing is ever pushed into the channel, it's only ever closed

heavycrystal (approved):

iamKunalGupta:
Nice branch name

PR #2644, opened 2025-02-28
[postgres] optimize uses of pgtype.Map

heavycrystal:


serprex:
If we aren't mutating this map seems like it could be a global variable

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -19,8 +19,8 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared"
 )
 
-func (c *PostgresConnector) postgresOIDToName(recvOID uint32) string {
-	if ty, ok := pgtype.NewMap().TypeForOID(recvOID); ok {
+func (c *PostgresConnector) postgresOIDToName(typeMap *pgtype.Map, recvOID uint32) string {
```
serprex, line 22: Don't need to pass typeMap parameters around when they're on the method receiver

serprex (approved):

PR #2646, opened 2025-02-28
feat: add lsn metric

iamKunalGupta:


serprex (approved):

Amogh-Bharadwaj (approved):

flow/connectors/postgres/cdc.go
```diff
@@ -626,6 +625,24 @@ func PullCdcRecords[Items model.Items](
 	}
 }
 
+func updateConsumedOffset(
```
serprex, line 628: wouldn't need to pass `p.otelManager` if this was a method

iamKunalGupta: nice catch, can also remove call to `NewPostgresMetadataFromCatalog`, but will make it just a method for now

PR #2660, opened 2025-03-04
PEERDB_IGNORE_DELETE

serprex:
This setting causes postgres/mysql to skip propagating deletes,
this may seem like a worse soft delete, but on ClickHouse it can be useful

TODO e2e test
TODO description should mention resync still wipes old data


serprex:
this isn't necessary since for postgres publication can filter out specific kinds of DML

PR #2661, opened 2025-03-04
mysql: bake in aws root

serprex:
allows mysql rds tls to work without having to explicitly set root ca of peer


stacks/flow.Dockerfile
```diff
@@ -20,7 +20,9 @@ ENV CGO_ENABLED=1
 RUN go build -ldflags="-s -w" -o /root/peer-flow
 
 FROM alpine:3.21@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c AS flow-base
+ADD https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem /usr/local/share/ca-certificates/global-aws-rds-bundle.pem
```
iskakaushik, line 23: WDYT about downloading the file in a separate build step (using tools like curl or wget) with proper integrity checks before copying it into the image.

serprex: In this case we'd have to also add curl. What is your concern here? Could skip network, copy cert into repo, & use `COPY`

serprex: https://docs.docker.com/reference/dockerfile/#add looks like `ADD` also supports checksum

iskakaushik (approved):

PR #2662, opened 2025-03-05
ClickHouse: Maintain normalize metadata to prevent redundant sync steps

Amogh-Bharadwaj:
Consider a mirror with 5 tables. Currently, if a push to one of those 5 tables fails, we will retry normalize, and this retry, the following redundant steps occur:
1) **We load the S3 files to the raw table again**: This can be expensive (time, memory, storage, cost etc.) in cases where syncs have gone way ahead of normalize. 
2) The remaining 4 tables will have these batches synced again. The inconveniences caused by this in terms of resources and data modelling is beyond the scope of this PR description.

Screenshot 2025-03-05 at 2 06 46 PM

This PR introduces two new columns in the `metadata_last_sync_state` column:
1) A bigint column which denotes the most recently synced batch id in raw table
2) A json column which tells us the recent batch id normalized to each target table of the mirror.

We use this to memoize our pushes to raw table and the final tables

Screenshot 2025-03-05 at 2 07 29 PM

#### Notes
1) This can be implemented for the other target peers as a follow up
2) An E2E test is included in this PR testing the above and getting it to work is in progress

PR #2673, opened 2025-03-06
ClickHouse: use retryable exec for normalise insert

Amogh-Bharadwaj:
Seems like we have facility to detect some retryable error codes when inserting into clickhouse, allowing us to retry on the spot rather than having normalise as a whole retry

Related [comment](https://github.com/PeerDB-io/peerdb/pull/2662#discussion_r1983113366)


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -278,6 +278,11 @@ func (c *ClickHouseConnector) exec(ctx context.Context, query string, args ...an
 	return chvalidate.Exec(ctx, c.logger, c.database, query, args...)
 }
 
+//nolint:unparam
```
serprex, line 281:
```suggestion
```

PR #2677, opened 2025-03-08
 feat(mysql): add support for geometry type

panbhag-ch:


flow/connectors/mysql/qvalue_convert.go
```diff
@@ -185,6 +189,60 @@ func QRecordSchemaFromMysqlFields(tableSchema *protos.TableSchema, fields []*mys
...
+	}
+	
+	// If we can't parse it as WKB, check if it's already in WKT format
+	strVal := string(data)
+	if isWKTFormat(strVal) {
+		// It's already in WKT format
+		return qvalue.QValueGeometry{Val: strVal}
+	}
+	
+	// Fallback to storing as string if parsing fails
```
serprex, line 242:
```suggestion
```
our result is same regardless of valid WKT or not

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -185,6 +189,60 @@ func QRecordSchemaFromMysqlFields(tableSchema *protos.TableSchema, fields []*mys
 	return qvalue.QRecordSchema{Fields: schema}, nil
 }
 
+// Helper function to convert MySQL geometry binary data to WKT format
+func geometryValueFromBytes(wkbData []byte) (string, error) {
+	// Try to parse it as WKB with the MySQL header
+	g, err := geom.NewGeomFromWKB(wkbData)
+	if err != nil && len(wkbData) > 4 {
```
serprex, line 196: my understanding is that this will always be the case

iskakaushik:
can we add an e2e test here please?

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -366,7 +425,22 @@ func QValueFromMysqlRowEvent(mytype byte, qkind qvalue.QValueKind, val any) (qva
 		case qvalue.QValueKindJSON:
 			return qvalue.QValueJSON{Val: val}, nil
 		case qvalue.QValueKindGeometry:
-			// TODO figure out mysql geo encoding
+			// For string representation of geometry
```
serprex, line 428: We can probably remove this code path, I think string is only really passed down from go-mysql for time types

panbhag-ch:
> can we add an e2e test here please?

added

flow/connectors/mysql/qvalue_convert.go
panbhag-ch: got it, for now removing one so that we can test

serprex (approved):

PR #2687, opened 2025-03-10
qrep: fetched bytes

serprex:
1. sqlserver source reports 0 bytes
2. PGPG reports 0 bytes (would need to wrap writer to count bytes)
3. MySQL estimates byte size of fields, ignores compression


iamKunalGupta:
`temporalWorkflowType="QRepPartitionWorkflow"` can be used to filter for initial load, will check isResync and inform as well

iamKunalGupta (approved):

PR #2694, opened 2025-03-11
Close pgx.Rows in queries to Postgres peer and catalog

Amogh-Bharadwaj:
Noticed there were a lot of places where we were not closing our connections to Postgres


serprex:
`CollectRows` handles closing: https://pkg.go.dev/github.com/jackc/pgx/v5#CollectRows

Amogh-Bharadwaj:
Related to connectors not being closed rather than the rows. Will be done in a PR soon

PR #2702, opened 2025-03-12
Add test for geometry for pg to clickhouse

panbhag-ch:


flow/e2e/clickhouse/peer_flow_ch_test.go
```diff
@@ -1146,6 +1146,175 @@ func (s ClickHouseSuite) Test_InitialLoadOnly_No_Primary_Key() {
...
+		line    string
+		lseg    string
+		box     string
+		path    string
+		polygon string
+		circle  string
+		id      int
+	}{
+		{
+			point:   "POINT(1.000000 2.000000)",
```
panbhag-ch, line 1249: Lets discuss on this. Why is this different from other formats

PR #2705, opened 2025-03-13
ClickHouse: Support for special characters in column names

### LINKED ISSUE #2701: support source column with spaces and other weird characters
Avro does not like spaces (and lots of other characters) in field names, so we need to support either via sanitizing or have a source <-> avro <-> destination mapping

Amogh-Bharadwaj:
Avro requires alphanumeric + underscore field names. However Postgres column names do not have the same restriction.

- This PR introduces a map of column names to an Avro compatible encoding of the same. Violating characters are replaced by underscore and the resulting string is suffixed with a serial number.
- This is utilized only for the initial load Avro loading since for CDC the columns will be that of our raw table columns.
- This PR only implements this support for ClickHouse target peers
- It also supports these columns for our normalize inserts by escaping the single quotes in column name argument of the JSONExtract* functions
- Functionally tested and the unprivileged columns E2E test has been adapted to include more columns to test this PR

Fixes #2701

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -375,23 +375,23 @@ func (c *ClickHouseConnector) NormalizeRecords(
 				case "Date32", "Nullable(Date32)":
 					projection.WriteString(fmt.Sprintf(
 						"toDate32(parseDateTime64BestEffortOrNull(JSONExtractString(_peerdb_data, '%s'),6)) AS `%s`,",
-						colName, dstColName,
+						strings.ReplaceAll(colName, "'", "''"), dstColName,
 					))
 					if enablePrimaryUpdate {
 						projectionUpdate.WriteString(fmt.Sprintf(
 							"toDate32(parseDateTime64BestEffortOrNull(JSONExtractString(_peerdb_match_data, '%s'),6)) AS `%s`,",
-							colName, dstColName,
+							strings.ReplaceAll(colName, "'", "''"), dstColName,
 						))
 					}
 				case "DateTime64(6)", "Nullable(DateTime64(6))":
 					projection.WriteString(fmt.Sprintf(
 						"parseDateTime64BestEffortOrNull(JSONExtractString(_peerdb_data, '%s'),6) AS `%s`,",
-						colName, dstColName,
+						strings.ReplaceAll(colName, "'", "''"), dstColName,
 					))
 					if enablePrimaryUpdate {
 						projectionUpdate.WriteString(fmt.Sprintf(
 							"parseDateTime64BestEffortOrNull(JSONExtractString(_peerdb_match_data, '%s'),6) AS `%s`,",
-							colName, dstColName,
+							strings.ReplaceAll(colName, "'", "''"), dstColName,
 						))
 					}
 				default:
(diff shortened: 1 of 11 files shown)


flow/connectors/utils/avro/avro_writer.go
```diff
@@ -127,19 +128,33 @@ func (p *peerDBOCFWriter) createOCFWriter(w io.Writer) (*goavro.OCFWriter, error
...
+	}
+
+	return avroFieldNames, nil
+}
+
 func (p *peerDBOCFWriter) writeRecordsToOCFWriter(ctx context.Context, env map[string]string, ocfWriter *goavro.OCFWriter) (int64, error) {
 	logger := internal.LoggerFromCtx(ctx)
-	schema, err := p.stream.Schema()
+
+	avroFieldNames, err := p.getAvroFieldNamesFromSchemaJSON()
```
Amogh-Bharadwaj, line 148: I approve of this even outside of this PR as using the very same field names set during schema definition for the OCF map append makes sense over getting the field names from Postgres schema

flow/connectors/clickhouse/normalize.go
```diff
@@ -375,23 +375,23 @@ func (c *ClickHouseConnector) NormalizeRecords(
 				case "Date32", "Nullable(Date32)":
 					projection.WriteString(fmt.Sprintf(
 						"toDate32(parseDateTime64BestEffortOrNull(JSONExtractString(_peerdb_data, '%s'),6)) AS `%s`,",
-						colName, dstColName,
+						strings.ReplaceAll(colName, "'", "''"), dstColName,
```
serprex, line 378: clickhouse doesn't use `''` escaping, you need to use `\`. We should create a utility function

https://github.com/ClickHouse/clickhouse-connect/blob/ca45ff68ece18211d2a5c47158f84e834bf6690f/clickhouse_connect/driver/binding.py#L108

flow/e2e/clickhouse/peer_flow_ch_test.go
```diff
@@ -1068,15 +1068,20 @@ func (s ClickHouseSuite) Test_Unprivileged_Postgres_Columns() {
 		CREATE TABLE IF NOT EXISTS %s (
 			id SERIAL PRIMARY KEY,
 			key TEXT NOT NULL,
-			"se'cret" TEXT
+			"se'cret" TEXT,
```
serprex, line 1071: note this is being excluded so doesn't fully test `'` handling

You may want to see the WeirdTable function which was test meant to look into these (but skipped columns since those didn't work)

flow/model/qvalue/avro_converter.go
```diff
@@ -71,6 +72,24 @@ func TruncateOrLogNumeric(num decimal.Decimal, precision int16, scale int16, tar
...
+	// subsequently contain only [A-Za-z0-9_]
+	if columnName == "" {
+		return "_"
+	}
+	// Ensure the first character is a letter or underscore
+	if columnName[0] >= '0' && columnName[0] <= '9' {
+		columnName = "_" + columnName
+	}
+	// Replace invalid characters with _
+	re := regexp.MustCompile(`[^A-Za-z0-9_]`)
```
serprex, line 88: this should be a global variable

flow/connectors/clickhouse/normalize.go
Amogh-Bharadwaj: > clickhouse doesn't use '' escaping
It does, I tested it locally and on CH Cloud

Screenshot 2025-03-13 at 7 09 36 PM

Amogh-Bharadwaj: But I'll go with `\` anyway since it's more common

serprex:
#2701

serprex (approved):

PR #2723, opened 2025-03-18
added e2e test for cutoff bug in schema change handling

heavycrystal:
schema change replay has a bug where in some edge cases, we blindly update the schema of a Postgres table and save it to catalog before we receive the `RelationMessage` from Postgres that would help us detect and apply the change on the destination

The root cause of this was in a function `applySchemaDeltas` we refreshed the schemas of *all* tables in catalog instead of the ones we saw changes for, so consider the following sequence of events

1. receive changes for table t2
2. schema change: add column for table t1
3. schema change: add column for table t2
4. receive changes for table t1 with new column
5. receive changes for table t2 with new column

If the current batch ends between 4 and 5. Postgres sends `RelationMessage` for t1 with the new column so we are able to add the column on destination, but not for t2 since we haven't seen changes with the new schema yet.

But the aforementioned function fetches the updated schema for t2 as well, and normalizes start failing because we're using a schema we haven't reconciled with the destination yet. This is non recoverable since even after we receive the `RelationMessage` for t2; we match it with the schema from catalog and see no differences to apply.

This test should fail atm, fixes in progress at #2717 and #2721.

### DIFF
--- a/flow/e2e/generic/generic_test.go
+++ b/flow/e2e/generic/generic_test.go
@@ -388,18 +388,10 @@ func (s Generic) Test_Schema_Changes_Cutoff_Bug() {
 	srcTableName2 := e2e.AttachSchema(s, srcTable2)
 	dstTableName2 := s.DestinationTable(dstTable2)
 
-	require.NoError(t, s.Source().Exec(t.Context(), fmt.Sprintf(`
-		CREATE TABLE IF NOT EXISTS %[1]s (
-			id SERIAL PRIMARY KEY,
-			c1 BIGINT
-		);
-		CREATE TABLE IF NOT EXISTS %[2]s (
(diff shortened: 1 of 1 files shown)


flow/e2e/generic/generic_test.go
```diff
@@ -374,3 +374,138 @@ func (s Generic) Test_Partitioned_Table() {
 	env.Cancel(t.Context())
 	e2e.RequireEnvCanceled(t, env)
 }
+
+func (s Generic) Test_Schema_Changes_Cutoff_Bug() {
+	t := s.T()
+
+	if _, ok := s.Source().(*e2e.MySqlSource); ok {
+		t.Skip("mysql connector does not support schema changes yet")
```
serprex, line 382: mysql supports schema changes

PR #2724, opened 2025-03-18
PEERDB_SOURCE_SCHEMA_AS_DESTINATION_COLUMN

serprex:
this can be useful when multiple sources ingest to a single table, which is now possible after #2534

another potential option would be `PEERDB_SOURCE_AS_DESTINATION_COLUMN` which would put the whole table identifier into this column, instead of extracting the schema


serprex:
This seems more suitable as a bool on the cdc flow proto, which would pass down to snapshot where qrep config would tweak query accordingly. As a setting it doesn't really make sense since as of now changing the setting will break mirrors until destination tables updated by user to remove or add `_peerdb_source_schema`

heavycrystal (approved):

PR #2755, opened 2025-03-25
remove sqlserver

serprex:
merge generic sql code into snowflake, its only consumer


heavycrystal (approved):
nice cleanup, brings joy

PR #2759, opened 2025-03-25
feat(maintenance): add signal to skip snapshot for specific flows

iamKunalGupta:
ref #2259 

To bypass maintenance for long running failing initial loads, we had to disable sync and maintenance in chart, and then remember to enable it in chart later when upgrading the next time.

This also caused laziness to skip upgrading altogether

This PR adds a new signal which can be used to bypass initial load checks for specified flows:
```json
{
  "skipped_snapshot_wait_flows": [
    "failing_flow_name_here"
  ]
}
```

### DIFF
--- a/flow/workflows/maintenance_flow.go
+++ b/flow/workflows/maintenance_flow.go
@@ -93,7 +93,7 @@ func startMaintenance(ctx workflow.Context, logger log.Logger) (*protos.StartMai
 		cancelCurrentChild()
 	})
 	workflow.Go(ctx, func(ctx workflow.Context) {
-		for {
+		for ctx.Err() == nil {
 			maintenanceSelector.Select(ctx)
 		}
 	})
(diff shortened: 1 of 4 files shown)


serprex (approved):

flow/workflows/maintenance_flow.go
```diff
@@ -72,19 +76,59 @@ func StartMaintenanceWorkflow(ctx workflow.Context, input *protos.StartMaintenan
...
+		for _, flow := range maintenanceSignal.SkippedSnapshotWaitFlows {
+			newSkippedFlows[flow] = struct{}{}
+		}
+		skippedFlows = newSkippedFlows
+		cancelCurrentChild()
+	})
+	workflow.Go(ctx, func(ctx workflow.Context) {
+		for {
+			maintenanceSelector.Select(ctx)
+		}
```
serprex, line 97: need to break loop when context done. you'll also need to add  `ctx.Done()` to `maintenanceSelect`

flow/workflows/maintenance_flow.go
```diff
@@ -72,19 +76,59 @@ func StartMaintenanceWorkflow(ctx workflow.Context, input *protos.StartMaintenan
...
+			var snapshotWaitCancelCtx workflow.Context
+			snapshotWaitCancelCtx, cancelCurrentChild = workflow.WithCancel(snapshotWaitCtx)
+			waitSnapshotsFuture := workflow.ExecuteActivity(snapshotWaitCancelCtx,
+				maintenance.WaitForRunningSnapshots,
+				skippedFlows,
+			)
+			if err := waitSnapshotsFuture.Get(snapshotWaitCtx, nil); err != nil {
+				// It is ok to always ignore cancellation and always retry
+				// if more information is needed, we can maintain some state whether we cancelled it or someone else
+				if errors.Is(err, workflow.ErrCanceled) {
```
serprex, line 123: should check if `snapshotWaitCtx` is canceled, in that case exit

iamKunalGupta: wouldn't cooperative cancellation always ensure that `snapshotWaitCtx` is also always cancelled if we take the `ErrCanceled` branch?

flow/workflows/maintenance_flow.go
iamKunalGupta: > I would suggest trying to do this without `workflow.Go`

Is that an antipattern? I am mainly trying to process signals in background an using cancellation of the activity as a downstream signalling mechanism

https://docs.temporal.io/develop/go/message-passing#signals

PR #2760, opened 2025-03-25
update columns endpoint to not join fields by colon into a string

serprex:


Amogh-Bharadwaj (approved):

serprex:
talking with amogh, we're good to replace this inplace

PR #2766, opened 2025-03-27
refactoring, map enums to LowCardinality(String) on CH

### LINKED ISSUE #2738: Add support for custom ENUM array type
Today when performing an initial load from postgres to clickhouse, custom enum types are detected as invalid (QValueKindInvalid) and mapped to string which is fine,
However, custom enum arrays are also mapped to string, we think they should be mapped Array(string).

Please allow mapping of custom enum array type to array_string 
Attached is a small temporary patch we have done in the meanwhile to allow us to correctly perform the mapping to array_string

[array_string.patch](https://github.com/user-attachments/files/19407555/array_string.patch)

serprex:
fixes #2738


Amogh-Bharadwaj (approved):

flow/connectors/postgres/qrep_query_executor_test.go
```diff
@@ -145,15 +145,7 @@ func TestAllDataTypes(t *testing.T) {
...
-	query = fmt.Sprintf("SELECT * FROM %s.test;", schemaName)
-	rows, err := qe.ExecuteQuery(t.Context(), query)
-	require.NoError(t, err, "error while executing query")
-	defer rows.Close()
-
-	// Use rows.FieldDescriptions() to get field descriptions
-	fieldDescriptions := rows.FieldDescriptions()
-
-	batch, err := qe.ProcessRows(ctx, rows, fieldDescriptions)
+	batch, err := qe.ExecuteAndProcessQuery(t.Context(), fmt.Sprintf("SELECT * FROM %s.test;", schemaName))
```
Amogh-Bharadwaj, line 148: QuoteIdentifier?

PR #2775, opened 2025-03-31
[mysql] check RDS/Aurora binlog retention hours

### LINKED ISSUE #2772: MySQL RDS validation: binlog retention hours
```sql
select value from mysql.rds_configuration where name = 'binlog retention hours'
```

default is 0, should be at least 24h

heavycrystal:
closes #2772 
also refactor some checks to be available to MariaDB as well

### DIFF
--- a/flow/connectors/mysql/validate.go
+++ b/flow/connectors/mysql/validate.go
@@ -143,7 +143,7 @@ func (c *MySqlConnector) CheckBinlogSettings(ctx context.Context, requireRowMeta
 		if binlogRetentionHours, err := strconv.Atoi(binlogRetentionHoursStr); err != nil {
 			return errors.New("failed to parse RDS/Aurora setting 'binlog retention hours': " + err.Error())
 		} else if binlogRetentionHours < 24 {
-			return errors.New("RDS/Aurora setting 'binlog retention hours' should be at least 24, currently " + fmt.Sprint(binlogRetentionHoursStr))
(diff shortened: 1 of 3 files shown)


flow/connectors/mysql/validate.go
```diff
@@ -117,10 +117,36 @@ func (c *MySqlConnector) CheckBinlogSettings(ctx context.Context, requireRowMeta
...
+			return errors.New("RDS/Aurora setting 'binlog retention hours' should be atleast 24, currently unset")
+		}
+		slog.Info("binlog retention hours", "binlogRetentionHours", binlogRetentionHoursStr)
+		if binlogRetentionHours, err := strconv.Atoi(binlogRetentionHoursStr); err != nil {
+			return errors.New("failed to parse RDS/Aurora setting 'binlog retention hours': " + err.Error())
+		} else if binlogRetentionHours < 24 {
+			return errors.New("RDS/Aurora setting 'binlog retention hours' should be at least 24, currently " + binlogRetentionHoursStr)
+		}
+	} else {
+		slog.Warn("binlog retention hours returned nothing, skipping Aurora/RDS binlog retention check")
```
serprex, line 149: I think this scenario means 0h retention

heavycrystal: in that case the query returns NULL which maps to empty string L139-L141

serprex (commented):
some e2e would be nice, can use API tests, can create `mysql.rds_configuration` table

serprex (approved):

PR #2778, opened 2025-03-31
be more precise about ignoring temporal application errors

serprex:
postgres connector errors with disconnect typed errors when it detects misconfiguration,
these errors should be propagated to alerting

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -276,7 +276,8 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	syncStartTime := time.Now()
 	if err := errGroup.Wait(); err != nil {
 		// don't log flow error for "replState changed" and "slot is already active"
-		if !(temporal.IsApplicationError(err) || shared.IsSQLStateError(err, pgerrcode.ObjectInUse)) {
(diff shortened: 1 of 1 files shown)


flow/activities/flowable_core.go
```diff
@@ -276,7 +276,8 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	syncStartTime := time.Now()
 	if err := errGroup.Wait(); err != nil {
 		// don't log flow error for "replState changed" and "slot is already active"
-		if !(temporal.IsApplicationError(err) || shared.IsSQLStateError(err, pgerrcode.ObjectInUse)) {
+		var applicationError *temporal.ApplicationError
+		if !((errors.As(err, &applicationError) && applicationError.Type() == "desync") || shared.IsSQLStateError(err, pgerrcode.ObjectInUse)) {
```
iamKunalGupta, line 280: shouldn't it be disconnect? ref https://github.com/PeerDB-io/peerdb/blob/0eac5fa55b7346d12213abd954778f3778db38ec/flow/connectors/postgres/postgres.go#L385-L395

serprex: this is checking for what to ignore, we want to ignore `desync` not `disconnect`

PR #2785, opened 2025-04-01
fix(maintenance): skip for not found workflows if older than 90 days

### LINKED ISSUE #2768: skip maintenance for not found workflows?
```json

{"message":"failed to query status in workflow with ID mirror_d1e3f532__8adb__4f79__9d00__01e44b6bcbfb-peerflow-27144d2c-06ce-4552-87e5-696b3a909702: workflow not found for ID: mirror_d1e3f532__8adb__4f79__9d00__01e44b6bcbfb-peerflow-27144d2c-06ce-4552-87e5-696b3a909702","source":"GoSDK","cause":{"message":"workflow not found for ID: mirror_d1e3f532__8adb__4f79__9d00__01e44b6bcbfb-peerflow-27144d2c-06ce-4552-87e5-696b3a909702","source":"GoSDK","applicationFailureInfo":{"type":"NotFound"}},"applicationFailureInfo":{"type":"wrapError"}}
```

This happens when an old DropFlow was unable to delete entry from catalog and even the original CDCFlow goes out of Temporal retention period

iamKunalGupta:
resolves #2768 

If workflow does not exist in temporal, it probably means that it was dropped long back and the cancelled CDCFlowWorkflow is beyond its retention period. This PR aims to guess this and skip pausing for such mirrors

### DIFF
--- a/flow/activities/maintenance_activity.go
+++ b/flow/activities/maintenance_activity.go
@@ -220,16 +220,17 @@ func (a *MaintenanceActivity) PauseMirrorIfRunning(ctx context.Context, mirror *
 				return false, wErr
 			}
 			logger.Info("Received response for ANY existing Workflows check", "len(executions)", len(response.Executions))
-			if len(response.Executions) != 0 {
-				foundWorkflowIds := make([]string, len(response.Executions))
-				for i, exec := range response.Executions {
-					logger.Info("Found existing CDCFlow", "workflowId", exec.GetExecution().GetWorkflowId())
-					foundWorkflowIds[i] = exec.GetExecution().GetWorkflowId()
-				}
-				logger.Warn("Found some existing CDCFlow, this is unexpected and should be investigated",
-					"foundWorkflows", foundWorkflowIds)
+			if len(response.Executions) == 0 {
+				logger.Warn("No existing workflows found, skipping pause")
 				return false, nil
 			}
+			foundWorkflowIds := make([]string, len(response.Executions))
+			for i, exec := range response.Executions {
+				logger.Info("Found existing CDCFlow", "workflowId", exec.GetExecution().GetWorkflowId())
+				foundWorkflowIds[i] = exec.GetExecution().GetWorkflowId()
+			}
+			logger.Warn("Found some existing CDCFlow, this is unexpected and should be investigated",
+				"foundWorkflows", foundWorkflowIds)
 		}
 		return false, err
 	}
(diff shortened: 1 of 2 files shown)


flow/activities/maintenance_activity.go
```diff
@@ -51,8 +52,10 @@ func (a *MaintenanceActivity) GetAllMirrors(ctx context.Context) (*protos.Mainte
 	maintenanceMirrorItems, err := pgx.CollectRows(rows, func(row pgx.CollectableRow) (*protos.MaintenanceMirror, error) {
 		var info protos.MaintenanceMirror
 		var createdAt time.Time
-		err := row.Scan(&info.MirrorId, &info.MirrorName, &info.WorkflowId, &createdAt, &info.IsCdc)
+		var updatedAt time.Time
+		err := row.Scan(&info.MirrorId, &info.MirrorName, &info.WorkflowId, &createdAt, &updatedAt, &info.IsCdc)
```
serprex, line 56:
```suggestion
		err := row.Scan(&info.MirrorId, &info.MirrorName, &info.WorkflowId, &info.MirrorCreatedAt, &info.MirrorUpdatedAt, &info.IsCdc)
```

flow/activities/maintenance_activity.go
```diff
@@ -201,6 +204,33 @@ func (a *MaintenanceActivity) PauseMirrorIfRunning(ctx context.Context, mirror *
 					return false, nil
 				}
 			}
+		} else if errors.As(err, &notFoundErr) && regexp.MustCompile("workflow not found for ID: (.+)").MatchString(notFoundErr.Message) &&
```
serprex, line 207: move regexp construction to global variable

flow/activities/maintenance_activity.go
iamKunalGupta: I'm not sure if it can be directly put to proto

serprex: It can, already does for a few fields

iamKunalGupta:
```
can't scan into dest[3] (col: created_at): cannot scan timestamp (OID 1114) in binary format into **timestamppb.Timestamp
```

serprex: oh, my eyes glazed over `timestamppb.New`

serprex (approved):

PR #2794, opened 2025-04-02
Support ClickHouse having extra columns compared to source for initial load

Amogh-Bharadwaj:
Currently to construct the column selectors for the Insert into Select to ClickHouse in initial load, we get the columns from ClickHouse. 
This results in initial load failing if there is an extra column on target vs source. 

The reason we were using the CH schema and not the source schema here was so that column exclusion was automatically taken care of. But this caused an unideal situation where we support source having extra columns for initial load and ClickHouse having extra columns for CDC normalize.

This PR re-wires things to make initial load use the source schema and gives it column exclusion info

### DIFF
--- a/flow/connectors/clickhouse/qrep.go
+++ b/flow/connectors/clickhouse/qrep.go
@@ -33,15 +33,11 @@ func (c *ClickHouseConnector) SyncQRepRecords(
 		slog.String("destinationTable", destTable),
 	)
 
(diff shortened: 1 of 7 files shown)


Amogh-Bharadwaj:
TODO: Add test with existing CH table with extra table

serprex (approved):

PR #2803, opened 2025-04-04
Draft: Do not review, add support for specific geometry types

panbhag-ch:


flow/connectors/mysql/qvalue_convert.go
```diff
@@ -154,8 +173,25 @@ func qkindFromMysqlColumnType(ct string) (qvalue.QValueKind, error) {
 		}
 	case "vector":
 		return qvalue.QValueKindArrayFloat32, nil
-	case "geometry", "point", "polygon", "linestring", "multipoint", "multipolygon", "geomcollection":
-		return qvalue.QValueKindGeometry, nil
+	case "geometry", "point", "linestring", "polygon", "multipoint", "multilinestring", "multipolygon", "geometrycollection":
+		switch ct {
```
iamKunalGupta, line 177: Earlier check is on `strings.ToLower(ct)` this check is on `ct`,
maybe we should instead merge with parent switch?

PR #2806, opened 2025-04-07
Formalize how we handle table inheritance for partitioned tables

heavycrystal:
Most users of partitioned tables we've encountered so far use Postgres "native" partitioning, in which the parent table is marked with a special relation kind (`relkind`) of `p`. This is handled best by setting `publish_via_partition_root` on the publication (we create publications like this, users are encouraged to) but as a fallback for pre PG13 we also support "lookup" of child to parent tables and using that information to manually implement child-to-parent table mapping. We also recently made this lookup dynamic, so new partitions encountered after we start CDC should also be handled.

There is also the possibility of child tables for a regular table, either via extensions doing this or because Postgres supports [table inheritance](https://www.postgresql.org/docs/current/ddl-inherit.html). Here the feature of child-to-parent mapping may not be desired since there is a chance of the "child" tables having a different schema and also mapping all partitions to the same parent table may not be desired. The remapping for regular tables was the default behaviour since https://github.com/PeerDB-io/peerdb/pull/2323 but now we're making this opt-out, toggled via a dynamic configuration setting. It is opt-out because the case of extensions using it for partitioning is more prevalent than being used for table inheritance, which is a bit of a niche feature.

In all cases initial load will move data from all the child tables/partitions of a table, because `SELECT` by default selects these. There is syntax for only selecting from the parent table (`SELECT ... FROM ONLY t1`) but the use for this is limited.

Also adding e2e tests for these various cases.

### DIFF
--- a/flow/e2e/generic/generic_test.go
+++ b/flow/e2e/generic/generic_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
@@ -500,8 +501,8 @@ func (s Generic) Test_Partitioned_Table_Without_Publish_Via_Partition_Root() {
 	}
 	conn := pgSource.PostgresConnector
 
-	srcTable := "test_partition"
-	dstTable := "test_partition_dst"
+	srcTable := "test_partition_noroot"
+	dstTable := "test_partition_noroot_dst"
 	srcSchemaTable := e2e.AttachSchema(s, srcTable)
 	srcPublicationName := fmt.Sprintf("%s_%s_pub", srcTable, s.Suffix())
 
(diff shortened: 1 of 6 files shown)


flow/e2e/generic/generic_test.go
```diff
@@ -489,3 +489,64 @@ func (s Generic) Test_Schema_Changes_Cutoff_Bug() {
 	env.Cancel(t.Context())
 	e2e.RequireEnvCanceled(t, env)
 }
+
+func (s Generic) Test_Partitioned_Table_Without_Publish_Via_Partition_Root() {
```
serprex, line 493: should test inherited table with & without setting

heavycrystal: working on it

serprex (approved):

PR #2819, opened 2025-04-07
chore: update dependencies

serprex:
adjust versioning rules so package.json doesn't get stuck on versions,
we'll trust packages to respect semver & lockfile to give us reproducibility


iamKunalGupta (approved):
> we'll trust packages to respect semver

let's see

PR #2828, opened 2025-04-09
ClickHouse normalize: Move fetching of number of parts inside table-wise loop

Amogh-Bharadwaj:
Currently, we fetch the dynamic config setting for normalization number of parts before we start iterating on the table INSERTs in Normalize in ClickHouse.
Practical utility has been seen in moving this inside the table-wise loop - because the change in the setting would be reflected from the next table onwards, instead of having to wait for the activity as a whole to retry.

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -292,12 +292,6 @@ func (c *ClickHouseConnector) NormalizeRecords(
 		slog.Int64("EndBatchID", req.SyncBatchID),
 		slog.Int("connections", parallelNormalize))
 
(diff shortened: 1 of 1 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -343,6 +335,12 @@ func (c *ClickHouseConnector) NormalizeRecords(
 	}
 
 	for _, tbl := range destinationTableNames {
+		numParts, err := internal.PeerDBClickHouseNormalizationParts(ctx, req.Env)
```
serprex, line 338: this is likely getting cached by dynconf code, need to adjust speed of setting

PR #2836, opened 2025-04-10
drop/resync: separate code path for completed workflows

serprex:


flow/connectors/mysql/validate.go
```diff
@@ -139,6 +139,8 @@ func (c *MySqlConnector) CheckBinlogSettings(ctx context.Context, requireRowMeta
 }
 
 func (c *MySqlConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.FlowConnectionConfigs) error {
+	return nil
```
heavycrystal, line 142: is this TODO?

heavycrystal (approved):

flow/connectors/mysql/validate.go
serprex: I screwed up & leaked code from testing earlier

PR #2849, opened 2025-04-14
feat(metrics): active flows and resource limits metrics

iamKunalGupta:


flow/activities/flowable.go
```diff
@@ -720,16 +721,23 @@ func (a *FlowableActivity) SendWALHeartbeat(ctx context.Context) error {
...
 	}
 
-	configs, err := pgx.CollectRows(rows, func(row pgx.CollectableRow) (*protos.FlowConnectionConfigs, error) {
+	infos, err := pgx.CollectRows(rows, func(row pgx.CollectableRow) (*flowInformation, error) {
 		var flowName string
 		var configProto []byte
-		err := rows.Scan(&flowName, &configProto)
+		var workflowID string
+		err := rows.Scan(&flowName, &configProto, &workflowID)
 		if err != nil {
```
serprex, line 741:
```suggestion
		if err := rows.Scan(&flowName, &configProto, &workflowID); err != nil {
```

flow/activities/flowable.go
```diff
@@ -774,58 +785,116 @@ func (a *FlowableActivity) RecordSlotSizes(ctx context.Context) error {
...
 				logger.Error("Failed to handle slot info", slog.Any("error", err))
 			}
-		}()
+		}(ctx)
+	}
+	if a.OtelManager != nil {
+		operationalFlows := filterOperationalFlowStatuses(infos)
+		operationFlowCount := len(operationalFlows)
+		a.OtelManager.Metrics.ActiveFlowsGauge.Record(ctx, int64(operationFlowCount))
+		for _, info := range operationalFlows {
```
serprex, line 845: why are we looping over operationalFlows here? we aren't using `info` outside of `a.GetFlowMetadata`

iamKunalGupta: the metric needs to be only emitted for operation/active flows.
basically `cpu_limit/active_flows` for every active flow

serprex: You can lift computing the metric values out of the loop

iamKunalGupta: make sense

iamKunalGupta: I'm also debating whether I should pull the data from cgroups or just use simple env vars which are injected at runtime via helm chart

iamKunalGupta: I went ahead with injected env

PR #2860, opened 2025-04-15
ci: try use eventhubs emulator

serprex:


serprex:
emulator being a pain fighting with redpanda over port 9092, annoying since we don't care about their kafka translation layer

closing in favor of #2862

PR #2872, opened 2025-04-18
feat(otel): run otel-col in tests

iamKunalGupta:
the main exporter can also now be run in tests to make sure there are no issues at runtime

### DIFF
--- a/flow/internal/env.go
+++ b/flow/internal/env.go
@@ -49,16 +49,7 @@ func GetEnvString(name string, defaultValue string) string {
 }
 
 func GetEnvBool(name string, defaultValue bool) bool {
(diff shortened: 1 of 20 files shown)


flow/internal/env.go
```diff
@@ -48,6 +48,19 @@ func GetEnvString(name string, defaultValue string) string {
 	return val
 }
 
+func GetEnvBool(name string, defaultValue bool) bool {
+	val, ok := os.LookupEnv(name)
+	if !ok {
+		return defaultValue
+	}
+
+	ret, err := strconv.ParseBool(val)
```
serprex, line 57: can use getEnvConvert

serprex (approved):

PR #2882, opened 2025-04-22
coverage, enforce subdirectory coverage

serprex:


flow/e2eshared/e2eshared.go
```diff
@@ -76,12 +76,7 @@ func ReadFileToBytes(path string) ([]byte, error) {
 	}
 	defer f.Close()
 
-	ret, err := io.ReadAll(f)
-	if err != nil {
-		return ret, fmt.Errorf("failed to read file: %w", err)
-	}
-
-	return ret, nil
+	return io.ReadAll(f)
```
heavycrystal, line 79: this pattern is so the error is "tagged" for logs, if all callsites of this function do it then this should be fine to change

flow/e2eshared/e2eshared_test.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/model/qvalue"
 )
 
+func TestReadMissingFileToBytes(t *testing.T) {
+	if _, err := ReadFileToBytes("missing"); err == nil {
+		t.Error("ReadFileToBytes expected to fail on missing files")
```
heavycrystal, line 12: is it possible to assert on the exact error? can dump the function into a playground and check

serprex: changed to check for `os.ErrNotExists`

heavycrystal (approved):

PR #2891, opened 2025-04-24
add field to override hostname for tls

serprex:
relevant when using a proxy endpoint which would otherwise have wrong hostname


iamKunalGupta:
This will also help for IAM Auth when using RDS Private Link as the actual DB hostname is required to generate the DB Token

iamKunalGupta (approved):

heavycrystal (approved):

PR #2893, opened 2025-04-25
Spin up individual workflow job for each table

alexstoick:
We're looking to sync a fairly large number of tables (~50k) and adding them in batches. The table creation is done sequentially and this ends up taking a very large amount of time. Additionally, it seems that everytime a new table is added the presence of _all_ tables is checked - something that can be optimised.

Let me know if the suggestion makes sense.

Thank you!

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -232,10 +232,7 @@ func processTableAdditions(
 		if err := childAdditionalTablesCDCFlowFuture.Get(childAdditionalTablesCDCFlowCtx, &res); err != nil {
 			return err
 		}
-		// gotta accumulate the results of the map
-		for xId, xStr := range res.SyncFlowOptions.SrcTableIdNameMapping {
-			accumlatedSrcTableIdNameMapping[xId] = xStr
-		}
+		maps.Copy(accumlatedSrcTableIdNameMapping, res.SyncFlowOptions.SrcTableIdNameMapping)
 	}
 
 	maps.Copy(state.SyncFlowOptions.SrcTableIdNameMapping, accumlatedSrcTableIdNameMapping)


flow/workflows/cdc_flow.go
```diff
@@ -201,36 +201,44 @@ func processTableAdditions(
...
+			nil,
+		)
+		var res *CDCFlowWorkflowResult
+		if err := childAdditionalTablesCDCFlowFuture.Get(childAdditionalTablesCDCFlowCtx, &res); err != nil {
+			return err
+		}
+		// gotta accumulate the results of the map
+		for xId, xStr := range res.SyncFlowOptions.SrcTableIdNameMapping {
+			accumlatedSrcTableIdNameMapping[xId] = xStr
+		}
```
serprex, line 238:
```suggestion
		maps.Copy(accumlatedSrcTableIdNameMapping, res.SyncFlowOptions.SrcTableIdNameMapping)
```
https://pkg.go.dev/maps#Copy

flow/workflows/cdc_flow.go
```diff
@@ -201,36 +201,44 @@ func processTableAdditions(
...
+			WorkflowID:        childAdditionalTableCDCFlowID,
+			ParentClosePolicy: enums.PARENT_CLOSE_POLICY_REQUEST_CANCEL,
+			RetryPolicy: &temporal.RetryPolicy{
+				MaximumAttempts: 20,
+			},
+			TypedSearchAttributes: mirrorNameSearch,
+			WaitForCancellation:   true,
+		}
+		childAdditionalTablesCDCFlowCtx := workflow.WithChildOptions(ctx, childAdditionalTablesCDCFlowOpts)
+		childAdditionalTablesCDCFlowFuture := workflow.ExecuteChildWorkflow(
```
serprex, line 225: not sure, does this have to be done serially? ie we wouldn't be able to run these in parallel with `BoundSelector`?

flow/workflows/cdc_flow.go
```diff
@@ -201,36 +201,41 @@ func processTableAdditions(
...
-		additionalTablesCfg,
-		nil,
-	)
-	var res *CDCFlowWorkflowResult
-	if err := childAdditionalTablesCDCFlowFuture.Get(childAdditionalTablesCDCFlowCtx, &res); err != nil {
-		return err
+	var accumlatedSrcTableIdNameMapping = make(map[uint32]string)
+
+	for _, table := range flowConfigUpdate.AdditionalTables {
+		additionalTableUUID := uuid.New().String()
```
serprex, line 207:
```suggestion
		additionalTableUUID := GetUUID(ctx)
```
`uuid.New()` is a side effect, direct use in workflow causes non-determinism

PR #2901, opened 2025-04-29
fix(codecov): status checks

iamKunalGupta:


codecov.yml
```diff
@@ -21,7 +21,6 @@ component_management:
   individual_components:
     - component_id: full
       paths:
-        - flow/datatypes/
-        - flow/e2eshared/
+        - flow/**
```
serprex, line 24: I tried splitting this out for segments which have 100% coverage (& those I'd like to have enforced)

PR #2906, opened 2025-05-05
Classify missing table on PG

Amogh-Bharadwaj:
If a Postgres table is dropped, then we get a `no rows in result set` error when a schema change (RelationMessage) is encountered.
This PR catches this specific error case and classifies it


serprex (approved):
@iamKunalGupta do unknown classifications get passed through as an error to slack without sending out notifications?

heavycrystal (approved):

flow/connectors/postgres/client.go
```diff
@@ -110,6 +108,9 @@ func (c *PostgresConnector) getRelIDForTable(ctx context.Context, schemaTable *u
 		 ON n.oid = c.relnamespace WHERE n.nspname=$1 AND c.relname=$2`,
 		schemaTable.Schema, schemaTable.Table).Scan(&relID)
 	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return 0, shared.ErrTableDoesNotExist
```
iamKunalGupta, line 112: do we want to include table name in the error information?
Also maybe we should wrap the original error as well

iamKunalGupta (approved):

iamKunalGupta:
> @iamKunalGupta do unknown classifications get passed through as an error to slack without sending out notifications?

Yes, anything unknown is auto sent to us, without sending out to customer

PR #2909, opened 2025-05-05
mysql: qrep partitioning

serprex:
mysql only has ? parameters, which interacts poorly with qrep templates

instead of using parameters, interpolate partition values into template

### DIFF
--- a/flow/connectors/mysql/qrep.go
+++ b/flow/connectors/mysql/qrep.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log/slog"
 	"math"
+	"strconv"
 	"text/template"
 
 	"github.com/go-mysql-org/go-mysql/mysql"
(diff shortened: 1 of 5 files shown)


flow/connectors/mysql/qrep.go
```diff
@@ -224,26 +218,59 @@ func (c *MySqlConnector) PullQRepRecords(
...
+						rangeEnd = strconv.FormatUint(uint64(uint16(x.IntRange.End)), 10)
+						unsigned = true
+					case qvalue.QValueKindUInt32:
+						rangeStart = strconv.FormatUint(uint64(uint32(x.IntRange.Start)), 10)
+						rangeEnd = strconv.FormatUint(uint64(uint32(x.IntRange.End)), 10)
+						unsigned = true
+					case qvalue.QValueKindUInt64:
+						rangeStart = strconv.FormatUint(uint64(x.IntRange.Start), 10)
+						rangeEnd = strconv.FormatUint(uint64(x.IntRange.End), 10)
+						unsigned = true
```
heavycrystal, line 250: default should error?

serprex: default is signed integers, we let existing logic handle that same as postgres

heavycrystal (approved):

PR #2924, opened 2025-05-09
[mysql] general fixes

heavycrystal:
1. MySQL connections were setting compression unconditionally, respect setting instead
2. remove unused Snowflake SQL
3. remove binlog/replication validation from peer creation so we can do initial load only mirrors for MySQL
4. make binlog/replication validation during mirror creation conditional on initial load only
5. change double quotes to backticks for statements we run on MySQL during initial load, brings us closer to supporting Vitess which can't handle ANSI_QUOTES


scriptforfivem:
Great implementation of this feature. I particularly like the approach to error handling.

serprex (approved):

PR #2926, opened 2025-05-12
Logs tab in UI: fix extractFromCloneName

Amogh-Bharadwaj:
In order to make it more apparent that a log from a "clone_" mirror is actually an initial load of its parent, in the UI logs tab we do regex parsing to get the parent name and say (initial load).
However the regex parsing doesn't work if the parent name has underscores. This PR fixes this via a stopgap regex fix.
Long term would want to introduce and populate a parent name column in the logs table similar to the qrep stats tables


ui/components/LogsTable.tsx
```diff
@@ -22,8 +22,11 @@ function colorForErrorType(errorType: string) {
 }
 
 function extractFromCloneName(mirrorOrCloneName: string) {
-  if (mirrorOrCloneName.includes('clone_')) {
-    return mirrorOrCloneName.split('_')[1] + ' (initial load)';
+  if (mirrorOrCloneName.startsWith('clone_')) {
+    return (
+      mirrorOrCloneName.substring(mirrorOrCloneName.indexOf('_') + 1) +
```
serprex, line 27: indexOf will always be 5

serprex (approved):

PR #2928, opened 2025-05-12
mysql: support full range of time

serprex:
> TIME values may range from '-838:59:59' to '838:59:59'

https://dev.mysql.com/doc/refman/8.4/en/time.html

### DIFF
--- a/flow/connectors/mysql/qvalue_convert.go
+++ b/flow/connectors/mysql/qvalue_convert.go
@@ -223,8 +223,82 @@ func processGeometryData(data []byte) qvalue.QValueGeometry {
 			return qvalue.QValueGeometry{Val: wkt}
 		}
 	}
-	strVal := string(data)
-	return qvalue.QValueGeometry{Val: strVal}
+	return qvalue.QValueGeometry{Val: string(data)}
+}
+
+// https://dev.mysql.com/doc/refman/8.4/en/time.html
+func processTime(str string) (time.Time, error) {
+	abs, isNeg := strings.CutPrefix(str, "-")
+	tpart, frac, _ := strings.Cut(abs, ".")
+
+	var nsec uint64
+	if frac != "" {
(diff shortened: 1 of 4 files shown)


Amogh-Bharadwaj (approved):

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -223,8 +223,81 @@ func processGeometryData(data []byte) qvalue.QValueGeometry {
...
+	abs, isNeg := strings.CutPrefix(str, "-")
+	tpart, frac, _ := strings.Cut(abs, ".")
+
+	var nsec uint64
+	if len(frac) > 0 {
+		fint, err := strconv.ParseUint(frac, 10, 64)
+		if err != nil {
+			return time.Time{}, err
+		}
+		switch len(frac) {
```
heavycrystal, line 240: how about something like?

```
if len(frac) <= 9 {
    nsec = fint * int(math.Pow10(9 - len(frac)))
} else {
    nsec = fint
}
```

serprex: I was going to complain about `math.Pow` perf overhead, but now I see `math.Pow10` is good

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -237,24 +237,9 @@ func processTime(str string) (time.Time, error) {
...
-			nsec = fint * 1000
-		case 7:
-			nsec = fint * 100
-		case 8:
-			nsec = fint * 10
-		default:
+		if len(frac) <= 9 {
+			nsec = fint * uint64(math.Pow10(9-len(frac)))
+		} else {
 			nsec = fint
```
serprex, line 243: really this case is an error, but we don't expect to ever get more than 6 digits anyways

heavycrystal (approved):

PR #2939, opened 2025-05-14
feat(wf-status): raise error if flow not found in catalog

iamKunalGupta:
During maintenance if snapshot flow is dropped, it can sometimes not update the temporal status.
This leads to missing flow entry in catalog once drop succeeds and GetWorkflowStatus falls back to temporal which shows stale status (snapshot).
If instead we get `pgx.ErrNoRows` we can raise the error further and assume there is a bigger issue at hand
 - in this case, maintenance would retry the activity and next fetch should not have the mirror in the list at all

### DIFF
--- a/flow/internal/workflow.go
+++ b/flow/internal/workflow.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 
+	"github.com/jackc/pgx/v5"
 	"go.temporal.io/sdk/client"
 
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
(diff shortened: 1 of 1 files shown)


serprex (approved):

flow/internal/workflow.go
```diff
@@ -33,6 +34,11 @@ func GetWorkflowStatus(ctx context.Context, pool shared.CatalogPool,
 	err := pool.QueryRow(ctx, "SELECT status FROM flows WHERE workflow_id = $1", workflowID).Scan(&flowStatus)
 	if err != nil || flowStatus == protos.FlowStatus_STATUS_UNKNOWN {
 		if err != nil {
+			if errors.Is(err, pgx.ErrNoRows) {
+				slog.Warn("workflowId not found in catalog, will raise an error",
+					slog.String("workflowId", workflowID))
+				return protos.FlowStatus_STATUS_UNKNOWN, fmt.Errorf("workflowId not found in catalog: %w", err)
```
heavycrystal, line 40: If only maintenance can hit this I think it's fine to return `STATUS_RUNNING` and log a warning; this can lead to multiple alerts at once and I think we can investigate it without blocking maintenance

PR #2945, opened 2025-05-19
[dropflow] move metadata_last_sync_state cleanup to last activity

heavycrystal:
This is to avoid race conditions where `metadata_last_sync_state` is removed by `DropFlowDestination` BEFORE `DropFlowSource` is able to cancel the goroutine that may insert data into the table.

If we hit this race condition during a resync, batchIDs and offsets being from an older mirror can cause many subtle issues and potential data inconsistency

### DIFF
--- a/flow/connectors/bigquery/bigquery.go
+++ b/flow/connectors/bigquery/bigquery.go
@@ -740,12 +740,11 @@ func (c *BigQueryConnector) SyncFlowCleanup(ctx context.Context, jobName string)
 	dataset := c.client.DatasetInProject(c.projectID, c.datasetID)
 	rawTableHandle := dataset.Table(c.getRawTableName(jobName))
 	// check if exists, then delete
-	_, err := rawTableHandle.Metadata(ctx)
-	if err == nil {
-		c.logger.Info("deleting raw table", slog.String("table", rawTableHandle.FullyQualifiedName()))
+	if _, err := rawTableHandle.Metadata(ctx); err == nil {
+		c.logger.Info("[bigquery] deleting raw table", slog.String("table", rawTableHandle.FullyQualifiedName()))
 		deleteErr := rawTableHandle.Delete(ctx)
 		if deleteErr != nil {
-			return fmt.Errorf("failed to delete raw table: %w", deleteErr)
+			return fmt.Errorf("[bigquery] failed to delete raw table: %w", deleteErr)
 		}
 	}
 
--- a/flow/connectors/bigquery/bigquery.go
+++ b/flow/connectors/bigquery/bigquery.go
@@ -737,15 +737,10 @@ func (c *BigQueryConnector) SetupNormalizedTable(
 }
 
 func (c *BigQueryConnector) SyncFlowCleanup(ctx context.Context, jobName string) error {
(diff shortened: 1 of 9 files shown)


flow/connectors/bigquery/bigquery.go
```diff
@@ -737,15 +737,10 @@ func (c *BigQueryConnector) SetupNormalizedTable(
...
-	if err != nil {
-		return fmt.Errorf("unable to clear metadata for sync flow cleanup: %w", err)
-	}
-
 	dataset := c.client.DatasetInProject(c.projectID, c.datasetID)
 	rawTableHandle := dataset.Table(c.getRawTableName(jobName))
 	// check if exists, then delete
-	_, err = rawTableHandle.Metadata(ctx)
+	_, err := rawTableHandle.Metadata(ctx)
 	if err == nil {
```
serprex, line 744:
```suggestion
	if _, err := rawTableHandle.Metadata(ctx); err == nil {
```

serprex (approved):

flow/connectors/snowflake/snowflake.go
```diff
@@ -615,14 +615,10 @@ func (c *SnowflakeConnector) CreateRawTable(ctx context.Context, req *protos.Cre
...
-	err := c.PostgresMetadata.SyncFlowCleanup(ctx, jobName)
-	if err != nil {
-		return fmt.Errorf("[snowflake drop mirror] unable to clear metadata for sync flow cleanup: %w", err)
-	}
 
 	// delete raw table if exists
 	rawTableIdentifier := getRawTableIdentifier(jobName)
-	_, err = c.execWithLogging(ctx, fmt.Sprintf(dropTableIfExistsSQL, c.rawSchema, rawTableIdentifier))
+	_, err := c.execWithLogging(ctx, fmt.Sprintf(dropTableIfExistsSQL, c.rawSchema, rawTableIdentifier))
 	if err != nil {
```
serprex, line 622:
```suggestion
	if _, err := c.execWithLogging(ctx, fmt.Sprintf(dropTableIfExistsSQL, c.rawSchema, rawTableIdentifier)); err != nil {
```

flow/connectors/external_metadata/store.go
```diff
@@ -215,10 +215,14 @@ func (p *PostgresMetadata) IsQRepPartitionSynced(ctx context.Context, req *proto
...
+	`, lastSyncStateTableName)
+}
+
 func (p *PostgresMetadata) SyncFlowCleanup(ctx context.Context, jobName string) error {
-	if _, err := p.pool.Exec(ctx,
-		`DELETE FROM `+lastSyncStateTableName+` WHERE job_name = $1`, jobName,
-	); err != nil {
+	if _, err := p.pool.Exec(ctx, GetSyncFlowCleanupQuery(), jobName); err != nil {
 		return err
 	}
```
iamKunalGupta, line 227: Does this affect pg as a destination?

iamKunalGupta (approved):

flow/connectors/external_metadata/store.go
heavycrystal: shouldn't, will be a noop

PR #2949, opened 2025-05-19
engines: ReplicatedReplacingMergeTree, ReplicatedMergeTree

serprex:


flow/connectors/clickhouse/normalize.go
```diff
@@ -159,14 +159,21 @@ func generateCreateTableSQLForNormalizedTable(
...
+	case protos.TableEngine_CH_ENGINE_REPLACING_MERGE_TREE:
 		engine = fmt.Sprintf("ReplacingMergeTree(%s)", peerdb_clickhouse.QuoteIdentifier(versionColName))
-	} else if tableMapping.Engine == protos.TableEngine_CH_ENGINE_MERGE_TREE {
+	case protos.TableEngine_CH_ENGINE_MERGE_TREE:
 		engine = "MergeTree()"
-	} else if tableMapping.Engine == protos.TableEngine_CH_ENGINE_NULL {
+	case protos.TableEngine_CH_ENGINE_REPLICATED_REPLACING_MERGE_TREE:
+		engine = fmt.Sprintf("ReplicatedReplacingMergeTree('/clickhouse/tables/{shard}/{database}/{table}','{replica}',%s)", peerdb_clickhouse.QuoteIdentifier(versionColName))
+	case protos.TableEngine_CH_ENGINE_REPLICATED_MERGE_TREE:
+		engine = fmt.Sprintf("ReplicatedMergeTree('/clickhouse/tables/{shard}/{database}/{table}','{replica}',%s)", peerdb_clickhouse.QuoteIdentifier(versionColName))
```
serprex, line 174: ideally we'd allow configuring these parameters

ui/app/mirrors/create/cdc/schemabox.tsx
```diff
@@ -245,6 +245,11 @@ export default function SchemaBox({
   const engineOptions = [
     { value: 'CH_ENGINE_REPLACING_MERGE_TREE', label: 'ReplacingMergeTree' },
     { value: 'CH_ENGINE_MERGE_TREE', label: 'MergeTree' },
+    {
+      value: 'CH_ENGINE_REPLICATED_REPLACING_MERGE_TREE',
+      label: 'ReplicatedReplacingMergeTree',
+    },
+    { value: 'CH_ENGINE_REPLICATED_MERGE_TREE', label: 'ReplicatedMergeTree' },
```
iamKunalGupta, line 252: I'm thinking whether we should add a tooltip for new users.
Maybe simply saying that works with ch cloud and what works with oss clusters

serprex: mirror validation could check for host being CHC

heavycrystal: should be possible with `CheckIfClickHouseCloudHasSharedMergeTreeEnabled`

serprex: That only runs when we're in clickpipes. Validation here is for OSS peerdb users with CHC destination

heavycrystal: yes I mean we can use that function outside of current codepath to get if a service is SMT which should mean we disable these engines

flow/connectors/clickhouse/normalize.go
```diff
@@ -159,14 +159,27 @@ func generateCreateTableSQLForNormalizedTable(
...
 		engine = "MergeTree()"
-	} else if tableMapping.Engine == protos.TableEngine_CH_ENGINE_NULL {
+	case protos.TableEngine_CH_ENGINE_REPLICATED_REPLACING_MERGE_TREE:
+		engine = fmt.Sprintf(
+			"ReplicatedReplacingMergeTree('/clickhouse/tables/{shard}/{database}/{table}','{replica}',%s)",
+			peerdb_clickhouse.QuoteIdentifier(versionColName),
+		)
+	case protos.TableEngine_CH_ENGINE_REPLICATED_MERGE_TREE:
+		engine = fmt.Sprintf(
+			"ReplicatedMergeTree('/clickhouse/tables/{shard}/{database}/{table}','{replica}',%s)",
```
serprex, line 178: Table name should be interpolated rather than template, as this has issues with rename which resync will run into

How this works with resync in general needs to be evaluated

ui/app/mirrors/create/cdc/schemabox.tsx
serprex: I think for now we can omit this, people using CHC should know this engine doesn't apply to them

PR #2951, opened 2025-05-19
sync flow: explicitly cancel & wait on cancellation before finishing

serprex:
this avoids activity living on past ContinueAsNew, which was causing races


iskakaushik (commented):
this looks OK to me, @serprex  would you mind adding some logging, putting a sleep (for testing) in SyncFlow and ensure the behavior matches what we want?

heavycrystal (approved):

Amogh-Bharadwaj (approved):

PR #2956, opened 2025-05-21
Extend API test for resync

Amogh-Bharadwaj:
## Overview
This PR extends the existing (now called) `TestEditTablesBeforeResync` test in `api_test.go`.
This test is meant to cover editing tables and resyncing after these table edits to see how the mirror behaves.

## Steps

The test can be divided into - setup, add table, remove table and resync. Relevant metadata checks are being done for the table edits.

### 1. Setup tables and running mirror
1. Create the first table called `original`.
2. Create the second table called `added`.
3. Insert a row into `original`.
4. Insert a row into `added`.
5. Create a mirror with only `original` in the table mapping.
6. Wait for initial load of this mirror with `original` to complete.
7. Check if equal tables for `original`.

### 2. Table addition
8. Pause the mirror.
9. Send a flow state change request to add the table - `added`.
10. Wait for `added` table's initial load to complete.
11. Wait for equal tables for `added`.
12. Check the table mapping in `config_proto` for this flow in catalog and verify that it has `original` and `added` and mirror is back to Running.
13. Check if initial load stats has 2 entries.
14. Insert a row into `added` in source.
15. Check if sync_batch_id and normalize_batch_id in `metadata_last_sync_state` table in catalog are set to expected values : 1 for both.

### 3. Table removal
16. Pause the mirror.
17. Send a flow state change request to remove the table - `original`.
18. Check if catalog's table mapping now only has the `added` table and mirror is back to Running.
19. In case of source peer being Postgres, check if the replication slot is active.

### 4. Resync
20. Pause the mirror.
21. Insert a row into the `added` table.
22. Send a flow state change request for resync.
23. Check if equal tables for `added`.
24. Check if mirror is back to Running.
25. Check if `InitialLoadSummary` which we call in UI - returns only the `added` table's initial load data.
26. Insert a row into `added`.
27. Check if sync_batch_id and normalize_batch_id in `metadata_last_sync_state` table in catalog are set to expected values : 1 for both.
28. Wait for equal tables for added.
29. Cancel the context.

### DIFF
--- a/flow/e2e/api/api_test.go
+++ b/flow/e2e/api/api_test.go
@@ -2,6 +2,7 @@ package e2e_api
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"slices"
 	"strings"
@@ -15,6 +16,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
 
 	connpostgres "github.com/PeerDB-io/peerdb/flow/connectors/postgres"
 	"github.com/PeerDB-io/peerdb/flow/e2e"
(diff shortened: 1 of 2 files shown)


flow/e2e/api/api_test.go
```diff
@@ -500,21 +582,77 @@ func (s Suite) TestAddTableBeforeResync() {
...
+		return valid && env.GetFlowStatus(s.t) == protos.FlowStatus_STATUS_RUNNING
 	})
 	e2e.RequireEqualTables(s.ch, "added", "id,val")
 
+	// Check initial load stats
+	initialLoadClientFacingDataAfterAddTable, err := s.InitialLoadSummary(s.t.Context(), &protos.InitialLoadSummaryRequest{
+		ParentMirrorName: flowConnConfig.FlowJobName,
+	})
+	require.NoError(s.t, err)
+	require.True(s.t, len(initialLoadClientFacingDataAfterAddTable.TableSummaries) == 2)
```
serprex, line 603: `require.Len`

flow/e2e/api/api_test.go
```diff
@@ -66,6 +69,85 @@ func (s Suite) DestinationTable(table string) string {
...
+	}
+
+	if len(config.TableMappings) != len(expectedSourceTableNames) {
+		return false, fmt.Errorf("expected %d table mappings, got %d", len(expectedSourceTableNames), len(config.TableMappings))
+	}
+	sourceTableNames := make([]string, len(config.TableMappings))
+	for i, tableMapping := range config.TableMappings {
+		sourceTableNames[i] = tableMapping.SourceTableIdentifier
+	}
+	return assert.ElementsMatch(s.t, expectedSourceTableNames, sourceTableNames), nil
```
serprex, line 148: this fails the test if it ever returns false

serprex (approved):

PR #2958, opened 2025-05-21
MySQL Mirror: set max execution time

Amogh-Bharadwaj:
MySQL has a setting - `max_execution_time` - after which queries are cancelled, including ours

This PR sets this setting to unlimited at a session level when connecting to MySQL

### DIFF
--- a/flow/connectors/mysql/mysql.go
+++ b/flow/connectors/mysql/mysql.go
@@ -182,6 +182,10 @@ func (c *MySqlConnector) connect(ctx context.Context) (*client.Conn, error) {
 		if _, err := conn.Execute("SET sql_mode = 'ANSI,NO_BACKSLASH_ESCAPES'"); err != nil {
 			return nil, fmt.Errorf("failed to set sql_mode to ANSI: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


flow/connectors/mysql/mysql.go
```diff
@@ -182,6 +182,10 @@ func (c *MySqlConnector) connect(ctx context.Context) (*client.Conn, error) {
 		if _, err := conn.Execute("SET sql_mode = 'ANSI,NO_BACKSLASH_ESCAPES'"); err != nil {
 			return nil, fmt.Errorf("failed to set sql_mode to ANSI: %w", err)
 		}
+		// set max_execution_time to unlimited
+		if _, err := conn.Execute("SET SESSION MAX_EXECUTION_TIME=0;"); err != nil {
```
heavycrystal, line 186: another one of those things where MySQL and Maria differ, they call it `max_statement_time`
Also older MySQL does not have this: https://github.com/sixhop/AutoMySQLBackup/issues/31

heavycrystal (approved):

PR #2960, opened 2025-05-21
Remove sync resync & drop api

serprex:


flow/cmd/handler.go
```diff
@@ -396,15 +393,13 @@ func (h *FlowRequestHandler) FlowStateChange(
...
-			} else {
-				changeErr = model.FlowSignalStateChange.SignalClientWorkflow(ctx, h.temporalClient, workflowID, "", req)
-			}
-		case protos.FlowStatus_STATUS_TERMINATED: // backwards compat, causes grpc timeouts
-			if currState != protos.FlowStatus_STATUS_TERMINATED {
-				changeErr = h.shutdownFlow(ctx, req.FlowJobName, req.DropMirrorStats, req.SkipDestinationDrop)
+		case protos.FlowStatus_STATUS_TERMINATING, protos.FlowStatus_STATUS_TERMINATED:
+			if currState != protos.FlowStatus_STATUS_TERMINATED && currState != protos.FlowStatus_STATUS_TERMINATING {
+				if currState == protos.FlowStatus_STATUS_COMPLETED {
+					changeErr = h.shutdownFlow(ctx, req.FlowJobName, req.DropMirrorStats, req.SkipDestinationDrop)
```
heavycrystal, line 399: why are we doing this? shouldn't we only need to delete from catalog?

heavycrystal (approved):

flow/cmd/handler.go
serprex: old code path for completed workflows deletes records synchronously since signals can't be used in this case

PR #2964, opened 2025-05-21
[WIP] Support numeric to string via destination_type in API

jgao54:
Work-in-progress, not ready for review (created PR to better understand CI flow)

Support `numeric` -> `String` conversion when  destination_type field is passed in to `POST api/v1/flows/cdc/create` API.


flow/connectors/clickhouse/qrep_avro_sync.go
```diff
@@ -255,3 +255,24 @@ func (s *ClickHouseAvroSyncMethod) writeToAvroFile(
...
+	for _, col := range columnSettings {
+		if col.SourceName == columnName {
+			destinationType = col.DestinationType
+		}
+	}
+
+	// TODO: check destination type is different from source type
+	switch destinationType {
+	case "String", "Nullable(String)":
+		return fmt.Sprintf("CAST(%s, '%s')", avroColumnName, destinationType)
```
serprex, line 274:
```suggestion
		return fmt.Sprintf("CAST(`%s`, '%s')", avroColumnName, EscapeStr(destinationType))
```

jgao54: oops good catch

ilidemi: Can column names contain backticks as well?

jgao54: hmmm yeah interesting! looks like it's possible in postgres with double quotes[[ref](https://www.postgresql.org/docs/7.0/syntax525.htm)]
```
postgres=# create table t1 ("a`" int);
CREATE TABLE
postgres=# insert into t1 ("a`") values (1);
INSERT 0 1
postgres=# select * from t1;
 a`
----
  1
(1 row)
```
but is not possible in Avro [[ref](https://avro.apache.org/docs/1.8.1/spec.html#names)]. so i suspect we already performed some kind of conversion here.

jgao54:
e2e test caught an issue (repro-ed locally as well) where if the decimal is larger than Decimal(72, 36), cast won't behave correctly.


> failed to sync records: QRepSync Error: code: 376, message: Cannot parse type Decimal(76, 38), expected non-empty binary data with size equal to or less than 32, got 49: (at row 10)

iiuc, CH is attempting to process this field from Avro as decimal before casting it to string?

flow/model/qvalue/qvalue_converter.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+func NumericToStringSchemaConversion(val QField) QField {
+	val.Type = QValueKindString
+	return val
+}
+
+func NumericToStringSchemaConversionNullable(val QField) QField {
+	val.Type = QValueKindString
+	val.Nullable = true
```
serprex, line 19: should preserve nullable from previous version

flow/model/qvalue/qvalue_converter.go
```diff
@@ -0,0 +1,25 @@
...
+
+type TypeConversion struct {
+	SchemaConversionFn func(QField) QField
+	ValueConversionFn  func(QValue) QValue
+	FromKind           QValueKind
+	ToKind             QValueKind
+}
+
+func NumericToStringSchemaConversion(val QField) QField {
+	val.Type = QValueKindString
```
serprex, line 13: `val.Nullable = false` will want to make sure nulls get mapped to empty string

flow/model/qvalue/qvalue_converter.go
```diff
@@ -0,0 +1,25 @@
+package qvalue
+
+import "github.com/shopspring/decimal"
+
+type TypeConversion struct {
```
serprex, line 5: this can be generic over `TypeConversion[TFrom, TTo]` then implement an interface which takes `Convert(from QValue) QValue` that handles downcasting. That way your methods can take concrete QValue types instead of having to call `.Value()` internally

jgao54: discussed this via DM, will handle this separately later

serprex (approved):

PR #2965, opened 2025-05-22
replace custom avro schema building logic with hamba schema

serprex:
less code, & now we directly pass schema instead of serializing our structs to json to deserialize to hamba schema


heavycrystal (commented):
~looks good, but would be nice to have assurance that Avro files are exact before and after this change for all types to prevent subtle behaviour changes~
EDIT: they won't be exact due to compression, but hopefully tests were good

iskakaushik (approved):

Issue #2968, opened 2025-05-22
Add test to make sure all metric gauges/counters are correctly initialized


iamKunalGupta:

serprex:
this is less of an issue with nop otel manager since we segfault on null gauges/counters

PR #2969, opened 2025-05-22
mysql: classify 1236. new gauge: CommitLag

serprex:


flow/alerting/classifier.go
```diff
@@ -296,6 +295,9 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 			1195, // ER_CRASHED_ON_REPAIR
 			1827: // ER_PASSWORD_FORMAT
 			return ErrorNotifyConnectivity, myErrorInfo
+		case 1236, // ER_MASTER_FATAL_ERROR_READING_BINLOG
+			1373: // ER_UNKNOWN_TARGET_BINLOG
```
heavycrystal, line 299: nit: are you aware of when this can happen?

flow/connectors/postgres/cdc.go
```diff
@@ -687,6 +687,7 @@ func processMessage[Items model.Items](
 			slog.Any("CommitLSN", msg.CommitLSN),
 			slog.Any("TransactionEndLSN", msg.TransactionEndLSN))
 		batch.UpdateLatestCheckpointID(int64(msg.CommitLSN))
+		p.otelManager.Metrics.CommitLagGauge.Record(ctx, time.Now().UTC().Sub(msg.CommitTime).Microseconds())
```
heavycrystal, line 690: is it guaranteed to be monotonically increasing? same for MySQL
if not it's worth documenting

flow/alerting/classifier.go
serprex: mariadb description is "Target log not found in binlog index" so I'm guessing it'll come up with binlog corruption or in certain failover situations

flow/connectors/postgres/cdc.go
serprex: For PG yes, for MySQL I'm not as sure in case replica is receiving GTIDs out of order. Drift should be minimal so not worried here

flow/connectors/mysql/cdc.go
```diff
@@ -393,6 +394,11 @@ func (c *MySqlConnector) PullRecords(
 		otelManager.Metrics.FetchedBytesCounter.Add(ctx, int64(len(event.RawData)))
 
 		switch ev := event.Event.(type) {
+		case *replication.GTIDEvent:
+			if ev.ImmediateCommitTimestamp > 0 {
+				otelManager.Metrics.CommitLagGauge.Record(ctx,
+					time.Now().UTC().Sub(time.UnixMicro(int64(ev.ImmediateCommitTimestamp))).Microseconds())
```
iamKunalGupta, line 400: Will need to initialize the gauge before using it.
I have an issue opened to add tests to fail in case we miss initializing #2968

heavycrystal (approved):

PR #2987, opened 2025-05-29
add frontend for fetching type conversion

jgao54:
Create frontend for [custom column type conversion.](https://github.com/PeerDB-io/peerdb/pull/2989).

It creates a dropdown of possible conversions determined by the backend. Right now only numeric -> string is supported.
If there are no available conversion, the option for `use custom column destination type` will be hidden.

Screenshot 2025-05-29 at 06 36 32@2x

Test: locally created a table that contains column with numeric type, create ClickPipe via UI, and see CH table get created with String type.

### DIFF
--- a/ui/app/mirrors/create/cdc/customColumnType.tsx
+++ b/ui/app/mirrors/create/cdc/customColumnType.tsx
@@ -75,7 +75,7 @@ export default function CustomColumnType({
 
   const destinationTypeOptions = useCallback(
     (col: string): { value: string; label: string }[] => {
-      return destinationTypeMapping[col] || [];
+      return destinationTypeMapping[col] ?? [];
     },
     [destinationTypeMapping]
   );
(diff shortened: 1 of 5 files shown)


ui/app/mirrors/create/cdc/tablemapping.tsx
```diff
@@ -103,7 +103,7 @@ export default function TablePicker({
           />
         </div>
       </div>
-      <div style={{ maxHeight: '70vh', overflow: 'scroll' }}>
```
jgao54, line 106: note: the maxHeight was blocking the UI list from being fully visible, so removed it here.

ui/app/mirrors/create/cdc/customColumnType.tsx
```diff
@@ -0,0 +1,336 @@
...
+          }
+        }
+        setDestinationTypeMapping(columnNameToDestinationTypes);
+      } catch (error) {
+        console.error('Error fetching type conversions:', error);
+      }
+    };
+
+    fetchTypeMappings();
+    // TODO: there is a bug where `columns` is getting re-rendered excessively
```
serprex, line 106: this should be fixed (potentially split out as separate PR)

Amogh-Bharadwaj (approved):

PR #2989, opened 2025-05-29
implement backend for fetching type conversions

jgao54:
Implement the backend to support [type conversion frontend](https://github.com/PeerDB-io/peerdb/pull/2987)


flow/connectors/core.go
```diff
@@ -70,6 +70,7 @@ type GetSchemaConnector interface {
 	GetColumns(ctx context.Context, schema string, table string) (*protos.TableColumnsResponse, error)
 	GetSchemas(ctx context.Context) (*protos.PeerSchemasResponse, error)
 	GetTablesInSchema(ctx context.Context, schema string, cdcEnabled bool) (*protos.SchemaTablesResponse, error)
+	GetColumnsTypeConversion() (*protos.ColumnsTypeConversionResponse, error)
```
jgao54, line 73: If we don't want to introduce this interface, we can just directly call a static function:

```
clickhouseconn.GetColumnsTypeConversion()
```

flow/connectors/clickhouse/qrep_avro_sync.go
```diff
@@ -373,6 +373,20 @@ var supportedDestinationTypes = map[string][]qvalue.TypeConversion{
...
 }
 
+func listSupportedTypeConversions() map[qvalue.QValueKind][]string {
+	typeConversions := make(map[qvalue.QValueKind][]string)
+
+	for dstType, l := range supportedDestinationTypes {
+		for _, conversion := range l {
+			if _, ok := typeConversions[conversion.FromKind()]; !ok {
+				typeConversions[conversion.FromKind()] = []string{}
+			}
```
serprex, line 383:
```suggestion
```
no need, `append(nil, elem)` works

PR #3007, opened 2025-06-04
[WIP] introduce direct source to destination type conversion

jgao54:
This PR introduce a direct type conversion from PG to CH. This has the benefit of not exposing qvalue, which ideally is hidden as implementation details inside peerdb.

- This logic is added to a shared package (`flow/shared/`) for easy import.
- Added a unit test to ensure that the direct type mapping (PG type -> CH type) returns the same result as indirect type mapping (PG type -> qkind -> CH type). Currently the postgres oids are hard-coded, as I did not find a pgx lib call that allows for exhaustively iterating through the pg types, but we can turn this into an integration test and use a real postgres backend to get exhaustive pg types. 
- Currently the supported destination types is hard-coded in flow/shared, this means it's possible that `ToSupportedDestinationTypesForPostgres` method in `flow/shared/clickhouse/type_conversion.go` and `supportedDestinationTypes` in `flow/connectors/clickhouse/type_conversion.go` are out-of-sync, resulting destination type conversion getting ignored if that happens. We could add a test to ensure this is in sync as well (open to other ideas).

TODO:
- [ ] test this e2e in CP.
- [ ] sort through the edge cases with type modifier and custom types
- [ ] replace old `fetchAllTypeConversions` with a new `fetchTypeConversionsForColumns` endpoint using direct type conversion
- [ ] update peerdb ui to use new logic
- [ ] remove qkind from column type once it's no longer needed

### DIFF
--- a/flow/connectors/mysql/qvalue_convert.go
+++ b/flow/connectors/mysql/qvalue_convert.go
@@ -15,7 +15,6 @@ import (
 	"github.com/shopspring/decimal"
 	geom "github.com/twpayne/go-geos"
 
-	"github.com/PeerDB-io/peerdb/flow/datatypes"
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
 	"github.com/PeerDB-io/peerdb/flow/model/qvalue"
 	"github.com/PeerDB-io/peerdb/flow/shared"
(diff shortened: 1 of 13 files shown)


jgao54:
Closing in favor of #3011

PR #3010, opened 2025-06-05
Map numeric arrays into string arrays

ilidemi:
Running E2E tests in CI hence draft PR

### DIFF
--- a/flow/connectors/postgres/qvalue_convert.go
+++ b/flow/connectors/postgres/qvalue_convert.go
@@ -585,13 +585,12 @@ func (c *PostgresConnector) parseFieldFromPostgresOID(
 		if items, ok := value.([]any); ok {
 			a := make([]string, len(items))
 			success := true
-		loop:
 			for i, item := range items {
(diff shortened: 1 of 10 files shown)


flow/connectors/postgres/qvalue_convert.go
```diff
@@ -577,6 +581,25 @@ func (c *PostgresConnector) parseFieldFromPostgresOID(
...
+				var ok bool
+				if numeric, ok = item.(pgtype.Numeric); !ok {
+					success = false
+					break
+				}
+				bytes, _ := numeric.MarshalJSON()
+				a[i] = shared.UnsafeFastReadOnlyBytesToString(bytes)
+			}
+			if success {
+				return qvalue.QValueArrayString{Val: a}, nil
```
serprex, line 599:
```suggestion
				return qvalue.QValueArrayNumeric{Val: a}, nil
```
code is in transition at the moment to have full numeric support & that this'll be decimal.Decimal slice anyways?

PR #3011, opened 2025-06-05
move type conversion related qvalue package to shared

jgao54:
Breaking out the task to more manageable steps to reduce risk of introducing regression.

For now I am focusing on the logic needed to get supported type conversions only (and not worrying about displaying the default type mapping, which is a lot hairier).

Approach is to focus on migrating `ListSupportedTypeConversions` (and packages it depends on) to shared, and refactor code so that the packages migrated to shared/ does not import packages from the rest of peerdb codebase.

Test: CI + locally tested a simple flow.


serprex:
To avoid name conflict, what's left in model/qvalue can be flattened into `model` package

jgao54:
There were also some conflicts with the model package in flow/model/ so named it to `types` instead

serprex (approved):

PR #3012, opened 2025-06-06
move pg -> qvalue logic to shared for type conversion

jgao54:
Follow-up to https://github.com/PeerDB-io/peerdb/pull/3011.

Move Postgres -> qvalue type mapping logic to shared package so we can determined qvalue. 

This gives us a way to use shared package to directly derive:
1) qvalue kind from postgres type using `PostgresOIDToQValueKind`
2) supported destination types from qvalue using `ListSupportedTypeConversions`

Test: CI + locally tested a simple flow.


flow/shared/postgres/type_conversion.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/lib/pq/oid"
+
+	"github.com/PeerDB-io/peerdb/flow/shared"
+	"github.com/PeerDB-io/peerdb/flow/shared/types"
+)
+
+type UnsupportedError struct {
```
serprex: Should implement `Error() string` method to fulfill error interface. & then error handler grabbing Message should call Error() instead. Can remove `Message` field entirely, constant can live in Error method

PR #3014, opened 2025-06-06
ch: skip rename for null table engine resync

serprex:
avoids issues where table renaming doesn't update materialized views

TODO update null table e2e with resync


serprex:
That's why we still do CREATE OR REPLACE

PR #3016, opened 2025-06-06
feat(classifier): recovery canceling statement as connectivity

iamKunalGupta:


heavycrystal:
can be thrown with another errcode, please catch that
```
	ereport(ERROR,
			(errcode(ERRCODE_T_R_DEADLOCK_DETECTED),
			 errmsg("canceling statement due to conflict with recovery"),
			 errdetail("User transaction caused buffer deadlock with recovery.")));
```

heavycrystal (approved):

PR #3023, opened 2025-06-09
move mysql qvalue logic to shared

jgao54:
Follow up to https://github.com/PeerDB-io/peerdb/pull/3012 -- move MySQL -> qvalue type mapping logic to shared package.

Test: CI + locally tested a simple flow with MySQL.


flow/connectors/mysql/cdc.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/otel_metrics"
 	"github.com/PeerDB-io/peerdb/flow/shared"
 	"github.com/PeerDB-io/peerdb/flow/shared/datatypes"
+	chmysql "github.com/PeerDB-io/peerdb/flow/shared/mysql"
```
serprex: `chmysql` would be better as `qmysql` since this maintains abstraction re destination types

serprex (approved):

PR #3031, opened 2025-06-10
ClickHouse: Maintain info on batch-wise pushed rows

Amogh-Bharadwaj:
This PR introduces a column in the `cdc_batches` stats table in catalog called `pushed_rows`.
In ClickHouse destination peer's `NormalizeRecords` , we are updating this column after every table is normalized by getting the count of the table's records count from raw table for the range of batches and inserting the same into the above stats table.

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -329,7 +329,7 @@ func (c *ClickHouseConnector) NormalizeRecords(
 	numParts = max(numParts, 1)
 
 	queries := make(chan NormalizeQueryGenerator)
-	rawTbl := c.GetRawTableName(req.FlowJobName)
+	rawTableName := c.GetRawTableName(req.FlowJobName)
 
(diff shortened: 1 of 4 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -493,6 +519,52 @@ func (c *ClickHouseConnector) getDistinctTableNamesInBatch(
...
+		peerdb_clickhouse.QuoteIdentifier(rawTbl))
+
+	rowCount := make(map[int64]uint64)
+	rows, err := c.query(ctx, q, tableName, batchIDBeingLoaded, syncBatchID)
+	if err != nil {
+		return nil, fmt.Errorf("error while querying raw table for row count: %w", err)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var count sql.NullInt64
```
serprex, line 550:
```suggestion
		var count int64
```
count won't be null

PR #3034, opened 2025-06-10
MySQL validation: Add check for MariaDB GTID set

### LINKED ISSUE #3032: MySQL validation: Handle MariaDB differences
Currently when validating replication permissions for MySQL source peer - in the case where the source peer is a MariaDB instance, we need to check the replication permissions differently.

Amogh-Bharadwaj:
Fixes #3032

### DIFF
--- a/flow/connectors/mysql/validate.go
+++ b/flow/connectors/mysql/validate.go
@@ -25,20 +25,26 @@ func (c *MySqlConnector) CheckSourceTables(ctx context.Context, tableNames []*ut
 }
 
 func (c *MySqlConnector) CheckReplicationConnectivity(ctx context.Context) error {
-	// GTID -> check GTID and error out if not enabled, check filepos as well
-	// AUTO -> check GTID and fall back to filepos check
-	// FILEPOS -> check filepos only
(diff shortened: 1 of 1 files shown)


flow/connectors/mysql/validate.go
```diff
@@ -25,20 +25,26 @@ func (c *MySqlConnector) CheckSourceTables(ctx context.Context, tableNames []*ut
 }
 
 func (c *MySqlConnector) CheckReplicationConnectivity(ctx context.Context) error {
-	// GTID -> check GTID and error out if not enabled, check filepos as well
-	// AUTO -> check GTID and fall back to filepos check
-	// FILEPOS -> check filepos only
-	if c.config.ReplicationMechanism != protos.MySqlReplicationMechanism_MYSQL_FILEPOS {
-		if _, err := c.GetMasterGTIDSet(ctx); err != nil {
```
heavycrystal, line 32: `GetMasterGTIDSet` already switches on flavor?
I'm not sure what the problem is

serprex: is issue that we need to use @@GLOBAL.gtid_current_pos in GetMasterGTIDSet?

Amogh-Bharadwaj:
Misidentified issue

PR #3036, opened 2025-06-10
change QValueTime/QValueTimeTZ to internally use time.Duration not time.Time

serprex:
MySQL should instead map `time` to `QValueTimestamp` since `QValueTime` maps to types on destinations where time is expected to be the time part of a datetime (ie, from 00:00 up to 24:00). Keep existing QValueTime handling to not break existing mirrors. Note these existing mirrors have a problem where in cdc they will have time be modulos 24 hours

Another issue was that postgres time values over cdc on clickhouse were coming in with current year instead of unix epoch. This is because we relied on `parseDateTime64BestEffort` which interprets times without dates as happening in current year (or previous year in some cases). Adjust normalization to have year be 1970 in order to be consistent between qrep & cdc

This also helps allow in the future mapping QValueTime to DateTime instead of DateTime64

Adjust tests to not be so permissive, which let these discrepancies exist, but which going forward we should prevent

### DIFF
--- a/flow/connectors/bigquery/merge_stmt_generator.go
+++ b/flow/connectors/bigquery/merge_stmt_generator.go
@@ -24,7 +24,7 @@ type mergeStmtGenerator struct {
 
 // generateFlattenedCTE generates a flattened CTE.
 func (m *mergeStmtGenerator) generateFlattenedCTE(dstTable string, normalizedTableSchema *protos.TableSchema) string {
-	// for each column in the normalized table, generate CAST + JSON_EXTRACT_SCALAR
+	// for each column in the normalized table, generate CAST + JSON_VALUE
 	// statement.
 	flattenedProjs := make([]string, 0, len(normalizedTableSchema.Columns)+3)
 
@@ -58,13 +58,13 @@ func (m *mergeStmtGenerator) generateFlattenedCTE(dstTable string, normalizedTab
 		// this needs to be handled again
 		// TODO add interval types again
 		// case model.ColumnTypeInterval:
-		// castStmt = fmt.Sprintf("MAKE_INTERVAL(0,CAST(JSON_EXTRACT_SCALAR(_peerdb_data, '$.%s.Months') AS INT64),"+
-		// 	"CAST(JSON_EXTRACT_SCALAR(_peerdb_data, '$.%s.Days') AS INT64),0,0,"+
-		// 	"CAST(CAST(JSON_EXTRACT_SCALAR(_peerdb_data, '$.%s.Microseconds') AS INT64)/1000000 AS  INT64)) AS %s",
+		// castStmt = fmt.Sprintf("MAKE_INTERVAL(0,CAST(JSON_VALUE(_peerdb_data, '$.%s.Months') AS INT64),"+
+		// 	"CAST(JSON_VALUE(_peerdb_data, '$.%s.Days') AS INT64),0,0,"+
+		// 	"CAST(CAST(JSON_VALUE(_peerdb_data, '$.%s.Microseconds') AS INT64)/1000000 AS  INT64)) AS %s",
 		// 	column.Name, column.Name, column.Name, column.Name)
 		// TODO add proper granularity for time types, then restore this
 		// case model.ColumnTypeTime:
-		// 	castStmt = fmt.Sprintf("time(timestamp_micros(CAST(JSON_EXTRACT(_peerdb_data, '$.%s.Microseconds')"+
+		// 	castStmt = fmt.Sprintf("time(timestamp_micros(CAST(JSON_VALUE(_peerdb_data, '$.%s.Microseconds')"+
 		// 		" AS int64))) AS %s",
 		// 		column.Name, column.Name)
 		default:
--- a/flow/connectors/bigquery/merge_stmt_generator.go
+++ b/flow/connectors/bigquery/merge_stmt_generator.go
@@ -52,21 +52,6 @@ func (m *mergeStmtGenerator) generateFlattenedCTE(dstTable string, normalizedTab
 		case types.QValueKindGeography, types.QValueKindGeometry, types.QValueKindPoint:
 			castStmt = fmt.Sprintf("CAST(ST_GEOGFROMTEXT(JSON_VALUE(_peerdb_data, '$.%s')) AS %s) AS `%s`",
 				column.Name, bqTypeString, shortCol)
-		// MAKE_INTERVAL(years INT64, months INT64, days INT64, hours INT64, minutes INT64, seconds INT64)
-		// Expecting interval to be in the format of {"Microseconds":2000000,"Days":0,"Months":0,"Valid":true}
-		// json.Marshal in SyncRecords for Postgres already does this - once new data-stores are added,
-		// this needs to be handled again
-		// TODO add interval types again
-		// case model.ColumnTypeInterval:
-		// castStmt = fmt.Sprintf("MAKE_INTERVAL(0,CAST(JSON_VALUE(_peerdb_data, '$.%s.Months') AS INT64),"+
-		// 	"CAST(JSON_VALUE(_peerdb_data, '$.%s.Days') AS INT64),0,0,"+
-		// 	"CAST(CAST(JSON_VALUE(_peerdb_data, '$.%s.Microseconds') AS INT64)/1000000 AS  INT64)) AS %s",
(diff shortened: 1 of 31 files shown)


flow/connectors/bigquery/merge_stmt_generator.go
```diff
@@ -58,13 +58,13 @@ func (m *mergeStmtGenerator) generateFlattenedCTE(dstTable string, normalizedTab
 		// this needs to be handled again
 		// TODO add interval types again
 		// case model.ColumnTypeInterval:
-		// castStmt = fmt.Sprintf("MAKE_INTERVAL(0,CAST(JSON_EXTRACT_SCALAR(_peerdb_data, '$.%s.Months') AS INT64),"+
-		// 	"CAST(JSON_EXTRACT_SCALAR(_peerdb_data, '$.%s.Days') AS INT64),0,0,"+
-		// 	"CAST(CAST(JSON_EXTRACT_SCALAR(_peerdb_data, '$.%s.Microseconds') AS INT64)/1000000 AS  INT64)) AS %s",
+		// castStmt = fmt.Sprintf("MAKE_INTERVAL(0,CAST(JSON_VALUE(_peerdb_data, '$.%s.Months') AS INT64),"+
```
heavycrystal: can just remove this? not sure we'll ever get to this

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -215,17 +215,17 @@ func processTime(str string) (time.Time, error) {
...
+		return 0, err
 	}
 
 	sec := hpart*3600 + mpart*60 + spart
 	if isNeg {
-		return time.Unix(-int64(sec), -int64(nsec)).UTC(), nil
+		return -time.Duration(sec)*time.Second - time.Duration(nsec), nil
 	}
-	return time.Unix(int64(sec), int64(nsec)).UTC(), nil
+	return time.Duration(sec)*time.Second + time.Duration(nsec), nil
```
heavycrystal: nit: do this for both cases and just return the negative of it if needed

flow/connectors/postgres/cdc.go
```diff
@@ -262,10 +262,8 @@ func (p *PostgresCDCSource) decodeColumnData(
 			if dtOid == oid.T_time || dtOid == oid.T_timetz ||
 				dtOid == oid.T_timestamp || dtOid == oid.T_timestamptz {
 				// indicates year is more than 4 digits or something similar,
-				// which you can insert into postgres,
-				// but not representable by time.Time
-				p.logger.Warn(fmt.Sprintf("Invalidated and hence nulled %s data: %s",
-					dt.Name, string(data)))
+				// which you can insert into postgres, but not representable by time.Time
+				p.logger.Warn("Invalidated time, nulled", slog.String("typeName", dt.Name), slog.String("value", string(data)))
```
heavycrystal:
```suggestion
				p.logger.Warn("Invalidate time for destination, nulled", slog.String("typeName", dt.Name), slog.String("value", string(data)))
```

flow/connectors/snowflake/client.go
```diff
@@ -231,9 +231,12 @@ func toQValue(kind types.QValueKind, val any) (types.QValue, error) {
 		if t, ok := val.(*sql.NullTime); ok {
 			if t.Valid {
 				tt := t.Time
-				// anchor on unix epoch, some drivers anchor on 0001-01-01
+				h, m, s := tt.Clock()
 				return types.QValueTime{
-					Val: time.Date(1970, time.January, 1, tt.Hour(), tt.Minute(), tt.Second(), tt.Nanosecond(), time.UTC),
```
heavycrystal, line 236: do tests confirm if this is safe in all cases?

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -75,7 +75,7 @@ func qkindFromMysql(field *mysql.Field) (types.QValueKind, error) {
 	case mysql.MYSQL_TYPE_DATETIME2:
 		return types.QValueKindTimestamp, nil
 	case mysql.MYSQL_TYPE_TIME2:
-		return types.QValueKindTime, nil
+		return types.QValueKindTimestamp, nil
```
heavycrystal: nit: club these cases together

flow/e2e/bigquery/bigquery_helper.go
```diff
@@ -187,8 +187,10 @@ func toQValue(bqValue bigquery.Value) (types.QValue, error) {
 	case civil.Date:
 		return types.QValueDate{Val: v.In(time.UTC)}, nil
 	case civil.Time:
-		tm := time.Unix(int64(v.Hour)*3600+int64(v.Minute)*60+int64(v.Second), int64(v.Nanosecond))
-		return types.QValueTime{Val: tm}, nil
+		return types.QValueTime{Val: time.Duration(v.Hour)*time.Hour +
```
heavycrystal, line 190: same question here

flow/e2e/postgres/peer_flow_pg_test.go
```diff
@@ -158,7 +158,7 @@ func (s PeerFlowE2ETestSuitePG) Test_Types_PG() {
 	// c36 lost tz info so does not compare equal
 	var c36 string
 	require.NoError(s.t, s.Conn().QueryRow(s.t.Context(), "select c36 from "+dstTableName).Scan(&c36))
-	require.Equal(s.t, "09:25:00+00", c36)
+	require.Equal(s.t, "06:25:00+00", c36)
```
heavycrystal, line 165: why did this need changing?

flow/model/qvalue/avro_converter.go
```diff
@@ -219,7 +219,7 @@ func QValueToAvro(
 	case types.QValueTime:
 		return c.processNullableUnion(c.processGoTime(v.Val))
 	case types.QValueTimeTZ:
-		return c.processNullableUnion(c.processGoTimeTZ(v.Val))
+		return c.processNullableUnion(c.processGoTime(v.Val))
```
heavycrystal, line 222: timezone?

flow/connectors/snowflake/client.go
serprex: yes

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -166,15 +166,15 @@ func processGeometryData(data []byte) types.QValueGeometry {
 }
 
 // https://dev.mysql.com/doc/refman/8.4/en/time.html
-func processTime(str string) (time.Time, error) {
+func processTime(str string) (time.Duration, error) {
```
jgao54, line 156: reading mysql doc, looks like mysql's `time` data type doesn't have the concept of timezone and just suggest best practice to treat i as UTC always.

curious thought why switch to Duration instead of keeping time.Time here as is. and then ~during normalization~ when writing to Avro  extract the hh/mm/ss from time.Time? 

EDIT: Nvm, reread code and understand better. with this change you are normalizing time to UTC when converting to Avro. this is fine to me but it does mean losing timezone info when writing to CH which may not be ideal iiuc?

serprex: There's no timezone info on time

We have a separate QValueTimeTZ for pg's `TIME WITH TIME ZONE` where we drop timezone info, which Postgres recommends not using, & which pgx decides not to support (hence we have to parse it ourselves)

flow/e2e/postgres/peer_flow_pg_test.go
serprex: should remove comment about tz info being lost

flow/e2e/bigquery/bigquery_helper.go
serprex: Yes

flow/connectors/mysql/qvalue_convert.go
serprex: case order is sorted by mysql type's byte value, but a more logical grouping seems fair (especially since elsewhere cases are logically grouped already)

flow/model/qvalue/avro_converter.go
serprex: Timezone is never (& was never) maintained. Postgres is responsible for collapsing timezone info now (which actually works, so that you get the UTC equivalent of the timetz value, rather than simply dropping timezone info, which breaks if the source has heterogeneous timezones on their values)

jgao54 (approved):

PR #3040, opened 2025-06-12
PG: Support more types as strings and string arrays

ilidemi:
Add a better catch-all into strings and string arrays. 

Types that are now not empty:
```
bytea[]
bit(n)[]
varbit[]
xml[]
time[]
interval[]
point[]
line[]
lseg[]
path[]
box[] -- still fails CDC, #3038 
polygon[]
circle[]
macaddr[]
cidr[]
inet[]
int4range[]
int8range[]
numrange[]
tsrange[]
tstzrange[]
```

Now readable:
```
bit(n): "{[160] 3 true}" -> "101"
varbit: "{[160] 3 true}" -> "101"
xml: "[60 105 116 101 109 62 100 97 116 97 60 47 105 116 101 109 62]"
  -> "<item>data</item>"
numrange: "{{15 -1 false finite true} {999 -1 false finite true} i i true}"
  -> "[1.5,99.9]"
nummultirange: "[{{11 -1 false finite true} {109 -1 false finite true} i i true} {{201 -1 false finite true} {309 -1 false finite true} i i true}]"
  -> "{[1.1,10.9],[20.1,30.9]}"
```

Easier to read/parse:
```
line: "{1 2 3 true}" -> "{1,2,3}"
lseg: "{[{1 1} {2 2}] true}" -> "[(1,1),(2,2)]"
path: "{[{1 1} {2 2} {3 1}] true true}" -> "((1,1),(2,2),(3,1))"
box: "{[{3 3} {1 1}] true}" -> "(3,3),(1,1)"
polygon: "{[{1 1} {2 2} {3 1}] true}" -> "((1,1),(2,2),(3,1))"
circle: "{{1 2} 3 true}" -> "<(1, 2), 3>"
int4range: "{1 101 i e true}" -> "[1,101)"
int8range: "{1 10000000001 i e true}" -> "[1,10000000001)"
tsrange: "{2023-01-01 00:00:00 +0000 UTC 2023-12-31 23:59:59 +0000 UTC i i true}"
  -> "[2023-01-01 00:00:00,2023-12-31 23:59:59]"
daterange: "{2023-01-01 00:00:00 +0000 UTC 2024-01-01 00:00:00 +0000 UTC i e true}"
  -> "[2023-01-01,2024-01-01)"
int4multirange: "[{1 11 i e true} {20 31 i e true}]" -> "{[1,11),[20,31)}"
int8multirange: "[{1 10000000001 i e true} {20000000000 30000000001 i e true}]"
  -> "{[1,10000000001),[20000000000,30000000001)}"
tsmultirange: "[{2023-01-01 00:00:00 +0000 UTC 2023-01-31 23:59:59 +0000 UTC i i true} {2023-03-01 00:00:00 +0000 UTC 2023-03-31 23:59:59 +0000 UTC i i true}]"
  -> "{[2023-01-01 00:00:00,2023-01-31 23:59:59],[2023-03-01 00:00:00,2023-03-31 23:59:59]}"
tstzmultirange: "[{2023-01-01 02:00:00 +0000 UTC 2023-01-31 23:59:59 +0000 UTC i i true} {2023-03-01 02:00:00 +0000 UTC 2023-03-31 23:59:59 +0000 UTC i i true}]"
  -> "{[2023-01-01 02:00:00Z,2023-01-31 23:59:59Z],[2023-03-01 02:00:00Z,2023-03-31 23:59:59Z]}"
datemultirange: "[{2023-01-01 00:00:00 +0000 UTC 2023-02-01 00:00:00 +0000 UTC i e true} {2023-03-01 00:00:00 +0000 UTC 2023-04-01 00:00:00 +0000 UTC i e true}]"
  -> "{[2023-01-01,2023-02-01),[2023-03-01,2023-04-01)}" 
```

Relying on pgx for conversions as much as possible. `QValueTSTZRange` doesn't need special handling anymore and was removed.

Inconsistencies:
* `point` is `"POINT(1.000000, 2.000000)"` and `point[]` is `["(1,2)", "(3,4)"]` like the rest of geometry
* `tstzrange` became `"[2023-01-01 02:00:00Z,2023-12-31 23:59:59Z]"` instead of `"[2023-01-01 02:00:00,2023-12-31 23:59:59]"` (note the Z's) - this makes it uniform with `tstzmultirange` and especially `tstzmultirange[]` which pgx already returns as a string we need to parse. Postgres destination can easily create `tzrange` from such strings, only if someone is doing manual parsing they'd see a difference.


serprex:
@Amogh-Bharadwaj who was using TSTZRange? I recall you added it for a pg<->pg customer while I PG type system was still being worked on & so they couldn't use that to preserve their types

We should be able to merge this after EOL, & ask anyone relying on TSTZRange preservation in postgres to postgres to use PG type system

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -208,6 +209,171 @@ func parseUUIDArray(value any) (types.QValue, error) {
...
+		case pgtype.NummultirangeOID:
+			typedM, err = multirangeToTyped[pgtype.Numeric](multirange)
+		case pgtype.DatemultirangeOID, pgtype.TsmultirangeOID, pgtype.TstzmultirangeOID:
+			typedM, err = multirangeToTyped[time.Time](multirange)
+		default:
+			err = errors.ErrUnsupported
+		}
+		if err == nil {
+			var buf []byte
+			buf, err = c.typeMap.Encode(oid, pgtype.TextFormatCode, typedM, nil)
```
serprex, line 327:
```suggestion
			buf, err := c.typeMap.Encode(oid, pgtype.TextFormatCode, typedM, nil)
```

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -463,6 +614,23 @@ func (c *PostgresConnector) parseFieldFromPostgresOID(
...
+			if typeData, ok := customTypeMapping[oid]; ok {
+				delim = typeData.Delim
+			}
+			arr := shared.ParsePgArrayStringToStringSlice(str, delim)
+			for i, itemStr := range arr {
+				if itemStr == "NULL" {
+					arr[i] = ""
+				}
+			}
+			return types.QValueArrayString{Val: arr}, nil
```
serprex, line 624: this is breaking things a bit returning QValueArrayString which won't have `Kind()` returning the original QValueKind

you should be able to have these all map to QValueArrayString, not introducing new pseudokinds, & in QValueArrayString check `oid` variable

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -246,51 +412,35 @@ func (c *PostgresConnector) parseFieldFromPostgresOID(
...
+						return nil, fmt.Errorf("failed to parse interval array: %w", err)
+					}
+					strs = append(strs, str)
+				} else {
+					success = false
+					break
+				}
+			}
+			if success {
+				return types.QValueArrayString{Val: strs}, nil
```
serprex:
```suggestion
				return types.QValueArrayInterval{Val: strs}, nil
```
(bug in existing code, was being fixed already in #3036)

flow/connectors/postgres/qvalue_convert.go
ilidemi: Makes sense, got confused by QValueKindInterval before your change

ilidemi:
Tested PG -> PG. By default, when tstzrange goes into a string column, Postgres succeeds converting the `Z` version to `::tstzrange` in a separate query. If the target column is pre-created as tstzrange, both main and this PR fail with
```
failed to sync records: failed to copy records into destination table: ERROR: COPY from stdin failed: unable to encode \"[2023-01-01 02:00:00,2023-12-31 23:59:59]\" into binary format for tstzrange (OID 3910): cannot find encode plan (SQLSTATE 57014)
```
so it's unlikely that's what the customer is doing.

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -208,6 +209,171 @@ func parseUUIDArray(value any) (types.QValue, error) {
...
+	if !interval.Valid {
+		return "", fmt.Errorf("invalid interval: %v", intervalObject)
+	}
+
+	return string(intervalJSON), nil
+}
+
+var ErrMismatchingRangeType = errors.New("mismatching range type")
+
+func rangeToTyped[T any](r pgtype.Range[any]) (any, error) {
```
serprex:
```suggestion
func rangeToTyped[T any](r pgtype.Range[any]) (pgtype.Range[*T], error) {
```

ilidemi: The `any` is there so that the call sites in `convertToString()` are one-liners

serprex: But then you're casting the result in `multirangeToTyped`

You should be able to assign typed values to variables typed as compatible interfaces so I'm not sure why you'd be running into issues where you're declaring `var typedR any` beforehand

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -473,13 +621,26 @@ func (c *PostgresConnector) parseFieldFromPostgresOID(
...
 			}
-			return types.QValueArrayString{Val: shared.ParsePgArrayStringToStringSlice(str, delim)}, nil
+			arr := shared.ParsePgArrayStringToStringSlice(str, delim)
+			switch oid {
+			case pgtype.TimetzArrayOID,
+				pgtype.Int4multirangeArrayOID, pgtype.Int8multirangeArrayOID,
+				pgtype.NummultirangeArrayOID, pgtype.TsmultirangeArrayOID,
+				pgtype.TstzmultirangeArrayOID, pgtype.DatemultirangeArrayOID:
+				for i, itemStr := range arr {
+					if itemStr == "NULL" {
```
serprex: [Pretty sure this applies to all arrays](https://www.postgresql.org/docs/current/arrays.html#ARRAYS-INPUT)

https://github.com/postgres/postgres/blob/6e951f279beab566efa267151d028c3684f6095e/src/backend/utils/adt/arrayfuncs.c#L741

Could be handled in `ParsePgArrayStringToStringSlice` irrespective of oid: #3040

ilidemi: We could have "NULL" literals from strings and god knows what other types, hence the whitelist on what pgx actually passes to us as a string and has NULL mean null. With #3051 it's nice and generic

flow/connectors/postgres/qvalue_convert.go
ilidemi: You are correct

PR #3041, opened 2025-06-12
ClickHouse: fix CDC for timestamp arrays

ilidemi:


serprex:
should include test in e2e

PR #3042, opened 2025-06-12
Implement basic snapshotting for mongo connector

jgao54:
- Implements basic snapshotting for `packed` mode.
- Assume Mongo's default ObjectId is used for `_id` field. 
- Full table copy only (partition mode not yet supported).
- Tested e2e for initial snapshot locally.

Notes:
- `Snapshot` read concern may indeed be problematic. Reproduced `SnapshotTooOld` error with minSnapshotHistoryWindowInSeconds = 5 on a 40MB+ collection with 3M+ documents. Using `Majority` read concern for now. 
- Bucketing took 5s+ on the document locally for simple ObjectId type, so need to test out how well this scales. (out-of-scope for this PR)

### DIFF
--- a/flow/connectors/mongo/qrep.go
+++ b/flow/connectors/mongo/qrep.go
@@ -3,7 +3,9 @@ package connmongo
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"log/slog"
 	"math"
 
 	"go.mongodb.org/mongo-driver/v2/bson"
@@ -34,7 +36,7 @@ func (c *MongoConnector) GetQRepPartitions(
 		}, nil
 	}
 	if last != nil && last.Range != nil {
-		return nil, fmt.Errorf("not implemented")
+		return nil, errors.ErrUnsupported
 	}
 	parseWatermarkTable, err := utils.ParseSchemaTable(config.WatermarkTable)
 	if err != nil {
@@ -106,14 +108,13 @@ func (c *MongoConnector) PullQRepRecords(
 	}
 
 	batchSize := config.NumRowsPerPartition
-	if config.NumRowsPerPartition == 0 || config.NumRowsPerPartition > math.MaxInt32 {
+	if config.NumRowsPerPartition <= 0 || config.NumRowsPerPartition > math.MaxInt32 {
 		batchSize = math.MaxInt32
 	}
+
 	// MongoDb will use the lesser of batchSize and 16MiB
 	// https://www.mongodb.com/docs/manual/reference/method/cursor.batchsize/
-	opts1 := options.Find().SetBatchSize(int32(batchSize))
-	opts := []options.Lister[options.FindOptions]{opts1}
-	cursor, err := collection.Find(ctx, filter, opts...)
+	cursor, err := collection.Find(ctx, filter, options.Find().SetBatchSize(int32(batchSize)))
 	if err != nil {
 		return 0, 0, fmt.Errorf("failed to query for records: %w", err)
 	}
@@ -133,8 +134,21 @@ func (c *MongoConnector) PullQRepRecords(
 		totalRecords += 1
 		totalBytes += bytes
 	}
-
 	close(stream.Records)
+	if err := cursor.Err(); err != nil {
+		if errors.Is(err, context.Canceled) {
+			c.logger.Warn("context canceled while reading documents",
+				slog.Any("partition", partition.PartitionId),
+				slog.String("watermark_table", config.WatermarkTable))
+		} else {
+			c.logger.Error("error while reading documents",
+				slog.Any("partition", partition.PartitionId),
+				slog.String("watermark_table", config.WatermarkTable),
+				slog.String("error", err.Error()))
+		}
+		return 0, 0, fmt.Errorf("cursor error: %w", err)
+	}
+
 	return totalRecords, totalBytes, nil
 }
 


flow/connectors/mongo/qrep.go
```diff
@@ -0,0 +1,148 @@
...
+			Min bson.ObjectID `bson:"min"`
+			Max bson.ObjectID `bson:"max"`
+		} `bson:"_id"`
+		Count int `bson:"count"`
+	}
+	if err = cursor.All(ctx, &results); err != nil {
+		log.Fatal("Failed to start snapshot for collection:", err)
+	}
+
+	partitionHelepr := utils.NewPartitionHelper(c.logger)
```
heavycrystal:
```suggestion
	partitionHelper := utils.NewPartitionHelper(c.logger)
```

flow/connectors/mongo/qrep.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/model"
+	"github.com/PeerDB-io/peerdb/flow/shared/types"
+)
+
+func (c *MongoConnector) GetQRepPartitions(
+	ctx context.Context,
+	config *protos.QRepConfig,
+	last *protos.QRepPartition,
+) ([]*protos.QRepPartition, error) {
```
heavycrystal, line 27: users may want to specify other columns they have created indexes on, just for reference

jgao54: oh, do you mean specifying column other than `_id` for for bucketing/partitioning? or something else

heavycrystal: columns other than `_id` could be specified

flow/connectors/mongo/qrep.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return 0, 0, fmt.Errorf("unable to parse watermark table: %w", err)
+	}
+	collection := c.client.Database(parseWatermarkTable.Schema).Collection(parseWatermarkTable.Table)
+
+	filter := bson.D{}
+	if !partition.FullTablePartition {
+		filter = toRangeFilter(partition.Range)
+	}
+	opts1 := options.Find().SetBatchSize(1000)
```
heavycrystal: replace this with `config.NumRowsPerPartition` so it's controllable

flow/connectors/mongo/qrep.go
```diff
@@ -0,0 +1,185 @@
...
+	var size int64
+
+	var qvalueId types.QValueString
+	for _, v := range doc {
+		if v.Key == "_id" {
+			switch oid := v.Value.(type) {
+			case bson.ObjectID:
+				qvalueId = types.QValueString{Val: oid.Hex()}
+			default:
+				return nil, 0, fmt.Errorf("invalid type for _id field")
```
heavycrystal:
```suggestion
				return nil, 0, fmt.Errorf("unsupported type for _id field")
```

flow/connectors/utils/partition.go
```diff
@@ -128,6 +129,20 @@ func createUIntPartition(start uint64, end uint64) *protos.QRepPartition {
 	}
 }
 
+func createObjectIdPartition(start bson.ObjectID, end bson.ObjectID) *protos.QRepPartition {
+	return &protos.QRepPartition{
+		PartitionId: uuid.New().String(),
+		Range: &protos.PartitionRange{
+			Range: &protos.PartitionRange_ObjectIdRange{
+				ObjectIdRange: &protos.ObjectIdPartitionRange{
+					Start: start.Hex(),
```
heavycrystal, line 138: nit: would it be more efficient if stored as bytes?

flow/connectors/mongo/qrep.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	parseWatermarkTable, err := utils.ParseSchemaTable(config.WatermarkTable)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse watermark table: %w", err)
+	}
+	collection := c.client.Database(parseWatermarkTable.Schema).Collection(parseWatermarkTable.Table)
+
+	// TODO: support partitioning with bucket > 1
+	// TODO: support partitioning by columns other than _id
+	numBuckets := 1
```
heavycrystal, line 49: should be `numDocuments / numRowsPerPartition`

flow/connectors/mongo/qrep.go
jgao54: [[doc](https://www.mongodb.com/docs/manual/reference/method/cursor.batchSize/)]
> By default, the initial batch size is the lesser of 101 documents or 16 mebibytes (MiB) worth of documents. Subsequent batches have a maximum size of 16 MiB. This option can enforce a smaller limit than 16 MiB, but not a larger one. When set, the batchSize is the lesser of batchSize documents or 16 MiB worth of documents.

it's likely that config.NumRowsPerPartition will exceed the cursor's 16MiB default batch size anyways for most real-world collections.  16MiB seems quite large for a single query batch. BatchSize doesn't directly map to MiB, but it doesn't looks like mongo lets us configure it by MiB, only by document count.

flow/connectors/mongo/qrep.go
jgao54: ah yeah, i want to scope this PR to just  a single snapshot/bucket (hence the `// TODO: support partitioning with bucket > 1` comment). 

Will follow-up with supporting `numDocuments/numRowsPerPartition` in a separate PR

flow/connectors/utils/partition.go
jgao54: For object ID, string is more convenient for building the bson document for building query filter. It only supports either string or ObjectId(<string>), but not bytes. So using Hex() here for convenience.

flow/connectors/mongo/qrep.go
jgao54: I tested this with and without configuring batchSize, and it looks like the networking roundtrip has a non-trivial cost in perf. So updated PR to use `config.NumRowsPerPartition` as you recommended here.  and 16 MiB can function as a ceiling, in the case when per partition size > 16MiB.

PR #3045, opened 2025-06-12
persist go module cache and build cache during local development

jgao54:
Avoid re-downloading go modules from scratch every time. Also include build cache as well to speed up build.


jgao54:
oops, just realize this Dockerfile is shared in prod as well.  probably want to separate out the base image for dev

stacks/flow.Dockerfile
```diff
@@ -17,7 +17,9 @@ RUN rm -f go.work*
 # build the binary from flow folder
 WORKDIR /root/flow
 ENV CGO_ENABLED=1
-RUN go build -o /root/peer-flow
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
```
heavycrystal, line 21: wouldn't this be dependent on many things like compiler version and flags? wondering if it's worth it

stacks/flow.Dockerfile
```diff
@@ -8,7 +8,7 @@ WORKDIR /root/flow
 COPY flow/go.mod flow/go.sum ./
 
 # download all the dependencies
-RUN go mod download
+RUN --mount=type=cache,target=/go/pkg/mod go mod download
```
heavycrystal, line 11: iirc this wouldn't work for some reason, maybe things are better now

serprex (approved):

stacks/flow.Dockerfile
jgao54: this made a pretty big difference for me when tested ` ./dev-peerdb.sh` locally:
- without it you get a cold build which is 200s+
- tested out switch between branches with this change and took 15-20s to build

I'll punt on merging this for now though -- want to separate out the mount logic to only local development. in prod we probably benefit from clean build in general -- but it speeds up docker builds for me when testing locally with this change.

PR #3047, opened 2025-06-13
ch: consolidate schema when destination table missing columns

serprex:
when processing schema changes, we refresh schema from source

this can bring in changes we haven't reached yet in cdc,
which results in attempting to insert into non-existent columns on destination

detect this scenario (error code 16 from clickhouse) & add columns reactively


serprex:
this looks a bit wrong since it skips checking table mappings, but in this scenario I doubt table mappings are relevant

PR #3050, opened 2025-06-13
[clickhouse] refactor building queries for SyncRecords

heavycrystal:


flow/connectors/clickhouse/qrep_avro_sync.go
```diff
@@ -35,30 +35,41 @@ func NewClickHouseAvroSyncMethod(
...
 
-	sessionTokenPart := ""
+	params := make([]any, 0, 5)
+	params = append(params, avroFileUrl, creds.AWS.AccessKeyID, creds.AWS.SecretAccessKey)
 	if creds.AWS.SessionToken != "" {
-		sessionTokenPart = fmt.Sprintf(", '%s'", creds.AWS.SessionToken)
+		params = append(params, creds.AWS.SessionToken)
 	}
+	params = append(params, "Avro")
+	return fmt.Sprintf("s3(%s)", strings.Repeat("?, ", len(params)-1)+"?"), params, nil
```
serprex:
```suggestion
	return fmt.Sprintf("s3(%s?)", strings.Repeat("?,", len(params)-1)), params, nil
```

ilidemi:
@heavycrystal wouldn't we still want to log the query and have the S3 path present in it (but not the credentials)? Let's say an Avro file has an issue and we actually want to retrieve it with proper access.

PR #3051, opened 2025-06-13
ParsePgArrayToStringSlice: map NULL to empty string

serprex:
noticed by @ilidemi in #3040

### DIFF
--- a/flow/shared/postgres.go
+++ b/flow/shared/postgres.go
@@ -283,6 +283,7 @@ func ParsePgArrayStringToStringSlice(data string, delim byte) []string {
 func ParsePgArrayToStringSlice(data []byte, delim byte) []string {
 	var result []string
 	var sb []byte
+	var null int
 	ps := psSearch2
 	for _, ch := range data {
 		switch ps {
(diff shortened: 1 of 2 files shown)


flow/shared/postgres.go
```diff
@@ -313,18 +317,34 @@ func ParsePgArrayToStringSlice(data []byte, delim byte) []string {
...
+				}
 				sb = sb[:0]
+				null = 0
 				if ch == '}' {
 					ps = psSearch2
 				} else {
 					ps = psSearch
 				}
 			} else {
+				if null == 1 && ch == 'U' {
```
serprex: potential ideas:
1. could wrap this in "if null != 0" since that should generally be the case & avoid comparisons. Not worth 2 lines to me
2. could avoid appending N,U,L,L into sb, having to backfill sb when it turns out we aren't parsing NULL. Seems complicated, & appending these characters isn't allocating since we're reusing buffer
3. at first I wanted to have psN/psNU/psNUL/psNULL to represent the 4 states with non-zero null, instead of null variable. Would need `goto psUnquoted` to not repeat `psUnquoted` logic tho, & I'm sure @iskakaushik would push back against my second attempt to get `goto` into the codebase *(also sketched it out, it was quite a bit more verbose)*

nvm I couldn't help it, having the handling be isolated, avoiding spaghetti of bookkeeping another piece of state, is too enticing

ilidemi:
If we want to map nulls in PG to nulls in CH in the future, that change would have to touch this layer again. But not today

PR #3060, opened 2025-06-17
mongo: implement connector interface for fetching schemas

jgao54:
Implement connector interface for fetching databases, tables, and columns. For now, ~we use hard-coded _destination_ columns as placeholders.~ return empty columns.

Open discussion: since Mongo is schemaless, fetching column types becomes nontrivial:
- scanning the entire dataset and infer data types is too expensive
- a more practical approach would be to sample a small subset of rows and give a best-effort data types.

### DIFF
--- a/flow/connectors/mongo/schema.go
+++ b/flow/connectors/mongo/schema.go
@@ -1,0 +1,97 @@
+package connmongo
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"go.mongodb.org/mongo-driver/v2/bson"
+)
+
+func (c *MongoConnector) GetAllTables(ctx context.Context) (*protos.AllTablesResponse, error) {
+	tableNames := make([]string, 0)
+
+	dbNames, err := c.getAllDatabaseNames(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get databases: %w", err)
+	}
+	tableNames = append(tableNames, dbNames...)
+	return &protos.AllTablesResponse{
+		Tables: tableNames,
+	}, nil
+}
+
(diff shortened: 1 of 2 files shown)


flow/connectors/mongo/schema.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"go.mongodb.org/mongo-driver/v2/bson"
+)
+
+func (c *MongoConnector) GetAllTables(ctx context.Context) (*protos.AllTablesResponse, error) {
+	tableNames := make([]string, 0)
+
+	dbNames, err := c.getAllDatabaseNames(ctx)
```
heavycrystal, line 14: TODO: needs to fetch all collections instead

flow/connectors/mongo/schema.go
```diff
@@ -0,0 +1,97 @@
...
+			// TODO: implement TableSize fetching
+			TableSize: "",
+		}
+		response.Tables = append(response.Tables, tableResp)
+	}
+
+	return &response, nil
+}
+
+// TODO: replace placeholder values, how should we displace source columns in mongodb?
```
heavycrystal: I don't think we can easily, need to check UX of other systems

flow/connectors/mongo/schema.go
jgao54: derp, copy-pasta bug

flow/connectors/mongo/schema.go
jgao54: Maybe we should just set Type to empty string for now for mongo,  not displaying type is better than displaying wrong type. wdty?

jgao54: Updated code to return empty columns for now so we can punt on this decision

heavycrystal (approved):

PR #3061, opened 2025-06-17
support pgvector types as float32 array

serprex:


flow/shared/postgres/type_conversion.go
```diff
@@ -133,6 +133,8 @@ func CustomTypeToQKind(typeData shared.CustomDataType) types.QValueKind {
 		return types.QValueKindGeography
 	case "hstore":
 		return types.QValueKindHStore
+	case "vector", "halfvec", "sparsevec":
```
serprex, line 136: sparsevec might better be passed along as Map(Int32, Float32) but that would require a new QValueKind

flow/connectors/postgres/qrep.go
```diff
@@ -435,6 +435,10 @@ func syncQRepRecords(
 		return 0, fmt.Errorf("failed to register hstore: %w", err)
 	}
 
+	if err := shared.RegisterPgVector(ctx, txConn); err != nil {
```
serprex: Could pull what pgvector's register is doing & only grab the non-array types, then merge with RegisterHStore so it's only one query

.github/workflows/flow.yml
```diff
@@ -135,7 +140,10 @@ jobs:
 
       - name: create hstore extension, increase logical replication limits, and setup catalog database
```
ilidemi:
```suggestion
      - name: create postgres extensions, increase logical replication limits, and setup catalog database
```

.github/workflows/flow.yml
```diff
@@ -135,7 +140,10 @@ jobs:
...
         run: >
-          docker exec "${{ job.services.catalog.id }}" psql -U postgres -c "CREATE EXTENSION hstore;"
+          docker exec "${{ job.services.catalog.id }}" apk add --no-cache build-base git &&
+          docker exec "${{ job.services.catalog.id }}" git clone --branch v0.8.0 https://github.com/pgvector/pgvector.git /tmp/pgvector &&
+          docker exec "${{ job.services.catalog.id }}" sh -c 'cd /tmp/pgvector && make with_llvm=no && make with_llvm=no install' &&
```
ilidemi, line 145: Is there any way to do a binary installation? CI will be taking 12 seconds longer from here on

serprex: need to have it for pg 15/16/17, & this is alpine so [no PGDG](https://www.postgresql.org/download/linux)

PR #3064, opened 2025-06-18
TEMPORAL_CLIENT_CERT_PATH, TEMPORAL_CLIENT_KEY_PATH

serprex:
for when you don't want credentials in your environment variables,
or you want to be able to refresh client certs without reconnecting


flow/cmd/cert.go
```diff
@@ -26,3 +32,91 @@ func parseTemporalCertAndKey(ctx context.Context) ([]tls.Certificate, error) {
...
+
+		clientOptions.ConnectionOptions = client.ConnectionOptions{
+			TLS: &tls.Config{
+				GetClientCertificate: func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+					if keyPair != nil {
+						if certInfo, err := os.Stat(certPath); err == nil && certInfo.ModTime().Equal(certMtime) {
+							if keyInfo, err := os.Stat(keyPath); err == nil && keyInfo.ModTime().Equal(keyMtime) {
+								return keyPair, nil
+							}
+						}
```
iamKunalGupta: Secrets are mounted as 1/2 level symlinks and os.Stat might not work

Anyways Temporal uses a grpc connection (so I assume single TCP connection) and secrets are mounted in a ramdisk in kubernetes as far as I know so I assume it should not be an issue to simply re-read

iamKunalGupta (changes requested):

flow/cmd/cert.go
serprex: This function can be called multiple times for a connection, particularly with TLS 1.3

flow/cmd/cert.go
```diff
@@ -26,3 +30,50 @@ func parseTemporalCertAndKey(ctx context.Context) ([]tls.Certificate, error) {
...
+					if err != nil {
+						return nil, fmt.Errorf("could not read temporal client cert %s: %w", certPath, err)
+					}
+
+					keyBytes, err := os.ReadFile(keyPath)
+					if err != nil {
+						return nil, fmt.Errorf("could not read temporal client key %s: %w", keyPath, err)
+					}
+
+					keyPairValue, err := tls.X509KeyPair(certBytes, keyBytes)
```
iamKunalGupta: since we are not caching, we can do https://pkg.go.dev/crypto/tls#LoadX509KeyPair
```suggestion
					keyPairValue, err := tls.LoadX509KeyPair(certPath, keyPath)
```

iamKunalGupta (approved):

PR #3065, opened 2025-06-18
mongo: always  convert  `_id` to string  + no parallel initial load

jgao54:
Previously we assume that `_id` must be ObjectId type in mongoDB. 

However, it's possible for `_id` to contain other data types. To start,  keep `_id` simple by always converting to a string type in destination table. 

This PR also removed partitioning logic for now -- will evaluate `$bucketAuto` on large collections to determine if  it scales well. 

Lastly, this PR moved conversion logic to `qvalue_convert.go` so it can be used by both snapshot and cdc.

Testing:
- [x] manual local e2e testing (will add e2e test in subsequent PR)

### DIFF
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -24,8 +24,10 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
 
-const DefaultDocumentKeyColumnName = "_id"
-const DefaultFullDocumentColumnName = "_full_document"
+const (
+	DefaultDocumentKeyColumnName  = "_id"
+	DefaultFullDocumentColumnName = "_full_document"
+)
 
 type MongoConnector struct {
 	*metadataStore.PostgresMetadata
@@ -297,7 +299,7 @@ func (c *MongoConnector) PullRecords(
 		if documentKey, found := changeDoc["documentKey"]; found {
 			if len(documentKey.(bson.D)) == 0 || documentKey.(bson.D)[0].Key != DefaultDocumentKeyColumnName {
 				// should never happen
-				return fmt.Errorf("invalid document key, expect _id")
+				return errors.New("invalid document key, expect _id")
 			}
 			id := documentKey.(bson.D)[0].Value
 			qValue, err := qValueStringFromKey(id)
@@ -319,7 +321,7 @@ func (c *MongoConnector) PullRecords(
 		} else {
 			// should never happen with fullDocument='UpdateLookup',
 			// fail loudly for now so we can investigate if it does
-			return fmt.Errorf("fullDocument field not found")
+			return errors.New("fullDocument field not found")
 		}
 
 		if operationType, ok := changeDoc["operationType"]; ok {
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -24,6 +24,9 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
 
+const DefaultDocumentKeyColumnName = "_id"
+const DefaultFullDocumentColumnName = "_full_document"
+
 type MongoConnector struct {
 	*metadataStore.PostgresMetadata
 	config *protos.MongoConfig
@@ -122,13 +125,13 @@ func (c *MongoConnector) GetTableSchema(
 ) (map[string]*protos.TableSchema, error) {
 	result := make(map[string]*protos.TableSchema, len(tableMappings))
 	idFieldDescription := &protos.FieldDescription{
-		Name:         "_id",
+		Name:         DefaultDocumentKeyColumnName,
 		Type:         string(types.QValueKindString),
 		TypeModifier: -1,
 		Nullable:     false,
 	}
 	dataFieldDescription := &protos.FieldDescription{
-		Name:         "_full_document",
+		Name:         DefaultFullDocumentColumnName,
 		Type:         string(types.QValueKindJSON),
 		TypeModifier: -1,
 		Nullable:     false,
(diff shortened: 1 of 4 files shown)


flow/connectors/mongo/qvalue_convert.go
```diff
@@ -0,0 +1,57 @@
...
+	case bson.Timestamp:
+		// https://www.mongodb.com/docs/v7.0/reference/bson-types/#timestamps
+		ts := (uint64(v.T) << 32) | uint64(v.I)
+		val = strconv.FormatUint(ts, 10)
+	case bson.DateTime:
+		val = strconv.FormatInt(int64(v), 10)
+	case bson.Binary:
+		val = base64.StdEncoding.EncodeToString(v.Data)
+	case bson.D:
+		jsonb, err := bson.MarshalExtJSON(key, true, true)
```
jgao54: We probably want to handle this better recursively similar to how we properly handle `qValueJSONFromDocument`.

flow/connectors/mongo/qvalue_convert.go
```diff
@@ -0,0 +1,57 @@
...
+			return types.QValueString{}, fmt.Errorf("error marshalling key '%v': %w", key, err)
+		}
+		val = string(jsonb)
+	default:
+		return types.QValueString{}, fmt.Errorf("unexpected key '%v' with format: %T", key, key)
+	}
+	return types.QValueString{Val: val}, nil
+}
+
+func qValueJSONFromDocument(document bson.D) (types.QValueJSON, error) {
```
jgao54: This converts to Extended JSON type which is not ideal, but is just a temporary solution until we figure out a more performant JSON representation in clickhouse

flow/connectors/mongo/mongo.go
```diff
@@ -289,18 +292,33 @@ func (c *MongoConnector) PullRecords(
...
+			}
+			id := documentKey.(bson.D)[0].Value
+			qValue, err := qValueStringFromKey(id)
+			if err != nil {
+				return fmt.Errorf("failed to convert _id to string: %w", err)
+			}
+			items.AddColumn(DefaultDocumentKeyColumnName, qValue)
+		} else {
+			// should never happen
+			return fmt.Errorf("documentKey field not found")
```
```suggestion
			return errors.New("documentKey field not found")
```

flow/connectors/mongo/qrep.go
```diff
@@ -29,56 +27,58 @@ func (c *MongoConnector) GetQRepPartitions(
...
-	}
 
 	partitionHelper := utils.NewPartitionHelper(c.logger)
-	for _, result := range results {
-		if err := partitionHelper.AddPartition(result.ID.Min, result.ID.Max); err != nil {
-			return nil, fmt.Errorf("failed to add partition: %w", err)
-		}
-	}
-
+	// TODO: test against large collection to evaluate performance of bucketing logic; exclude partition logic for now.
```
serprex: TODO or remove code altogether

flow/connectors/mongo/qrep.go
```diff
@@ -181,31 +184,34 @@ func toRangeFilter(partitionRange *protos.PartitionRange) (bson.D, error) {
...
+	qvalueDoc, err := qValueJSONFromDocument(doc)
 	if err != nil {
-		return nil, 0, fmt.Errorf("failed to convert doc to json: %w", err)
+		return nil, 0, err
 	}
-	qvalueDoc := types.QValueJSON{Val: string(jsonb), IsArray: false}
-	size += int64(len(qvalueDoc.Val))
 	qValues = append(qValues, qvalueDoc)
 
+	// TODO: more accurate size calculation
```
serprex: TODO. also, why isn't this accurate? comment could say whether this will tend to be over/under

aside, but for mysql I'm intending to put a meter on the connection to count bytes on the wire, maybe you want that here too

flow/connectors/mongo/qvalue_convert.go
serprex: should put comment, with link to docs about extended json, why can't we give normal json?

flow/connectors/mongo/qrep.go
jgao54: ah, i was actually not sure how this size calculation is used (i.e. billing vs. custom metrics), and how mysql/pg measure this.

Currently, since this is using extendedJsonFormat it is more "bloated" in a way vs the raw byte representation (bson.raw). so not sure how we should implement this

flow/connectors/mongo/qrep.go
jgao54: will remove!

flow/connectors/mongo/qvalue_convert.go
jgao54: switching to json.Marshall seems sufficient for now

flow/connectors/mongo/qrep.go
jgao54: re: meter, good idea, will punt on it for now and do it via an observability-focused PR later!

flow/connectors/mongo/mongo.go
```diff
@@ -289,18 +292,34 @@ func (c *MongoConnector) PullRecords(
...
 		items := model.NewRecordItems(2)
-		items.AddColumn("_id",
-			types.QValueString{Val: documentKey[0].Value.(bson.ObjectID).Hex()})
-		if fullDocument, ok := changeDoc["fullDocument"]; ok {
-			fullDocJSON, err := bson.MarshalExtJSON(fullDocument, true, true)
+
+		if documentKey, found := changeDoc["documentKey"]; found {
+			if len(documentKey.(bson.D)) == 0 || documentKey.(bson.D)[0].Key != DefaultDocumentKeyColumnName {
+				// should never happen
+				return fmt.Errorf("invalid document key, expect _id")
```
```suggestion
				return errors.New("invalid document key, expect _id")
```

flow/connectors/mongo/mongo.go
```diff
@@ -289,18 +292,34 @@ func (c *MongoConnector) PullRecords(
...
-				return fmt.Errorf("failed to marshal fullDocument to JSON: %w", err)
+				return fmt.Errorf("failed to convert fullDocument to JSON: %w", err)
 			}
-			items.AddColumn("_full_document", types.QValueJSON{Val: string(fullDocJSON)})
+			items.AddColumn(DefaultFullDocumentColumnName, qValue)
 		} else {
-			items.AddColumn("_full_document", types.QValueJSON{Val: "{}"})
+			// should never happen with fullDocument='UpdateLookup',
+			// fail loudly for now so we can investigate if it does
+			return fmt.Errorf("fullDocument field not found")
```
```suggestion
			return errors.New("fullDocument field not found")
```

PR #3066, opened 2025-06-18
introduce mirror versioning, put pgvector support behind version gate

serprex:
this avoids breaking existing setups build on pgvector types coming through as strings

this system can be used in the future to maintain a chronology of breaking changes & not apply new behavior to old mirrors

it's meant to be a single dimension so that there isn't the exponential explosion of feature combinations we have with the dynconf system. We've previously tried to leverage by having breaking changes only be introduced as optional features, but this hits two problems: old behavior has to be default, & it's a complicated mechanism designed for flexible configuration, rather than a means to gate backwards incompatible changes from existing mirrors

this represents a point in PeerDB's life where we have people who just want the thing they tested (+ security fixes, etc)

infra for managing this system will be future work, outlined in #3095


ilidemi (approved):

flow/activities/flowable_core.go
```diff
@@ -264,6 +266,8 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 			StagingPath:            config.CdcStagingPath,
 			Script:                 config.Script,
 			TableNameSchemaMapping: tableNameSchemaMapping,
+			Env:                    config.Env,
```
ilidemi, line 269: Could this have a risk of something being in env that would break the code or change the behavior?

serprex: Minimal risk. Customers where this matters have tended to have us passing settings through global settings, where this won't change anything. Wide spread use is nullable, which we haven't noticed this error because source pretty much drives that

heavycrystal (approved):

PR #3068, opened 2025-06-19
[mongo] save a resumeToken before starting snapshot

heavycrystal:
so we don't lose changes that happen during snapshot
deduplication is accepted

### DIFF
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -195,8 +195,9 @@ func (c *MongoConnector) SetupReplication(ctx context.Context, input *protos.Set
 		} else if errors.Is(err, context.Canceled) {
 			c.logger.Info("SetupReplication context canceled, stopping change stream")
 		} else {
-			return model.SetupReplicationResult{}, fmt.Errorf("change stream error: %w", err)
+			c.logger.Error("PullRecords change stream error", "error", err)
 		}
+		return model.SetupReplicationResult{}, fmt.Errorf("change stream error: %w", err)
 	} else if resumeToken != nil {
 		if err := c.SetLastOffset(ctx, input.FlowJobName, model.CdcCheckpoint{
 			Text: base64.StdEncoding.EncodeToString(resumeToken),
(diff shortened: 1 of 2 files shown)


flow/connectors/mongo/mongo.go
```diff
@@ -166,7 +166,39 @@ func (c *MongoConnector) FinishExport(any) error {
...
+	var resumeToken bson.Raw
+	changeStream.ResumeToken()
+	for changeStream.Next(ctx) {
+		resumeToken = changeStream.ResumeToken()
+		if resumeToken != nil {
+			c.SetLastOffset(ctx, input.FlowJobName, model.CdcCheckpoint{
+				Text: base64.StdEncoding.EncodeToString(resumeToken),
+			})
+		}
+	}
```
jgao54: I think it should be sufficient to make a single call to `changeStream.ResumeToken()` and save it. curious what's the reason for iterating with `changeStream.Next()`?

heavycrystal: resumeToken looks like it can be `nil`, my understanding is that this is when we first start it and there are no "changes" yet but needs to be tested

jgao54: Ah i see, `changestream.Next(ctx)` caches the resumeToken. 

Nit: it looks like line 181 before the for-loop could be removed; and iiuc, we want to break when resumeToken is found.

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,391 @@
...
+	if err := changeStream.Err(); err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			c.logger.Info("SetupReplication context deadline exceeded, stopping change stream")
```
jgao54, line 193: this logs "stopping change stream", but function does not early returns, same with the "else if" branch below.

flow/connectors/mongo/mongo.go
```diff
@@ -47,6 +47,7 @@ func NewMongoConnector(ctx context.Context, config *protos.MongoConfig) (*MongoC
 		SetAppName("PeerDB Mongo Connector").
 		SetReadPreference(readpref.Primary()).
 		SetCompressors([]string{"zstd", "snappy"}).
+		SetDirect(true).
```
jgao54: I think this only works for specifying a single host in connection uri, which is fine for now. 
Later I suspect we'll need to support multiple hosts or srv uri  (with read preference set to Secondary to avoid hitting the primary)

jgao54 (approved):

PR #3069, opened 2025-06-19
Truncate and log unbounded numerics that don't fit into destination type

ilidemi:
While at it, unified the four places where we were deciding on which type to use at the destination.

Attempted to support truncation from wider to narrower decimals in case the customer creates such columns, but we seem to royally not support diverging schemas for numerics. If anyone in the future has to implement this, hit me up for a bunch of test cases.


flow/activities/flowable_core.go
```diff
@@ -264,6 +264,7 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 			StagingPath:            config.CdcStagingPath,
 			Script:                 config.Script,
 			TableNameSchemaMapping: tableNameSchemaMapping,
+			Env:                    config.Env,
```
ilidemi: Chat is this dangerous

serprex: spotted this miss in #3066 too

flow/activities/flowable_core.go
```diff
@@ -264,10 +264,14 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
...
 		})
 		if err != nil {
 			return a.Alerter.LogFlowError(ctx, flowName, fmt.Errorf("failed to push records: %w", err))
 		}
+		for _, message := range res.Messages {
+			a.Alerter.LogFlowInfo(ctx, flowName, message)
```
ilidemi: This seems like a warn material but there's no warn method. Error type is a string so should be easy to handle on PeerDB side but would need to check the UI handling

flow/e2e/clickhouse/peer_flow_ch_test.go
```diff
@@ -731,6 +734,229 @@ func (s ClickHouseSuite) Test_Unbounded_Numeric_Without_FF() {
 	s.testNumericFF(false)
 }
 
+func (s ClickHouseSuite) testNumericTruncation(unbNumAsStringFf bool) {
+	var pgSource *e2e.PostgresSource
+	var ok bool
+	if pgSource, ok = s.source.(*e2e.PostgresSource); !ok {
+		s.t.Skip("todo: only applies to postgres for now")
```
serprex: mysql doesn't have truncation issue

flow/model/numeric_truncator.go
```diff
@@ -0,0 +1,121 @@
+package model
+
+import (
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/model/qvalue"
+	"github.com/PeerDB-io/peerdb/flow/shared/types"
+)
+
+type StreamNumericTruncator struct {
+	TruncatorsByTable map[string]*CdcTableNumericTruncator
```
serprex: Seems like you could avoid a lot of `== nil` checks by making these maps by value instead of pointer

`map` in go is just a pointer, so having pointer to `struct { map }` is an unnecessary pointer-to-pointer

flow/e2e/clickhouse/peer_flow_ch_test.go
```diff
@@ -731,6 +734,229 @@ func (s ClickHouseSuite) Test_Unbounded_Numeric_Without_FF() {
...
+		sb.WriteString(fmt.Sprintf(",\ncol%d_arr %s[]", i, tc.SrcType))
+	}
+	sb.WriteString(")")
+
+	createQuery := sb.String()
+	_, err := s.Conn().Exec(s.t.Context(), createQuery)
+	require.NoError(s.t, err)
+
+	sb.Reset()
+	sb.WriteString(fmt.Sprintf("INSERT INTO %s(", srcFullName))
```
serprex:
```suggestion
	fmt.Fprintf(&sb, "INSERT INTO %s(", srcFullName)
```

flow/activities/flowable_core.go
ilidemi: Added warnings

flow/model/numeric_truncator.go
ilidemi: Removed unnecessary indirection for stream. Table is now a proper struct. Nil checks sadly have to persist because the return values are properly constructed.

flow/alerting/alerting.go
```diff
@@ -434,8 +434,10 @@ func (a *Alerter) LogNonFlowEvent(ctx context.Context, eventType telemetry.Event
 }
 
 // logFlowErrorInternal pushes the error to the errors table and emits a metric as well as a telemetry message
-func (a *Alerter) logFlowErrorInternal(ctx context.Context, flowName, errorType string, inErr error, inErrWithStack string) error {
+func (a *Alerter) logFlowErrorInternal(ctx context.Context, flowName, errorType string, inErr error, loggerFunc func(string, ...any)) error {
 	logger := internal.LoggerFromCtx(ctx)
+	inErrWithStack := fmt.Sprintf("%+v", inErr)
+	loggerFunc(inErr.Error(), slog.Any("stack", inErrWithStack))
```
serprex:
```suggestion
	loggerFunc(inErr.Error(), slog.String("stack", inErrWithStack))
```

flow/alerting/classifier.go
```diff
@@ -61,6 +61,10 @@ func (e ErrorSource) String() string {
 	return string(e)
 }
 
+func TableColumnErrorSource(destinationTable, destinationColumn string) ErrorSource {
+	return ErrorSource(fmt.Sprintf("column:%s.%s", destinationTable, destinationColumn))
```
iamKunalGupta: maybe we should prefix with `"avroConverter:column:"` in the error source?

flow/alerting/alerting.go
```diff
@@ -498,9 +499,23 @@ func (a *Alerter) LogFlowError(ctx context.Context, flowName string, inErr error
 	a.otelManager.Metrics.ErrorsEmittedCounter.Add(ctx, 1, errorAttributeSet)
 	a.otelManager.Metrics.ErrorEmittedGauge.Record(ctx, 1, errorAttributeSet)
 
+	return nil
+}
+
+func (a *Alerter) LogFlowError(ctx context.Context, flowName string, inErr error) error {
+	logger := internal.LoggerFromCtx(ctx)
+	err := a.logFlowErrorInternal(ctx, flowName, "error", inErr, logger.Error)
+	if err != nil {
```
iamKunalGupta: This will always be nil as `logFlowErrorInternal` never errors out. It is assumed that this part of the code should not interrupt normal operations

flow/connectors/clickhouse/qrep_avro_sync.go
```diff
@@ -124,43 +124,45 @@ func (s *ClickHouseAvroSyncMethod) SyncQRepRecords(
 	config *protos.QRepConfig,
 	partition *protos.QRepPartition,
 	stream *model.QRecordStream,
-) (int64, error) {
+) (int64, []error, error) {
```
iamKunalGupta: maybe alias `[]error` as `QrepWarnings` or something similar to be more clear?

PR #3080, opened 2025-06-23
mysql: more accurate bean counting

serprex:
cdc: only count row events which aren't ignored
qrep: meter connection rather than making terrible estimate based on decoded row values

based on talking upstream (https://github.com/go-mysql-org/go-mysql/issues/1044) this was the recommended approach

### DIFF
--- a/flow/connectors/mysql/metered_dialer.go
+++ b/flow/connectors/mysql/metered_dialer.go
@@ -1,5 +1,8 @@
 package connmysql
 
+/* go-mysql does not expose raw bytes for streaming selects,
+ * thus this allows accurately measuring fetched bytes */
+
 import (
 	"context"
 	"net"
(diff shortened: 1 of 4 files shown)


flow/connectors/mysql/metered_dialer.go
```diff
@@ -0,0 +1,37 @@
+package connmysql
```
heavycrystal, line 1: can we make this shared? Mongo could use this as well

serprex: was going to do that at start, but then `client.Dialer` comes from mysql package (amusingly pgx has the same Dialer interface)

MeteredConn can be copied over shared in mongo PR depending on interface used there / whether this ultimately proves useful for mongo

iskakaushik (approved):

PR #3091, opened 2025-06-23
use net.JoinHostPort

serprex:
properly handles IPv6 address as host

### DIFF
--- a/flow/shared/string.go
+++ b/flow/shared/string.go
@@ -1,8 +1,12 @@
 package shared
 
 import (
+	"net"
 	"regexp"
+	"strconv"
 	"unsafe"
+
+	"golang.org/x/exp/constraints"
 )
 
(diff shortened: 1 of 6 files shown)


flow/shared/string.go
```diff
@@ -25,3 +29,7 @@ func ReplaceIllegalCharactersWithUnderscores(s string) string {
 func IsValidReplicationName(s string) bool {
 	return reLegalIdentifierLower.MatchString(s)
 }
+
+func JoinHostPort[I constraints.Integer](host string, port I) string {
```
heavycrystal, line 33: nit: instead of importing an `x/exp` package can we define our own constraint?

serprex: perhaps, but we already use `x/exp` elsewhere so this isn't a new dependency

PR #3070, opened 2025-06-19
mysql: update checkpoint in catalog once an hour while waiting for records

serprex:
avoids pause on inactive mirror resuming with invalid binlog errors

PR #3092, opened 2025-06-23
mongo

jgao54:
Merge mongo branch to main to keep branch manageable. Note that some of these commits have been reviewed when merged into `mongo`.

This PR implements mongo connector with packed mode and handles:
- initial snapshot
- CDC for insert/update/replace/delete event implemented

Testing:
- added e2e tests
- ran peerdb mirror from mongo -> clickhouse locally

UI PR is separately created [here](https://github.com/PeerDB-io/peerdb/pull/3054) if you want to try this locally.

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -69,7 +69,7 @@ func (a *FlowableActivity) CheckConnection(
 	config *protos.SetupInput,
 ) error {
 	ctx = context.WithValue(ctx, shared.FlowNameKey, config.FlowName)
-	conn, err := connectors.GetByNameAs[connectors.CDCPullConnector](ctx, config.Env, a.CatalogPool, config.PeerName)
+	conn, err := connectors.GetByNameAs[connectors.Connector](ctx, config.Env, a.CatalogPool, config.PeerName)
 	if err != nil {
 		if errors.Is(err, errors.ErrUnsupported) {
 			return nil
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -69,7 +69,7 @@ func (a *FlowableActivity) CheckConnection(
 	config *protos.SetupInput,
 ) error {
 	ctx = context.WithValue(ctx, shared.FlowNameKey, config.FlowName)
-	conn, err := connectors.GetByNameAs[connectors.CDCSyncConnector](ctx, config.Env, a.CatalogPool, config.PeerName)
+	conn, err := connectors.GetByNameAs[connectors.CDCPullConnector](ctx, config.Env, a.CatalogPool, config.PeerName)
 	if err != nil {
 		if errors.Is(err, errors.ErrUnsupported) {
 			return nil
--- a/flow/cmd/peer_data.go
+++ b/flow/cmd/peer_data.go
@@ -94,8 +94,8 @@ func (h *FlowRequestHandler) ListPeers(
 ) (*protos.ListPeersResponse, error) {
 	query := "SELECT name, type FROM peers"
 	if internal.PeerDBOnlyClickHouseAllowed() {
-		// only postgres, mysql, and clickhouse
-		query += " WHERE type IN (3, 7, 8)"
+		// only postgres, mysql, mongo,and clickhouse
+		query += " WHERE type IN (2, 3, 7, 8)"
 	}
 	rows, err := h.pool.Query(ctx, query)
 	if err != nil {
@@ -114,10 +114,10 @@ func (h *FlowRequestHandler) ListPeers(
 	sourceItems := make([]*protos.PeerListItem, 0, len(peers))
 	destinationItems := make([]*protos.PeerListItem, 0, len(peers))
 	for _, peer := range peers {
-		if peer.Type == protos.DBType_POSTGRES || peer.Type == protos.DBType_MYSQL {
+		if peer.Type == protos.DBType_POSTGRES || peer.Type == protos.DBType_MYSQL || peer.Type == protos.DBType_MONGO {
 			sourceItems = append(sourceItems, peer)
 		}
-		if peer.Type != protos.DBType_MYSQL && (!internal.PeerDBOnlyClickHouseAllowed() || peer.Type == protos.DBType_CLICKHOUSE) {
+		if peer.Type != protos.DBType_MYSQL && peer.Type != protos.DBType_MONGO && (!internal.PeerDBOnlyClickHouseAllowed() || peer.Type == protos.DBType_CLICKHOUSE) {
 			destinationItems = append(destinationItems, peer)
 		}
 	}
--- a/flow/cmd/peer_data.go
+++ b/flow/cmd/peer_data.go
@@ -117,7 +117,8 @@ func (h *FlowRequestHandler) ListPeers(
 		if peer.Type == protos.DBType_POSTGRES || peer.Type == protos.DBType_MYSQL || peer.Type == protos.DBType_MONGO {
 			sourceItems = append(sourceItems, peer)
 		}
-		if peer.Type != protos.DBType_MYSQL && peer.Type != protos.DBType_MONGO && (!internal.PeerDBOnlyClickHouseAllowed() || peer.Type == protos.DBType_CLICKHOUSE) {
+		if peer.Type != protos.DBType_MYSQL &&
+			peer.Type != protos.DBType_MONGO && (!internal.PeerDBOnlyClickHouseAllowed() || peer.Type == protos.DBType_CLICKHOUSE) {
 			destinationItems = append(destinationItems, peer)
 		}
 	}
--- a/flow/cmd/peer_data.go
+++ b/flow/cmd/peer_data.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"fmt"
 	"log/slog"
 	"time"
 
@@ -95,7 +96,8 @@ func (h *FlowRequestHandler) ListPeers(
 	query := "SELECT name, type FROM peers"
 	if internal.PeerDBOnlyClickHouseAllowed() {
 		// only postgres, mysql, mongo,and clickhouse
-		query += " WHERE type IN (2, 3, 7, 8)"
+		query += fmt.Sprintf(" WHERE type IN (%d,%d,%d,%d)",
+			protos.DBType_POSTGRES, protos.DBType_MYSQL, protos.DBType_MONGO, protos.DBType_CLICKHOUSE)
 	}
 	rows, err := h.pool.Query(ctx, query)
 	if err != nil {
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -70,7 +70,6 @@ func (c *MongoConnector) Close() error {
 }
 
 func (c *MongoConnector) ConnectionActive(ctx context.Context) error {
-	// Use a ping command to check if the connection is active
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
@@ -112,7 +111,10 @@ func (c *MongoConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.F
 	if _, ok := result["set"]; !ok {
 		return errors.New("MongoDB is not configured as a replica set, which is required for CDC")
 	}
-	if myState, ok := result["myState"].(int32); !ok || myState != 1 {
+
+	if myState, ok := result["myState"]; !ok {
+		return errors.New("myState not found in response")
+	} else if myStateInt, ok := myState.(int32); !ok || myStateInt != 1 {
 		return fmt.Errorf("MongoDB is not the primary node in the replica set, current state: %d", myState)
 	}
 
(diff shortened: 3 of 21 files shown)


flow/activities/flowable.go
```diff
@@ -69,7 +69,7 @@ func (a *FlowableActivity) CheckConnection(
 	config *protos.SetupInput,
 ) error {
 	ctx = context.WithValue(ctx, shared.FlowNameKey, config.FlowName)
-	conn, err := connectors.GetByNameAs[connectors.CDCSyncConnector](ctx, config.Env, a.CatalogPool, config.PeerName)
+	conn, err := connectors.GetByNameAs[connectors.CDCPullConnector](ctx, config.Env, a.CatalogPool, config.PeerName)
```
```suggestion
	conn, err := connectors.GetByNameAs[connectors.Connector](ctx, config.Env, a.CatalogPool, config.PeerName)
```

flow/cmd/peer_data.go
```diff
@@ -94,8 +94,8 @@ func (h *FlowRequestHandler) ListPeers(
 ) (*protos.ListPeersResponse, error) {
 	query := "SELECT name, type FROM peers"
 	if internal.PeerDBOnlyClickHouseAllowed() {
-		// only postgres, mysql, and clickhouse
-		query += " WHERE type IN (3, 7, 8)"
+		// only postgres, mysql, mongo,and clickhouse
+		query += " WHERE type IN (2, 3, 7, 8)"
```
serprex:
```suggestion
		query += fmt.Sprintf(" WHERE type IN (%d,%d,%d,%d)",
			protos.DBType_POSTGRES, protos.DBType_MYSQL, protos.DBType_MONGO, protos.DBType_CLICKHOUSE)
```

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+		// Use a timeout to ensure the disconnect operation does not hang indefinitely
+		timeout, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+		return c.client.Disconnect(timeout)
+	}
+	return nil
+}
+
+func (c *MongoConnector) ConnectionActive(ctx context.Context) error {
+	// Use a ping command to check if the connection is active
```
```suggestion
```

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+	}).Decode(&result); err != nil {
+		return fmt.Errorf("failed to get replica set status: %w", err)
+	}
+
+	// Check if this is a replica set
+	if _, ok := result["set"]; !ok {
+		return errors.New("MongoDB is not configured as a replica set, which is required for CDC")
+	}
+	if myState, ok := result["myState"].(int32); !ok || myState != 1 {
+		return fmt.Errorf("MongoDB is not the primary node in the replica set, current state: %d", myState)
```
serprex, line 120: myState will be 0 here when it failed to cast to int32. Should have !ok error separate from branch using myState

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+func (c *MongoConnector) ExportTxSnapshot(context.Context, map[string]string) (*protos.ExportTxSnapshotOutput, any, error) {
+	return nil, nil, nil
+}
+
+func (c *MongoConnector) FinishExport(any) error {
+	return nil
+}
+
+func (c *MongoConnector) SetupReplication(ctx context.Context, input *protos.SetupReplicationInput) (model.SetupReplicationResult, error) {
+	// TODO: need to think of the situation where the database is idle and we don't "get" a resume token.
```
serprex: TODO

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+	if err != nil {
+		return model.SetupReplicationResult{}, fmt.Errorf("failed to start change stream for storing initial resume token: %w", err)
+	}
+	defer changeStream.Close(ctx)
+
+	var resumeToken bson.Raw
+	c.logger.Info("SetupReplication started, waiting for initial resume token",
+		slog.String("flowJobName", input.FlowJobName))
+	for resumeToken = changeStream.ResumeToken(); resumeToken == nil && changeStream.Next(ctx); {
+	}
```
```suggestion
	c.logger.Info("SetupReplication started, waiting for initial resume token",
		slog.String("flowJobName", input.FlowJobName))
	var resumeToken bson.Raw
	for {
		resumeToken = changeStream.ResumeToken()
		if resumeToken != nil {
			break
		} else if !changeStream.Next(ctx) {
			break
		}
	}
```

as is resumeToken never updates & the loop goes on until changeStream completes, leaving resumeToken nil

if we reach end of changeStream without resumeToken is this an error? code suggestion splits code paths since I'm not sure if maybe one of those paths should error. Or at least log warning

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+
+func (c *MongoConnector) ExportTxSnapshot(context.Context, map[string]string) (*protos.ExportTxSnapshotOutput, any, error) {
+	return nil, nil, nil
+}
+
+func (c *MongoConnector) FinishExport(any) error {
+	return nil
+}
+
+func (c *MongoConnector) SetupReplication(ctx context.Context, input *protos.SetupReplicationInput) (model.SetupReplicationResult, error) {
```
serprex, line 168: not a stub

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+func (c *MongoConnector) PullRecords(
+	ctx context.Context,
+	catalogPool shared.CatalogPool,
+	otelManager *otel_metrics.OtelManager,
+	req *model.PullRecordsRequest[model.RecordItems],
+) error {
+	defer req.RecordStream.Close()
+	c.logger.Info("[started] PullRecords for mirror "+req.FlowJobName,
+		slog.Any("table_mapping", req.TableNameMapping),
+		slog.Int("max_batch_size", int(req.MaxBatchSize)),
```
```suggestion
		slog.Uint64("max_batch_size", uint64(req.MaxBatchSize)),
```

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+		slog.Duration("idle_timeout", req.IdleTimeout))
+
+	changeStreamOpts := options.ChangeStream().
+		SetComment("PeerDB changeStream for mirror " + req.FlowJobName).
+		SetFullDocument(options.UpdateLookup).
+		SetFullDocumentBeforeChange(options.WhenAvailable)
+	if req.LastOffset.Text != "" {
+		// If we have a last offset, we resume from that point
+		c.logger.Info("Resuming change stream from offset: " + req.LastOffset.Text)
```
```suggestion
		c.logger.Info("[mongo] resuming change stream", slog.String("resumeToken", req.LastOffset.Text))
```
best to use the source's terminology in logs etc

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+	var recordCount uint32
+	defer func() {
+		if recordCount == 0 {
+			req.RecordStream.SignalAsEmpty()
+		}
+		c.logger.Info(fmt.Sprintf("[finished] PullRecords streamed %d records", recordCount))
+	}()
+	// before first record, we wait indefinitely so give it ctx
+	// after first record, we wait for idle timeout
+	getCtx := ctx
```
serprex, line 307: should update this with improvements in #3070

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+			c.logger.Warn("operationType field not found")
+			continue
+		} else if operationType != "insert" && operationType != "update" && operationType != "replace" && operationType != "delete" {
+			continue
+		}
+
+		clusterTime := changeDoc["clusterTime"].(bson.Timestamp)
+		clusterTimeNanos := time.Unix(int64(clusterTime.T), 0).UnixNano()
+
+		sourceTableName := changeDoc["ns"].(bson.D)[0].Value.(string) + "." + changeDoc["ns"].(bson.D)[1].Value.(string)
```
```suggestion
		sourceTableName := fmt.Sprintf("%s.%s", changeDoc["ns"].(bson.D)[0].Value, changeDoc["ns"].(bson.D)[1].Value)
```

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+			case "insert":
+				if err := addRecord(ctx, &model.InsertRecord[model.RecordItems]{
+					BaseRecord:           model.BaseRecord{CommitTimeNano: clusterTimeNanos},
+					Items:                items,
+					SourceTableName:      sourceTableName,
+					DestinationTableName: destinationTableName,
+				}); err != nil {
+					return fmt.Errorf("failed to add insert record: %w", err)
+				}
+			case "update", "replace":
```
serprex, line 390: looks like we're not going to really support update if I'm reading this right: update is for partial modifications, whereas replace is a complete update

is this something the user needs to configure?

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,431 @@
...
+				}
+			default:
+				return fmt.Errorf("unsupported operationType: %s", operationType)
+			}
+		}
+	}
+	if resumeToken := changeStream.ResumeToken(); resumeToken != nil {
+		// Update the last offset with the resume token
+		req.RecordStream.UpdateLatestCheckpointText(base64.StdEncoding.EncodeToString(resumeToken))
+		c.logger.Info("Updated last offset to: " + req.LastOffset.Text)
```
```suggestion
		c.logger.Info("[mongo] latest resume token", slog.String("resumeToken", req.LastOffset.Text))
```
"updated" is a bit misleading: that should be reserved for when this is persisted to catalog

flow/connectors/mongo/schema.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (c *MongoConnector) GetColumns(ctx context.Context, version uint32, schema string, table string) (*protos.TableColumnsResponse, error) {
+	return &protos.TableColumnsResponse{
+		Columns: []*protos.ColumnsItem{},
+	}, nil
+}
+
+// Get all database names, but excluding MongoDB's default databases
+func (c *MongoConnector) getAllDatabaseNames(ctx context.Context) ([]string, error) {
+	// TODO: investigate why query fails when this logic is added to the filter
```
serprex: TODO

flow/connectors/mongo/schema.go
```diff
@@ -0,0 +1,94 @@
...
+	excludedDatabaseSet := map[string]bool{
+		"config": true,
+		"admin":  true,
+		"local":  true,
+	}
```
serprex:
```suggestion
	excludedDatabaseSet := map[string]struct{}{
		"config": {},
		"admin":  {},
		"local":  {},
	}
```

flow/connectors/mongo/schema.go
```diff
@@ -0,0 +1,94 @@
...
+	filter := bson.D{}
+	dbs, err := c.client.ListDatabaseNames(ctx, filter)
+	if err != nil {
+		return nil, err
+	}
+	var filteredDbNames []string
```
serprex:
```suggestion
	filteredDbNames := make([]string, 0, len(dbs))
```

flow/connectors/mongo/mongo.go
jgao54: > update is for partial modifications, whereas replace is a complete update

correct. 
We are supporting both. Regardless whether the change event is update or replace, mongo will perform a fullDocument lookup after the change event. 

> is this something the user needs to configure?

User doesn't need to configure on their end, but we need to explicitly enable it in our code  [[ref](https://www.mongodb.com/docs/manual/changeStreams/#lookup-full-document-for-update-operations)]:
```
SetFullDocument(options.UpdateLookup)
```

The caveat is that unlike postgres/mysql, this result does not necessarily provide the document right after the update is applied, as there may be other writes that happened after the update but before the lookup.

flow/connectors/mongo/mongo.go
jgao54: Reading method doc for `changestream.Next()`:
> If Next returns false, subsequent calls will also return false.

It looks like errors in changestream are unrecoverable.  So seem to suggest we should just return the changestream error whenever changestream.Next() returns false.

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,428 @@
...
+
+	// Check if this is a replica set
+	if _, ok := result["set"]; !ok {
+		return errors.New("MongoDB is not configured as a replica set, which is required for CDC")
+	}
+
+	if myState, ok := result["myState"]; !ok {
+		return errors.New("myState not found in response")
+	} else if myStateInt, ok := myState.(int32); !ok || myStateInt != 1 {
+		return fmt.Errorf("MongoDB is not the primary node in the replica set, current state: %d", myState)
```
serprex, line 120: will get odd log where we say 0 when myState isn't int32, need to split this up

flow/connectors/mongo/mongo.go
heavycrystal: isn't that what current code is doing? ```	if err := changeStream.Err(); err != nil {```

flow/connectors/mongo/mongo.go
heavycrystal: to be done as a follow up

flow/connectors/mongo/mongo.go
jgao54: removed this comment, iirc @heavycrystal tested this and did not find this to be an issue. Although i'm not sure how the magic here worked under the hood -- i.e. if there's no change events, what populated the resume token here?

flow/connectors/mongo/mongo.go
jgao54: Created https://github.com/PeerDB-io/peerdb/issues/3104 to track

flow/connectors/mongo/mongo.go
heavycrystal: the resumeToken is an encoded BSON document that had a pointer into oplog iirc, I guess that's still possible when idle

jgao54: Looking at the [code](https://github.com/mongodb/mongo-go-driver/blob/0e23de1ff0c636bc109d5f12fdf14d55d2b3b4ed/mongo/change_stream.go#L409-L430)/[doc](https://www.mongodb.com/docs/manual/changeStreams/#resume-tokens-from-change-events), resume token is the `_id` field of the change stream event.  So what's not clear to me is if there's no change event, what alternative mechanism populated resumeToken. I would have expected resumeToken to be nil when there's no change event (there's also  aggregation/GetMore that generate resumeToken, although neither of which seems relevant here)

flow/connectors/mongo/mongo.go
jgao54: I think the bug @serprex  is pointing out in the original code is that `for resumeToken = changeStream.ResumeToken()` will assign the value to `resumeToken` _once_ only; where as the proposed change calls ResumeToken() on every iteration, which is what we want.

flow/connectors/mongo/mongo.go
```diff
@@ -0,0 +1,433 @@
...
+		// If we have a last offset, we resume from that point
+		c.logger.Info("[mongo] resuming change stream", slog.String("resumeToken", req.LastOffset.Text))
+		resumeTokenBytes, err := base64.StdEncoding.DecodeString(req.LastOffset.Text)
+		if err != nil {
+			return fmt.Errorf("failed to parse last offset: %w", err)
+		}
+		changeStreamOpts.SetResumeAfter(bson.Raw(resumeTokenBytes))
+	}
+
+	changeStream, err := c.client.Watch(ctx, mongo.Pipeline{}, changeStreamOpts)
```
jgao54, line 286: Reading this code, it looks like we are creating a new changestream every time `PullRecords` gets called. I think ideally we'd want to create a long-lived changestream object once, and call changestream.Next(...) on it each time `PullRecords` is called.  There's [concerns](https://www.mongodb.com/docs/manual/changeStreams/#change-stream-performance-considerations) around having too many changestream opened against the database.

jgao54: ah, separately, according to the link:
> Each change stream uses a connection and a [getMore](https://www.mongodb.com/docs/manual/reference/command/getMore/#std-label-manual-reference-commands-getMore) operation on the change stream for the period of time that it waits for the next event.

I think this may be the source of `resumeToken` when there's no change event.

jgao54: tracked: https://github.com/PeerDB-io/peerdb/issues/3106

flow/connectors/mongo/mongo.go
heavycrystal: ah, good catch

PR #3093, opened 2025-06-24
[ignore] mongo: testing e2e

jgao54:
testing e2e for mongo connector


flow/connectors/mongo/qrep.go
```diff
@@ -141,7 +141,7 @@ func toRangeFilter(partitionRange *protos.PartitionRange) (bson.D, error) {
 	}
 }
 
-func qValuesFromDocument(doc bson.D) ([]types.QValue, int64, error) {
+func (c *MongoConnector) QValuesFromDocument(doc bson.D) ([]types.QValue, int64, error) {
```
serprex:
```suggestion
func QValuesFromDocument(doc bson.D) ([]types.QValue, int64, error) {
```
doesn't use connector at all

flow/connectors/mongo/qrep.go
```diff
@@ -111,7 +111,7 @@ func (c *MongoConnector) PullQRepRecords(
 	return totalRecords, totalBytes, nil
 }
 
-func getDefaultSchema() types.QRecordSchema {
+func (c *MongoConnector) GetDefaultSchema() types.QRecordSchema {
```
serprex:
```suggestion
func GetDefaultSchema() types.QRecordSchema {
```

.github/workflows/flow.yml
```diff
@@ -228,6 +242,8 @@ jobs:
           ELASTICSEARCH_TEST_ADDRESS: http://localhost:9200
           CI_PG_VERSION: ${{ matrix.db-version.pg }}
           CI_MYSQL_VERSION: ${{ matrix.db-version.mysql }}
+          CI_MONGO_VERSION: 8.0.10
+          CI_MONGO_URI: mongodb://localhost:27017/?replicaSet=rs0
```
heavycrystal, line 246: nit: I'd prefer this has a username/password so we test authentication
and why does `replicaSet` need specifying?

.github/workflows/flow.yml
```diff
@@ -228,6 +242,8 @@ jobs:
           ELASTICSEARCH_TEST_ADDRESS: http://localhost:9200
           CI_PG_VERSION: ${{ matrix.db-version.pg }}
           CI_MYSQL_VERSION: ${{ matrix.db-version.mysql }}
+          CI_MONGO_VERSION: 8.0.10
```
heavycrystal, line 245: what's the point of this if version is hardcoded in the `docker run` command?

flow/e2e/mongo/mongo.go
```diff
@@ -0,0 +1,79 @@
...
+	db := s.conn.Client().Database(GetTestDatabase(suffix))
+	_ = db.Drop(t.Context())
+}
+
+func (s *MongoSource) Connector() connectors.Connector {
+	return s.conn
+}
+
+func (s *MongoSource) Exec(ctx context.Context, sql string) error {
+	return errors.New("not supported")
```
heavycrystal: nit: use `errors.ErrUnsupported` instead of bespoke ones

.github/workflows/flow.yml
jgao54: Starting with this, but plan to add more version later to the CI matrix to cover all the versions we want to support :)

.github/workflows/flow.yml
jgao54: For mongo, it's recommended to include replicaSet in the connection string URI as best practice for Replica Set setup.  Without it, Mongo will treat the connection as if connecting to a standalone instance, which won't work when we switch read preference from primary to Secondary/SecondaryPreferred.

re: auth. I was fiddling with username/password earlier but encountered some issues. Created [a ticket](https://github.com/PeerDB-io/peerdb/issues/3100) to look into authentication in a [subsequent PR](https://github.com/PeerDB-io/peerdb/pull/3102).

PR #3094, opened 2025-06-24
ch cleanup

serprex:
1. quote identifiers in rename
2. don't use sql.Null for non-nullable results

### DIFF
--- a/flow/connectors/clickhouse/cdc.go
+++ b/flow/connectors/clickhouse/cdc.go
@@ -2,8 +2,6 @@ package connclickhouse
 
 import (
 	"context"
-	"database/sql"
-	"errors"
 	"fmt"
 	"log/slog"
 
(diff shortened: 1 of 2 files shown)


flow/connectors/clickhouse/cdc.go
```diff
@@ -54,8 +48,7 @@ func (c *ClickHouseConnector) CreateRawTable(ctx context.Context, req *protos.Cr
 		_peerdb_data String,
 		_peerdb_record_type Int,
 		_peerdb_match_data String,
-		_peerdb_batch_id Int64,
-		_peerdb_unchanged_toast_columns String
```
ilidemi: Intended?

serprex: odd, `_peerdb_batch_id` isn't removed on my system, not sure what GitHub's doing

serprex: yes, updated PR description

serprex: reverted

ilidemi (approved):

PR #3102, opened 2025-06-25
mongo: test mongo versions 6.0/7.0/8.0 + authentication

jgao54:
- Test mongo versions 6.0/7.0/8.0 pass CI
- Create dedicated user for changestream and test authentication works as expected.

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -263,7 +263,6 @@ jobs:
           ELASTICSEARCH_TEST_ADDRESS: http://localhost:9200
           CI_PG_VERSION: ${{ matrix.db-version.pg }}
           CI_MYSQL_VERSION: ${{ matrix.db-version.mysql }}
-          CI_MONGO_VERSION: ${{ matrix.db-version.mongo }}
           CI_MONGO_URI: mongodb://ciuser:cipass@localhost:27017/?replicaSet=rs0&authSource=admin
           ENABLE_OTEL_METRICS: ${{ (matrix.db-version.pg == '16' || matrix.db-version.mysql == 'mysql-pos') && 'true' || 'false' }}
           OTEL_EXPORTER_OTLP_METRICS_ENDPOINT: http://localhost:4317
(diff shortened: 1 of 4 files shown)


.github/workflows/flow.yml
```diff
@@ -131,18 +131,39 @@ jobs:
...
+              roles: ["root"]
+            })'
+
+          echo "create non-admin user for tests"
+          docker exec mongo mongosh -u admin -p admin --eval '
+            db = db.getSiblingDB("admin");
+            db.createUser({
+              user: "ciuser",
+              pwd: "cipass",
+              roles: ["readWriteAnyDatabase"]
```
jgao54: `readAnyDatabases` may be the less privileged role we need to make mongo connector work as expected. since we need `ListDatabases` privileges during discovery. [[ref](https://www.mongodb.com/docs/manual/reference/built-in-roles/#all-database-roles)]

~for CI, we use `readWriteAnyDatabase` since we need to insert records for testing.~
Update: use admin user for writing -- so we can keep change stream user scoped to readonly.

jgao54: Ah nvm, user can provision just the [`clusterMonitoring`](https://www.mongodb.com/docs/manual/reference/built-in-roles/#mongodb-authrole-clusterMonitor)  role along with read role on specific databases if they wish to scope the permissions -- but for e2e our dbs names are randomly generated so we can just `readAnyDatabase` for now.

.github/workflows/flow.yml
```diff
@@ -242,8 +263,8 @@ jobs:
           ELASTICSEARCH_TEST_ADDRESS: http://localhost:9200
           CI_PG_VERSION: ${{ matrix.db-version.pg }}
           CI_MYSQL_VERSION: ${{ matrix.db-version.mysql }}
-          CI_MONGO_VERSION: 8.0.10
-          CI_MONGO_URI: mongodb://localhost:27017/?replicaSet=rs0
+          CI_MONGO_VERSION: ${{ matrix.db-version.mongo }}
```
serprex:
```suggestion
```
checking code, this isn't necessary

heavycrystal (approved):

PR #3108, opened 2025-06-27
mongo: handle large mongo document

jgao54:
This PR handles two issues w.r.t. large event (separated by commits, recommend review them separately):

1) Mongo limits document size to 16MB, trying to insert even one byte larger will result in error:
>  [object to insert too large. size in bytes: xxx, max size: xxx]

With this information I was able to generate document in test that is exactly the maximum size that we can write to mongodb -- and we should support up to this size in our CDC connector without errors.

Currently we seem to be imposing snowflake's 15MB limit in the shared `MarshalJSONWithOptions` function, so the e2e test  added to this PR was failing because the json got set to "{}". With this PR, Mongo is made an exception (although i think the current implementation is sloppy since it's based on source data type rather than destination datatypes).  Looking for feedback on a better solution, or if we can get ride of the 15MB limit altogether in `MarshalJSONWithOptions` given `QValueToAvro` seem to handle snowflake's edge cases already -- but not sure if I'm missing something)

2) With above change, inserts/replace/delete events processed successfully during CDC, but update events was still failing because it was double the size as the other events and change stream also has a 16MB limitation on event size (because the value was present at least twice in the event: once in the `fullDocument` field and once in the `updateDescription` field, not to mention it could also be present in the `fullDocumentBeforeChange` field if available). So the PR:
- set `fullDocumentBeforeChange` to `off` always, which we are not using
- make the pipeline filter out `updateDescription` field, which we are not using
-  restrict to only event type we care about so far (insert/update/delete/replace) and only to fields we need (operatinType/clusterTime/documentKey/fullDocument/ns) so far.

Intentionally not using `$changeStreamSplitLargeEvent` for now as it seems quite discouraged by [mongo doc](https://www.mongodb.com/docs/manual/reference/operator/aggregation/changeStreamSplitLargeEvent/) -- we can support it if current implementation is no longer sufficient.

Test: e2e test passing with this change.

### DIFF
--- a/flow/model/record_items.go
+++ b/flow/model/record_items.go
@@ -25,14 +25,21 @@ func ItemsToJSON(items Items) (string, error) {
 
 // encoding/gob cannot encode unexported fields
 type RecordItems struct {
-	ColToVal   map[string]types.QValue
-	NoTruncate bool
+	ColToVal               map[string]types.QValue
+	TruncateThresholdBytes int
 }
 
-func NewRecordItems(capacity int, noTruncate bool) RecordItems {
+func NewRecordItems(capacity int) RecordItems {
 	return RecordItems{
-		ColToVal:   make(map[string]types.QValue, capacity),
-		NoTruncate: noTruncate,
+		ColToVal:               make(map[string]types.QValue, capacity),
+		TruncateThresholdBytes: 15 * 1024 * 1024,
+	}
+}
+
+func NewMongoRecordItems(capacity int) RecordItems {
+	return RecordItems{
+		ColToVal:               make(map[string]types.QValue, capacity),
+		TruncateThresholdBytes: 16 * 1024 * 1024,
 	}
 }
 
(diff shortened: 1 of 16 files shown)


flow/model/record_items.go
```diff
@@ -25,12 +25,14 @@ func ItemsToJSON(items Items) (string, error) {
 
 // encoding/gob cannot encode unexported fields
 type RecordItems struct {
-	ColToVal map[string]types.QValue
+	ColToVal   map[string]types.QValue
+	NoTruncate bool
```
serprex: This could be size limit instead of bool. Let's go larger than 15MB for non SF connectors

flow/pua/stream_adapter.go
```diff
@@ -18,7 +18,7 @@ func AttachToStream(ls *lua.LState, lfn *lua.LFunction, stream *model.QRecordStr
 		}
 		output.SetSchema(schema)
 		for record := range stream.Records {
-			row := model.NewRecordItems(len(record))
+			row := model.NewRecordItems(len(record), false)
```
serprex: This can be true, makes it clearer target stream responsible

flow/model/record_items.go
jgao54: In this PR, i only set the callsite in mongo to true, everything else is false, to maintains backwards compatibility.  if we go larger than 15MB for all non SF connectors, we'll break back compatibility (in the sense that data that was truncated are now available), is this okay?

flow/pua/stream_adapter.go
jgao54: At each of the call site for `NewRecordItems` the information about target DWH is not always available. is there an easy way to wire this information through to the callsite?

flow/model/record_items.go
serprex: yes

flow/pua/stream_adapter.go
serprex: I meant this specific callsite

PR #3112, opened 2025-06-27
qrep: continuous ingestion metrics

serprex:
previous addition of qrep metrics would only report bytes sent after partition sent

this doesn't work well with MySQL's large unpartitioned initial loads. Report bytes sent throughout process

Rows not included continuously to keep records sent tied to successful partitions


serprex:
this ended up being handled with metered connections

PR #3119, opened 2025-06-28
mongo: allow write to JSON in CH when FF is enabled

jgao54:
Currently QValueJson gets converted to String in ClickHouse. For Mongo we want to be able to write `_full_document` as JSON type. This PR will default to JSON in CH if these are all true:
1) Schema is QValueJson
2) CH version supports JSON (25.3 is when JSON is first introduced)
3) JSON is enabled by FF (default to false)

Test: 
- e2e test
- tested locally, with FF on (as JSON) and off (as String)

Next step: look into enabling this setting by default -- but for Mongo only.

### DIFF
--- a/flow/connectors/clickhouse/clickhouse.go
+++ b/flow/connectors/clickhouse/clickhouse.go
@@ -33,7 +33,7 @@ type ClickHouseConnector struct {
 	logger        log.Logger
 	config        *protos.ClickhouseConfig
 	credsProvider *utils.ClickHouseS3Credentials
-	chVersion     string
+	chVersion     *chproto.Version
 }
 
 func NewClickHouseConnector(
@@ -109,7 +109,7 @@ func NewClickHouseConnector(
 			Provider:   credentialsProvider,
 			BucketPath: awsBucketPath,
 		},
-		chVersion: clickHouseVersion.Version.String(),
+		chVersion: &clickHouseVersion.Version,
 	}
 
 	if credentials.AWS.SessionToken != "" {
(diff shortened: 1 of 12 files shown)


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -104,15 +109,12 @@ func NewClickHouseConnector(
 			Provider:   credentialsProvider,
 			BucketPath: awsBucketPath,
 		},
+		chVersion: clickHouseVersion.Version.String(),
```
serprex: Can't store 
```suggestion
		chVersion: clickHouseVersion.Version,
```
why does this need to be stored as string?

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -361,6 +363,10 @@ func (c *ClickHouseConnector) processTableComparison(dstTableName string, srcSch
 }
 
 func (c *ClickHouseConnector) GetVersion(ctx context.Context) (string, error) {
+	if c.chVersion != "" {
+		return c.chVersion, nil
+	}
+
```
serprex, line 369: a few lines down `slog.Any("version", clickhouseVersion.DisplayName)` could be `slog.String("version", clickhouseVersion.DisplayName)`

flow/connectors/clickhouse/clickhouse.go
jgao54: it doesn't, we can store it as Version :)

serprex (approved):

PR #3123, opened 2025-06-30
Add `AUTO_COMPRESS` environment variable

alexstoick:
This allows us to skip the auto-compression enabled by default for
Snowflake. We require this, as Snowflake seems to struggle with certain
files that get auto-compressed and then fail to decompress successfully
with the following error:

```
Invalid data encountered during decompression for file: 'not_file',
compression type used: 'ZSTD', cause: 'Data corruption detected'
```

### DIFF
--- a/flow/connectors/snowflake/qrep_avro_sync.go
+++ b/flow/connectors/snowflake/qrep_avro_sync.go
@@ -202,10 +202,9 @@ func (s *SnowflakeAvroSyncHandler) putFileToStage(ctx context.Context, avroFile 
 
 	autoCompressionStr := ""
 	if autoCompression := internal.GetEnvString("PEERDB_SNOWFLAKE_AUTO_COMPRESSION", ""); autoCompression != "" {
-		autoCompressionStr = "AUTO_COMPRESS=" + autoCompression
+		autoCompressionStr = " AUTO_COMPRESS=" + autoCompression
 	}
-
-	putCmd := fmt.Sprintf("PUT file://%s @%s %s", avroFile.FilePath, stage, autoCompressionStr)
+	putCmd := fmt.Sprintf("PUT file://%s @%s%s", avroFile.FilePath, stage, autoCompressionStr)
 
 	if _, err := s.ExecContext(ctx, putCmd); err != nil {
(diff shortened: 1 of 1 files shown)


flow/connectors/snowflake/qrep_avro_sync.go
```diff
@@ -199,7 +200,12 @@ func (s *SnowflakeAvroSyncHandler) putFileToStage(ctx context.Context, avroFile
 		return nil
 	}
 
-	putCmd := fmt.Sprintf("PUT file://%s @%s", avroFile.FilePath, stage)
+	autoCompressionStr := ""
+	if autoCompression := internal.GetEnvString("PEERDB_SNOWFLAKE_AUTO_COMPRESSION", ""); autoCompression != "" {
+		autoCompressionStr = "AUTO_COMPRESS=" + autoCompression
+	}
+
+	putCmd := fmt.Sprintf("PUT file://%s @%s %s", avroFile.FilePath, stage, autoCompressionStr)
```
serprex:
```suggestion
		autoCompressionStr = " AUTO_COMPRESS=" + autoCompression
	}
	putCmd := fmt.Sprintf("PUT file://%s @%s%s", avroFile.FilePath, stage, autoCompressionStr)
```

flow/connectors/snowflake/qrep_avro_sync.go
```diff
@@ -199,7 +200,12 @@ func (s *SnowflakeAvroSyncHandler) putFileToStage(ctx context.Context, avroFile
 		return nil
 	}
 
-	putCmd := fmt.Sprintf("PUT file://%s @%s", avroFile.FilePath, stage)
+	autoCompressionStr := ""
+	if autoCompression := internal.GetEnvString("PEERDB_SNOWFLAKE_AUTO_COMPRESSION", ""); autoCompression != "" {
```
serprex, line 204: this should be wired into dynconf

alexstoick: Yep! There is a slack thread here with some more conversation as well: https://peerdb-public.slack.com/archives/C05BMVAM9FD/p1751283629386039

PR #3125, opened 2025-06-30
MySQL: Fix mixed case for schema changes

Amogh-Bharadwaj:
In our handler to process alter table commands (`processAlterTableQuery`), we use a `.String()` method to convert the column object we get to string to add it to our state of added columns.
Turns out this method actually converts the column to lowercase before giving it to us:
https://github.com/pingcap/tidb/blob/1962f220e9d40608867fa172c91e5ad9b67c0865/pkg/parser/ast/expressions.go#L546

This PR instead uses a different method - `.OrigColName()` - which gives us the column name as is.
https://github.com/pingcap/tidb/blob/1962f220e9d40608867fa172c91e5ad9b67c0865/pkg/parser/ast/expressions.go#L558

Also this PR introduces a test for MySQL schema changes in `mysql_test.go`, and it differs from the generic test in how an insert is not performed after column addition and the quoting of identifiers is different


flow/connectors/mysql/cdc.go
```diff
@@ -698,8 +698,10 @@ func (c *MySqlConnector) processAlterTableQuery(ctx context.Context, catalogPool
 					typmod = datatypes.MakeNumericTypmod(int32(precision), int32(scale))
 				}
 
+				slog.Info("[processAlterTableQuery] adding column",
+					slog.String("columnName", col.Name.String()))
```
serprex:
```suggestion
					slog.String("columnName", col.Name.OrigColName()))
```

serprex (approved):

PR #3130, opened 2025-07-01
Respect `PEERDB_NULLABLE` for custom type mappings

dungeon-master-666:
Currently setting `PEERDB_NULLABLE` does not propagate nullability of source column to clickhouse when specifying custom column types. As result mapping `NUMERIC NULL` from postgres to `String` in CH creates column of type `String` instead of `Nullable(String)`. This PR aims to fix it.

### DIFF
--- a/protos/flow.proto
+++ b/protos/flow.proto
@@ -21,6 +21,7 @@ message ColumnSetting {
   string destination_type = 3;
   int32 ordering = 4;
   bool nullable_enabled = 5;
+  bool auto_detect_nullable = 6;
 }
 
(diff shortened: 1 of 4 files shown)


protos/flow.proto
```diff
@@ -21,6 +21,7 @@ message ColumnSetting {
   string destination_type = 3;
   int32 ordering = 4;
   bool nullable_enabled = 5;
+  bool auto_detect_nullable = 6;
```
serprex, line 24: this shouldn't be necessary

PR #3142, opened 2025-07-02
mysql: stat activity

serprex:
performance schema was added in 5.5

### DIFF
--- a/flow/connectors/mysql/mysql.go
+++ b/flow/connectors/mysql/mysql.go
@@ -392,3 +392,30 @@ func (c *MySqlConnector) GetVersion(ctx context.Context) (string, error) {
 	}
 	return "", errors.New("failed to connect")
 }
+
+func (c *MySqlConnector) StatActivity(
+	ctx context.Context,
+	req *protos.PostgresPeerActivityInfoRequest,
+) (*protos.PeerStatResponse, error) {
+	rs, err := c.Execute(ctx, "select ID, COMMAND, STATE, TIME, INFO from performance_schema.processlist WHERE USER=?", c.config.User)
+	if err != nil {
+		return nil, err
+	}
+
+	statInfoRows := make([]*protos.StatInfo, len(rs.Values))
+	for idx, row := range rs.Values {
+		statInfoRows[idx] = &protos.StatInfo{
+			Pid:           row[0].AsInt64(),
+			WaitEvent:     string(row[1].AsString()),
+			WaitEventType: "",
+			QueryStart:    "",
+			Query:         string(row[4].AsString()),
+			Duration:      float32(row[3].AsUint64()),
+			State:         string(row[2].AsString()),
+		}
+	}
+
+	return &protos.PeerStatResponse{
+		StatData: statInfoRows,
+	}, nil
+}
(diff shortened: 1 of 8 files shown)


flow/connectors/mysql/mysql.go
```diff
@@ -392,3 +392,30 @@ func (c *MySqlConnector) GetVersion(ctx context.Context) (string, error) {
 	}
 	return "", errors.New("failed to connect")
 }
+
+func (c *MySqlConnector) StatActivity(
+	ctx context.Context,
+	req *protos.PostgresPeerActivityInfoRequest,
+) (*protos.PeerStatResponse, error) {
+	rs, err := c.Execute(ctx, "select ID, COMMAND, STATE, TIME, INFO from performance_schema.processlist WHERE USER=?", c.config.User)
```
serprex: You might expect this to have how long the connection has been open, but no, we only get how many seconds the connection has been in `STATE`

heavycrystal:
```
MariaDB [mysql]> select ID, COMMAND, STATE, TIME, INFO from performance_schema.processlist;
ERROR 1146 (42S02): Table 'performance_schema.processlist' doesn't exist
```

tale as old as time, I see `information_schema.processlist` which is [deprecated ](https://dev.mysql.com/doc/refman/8.4/en/information-schema-processlist-table.html)but might work better

serprex: looks like mariadb has [threads table](https://mariadb.com/docs/server/reference/sql-statements/administrative-sql-statements/system-tables/performance-schema/performance-schema-tables/performance-schema-threads-table)

heavycrystal: So does [MySQL](https://dev.mysql.com/doc/refman/8.4/en/performance-schema-threads-table.html), but with a different schema.

flow/connectors/mysql/mysql.go
```diff
@@ -392,3 +392,40 @@ func (c *MySqlConnector) GetVersion(ctx context.Context) (string, error) {
...
+	ctx context.Context,
+	req *protos.PostgresPeerActivityInfoRequest,
+) (*protos.PeerStatResponse, error) {
+	rs, err := c.Execute(ctx, "select ID,COMMAND,STATE,TIME,INFO from performance_schema.processlist WHERE USER=?", c.config.User)
+	if err != nil {
+		var myErr *mysql.MyError
+		if errors.As(err, &myErr) && myErr.Code == 1146 && myErr.State == "42S02" { // ER_NO_SUCH_TABLE
+			// mariadb
+			rs, err = c.Execute(ctx,
+				"select PROCESSLIST_ID,PROCESSLIST_COMMAND,PROCESSLIST_STATE,PROCESSLIST_TIME,PROCESSLIST_INFO"+
```
Amogh-Bharadwaj:
```suggestion
				"SELECT PROCESSLIST_ID,PROCESSLIST_COMMAND,PROCESSLIST_STATE,PROCESSLIST_TIME,PROCESSLIST_INFO"+
```

flow/connectors/mysql/mysql.go
```diff
@@ -392,3 +392,40 @@ func (c *MySqlConnector) GetVersion(ctx context.Context) (string, error) {
...
+					" from performance_schema.threads WHERE USER=?", c.config.User)
```
Amogh-Bharadwaj:
```suggestion
					" FROM performance_schema.threads WHERE USER=?", c.config.User)
```

PR #3145, opened 2025-07-03
API endpoint for total rows synced by mirror

Amogh-Bharadwaj:
Currently to get the total rows synced by a mirror across CDC and initial load, clients of our API would have to call `CDCTableTotalCounts` and `InitialLoadSummary` and in the case of the latter they would have to sum the `NumRowsSynced` field of all entries. Not to mention both these API calls would be overfetching if we just want a single count.

This PR introduces a new endpoint dedicated to getting the sum total rows synced by CDC and initial load.
An API test has been included.

### DIFF
--- a/flow/cmd/mirror_status.go
+++ b/flow/cmd/mirror_status.go
@@ -613,7 +613,7 @@ func (h *FlowRequestHandler) TotalRowsSyncedByCDCMirror(
 	if !req.ExcludeInitialLoad {
 		err := h.pool.QueryRow(ctx, `SELECT SUM(rows_in_partition) FILTER (WHERE end_time IS NOT NULL) AS NumRowsSynced
 		FROM peerdb_stats.qrep_partitions
-		WHERE flow_name = $1`, req.FlowJobName).Scan(&totalRowsInitialLoad)
+		WHERE parent_mirror_name = $1`, req.FlowJobName).Scan(&totalRowsInitialLoad)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


ilidemi (approved):

flow/cmd/mirror_status.go
```diff
@@ -595,6 +595,37 @@ func (h *FlowRequestHandler) CDCBatches(ctx context.Context, req *protos.GetCDCB
...
+		WHERE flow_name = $1`, req.FlowJobName).Scan(&totalRowsCDC)
+		if cdcErr != nil {
+			return nil, fmt.Errorf("unable to get total rows synced via CDC for mirror %s: %w", req.FlowJobName, cdcErr)
+		}
+	}
+
+	if !req.ExcludeInitialLoad {
+		err := h.pool.QueryRow(ctx, `SELECT SUM(rows_in_partition) FILTER (WHERE end_time IS NOT NULL) AS NumRowsSynced
+		FROM peerdb_stats.qrep_partitions
+		WHERE parent_mirror_name = $1`, req.FlowJobName).Scan(&totalRowsInitialLoad)
```
serprex:
```suggestion
		err := h.pool.QueryRow(ctx, `SELECT SUM(rows_in_partition) AS NumRowsSynced
		FROM peerdb_stats.qrep_partitions
		WHERE parent_mirror_name = $1 AND end_time IS NOT NULL`, req.FlowJobName).Scan(&totalRowsInitialLoad)
```

PR #3151, opened 2025-07-06
MySQL: Move binary format fetch outside

Amogh-Bharadwaj:
During a parallel initial load from MySQL to ClickHouse, it was observed that the PeerDBBinaryFormat call in QValueToAvro which gets the value of the binary format dynamic setting from catalog was taking ~60% of CPU utilisation of the flow-worker pod.
Makes sense - making a call to Postgres for every single record can be expensive
This PR moves this fetch outside the loop on the pulled records.

Also some lint in `cdc_flow.go`

### DIFF
--- a/flow/model/qvalue/avro_converter.go
+++ b/flow/model/qvalue/avro_converter.go
@@ -198,6 +198,7 @@ func QValueToAvro(
 		Stat:                     stat,
 		TargetDWH:                targetDWH,
 		UnboundedNumericAsString: unboundedNumericAsString,
+		binaryFormat:             binaryFormat,
 	}
 
 	switch v := value.(type) {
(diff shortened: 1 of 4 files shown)


serprex (approved):

flow/model/qvalue/avro_converter.go
```diff
@@ -179,12 +179,14 @@ type QValueAvroConverter struct {
 	Stat                     *NumericStat
 	TargetDWH                protos.DBType
 	UnboundedNumericAsString bool
+	binaryFormat             internal.BinaryFormat
 }
 
 func QValueToAvro(
 	ctx context.Context, env map[string]string,
```
serprex: why do we still need to pass env?

PR #3160, opened 2025-07-07
allow editing initial load settings

heavycrystal:


flow/workflows/cdc_flow.go
```diff
@@ -611,6 +637,10 @@ func CDCFlowWorkflow(
 		}
 
 		snapshotFlowCtx := workflow.WithChildOptions(ctx, childSnapshotFlowOpts)
+		// now snapshot parameters are also part of the state, but until we finish snapshot they wouldn't be modifiable
+		// so we can use the same cfg for snapshot flow, and then rely on being state saved to catalog
```
serprex:
```suggestion
		// so we can use the same cfg for snapshot flow, and then rely on state being saved to catalog
```

flow/workflows/cdc_flow.go
```diff
@@ -611,6 +637,10 @@ func CDCFlowWorkflow(
...
+		// during any operation that triggers another snapshot (INCLUDING add tables)
+		// could fail for very weird Temporal resets
```
serprex: is this part of sentence above? grammar too weird, but not sure what this is trying to say

heavycrystal: it is part of sentence above, was trying to avoid `lll` but cut off weirdly

PR #3166, opened 2025-07-07
mongo: mongo version check during validation

jgao54:
- Check mongo version for validation.
- Small refactors to use bson struct tags for deserialization.
- Extract shared logic to shared/mongo/validation.go
- Default read preference mode to SecondaryPreferred, which read from primary only if secondary does not exists -- we should still support read from primary via uri if user prefers, will work on this as a subsequent PR as we separate out username/password from rest of uri.

Test: locally tested deserialization.

### DIFF
--- a/flow/shared/mongo/validation.go
+++ b/flow/shared/mongo/validation.go
@@ -1,0 +1,46 @@
+package mongo
+
+import (
+	"context"
+	"fmt"
+
+	"go.mongodb.org/mongo-driver/v2/bson"
+	"go.mongodb.org/mongo-driver/v2/mongo"
+)
+
+const MinSupportedVersion = "5.1.0"
+
+type BuildInfo struct {
+	Version string `bson:"version"`
+}
+
+type ReplSetGetStatus struct {
+	Set     string `bson:"set"`
+	MyState int    `bson:"myState"`
+}
+
+func GetBuildInfo(ctx context.Context, client *mongo.Client) (*BuildInfo, error) {
+	singleResult := client.Database("admin").RunCommand(ctx, bson.D{bson.E{Key: "buildInfo", Value: 1}})
+	if singleResult.Err() != nil {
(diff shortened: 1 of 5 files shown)


flow/go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/eventhub/armeventhub v1.3.0
 	github.com/ClickHouse/ch-go v0.66.1
 	github.com/ClickHouse/clickhouse-go/v2 v2.37.2
+	github.com/Masterminds/semver v1.5.0
```
serprex: I just got rid of this..

flow/shared/mongo/validation.go
```diff
@@ -0,0 +1,59 @@
...
+	buildInfo, err := GetBuildInfo(ctx, client)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get mongo version: %w", err)
+	}
+	parsedVersion, err := semver.NewVersion(buildInfo.Version)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse mongo version %w", err)
+	}
+	compareVersion := semver.MustParse(versionString)
+	return parsedVersion.Compare(compareVersion), nil
```
serprex: Since we're only comparing triplets, seems like we could avoid full semver, see https://github.com/go-mysql-org/go-mysql/pull/1028

flow/shared/mongo/validation.go
```diff
@@ -0,0 +1,59 @@
+package mongo
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Masterminds/semver"
```
serprex:
```suggestion
	"github.com/Masterminds/semver/v3"
```

flow/shared/mongo/validation.go
jgao54: sounds good, no preference here, didn't see the commit to remove :)

serprex (approved):

PR #3168, opened 2025-07-08
mongo: add validation for (1) storage engine (2) oplog retention

jgao54:
>You can specify the minimum number of hours to preserve an oplog entry where [mongod](https://www.mongodb.com/docs/manual/reference/program/mongod/#mongodb-binary-bin.mongod) only removes an oplog entry if both of the following criteria are met:
-The oplog has reached the [maximum configured size.](https://www.mongodb.com/docs/manual/core/replica-set-oplog/#std-label-replica-set-oplog-sizing)
-The oplog entry is older than the configured number of hours based on the host system clock.
By default MongoDB does not set a minimum oplog retention period and automatically truncates the oplog starting with the oldest entries to maintain the configured maximum oplog size.

Seems like a good idea to require a min oplog retention (instead of just relying on max oplog size).

Also add a validation for storage engne, as 'wiredTiger' is a prerequisite for changestream to work.

Test: tested e2e locally


flow/connectors/mongo/validate.go
```diff
@@ -30,7 +31,20 @@ func (c *MongoConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.F
...
 	if err != nil {
-		return fmt.Errorf("failed to get replica set status: %w", err)
+		return err
+	}
+
+	serverStatus, err := shared_mongo.GetServerStatus(ctx, c.client)
+	if err != nil {
+		return err
+	}
+	if serverStatus.StorageEngine.Name != "wiredTiger" {
```
heavycrystal, line 41: nit: it looks like with the minimum version we support `wiredTiger` is the only option?

jgao54: There's also [in-memory storage engine.](https://www.mongodb.com/docs/manual/core/inmemory/#std-label-storage-inmemory)

PR #3169, opened 2025-07-08
Validate tables on ClickHouse: validate in batches of 200 tables

Amogh-Bharadwaj:
Sometimes mirrors can have tables in the thousands, and this particular query we run on ClickHouse run into the max_query_size limit. 
This PR makes it so that if validation sees if the number of tables is more than 200 and if so validates in batches of 200 tables.

### DIFF
--- a/flow/shared/clickhouse/validation.go
+++ b/flow/shared/clickhouse/validation.go
@@ -33,37 +33,47 @@ func CheckIfClickHouseCloudHasSharedMergeTreeEnabled(ctx context.Context, logger
 func CheckIfTablesEmptyAndEngine(ctx context.Context, logger log.Logger, conn clickhouse.Conn,
 	tables []string, initialSnapshotEnabled bool, checkForCloudSMT bool, allowNonEmpty bool,
 ) error {
-	queryTables := make([]string, 0, len(tables))
-	for _, table := range tables {
-		queryTables = append(queryTables, QuoteLiteral(table))
-	}
-	rows, err := Query(ctx, logger, conn,
-		fmt.Sprintf("SELECT name,engine,total_rows FROM system.tables WHERE database=currentDatabase() AND name IN (%s)",
-			strings.Join(queryTables, ",")))
-	if err != nil {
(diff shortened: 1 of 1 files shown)


flow/shared/clickhouse/validation.go
```diff
@@ -82,24 +82,49 @@ func GetTableColumnsMapping(ctx context.Context, logger log.Logger, conn clickho
 	for _, table := range tables {
 		queryTables = append(queryTables, QuoteLiteral(table))
 	}
+	obtainSchemaInBatches := len(queryTables) > 200
+	if obtainSchemaInBatches {
```
serprex: this can be unconditional

flow/shared/clickhouse/validation.go
```diff
@@ -82,24 +82,49 @@ func GetTableColumnsMapping(ctx context.Context, logger log.Logger, conn clickho
 	for _, table := range tables {
```
serprex: `queryTables` could be kept to at most 200 capacity & reused if we had QuoteLiteral in chunking

serprex (approved):

flow/shared/clickhouse/validation.go
```diff
@@ -78,28 +78,47 @@ func GetTableColumnsMapping(ctx context.Context, logger log.Logger, conn clickho
 	tables []string,
 ) (map[string][]ClickHouseColumn, error) {
 	tableColumnsMapping := make(map[string][]ClickHouseColumn, len(tables))
-	queryTables := make([]string, 0, len(tables))
-	for _, table := range tables {
-		queryTables = append(queryTables, QuoteLiteral(table))
+	queryTables := make([]string, 0, min(len(tables), 200))
+
+	for i := 0; i < len(tables); i += 200 {
```
iskakaushik: maybe create a util called utils.IterBatches(iterable, chunkSize, func)

serprex: especially since there's a 2nd function that wants to reuse this already

serprex: https://pkg.go.dev/slices#Chunk

PR #3170, opened 2025-07-08
Allow updating of the `snapshot` variables for mirror

alexstoick:
As part of our use-case we have tables of varying sizes and we would like
to be able to tweak the snapshot sizes/concurrency to ensure a smooth
initial snapshot.

This would also allow us to bump the concurrency for the initial sync if
we need to do this in lower traffic hours vs peak traffic.


heavycrystal:
duplicate of #3160, which is merged

PR #3172, opened 2025-07-08
vitess: only allow during initial load only

heavycrystal:


flow/connectors/mysql/validate.go
```diff
@@ -94,6 +94,18 @@ func (c *MySqlConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.F
 		return fmt.Errorf("unable to establish replication connectivity: %w", err)
 	}
 
+	for conn, err := range c.withRetries(ctx) {
+		if err != nil {
+			return err
+		}
+
+		if isVitess, err := peerdb_mysql.CheckIfVitess(conn, c.logger); err != nil {
+			return fmt.Errorf("failed to check if Vitess: %w", err)
```
serprex: this will end up saying `failed to check if Vitess: failed to check if Vitess:`

serprex (approved):

PR #3178, opened 2025-07-08
mongo: update config with fields for username/password/tls

jgao54:
- Separate out username and password to separate fields.
- Add default client options - currently give user override options via connection string, we can also choose to ignore user options always use default 
- Add TLS support
- Apply corresponding changes to frontend.

TODO: verify TLS works as expected

### DIFF
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -2,6 +2,7 @@ package connmongo
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -11,7 +12,9 @@ import (
 	"go.mongodb.org/mongo-driver/v2/bson"
 	"go.mongodb.org/mongo-driver/v2/mongo"
 	"go.mongodb.org/mongo-driver/v2/mongo/options"
+	"go.mongodb.org/mongo-driver/v2/mongo/readconcern"
 	"go.mongodb.org/mongo-driver/v2/mongo/readpref"
+	"go.mongodb.org/mongo-driver/v2/x/mongo/driver/connstring"
 	"go.temporal.io/sdk/log"
 
 	metadataStore "github.com/PeerDB-io/peerdb/flow/connectors/external_metadata"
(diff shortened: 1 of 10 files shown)


flow/connectors/mongo/mongo.go
```diff
@@ -405,3 +409,41 @@ func defaultPipeline() mongo.Pipeline {
...
+		SetAuth(options.Credential{
+			Username: config.Username,
+			Password: config.Password,
+		})
+
+	if connStr.Compressors != nil {
+		clientOptions.SetCompressors([]string{"zstd", "snappy"})
+	}
+	if connStr.ReadPreference == "" {
+		clientOptions.SetReadPreference(readpref.SecondaryPreferred())
```
heavycrystal, line 414: do we want this? is this recommended somewhere?

flow/connectors/mongo/mongo.go
```diff
@@ -405,3 +409,41 @@ func defaultPipeline() mongo.Pipeline {
...
+		// always use majority read concern for correctness
+		SetReadConcern(readconcern.Majority())
+
+	// allow user to override with other read preference
+	if connStr.ReadPreference == "" {
+		clientOptions.SetReadPreference(readpref.SecondaryPreferred())
+	}
+
+	if !config.DisableTls && len(connStr.Hosts) > 0 {
+		tlsConfig, err := shared.CreateTlsConfig(tls.VersionTLS12, config.RootCa, connStr.Hosts[0], config.TlsHost, false)
```
heavycrystal: I'm not sure about using `.Hosts[0]` here if provided a list of hosts. What does the connector do here? Can tls not work if we connect to a different host?

jgao54: we can remove this and require user to always pass in tlsHost

flow/connectors/mongo/mongo.go
jgao54: discussed offline, defaulting to secondary for CDC to avoid affecting production traffic -- but give user ability to override if they wish.

jgao54: discussed offline, defaulting to secondary for CDC to avoid affecting production traffic -- but give user ability to override if they wish.

PR #3180, opened 2025-07-08
mysql: report bytes while ingestion still ongoing

serprex:
this avoids full table partitions giving off little information & appearing stuck


ilidemi:
Fetched bytes vs synced rows is a nice balance, no need to report WIP rows imo (hopefully the UI can do the rows for initial load too). Alternatively, report fetched and synced for both but that's more UI work.

image

PR #3186, opened 2025-07-09
Fix null resync

serprex:
resync on null table engine was deleting table schema mapping data

don't unconditionally add `public.` to unused columns in flows
this was adding `public.` in front of clickhouse table names

### DIFF
--- a/flow/e2e/clickhouse/peer_flow_ch_test.go
+++ b/flow/e2e/clickhouse/peer_flow_ch_test.go
@@ -2087,7 +2087,7 @@ func (s ClickHouseSuite) Test_NullEngine() {
 
 	require.NoError(s.t, s.source.Exec(s.t.Context(),
 		fmt.Sprintf(`insert into %s values (3,'cdcresync',1)`, srcFullName)))
-	e2e.EnvWaitForEqualTablesWithNames(env, s, "waiting on initial", srcTableName, "nulltarget", "id,\"key\"")
+	e2e.EnvWaitForEqualTablesWithNames(env, s, "waiting on insert after resync", srcTableName, "nulltarget", "id,\"key\"")
 
 	ch, err = connclickhouse.Connect(s.t.Context(), nil, chPeer)
 	require.NoError(s.t, err)


flow/e2e/clickhouse/peer_flow_ch_test.go
```diff
@@ -2085,6 +2085,10 @@ func (s ClickHouseSuite) Test_NullEngine() {
 	e2e.SetupCDCFlowStatusQuery(s.t, env, flowConnConfig)
 	e2e.EnvWaitForEqualTablesWithNames(env, s, "waiting on initial", srcTableName, "nulltarget", "id,\"key\"")
 
+	require.NoError(s.t, s.source.Exec(s.t.Context(),
+		fmt.Sprintf(`insert into %s values (3,'cdcresync',1)`, srcFullName)))
+	e2e.EnvWaitForEqualTablesWithNames(env, s, "waiting on initial", srcTableName, "nulltarget", "id,\"key\"")
```
jgao54: nit: waiting on _insert_, but can skip if time sensitive

flow/connectors/eventhub/hub_batches.go
```diff
@@ -106,7 +106,7 @@ func (h *HubBatches) sendBatch(
 	}
 
 	internal.LoggerFromCtx(ctx).Info("sendBatch",
-		slog.Int("events sent", int(events.NumEvents())), slog.String("event hub topic", tblName.ToString()))
+		slog.Int64("events sent", int64(events.NumEvents())), slog.String("event hub topic", tblName.ToString()))
```
jgao54, line 109: perhaps Uint64 better here?

flow/activities/flowable.go
```diff
@@ -1229,7 +1231,7 @@ func (a *FlowableActivity) RemoveFlowDetailsFromCatalog(
 			logger.Warn("flow entry not found in catalog, 0 records deleted")
 		} else {
 			logger.Info("flow entries removed from catalog",
-				slog.Int("rowsAffected", int(ct.RowsAffected())))
+				slog.Int64("rowsAffected", ct.RowsAffected()))
```
jgao54, line 1234: nit: similar to the other comment, perhaps Uint64 better here?

flow/connectors/eventhub/hub_batches.go
jgao54: nvm, i see that NumEvents() returns int32.

flow/cmd/handler.go
```diff
@@ -81,8 +73,8 @@ func (h *FlowRequestHandler) createCdcJobEntry(ctx context.Context,
 		source_table_identifier, destination_table_identifier) VALUES ($1, $2, $3, $4, $5, $6, $7)
 		`, workflowID, req.ConnectionConfigs.FlowJobName, sourcePeerID, destinationPeerID,
 			"Mirror created via GRPC",
-			schemaForTableIdentifier(v.SourceTableIdentifier, sourePeerType),
```
jgao54, line 84: ooc, why did this get removed?

flow/connectors/eventhub/hub_batches.go
serprex: `NumEvents` returns `int32`

flow/cmd/handler.go
serprex: updated PR description, but these columns no longer seem to be used (so are only informational) & `public.` only really makes sense for a couple databases

PR #3190, opened 2025-07-10
Convert PG Composite Types to Tuples in CH

dungeon-master-666:
Currently custom composite types from postgres are saved as strings in CH which makes it inconvenient to access/filter/sort/etc them. This PR adds functionality converting composite types to Tuple. 

1. Nested composite types and arrays of composites are supported
2. Clickhouse doesn't have `Nullable(Tuple(...))` so instead every field of tuple is created `Nullable` (if `PEERDB_NULLABLE` is enabled)
3. Apparently this feature should be wrapped in setting to make existing mirrors won't break.
4. I tested only the direct scenario, PG -> CH mirror. PG -> PG and others were not tested

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -15,6 +15,7 @@ import (
 	"github.com/jackc/pgerrcode"
 	"github.com/jackc/pglogrepl"
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/jackc/pgx/v5/pgtype"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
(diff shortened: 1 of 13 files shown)


flow/shared/postgres.go
```diff
@@ -60,6 +61,74 @@ func GetCustomDataTypes(ctx context.Context, conn *pgx.Conn) (map[uint32]CustomD
...
+	rows, err := conn.Query(ctx, fmt.Sprintf(`
+		SELECT
+			att.attnum                              AS ordinal_position,
+			att.attname                             AS field,
+			att.atttypid                            AS field_type_oid,
+			pt.typname                              AS field_type_name,
+			att.atttypmod       					AS type_modifier,  
+			att.attnotnull                          AS not_null,
+			pg_catalog.col_description(att.attrelid,
+									att.attnum)  AS comment
```
serprex, line 81:
```suggestion
		SELECT
			att.attnum                              AS ordinal_position,
			att.attname                             AS field,
			att.atttypid                            AS field_type_oid,
			pt.typname                              AS field_type_name,
			att.atttypmod                           AS type_modifier,  
			att.attnotnull                          AS not_null,
			pg_catalog.col_description(att.attrelid, att.attnum)  AS comment
```

dungeon-master-666:
Hi @serprex @heavycrystal, it's ready for reviewing. I've added a new version `CompositeTypeAsTuple` for backward compatibility and fixed several issues with arrays of composites and nested tuples.

flow/connectors/postgres/postgres.go
```diff
@@ -140,13 +141,29 @@ func ParseConfig(connectionString string, pgConfig *protos.PostgresConfig) (*pgx
...
 		c.customTypeMapping = customTypeMapping
+
+		if version >= shared.InternalVersion_CompositeTypeAsTuple {
+			var compositeTypeNames []string
+			for _, typeData := range customTypeMapping {
+				if typeData.Type == 'c' && typeData.Delim == 0 { // Only composite types
+					compositeTypeNames = append(compositeTypeNames, typeData.Name)
+				}
+			}
+			types, err := c.conn.LoadTypes(ctx, compositeTypeNames)
```
serprex, line 159: this is failing in CI for #3384
```
failed to fetch custom type mapping: failed to load composite types: Unknown field for composite type "geometry_dump":  field "geom" (OID 18053) is not already registered.
```
seems like there's a mess with having to make sure dependencies are registered before loading. Looking at `LoadTypes` this can probably be cleaned up to query catalog ourselves for composite types specifically & figure out how to register types in correct order

dungeon-master-666: Weird, `LoadTypes` already has some logic for registering children types before parents: https://github.com/jackc/pgx/blob/master/derived_types.go#L82-L84

PR #3196, opened 2025-07-11
mongo: make sure events from irrelevant tables are filtered out

jgao54:
Previously , we were processing all change stream events from all tables. We ended up getting confusing logs that shows we processed N records when they were coming from tables we are not tracking (and these events got dropped later when there's no destination table mapped). More importantly, having changestream process/deserialize all events from all databases is very expensive. 

This PR addresses this by making sure the changestream pipeline only let relevant events pass through. This should help us with throughput quite a bit (although I have yet to benchmark this, plan to that next).

Test:  e2e integration tests + running pipe locally.

### DIFF
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -136,7 +137,12 @@ func (c *MongoConnector) SetupReplication(ctx context.Context, input *protos.Set
 		SetComment("PeerDB changeStream").
 		SetFullDocument(options.UpdateLookup).
 		SetFullDocumentBeforeChange(options.Off)
-	changeStream, err := c.client.Watch(ctx, defaultPipeline(), changeStreamOpts)
+
+	pipeline, err := createPipeline(nil)
+	if err != nil {
+		return model.SetupReplicationResult{}, fmt.Errorf("failed to create changestream pipeline: %w", err)
+	}
+	changeStream, err := c.client.Watch(ctx, pipeline, changeStreamOpts)
 	if err != nil {
 		return model.SetupReplicationResult{}, fmt.Errorf("failed to start change stream for storing initial resume token: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


flow/connectors/mongo/mongo.go
```diff
@@ -216,7 +222,9 @@ func (c *MongoConnector) PullRecords(
 	changeStreamOpts := options.ChangeStream().
 		SetComment("PeerDB changeStream for mirror " + req.FlowJobName).
 		SetFullDocument(options.UpdateLookup).
-		SetFullDocumentBeforeChange(options.Off)
+		SetFullDocumentBeforeChange(options.Off).
+		SetMaxAwaitTime(req.IdleTimeout)
```
jgao54, line 226: Reading: https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.md#why-do-we-need-to-expose-the-postbatchresumetoken

IIUC, this setting should help us resolve the stale resume token when there are not events. when Max Await Time is reached and we haven't received a record, `Next`  should return false (but without errors) and the latest resumeToken from `GetMore` should get populated which we can store.

flow/connectors/mongo/mongo.go
```diff
@@ -404,26 +422,60 @@ func (c *MongoConnector) PullRecords(
...
 
-func defaultPipeline() mongo.Pipeline {
-	return mongo.Pipeline{
-		{{Key: "$match", Value: bson.D{
+func createPipeline(tableNameMapping map[string]model.NameAndExclude) (mongo.Pipeline, error) {
+	pipeline := mongo.Pipeline{}
+
+	// filter out events from tables that are not in the mapping
+	if tableNameMapping != nil {
+		dbCollMap := make(map[string][]string)
```
heavycrystal, line 430:
```suggestion
		dbCollMap := make(map[string][]string, len(tableNameMapping))
```

flow/connectors/mongo/mongo.go
```diff
@@ -404,26 +422,60 @@ func (c *MongoConnector) PullRecords(
...
+		for dbAndTable := range tableNameMapping {
+			parts := strings.SplitN(dbAndTable, ".", 2)
```
heavycrystal, line 432: not related to this PR but wondering how we handle collections/tables with a period in the name

heavycrystal (approved):

flow/connectors/mongo/mongo.go
jgao54: we don't know the length in advance here, since tableNameMapping keys are strings that include both db and table, and dbs may not be unique

flow/connectors/mongo/mongo.go
jgao54: should be ok: https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Restrictions-on-Database-Names-for-Unix-and-Linux-Systems

PR #3197, opened 2025-07-11
feat: include xid, lsn and commit time for avro destinations

pratheekrebala:
Following up on [this Slack conversation](https://peerdb-public.slack.com/archives/C05BMVAM9FD/p1751907236836079), this is a WIP PR that:

- Updates the `BaseRecord` to include the Xid for the postgres record stream.
- Makes the Xid available to Lua scripts as `transaction_id`
- Update the avro schema to include `_peerdb_origin` column which contains the Xid along with LSN and commit time. (LSN and Commit time were previously only available to the Lua script.

To do:

- test how this interacts with BigQuery and Snowflake connectors that depend on the `RecordsToRawTableStream` function.  
- gate this behind a env variable

### DIFF
--- a/flow/connectors/utils/stream.go
+++ b/flow/connectors/utils/stream.go
@@ -98,10 +98,10 @@ func recordToQRecordOrError[Items model.Items](
 
 		originItemsJSON := ""
 		if includeOriginMeta {
-			originItems := appendOriginMeta(model.NewRecordItems(3), record)
+			originItems := getOriginMeta(record)
 
 			var err error
-			originItemsJSON, err = model.ItemsToJSON(originItems.(model.RecordItems))
+			originItemsJSON, err = model.ItemsToJSON(originItems)
 			if err != nil {
 				return nil, fmt.Errorf("failed to serialize origin items to JSON: %w", err)
 			}
@@ -128,7 +128,8 @@ func recordToQRecordOrError[Items model.Items](
 
 		oldItems := typedRecord.OldItems
 		if includeOriginMeta {
-			oldItems = appendOriginMeta(typedRecord.OldItems, record).(Items)
+			originMeta := getOriginMeta(record)
+			oldItems.UpdateIfNotExists(originMeta)
 		}
 
 		oldItemsJSON, err := model.ItemsToJSON(oldItems)
@@ -150,7 +151,9 @@ func recordToQRecordOrError[Items model.Items](
 
 		matchedItemsJSON := itemsJSON
 		if includeOriginMeta {
-			matchedItems := appendOriginMeta(typedRecord.Items, record)
+			originMeta := getOriginMeta(record)
+			matchedItems := typedRecord.Items
+			matchedItems.UpdateIfNotExists(originMeta)
 			var err error
 
 			matchedItemsJSON, err = model.ItemsToJSON(matchedItems)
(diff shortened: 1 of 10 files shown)


flow/connectors/utils/stream.go
```diff
@@ -59,6 +60,11 @@ func RecordsToRawTableStream[Items model.Items](
 				Type:     types.QValueKindString,
 				Nullable: true,
 			},
+			{
```
serprex: should be in match data when enabled, see `PEERDB_SOURCE_SCHEMA_AS_DESTINATION_COLUMN`

pratheekrebala: Done @serprex! I reverted the change to this QRecordStream schema and added the following fields to the `_peerdb_match_data` when the `PEERDB_ORIGIN_META_AS_DESTINATION_COLUMN` is set to true.

- `_peerdb_origin_transaction_id`
- `_peerdb_origin_checkpoint_id`
- `_peerdb_origin_commit_time_nano`

flow/connectors/utils/stream.go
```diff
@@ -155,6 +188,23 @@ func InitialiseTableRowsMap(tableMaps []*protos.TableMapping) map[string]*model.
 	return tableNameRowsMapping
 }
 
+func appendOriginMeta[Items model.Items](items model.Items, originMeta model.Record[Items]) model.Items {
```
serprex:
```suggestion
func appendOriginMeta[Items model.Items](items Items, originMeta model.Record[Items]) Items {
```

flow/connectors/utils/stream.go
```diff
@@ -155,6 +188,23 @@ func InitialiseTableRowsMap(tableMaps []*protos.TableMapping) map[string]*model.
...
+	recordItems, ok := items.(model.RecordItems)
```
serprex: what stops this from working with PG type system vs Q type system?

pratheekrebala: I'm not sure I entirely follow the question!

serprex: This generic is to support our 2 type systems: Q type system & PG type system. PG type system doesn't use `model.RecordItems` because it uses a much simpler representation (pg's text representation) to maintain 1:1 full compat for pg to pg replication

Your code would not append when used with the pg to pg type system

pratheekrebala: Ah! I didn't entirely understand the nuance there earlier. I updated the `appendOriginMeta` function to use the generic `UpdateIfNotExists` function instead

Although, in this case, would pg -> pg records even pass through `RecordsToRawTableStream`? My understanding was that only the BigQuery, Clickhouse, Snowflake and S3 connectors would call this function.... (since the PG destinations use logic from [syncRecordsCore](https://github.com/PeerDB-io/peerdb/blob/52d22fdf66b0bf756c275385ee77a6245f8da1ff/flow/connectors/postgres/postgres.go#L497))

I'm sorry if I'm entirely misunderstanding.....

flow/connectors/utils/stream.go
```diff
@@ -84,22 +85,36 @@ func RecordsToRawTableStream[Items model.Items](
...
+
 		itemsJSON, err := model.ItemsToJSON(preprocessedItems)
 		if err != nil {
 			return nil, fmt.Errorf("failed to serialize insert record items to JSON: %w", err)
 		}
 
 		entries[3] = types.QValueString{Val: itemsJSON}
 		entries[4] = types.QValueInt64{Val: 0}
-		entries[5] = types.QValueString{Val: ""}
+		entries[5] = types.QValueString{Val: originItemsJSON}
```
serprex: This should be kept empty, inserts have no old items

flow/connectors/utils/stream.go
serprex: Right. You should be adding this info in source records, not in destination

PR #3200, opened 2025-07-11
mongo: add read preference to mongo config

jgao54:
Implement backend to support passing ReadPreference as an explicit mongo config setting via UI

### DIFF
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -411,9 +411,19 @@ func parseAsClientOptions(config *protos.MongoConfig) (*options.ClientOptions, e
 		// always use majority read concern for correctness
 		SetReadConcern(readconcern.Majority())
 
-	// allow user to override with other read preference
-	if connStr.ReadPreference == "" {
+	// TODO: remove the empty string option once it's wired through the UI
+	if config.ReadPreference == "" {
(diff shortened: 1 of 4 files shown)


flow/connectors/mongo/mongo.go
```diff
@@ -411,9 +411,19 @@ func parseAsClientOptions(config *protos.MongoConfig) (*options.ClientOptions, e
 		// always use majority read concern for correctness
 		SetReadConcern(readconcern.Majority())
 
-	// allow user to override with other read preference
-	if connStr.ReadPreference == "" {
+	// TODO: remove the empty string option once it's wired through the UI
```
Amogh-Bharadwaj: I was wondering what if a PeerDB API user doesn't specify the field. Do we want to assert that this has a value in peer validation?

Amogh-Bharadwaj (approved):

flow/connectors/mongo/mongo.go
jgao54: hmm, good catch -- will add assertion for now and we could introduce a default later on if we find it to be useful

jgao54: If i add the assertion now it will not be easy to test pipe without the UI selection today :)  I think when we have the UI we can add an assertion

PR #3202, opened 2025-07-11
Timezone cleanup

serprex:
1. if-err cleanup
2. don't wait on sync to cancel when pausing, this was causing pausing to sometimes take 10+ seconds
3. set timezone UTC on postgres connections
4. parse datetime as UTC in ClickHouse

4 addresses this:
```
select parseDateTimeBestEffort('2025-07-11 19:30:02.313462+0000')

SELECT parseDateTimeBestEffort('2025-07-11 19:30:02.313462+0000')

Query id: 0d4bbad2-fdab-414f-85c8-2c80c61ad0ab

┌─parseDateTimeBestEffort('2025-07-11 19:30:02.313462+0000')─┐
│                                        2025-07-11 15:30:02 │
└────────────────────────────────────────────────────────────┘
```
where the ClickHouse server has a non-UTC timezone. The problem is our timezone hygiene isn't clear enough bringing data from postgres & ingesting into ClickHouse to handle this, between initial loads & cdc, that we should instead maintain a policy of "peerdb transmits times as UTC"

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -73,10 +73,10 @@ func NewPostgresConnector(ctx context.Context, env map[string]string, pgConfig *
 		return nil, err
 	}
 
-	runtimeParams := connConfig.Config.RuntimeParams
-	runtimeParams["idle_in_transaction_session_timeout"] = "0"
-	runtimeParams["statement_timeout"] = "0"
-	runtimeParams["DateStyle"] = "ISO, DMY"
+	connConfig.Config.RuntimeParams["timezone"] = "UTC"
+	connConfig.Config.RuntimeParams["idle_in_transaction_session_timeout"] = "0"
+	connConfig.Config.RuntimeParams["statement_timeout"] = "0"
+	connConfig.Config.RuntimeParams["DateStyle"] = "ISO, DMY"
 
 	tunnel, err := utils.NewSSHTunnel(ctx, pgConfig.SshConfig)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


ilidemi (approved):

flow/connectors/postgres/postgres.go
```diff
@@ -73,10 +73,10 @@ func NewPostgresConnector(ctx context.Context, env map[string]string, pgConfig *
 		return nil, err
 	}
 
-	runtimeParams := connConfig.Config.RuntimeParams
-	runtimeParams["idle_in_transaction_session_timeout"] = "0"
-	runtimeParams["statement_timeout"] = "0"
-	runtimeParams["DateStyle"] = "ISO, DMY"
+	connConfig.Config.RuntimeParams["timezone"] = "UTC"
```
jgao54, line 76: for my understanding, this runtime param updates the timezone of the client session,  does this setting impact data received from export snapshot and logical replication as well when TZ is specified? (if so, that's really convenient that the client just handles the timezone conversion)

serprex: https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-TIMEZONE

overall I'm trying to set UTC everywhere. this affects display, which is what pgx will eventually parse. probably also want to set this on catalog connections, but haven't been able to reproduce batch times having skew (ie, all batches are 7h because start time is PDT but end time is UTC, except that somehow the timezone info is gone so on ui we just get 7h batch durations)

jgao54: ah gotcha, this is specifically for timestamp data from query results

jgao54 (approved):

PR #3203, opened 2025-07-11
[mongo] try to save resumeToken even when idle

heavycrystal:


flow/connectors/mongo/mongo.go
```diff
@@ -268,100 +267,125 @@ func (c *MongoConnector) PullRecords(
...
-					BaseRecord:           model.BaseRecord{CommitTimeNano: clusterTimeNanos},
-					Items:                items,
-					SourceTableName:      sourceTableName,
-					DestinationTableName: destinationTableName,
-				}); err != nil {
-					return fmt.Errorf("failed to add delete record: %w", err)
+			}
+		}
+		// this branch will almost always be hit with the .Next() API
+		if err := changeStream.Err(); err != nil {
```
jgao54: This makes me wonder if we should even attempt to prolong the lifetime of PullRecord when we get a context deadline exceeded error, and instead when idleTimeout is reached we just exit and let the next PullRecord continue from where this left off.

jgao54: Because the context timeout reset below is essentially not doing anything

flow/connectors/mongo/mongo.go
```diff
@@ -371,7 +395,7 @@ func defaultPipeline() mongo.Pipeline {
 	return mongo.Pipeline{
 		{{Key: "$match", Value: bson.D{
 			{Key: "operationType", Value: bson.D{
-				{Key: "$in", Value: bson.A{"insert", "update", "replace", "delete"}},
+				{Key: "$in", Value: bson.A{"insert", "update", "replace", "delete", "invalidate"}},
```
jgao54: nit: is this intentionally added to fail loudly if we do get it?  (since deployment level change stream doesn't get invalidate event)

flow/connectors/mongo/mongo.go
```diff
@@ -246,10 +246,9 @@ func (c *MongoConnector) PullRecords(
 		}
 		c.logger.Info(fmt.Sprintf("[finished] PullRecords streamed %d records", recordCount))
 	}()
-	// before first record, we wait indefinitely so give it ctx
+	// before first record, we wait for an hour and reset so we can periodically save resumeToken
 	// after first record, we wait for idle timeout
-	getCtx := ctx
-	var cancelTimeout context.CancelFunc
+	getCtx, cancelTimeout := context.WithTimeout(ctx, time.Hour)
```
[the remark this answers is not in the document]
serprex: time.Hour isn't a magic number, it's an hour. `time.Duration(3600000000000)` would be magic number

flow/connectors/mongo/mongo.go
serprex: need to avoid creating empty batches if you do that

jgao54: cc @serprex  -- i know we do this in mysql, what's the benefit of prolonging an existing activity when idle (comparing to starting a new activity at the next sync interval)

jgao54: discussed offline with @serprex -- it will lead to cluttered UI and wasted effort processing empty batches. and since postgres require connection to stay open, it would be better to handle all connectors consistently.

I'm wondering if the best option here is to just call changestream.Next() without an idleTimeout deadline -- but in a separate goroutine where we pass decoded events back via a channel. This will allow us to prolong Activity lifetime in the main loop, something like:
```
select {
   case <- ctx.done():
          break
   case <-time.After(config.idleTimeout):
         saveResumeToken()
   case event :=  <-eventCh:
         // process event or error
}
```

serprex: That'd work, so long as:
1. this cannot be done once a record is received, so `saveResumeToken` here must not race with resume token related to receiving first record
2. `saveResumeToken` must be threadsafe

Other option is to have a `time.Ticker` that's used in main loop, but you want to stop the ticker once messages are being received

In MySQL I went the context timeout route since there's symmetry where before first record we're waiting on hour timer & after first record we're waiting on sync interval timer

jgao54: good catch. it looks like changestream API isn't thread-safe when it comes to reading/write the `cs.resumeToken` field. 

I think this means we are left with options of:
- create a new changestream instance when context times out
- use TryNext instead of Next which is non-blocking 

I think second approach means we need to keep polling on TryNext, which i am not a fan.

jgao54: Something like this: https://github.com/PeerDB-io/peerdb/pull/3205/files?diff=split&w=1, this also avoids the double for-loop, which is not ideal

jgao54: Discussed with @heavycrystal  -- merged change linked above.

flow/connectors/mongo/mongo.go
jgao54: Going to delete for now, can be introduced separately if needed.  the data contains different fields than the other operationType so would encounter deserialization issue today since it's not handled.

flow/connectors/mongo/mongo.go
```diff
@@ -229,46 +231,89 @@ func (c *MongoConnector) PullRecords(
...
-			// after the first record, we switch to a timeout context
-			getCtx, cancelTimeout = context.WithTimeout(ctx, req.IdleTimeout)
+			timeoutCtx, cancelTimeout = context.WithTimeout(ctx, req.IdleTimeout)
 		}
 		return nil
 	}
 
-	for recordCount < req.MaxBatchSize && changeStream.Next(getCtx) {
+	recreateChangeStream := func() error {
+		if err := changeStream.Close(timeoutCtx); err != nil {
```
serprex: seems odd passing timeoutCtx when we're calling this after hitting DeadlineExceeded already

jgao54: ah,  should just be `ctx` here.

PR #3204, opened 2025-07-12
Add OpenSearch support via Elasticsearch connector shim

sabino:
## Summary

This PR adds OpenSearch compatibility by reusing the existing Elasticsearch destination connector instead of maintaining a separate OpenSearch connector.

The implementation follows the review direction from the PR discussion: OpenSearch is treated as another backend for the existing Elasticsearch integration, with backend-specific behavior hidden behind a small client shim.

## Implementation

- Reworked the Elasticsearch connector so it detects whether the configured endpoint is Elasticsearch or OpenSearch and then uses the appropriate client internally.
- Removed the separate OpenSearch connector path and its duplicate connector registration.
- Kept the user-facing peer type and config surface centered on the existing Elasticsearch connector rather than introducing a second parallel destination type.
- Reused the existing Elasticsearch end-to-end coverage for both backends instead of maintaining a second OpenSearch-specific test suite.
- Updated CI to start an OpenSearch service as an additional backend and to exercise the shared destination test path against it.
- Added a small backend-detection unit test for the shim layer.

## Why This Approach

The original PR copied the Elasticsearch connector into a separate OpenSearch implementation. That worked functionally, but it duplicated nearly all of the destination logic and test coverage.

This version keeps the code easier to maintain by:

- having one destination connector implementation,
- isolating backend differences in one shim layer,
- avoiding duplicate CDC and QRep logic, and
- keeping Elasticsearch and OpenSearch behavior covered through the same test surface.

## Validation

Locally validated with:

- generated protobuf and Go code regeneration,
- compile-only Go package checks for the affected connector and e2e packages,
- unit tests covering backend detection for the shim, and
- targeted local golangci-lint on the affected Flow packages.

## Notes

The upstream PR environment still has gated jobs that depend on repository-side approval for fork-originated `pull_request_target` workflows. This PR is now ready for review, and the branch contains the merge from current `main` plus the follow-up fixes needed to satisfy local compile and lint validation.


.github/workflows/flow.yml
```diff
@@ -63,6 +63,24 @@ jobs:
           MINIO_API_PORT_NUMBER: 9999
           AWS_EC2_METADATA_DISABLED: true
           MINIO_DEFAULT_BUCKETS: peerdb
+      opensearch:
+        image: opensearchproject/opensearch:2.13.0
```
serprex:
```suggestion
        image: opensearchproject/opensearch:3
```

heavycrystal (commented):
At a high-level, it seems to me that the connector is a direct adaptation of the ES connector with a different library. Is there differences besides the library where having an entirely separate connector is needed?

If not, it seems that OpenSearch can be a "flavour" of the ES connector (similar to how MySQL/MariaDB works). Code should switch on what library to use underneath via a shim or similar

heavycrystal:
> @heavycrystal yes! Thanks for the review.
> 
> You're absolutely right that this shares a lot of ground with the existing Elasticsearch connector. [I even brought this up in the Community Slack before](https://peerdb-public.slack.com/archives/C05BMVAM9FD/p1751636448758989) implementing it, as my initial thought was also to shim the client based on the backend selecting OpenSearch or Elasticsearch dynamically.
> 
> At the moment, I don't have strong opinions either way. That said, my reasoning for keeping them separate for now is that it helps clarify and isolate the nuances between the two. Especially as their APIs and behaviors are already diverging in subtle ways. For example, the client libraries differ, and certain API endpoints and auth mechanisms are evolving independently.
> 
> That said, [OpenSearch still claims compatibility with Elasticsearch 7.10 clients](https://opensearch.org/faq/#:~:text=Is%20OpenSearch%20wire%2Dcompatible%20with%20Elasticsearch%3F), so it might indeed make sense to unify them into a single connector for now. Using a client shim and perhaps probing the server version or exposing a configuration flag to let the user choose.
> 
> Happy to revisit the structure based on your thoughts.

I think it would be easier to maintain if it was a single connector with shims and switches for different behaviour. I don't anticipate drastic changes happening, and we can always revisit this if they do. You can take a look at the work we did for MySQL/MariaDB for example, there's quite a few differences but it's still a single connector. 

You should be able to reuse the same tests and just test both flavours, let me know if this works

serprex (approved):

PR #3209, opened 2025-07-14
PEERDB_ORIGIN_METADATA_AS_DESTINATION_COLUMN

serprex:
closes #3197

on non-queues this info will only end up in raw table for debugging purposes, but for queues/s3 this is sufficient

### DIFF
--- a/flow/e2e/s3/cdc_s3_test.go
+++ b/flow/e2e/s3/cdc_s3_test.go
@@ -20,11 +20,11 @@ func (s PeerFlowE2ETestSuiteS3) attachSuffix(input string) string {
 	return fmt.Sprintf("%s_%s", input, s.suffix)
 }
 
-func (s PeerFlowE2ETestSuiteS3) Test_Complete_Simple_Flow_S3() {
+func (s PeerFlowE2ETestSuiteS3) Test_Simple() {
 	tc := e2e.NewTemporalClient(s.t)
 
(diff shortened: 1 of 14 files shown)


flow/e2e/s3/cdc_s3_test.go
```diff
@@ -80,3 +80,52 @@ func (s PeerFlowE2ETestSuiteS3) Test_Complete_Simple_Flow_S3() {
...
+	e2e.EnvNoError(s.t, env, err)
+
+	e2e.EnvWaitFor(s.t, env, 2*time.Minute, "waiting for blobs", func() bool {
+		ctx, cancel := context.WithTimeout(s.t.Context(), 25*time.Second)
+		defer cancel()
+		files, err := s.s3Helper.ListAllFiles(ctx, flowJobName)
+		e2e.EnvNoError(s.t, env, err)
+		s.t.Logf("Files in Test_Complete_Simple_Flow_S3 %s: %d", flowJobName, len(files))
+		return len(files) == 4
+	})
```
serprex, line 122: this tests that we write something to s3, but ideally test pulls avro from s3 & checks for columns

maybe kafka test would be better place, could also test scripting then

pratheekrebala:
I have tested this PR with pg -> pg, pg -> s3 and pg -> redpanda mirrors and all of them worked as expected! Thanks again for working on getting these changes merged!

heavycrystal (approved):

PR #3217, opened 2025-07-15
mongo: use enum for readPreference instead of string

jgao54:
Use enum instead of string for readPreference to match driver; more compact for wiring over the network as well.

### DIFF
--- a/protos/peers.proto
+++ b/protos/peers.proto
@@ -63,7 +63,7 @@ message PubSubConfig {
 }
 
 enum ReadPreference {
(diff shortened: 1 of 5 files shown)


protos/peers.proto
```diff
@@ -62,6 +62,15 @@ message PubSubConfig {
   GcpServiceAccount service_account = 1;
 }
 
+enum ReadPreference {
+  NULL = 0;
```
heavycrystal: nit: should be `UNKNOWN`

PR #3224, opened 2025-07-16
mongo: log a warning for unsupported events

jgao54:
Recommend review the commits separately:
- move cdc logic to a separate cdc.go file, no logic change in [this commit](https://github.com/PeerDB-io/peerdb/pull/3224/commits/6b8e59d6715ab7429797082f4758d764889e0cdf).
- Instead of filter out specific events,  process all events corresponding to watched tables, and log a warning for unsupported operations and continue on [in this commit](https://github.com/PeerDB-io/peerdb/pull/3224/commits/d5000ee0e57ca807b36f9644899e21b98e5bda04).

Reason for logging a warning instead of erroring out after thinking about this more:
- a pipe can contain multiple tables, so seems preferable to allow normal CDC of other tables when one table encounters an unsupported operation.
- we could technically support rename tables by mapping new name to old name and continue to use the same CH table (this means we need to persist the new-to-old table name mapping which is not super trivial), or bootstrap to a new destination table with matching names (this can error if the renamed destination table already exists). For now, just log warning and let user decide on how best to handle it when a table is renamed.

Testing: manual e2e testing to confirm that:
- renaming a collection will generate a `rename` event in the changestream, and log a warnng
- dropping a collection will generate a `drop` event in the changestream, and log a warning
- dropping a database will generate a `drop` event for each watched table in the database, and log a warning for each watched table.
- createCollection/createIndex/dropIndex/modify do not generate a change event -- this requires setting `showExpandedEvents` to true if for whatever reason we want to support it later.
- `refineCollectionShardKey`, `shardCollection`, `reshardCollection` events are only applicable to sharded cluster, but log an warning just the same since we will be supporting it soon.


flow/connectors/mongo/cdc.go
```diff
@@ -224,9 +231,16 @@ func (c *MongoConnector) PullRecords(
 			return fmt.Errorf("failed to decode change stream document: %w", err)
 		}
 
-		if _, ok := changeDoc["operationType"]; !ok {
+		if op, ok := changeDoc["operationType"]; !ok {
 			c.logger.Warn("operationType field not found")
 			continue
+		} else if _, supported := SupportedOperations[op.(string)]; !supported {
```
serprex: this logic overlaps with the switch below, this could live in that switch's default case

jgao54: It's not super obvious, but these unsupported types don't have documentKey/fullDocument. 

Pipeline projection doesn't require all fields to exist. So checking operation type first here.

serprex: It may make sense to have `SupportedOperations` be `map[string]func()` which handles the cases then, removing unreachable default case

Logic is currently confused since below we check `op.(string)` cast, but up here we panic when it isn't a string

serprex (approved):

PR #3225, opened 2025-07-16
run CI in local container

serprex:
currently tests run in container, but ideally we'd run go test with just the environment variables in local-ci.sh script, would avoid downloading everything

### DIFF
--- a/docker-compose-e2e.yml
+++ b/docker-compose-e2e.yml
@@ -1,0 +1,278 @@
+name: peerdb-quickstart-dev
+
+x-minio-config: &minio-config
+  PEERDB_CLICKHOUSE_AWS_CREDENTIALS_AWS_ACCESS_KEY_ID: _peerdb_minioadmin
+  PEERDB_CLICKHOUSE_AWS_CREDENTIALS_AWS_SECRET_ACCESS_KEY: _peerdb_minioadmin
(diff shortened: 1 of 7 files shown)


docker-compose-e2e.yml
```diff
@@ -0,0 +1,278 @@
...
+      export MINIO_ROOT_USER=$$PEERDB_CLICKHOUSE_AWS_CREDENTIALS_AWS_ACCESS_KEY_ID;
+      export MINIO_ROOT_PASSWORD=$$PEERDB_CLICKHOUSE_AWS_CREDENTIALS_AWS_SECRET_ACCESS_KEY;
+      minio server /data --console-address=":36987" &
+      sleep 2;
+      /usr/bin/mc alias set myminiopeerdb http://minio:9000 $$MINIO_ROOT_USER $$MINIO_ROOT_PASSWORD;
+      /usr/bin/mc mb myminiopeerdb/$$PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME;
+      wait
+      "
+
+  mysql:
```
serprex: currently not running mysql tests here, but that'd be a followup

iskakaushik:
i'd love to see all the go code run from source rather than flow-api and snapshot worker be images.

PR #3227, opened 2025-07-16
mongo: always return full table partition

jgao54:
@Amogh-Bharadwaj  noticed an issue where initial load does not populate stats to db.

- Fix a bug where `nil` partition is returned.
- Make sure mongo also populate `qrep_partitions` table

Test: ran test locally and see that qrep_partition populated for mongo:
```
postgres=# select rows_synced, parent_mirror_name from qrep_partitions where parent_mirror_name = 'stats3';
 rows_synced | parent_mirror_name
-------------+--------------------
           4 | stats3
```

### DIFF
--- a/flow/connectors/utils/monitoring/monitoring.go
+++ b/flow/connectors/utils/monitoring/monitoring.go
@@ -18,6 +18,7 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/internal"
 	"github.com/PeerDB-io/peerdb/flow/model"
 	"github.com/PeerDB-io/peerdb/flow/shared"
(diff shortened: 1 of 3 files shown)


flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -267,11 +268,12 @@ func AppendSlotSizeInfo(
 	return nil
 }
 
-func isMySQLFullTablePartition(partition *protos.QRepPartition) bool {
+func supportsStatsForFullTablePartition(partition *protos.QRepPartition) bool {
```
jgao54, line 271: conservative change, but wondering if we could delete this function altogether.

serprex (approved):

PR #3231, opened 2025-07-17
[mongo] SSH tunnelling and metered connection

### LINKED ISSUE #3110: mongo: metered connection to track bytes

### LINKED ISSUE #3175: mongo: SSH tunnelling

heavycrystal:
uses a goroutine to periodically report during CDC

closes #3110 
closes #3175


flow/connectors/mongo/cdc.go
```diff
@@ -161,17 +160,30 @@ func (c *MongoConnector) PullRecords(
...
 		cancelTimeout()
+		cancelReportBytes()
+	}()
+
+	go func() {
+		for {
+			select {
+			case <-reportBytesCtx.Done():
+				return
+			case <-time.After(10 * time.Second):
```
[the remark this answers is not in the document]
heavycrystal: fixed in newer Go:

```
Before Go 1.23, this documentation warned that the underlying [Timer] would not be recovered by the garbage collector until the timer fired, and that if efficiency was a concern, code should use NewTimer instead and call [Timer.Stop] if the timer is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers. There is no reason to prefer NewTimer when After will do.
```

ui/components/PeerForms/MongoForm.tsx
```diff
@@ -158,6 +171,74 @@ export default function MongoForm({
...
+                  type={sshParam.type}
+                  defaultValue={
+                    (sshConfig as SSHConfig)[
+                      sshParam.label === 'SSH Private Key'
+                        ? 'privateKey'
+                        : sshParam.label === "Host's Public Key"
+                          ? 'hostKey'
+                          : (sshParam.label.toLowerCase() as keyof SSHConfig)
+                    ] || ''
+                  }
```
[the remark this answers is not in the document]
heavycrystal: will be fixed as a follow up

flow/connectors/mongo/cdc.go
```diff
@@ -161,17 +160,30 @@ func (c *MongoConnector) PullRecords(
...
 	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, time.Hour)
-	//nolint:gocritic // cancelTimeout is rebound, do not defer cancelTimeout()
+	reportBytesCtx, cancelReportBytes := context.WithCancel(ctx)
+
 	defer func() {
 		cancelTimeout()
+		cancelReportBytes()
+	}()
+
+	go func() {
```
jgao54: nit: can use shared.Interval ([example](https://github.com/PeerDB-io/peerdb/blob/6355207c115db673f18ecfb11ed9d200484d950e/flow/connectors/mongo/qrep.go#L55-L61))

flow/activities/flowable_core.go
```diff
@@ -453,6 +453,7 @@ func replicateQRepPartition[TRead any, TWrite StreamCloser, TSync connectors.QRe
 			return a.Alerter.LogFlowError(ctx, config.FlowJobName, fmt.Errorf("[qrep] failed to pull records: %w", err))
 		}
 
+		// should only be effective for Postgres source, others report within connector
```
serprex: this is still used by mysql since it returns whatever wasn't reported by goroutine yet at end

flow/connectors/mongo/metered_dialer.go
```diff
@@ -0,0 +1,45 @@
...
+	innerDialer client.Dialer
+}
+
+func (md *meteredDialer) DialContext(ctx context.Context, network, address string) (net.Conn, error) {
+	conn, err := md.innerDialer(ctx, network, address)
+	if err != nil {
+		return nil, err
+	}
+	return &meteredConn{
+		NoDeadlineConn: utils.NoDeadlineConn{Conn: conn},
```
serprex: NoDeadline should only be used with ssh

flow/connectors/mongo/metered_dialer.go
```diff
@@ -0,0 +1,45 @@
...
+		bytesRead:      md.bytesRead,
+	}, nil
+}
+
+func newMeteredDialer(bytesRead *atomic.Int64, innerDialer client.Dialer) meteredDialer {
```
serprex: can make this shared, then have mysql reuse this with to keep reference to `client.Dialer` encapsulated

flow/connectors/mongo/metered_dialer.go
```diff
@@ -0,0 +1,45 @@
+package connmongo
+
+import (
+	"context"
+	"net"
+	"sync/atomic"
+
+	"github.com/go-mysql-org/go-mysql/client"
```
serprex:
```suggestion
```

flow/connectors/mongo/metered_dialer.go
```diff
@@ -0,0 +1,45 @@
+package connmongo
```
jgao54: could we consolidate MeteredDialer with mysql's metered_dialer? it looks like there's a lot of overlap between the two

flow/connectors/mongo/metered_dialer.go
heavycrystal: was being used to pass dialer in, switching to own type

flow/connectors/mongo/cdc.go
heavycrystal: good catch

PR #3239, opened 2025-07-18
allow configuring clickhouse destination tables' PARTITION BY

serprex:

PR #3245, opened 2025-07-18
refactor custom order by ui

serprex:
split out from #3239

cuts down on having state in component which needs to be consolidated upstream with effects

### DIFF
--- a/ui/app/mirrors/create/cdc/sortingkey.tsx
+++ b/ui/app/mirrors/create/cdc/sortingkey.tsx
@@ -3,7 +3,7 @@ import {
   Dispatch,
   SetStateAction,
   useCallback,
-  useEffect,
+  useMemo,
   useState,
 } from 'react';
 import ReactSelect from 'react-select';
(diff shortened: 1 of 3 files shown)


ui/app/mirrors/create/cdc/sortingkey.tsx
```diff
@@ -30,81 +31,113 @@ interface SortingKeysProps {
...
+  const filtered = tableRow.columns.filter((col) => valueFunc(col) > 0);
+  filtered.sort((r1, r2) => valueFunc(r1) - valueFunc(r2));
+  return filtered;
+}
+
+function UpdatedRows(
+  rows: TableMapRow[],
+  tableRow: TableMapRow,
+  colName: string,
+  update: (setting: ColumnSetting) => ColumnSetting
```
serprex, line 47: these functions are a bit out of place here since it'd be simple enough to specialize, but in #3239 this is used to handle a symmetrical field to `ordering` named `partitioning`

Amogh-Bharadwaj (approved):

Issue #3234, opened 2025-07-17
Mark mirror as failed in certain scenarios


iamKunalGupta:
- [ ] When retries exceed max number during Setup Flow
- [ ] When mirror in Snapshot errors out with Invalid Snapshot Identifier

Resyncing should work for these failed states

PR #3255, opened 2025-07-22
update workflow status to FAILED when completing with error

serprex:
or TERMINATED when workflow context canceled (probably means canceled through temporal ui)

partially addresses #3234

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -521,17 +521,26 @@ func CDCFlowWorkflow(
 
 	originalRunID := workflow.GetInfo(ctx).OriginalRunID
 
-	var err error
-	ctx, err = GetFlowMetadataContext(ctx, &protos.FlowContextMetadataInput{
-		FlowName:        cfg.FlowJobName,
-		SourceName:      cfg.SourceName,
-		DestinationName: cfg.DestinationName,
-		Status:          state.CurrentFlowStatus,
-		IsResync:        cfg.Resync,
-	})
-	if err != nil {
-		state.updateStatus(ctx, logger, protos.FlowStatus_STATUS_FAILED)
-		return state, fmt.Errorf("failed to get flow metadata context: %w", err)
+	for {
+		if err := ctx.Err(); err != nil {
+			state.updateStatus(ctx, logger, protos.FlowStatus_STATUS_TERMINATED)
+			return state, err
+		}
+
+		var err error
+		ctx, err = GetFlowMetadataContext(ctx, &protos.FlowContextMetadataInput{
+			FlowName:        cfg.FlowJobName,
+			SourceName:      cfg.SourceName,
+			DestinationName: cfg.DestinationName,
+			Status:          state.CurrentFlowStatus,
+			IsResync:        cfg.Resync,
+		})
+		if err != nil {
+			logger.Error("failed to GetFlowMetadataContext", slog.Any("error", err))
+			continue
+		} else {
+			break
+		}
 	}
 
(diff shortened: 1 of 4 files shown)


flow/workflows/cdc_flow.go
```diff
@@ -528,6 +530,7 @@ func CDCFlowWorkflow(
 		IsResync:        cfg.Resync,
 	})
 	if err != nil {
+		syncStatusToCatalog(ctx, workflow.GetLogger(ctx), protos.FlowStatus_STATUS_FAILED)
```
iamKunalGupta: This should not fail, for e.g. if catalog is down temporarily

flow/workflows/cdc_flow.go
```diff
@@ -616,6 +619,7 @@ func CDCFlowWorkflow(
 				return nil, err
 			}
 			if setupFlowError != nil {
+				syncStatusToCatalog(ctx, workflow.GetLogger(ctx), protos.FlowStatus_STATUS_FAILED)
```
iamKunalGupta: should we update the status on line 617 for TerminateSignal? similarly at 720

iamKunalGupta (commented):
also small q: should we be calling `updateStatus` instead? Might allow for temporal status to be consistent (although the status is terminal)

flow/workflows/cdc_flow.go
heavycrystal: won't terminate remove stuff from catalog to need to update?

serprex: arguably reads nice, but we're discarding state after this so it isn't useful

iamKunalGupta: If drop flow source or destination is hung, then catalog state cleanup can take time

flow/workflows/cdc_flow.go
serprex: what should we do? log warning & continue? retry?

iamKunalGupta: we can do LogFlowError and retry as bulk of alerting/monitoring relies on this information being available

serprex: with catalog down I'm not sure LogFlowError works, but will log error & retry

flow/workflows/cdc_flow.go
```diff
@@ -519,16 +521,26 @@ func CDCFlowWorkflow(
...
+		ctx, err = GetFlowMetadataContext(ctx, &protos.FlowContextMetadataInput{
+			FlowName:        cfg.FlowJobName,
+			SourceName:      cfg.SourceName,
+			DestinationName: cfg.DestinationName,
+			Status:          state.CurrentFlowStatus,
+			IsResync:        cfg.Resync,
+		})
+		if err != nil {
+			logger.Error("failed to GetFlowMetadataContext", slog.Any("error", err))
+			continue
```
iamKunalGupta, line 540: we should probably add a sleep here to prevent overloading catalog

serprex: tbh this will rarely happen because temporal will be retrying the local activity in this scenario, `Get`'s failure would probably be rooted in type mismatch which we can't really do anything about besides not have type errors

StartToCloseTimeout leads to temporal retry happening, not activity failure

flow/workflows/cdc_flow.go
ilidemi: I have a change that moves UpdateFlowStatusInCatalogActivity to a non-local activity, seems ok to retry until catalog recovers

iamKunalGupta (approved):

PR #3256, opened 2025-07-22
remove custom sync

serprex:
flakey test, feature is a bit of a hack (especially after parallel sync normalize changes)


protos/flow.proto
```diff
@@ -420,7 +419,6 @@ message CDCFlowConfigUpdate {
   repeated TableMapping additional_tables = 1;
   uint32 batch_size = 2;
   uint64 idle_timeout = 3;
-  int32 number_of_syncs = 4;
```
ilidemi: Do these need to be reserved?

serprex: arguably yes, but only when they're non-zero, which they never should be, assuming this feature wasn't being used

if we want to handle getting rid of this gradually, we need to add a code path that zeros out this field at the start of cdc_flow for the next few releases. instead I'd suggest anyone who was using this before hit the endpoint to zero it out themselves before upgrading

serprex: added migration logic

ilidemi (approved):

Issue #3246, opened 2025-07-21
mongo: Resume token error after pause, insert and resume


Amogh-Bharadwaj:
Came across this error:
```

failed to create change stream: (ChangeStreamFatalError) PlanExecutor error during aggregation :: caused by :: cannot resume stream; the resume token was not found. {_data: "82687E82DE000000012B042C0100296E5A1004FEA7A1FFE23C4FD88070BFA43C9C995C463C6F7065726174696F6E54797065003C696E736572740046646F63756D656E744B65790046645F69640064687E82DC122A414A3ECD77F8000004"}
--
```

during CDC when testing the MongoDB mirror.
Steps to reproduce:
<redacted>

jgao54:
Minimum repro:

```
const p1 = [ { "$match": { "$and": [ {"ns.db": "test"}, {"ns.coll": {"$in": ["t1", "t2"]}} ] } } ];
const changeStream1 = db.watch(p1);
db.t1.insertOne({"a": 1});
const event = changeStream1.next();

// pause and delete a table

const p2 = [ { "$match": { "$and": [ {"ns.db": "test"}, {"ns.coll": {"$in": ["t2"]}} ] } } ];
const changeStream2 = db.watch(p2, {resumeAfter: event._id});
db.t2.insertOne({"b": 1});
changeStream2.next();
MongoServerError[ChangeStreamFatalError]: PlanExecutor error during aggregation :: caused by :: cannot resume stream; the resume token was not found. {_data: "82687F3418000000012B042C0100296E5A100402029C35AFFD457AA3093B44F7D71C6D463C6F7065726174696F6E54797065003C696E736572740046646F63756D656E744B65790046645F69640064687F3418245CA5A3B5F47124000004"}
```

Turns out this is expected behavior of Mongo change stream. We fail to resume because the delete table is filtered out of the changestream so mongo fails to reconcile to the resumeToken from the deleted table. Working on a fix where we store latest resume token from each table.

PR #3259, opened 2025-07-23
mongo: handle resume token from table not in changestream

jgao54:
See https://github.com/PeerDB-io/peerdb/issues/3246 for problem description.

Evaluated a few options:
1. persist previous table mapping so we can gracefully remove table from pipeline when we detect a table is deleted by doing so after changestream has been successfully resumed
- pro: relatively simple to implement
- con: need to store table mapping to catalog which may require db schema change

2. persist per-table resume tokens to `lastCheckpointText`, so we can always resume to a different table's token that exists in the pipeline
- pro: we can re-use `lastCheckpointText` without changing db schema
- con:  overhead of ser/deser resume tokens to/from string on checkpoint; more complex logic to find the "next best resume token" if last resumeToken doesn't work

3. decode resume token to extract clusterTime so we can set `ResumeAtOperationTime` instead
- pro: cleanest solution, as no additional data is needed other than the resume token itself that we already store at `lastCheckpointText`
- con: no existing mongo/bson library to decode resumeToken, need custom implementation

I ended up going with approach 3 because:
- The fact clusterTime is the first item to decode from resumeToken makes the implementation quite simple
- I expect the resumeToken format to be relatively stable so minimum maintenance should be needed
- Minimum performance overhead with this approach, since we only need to handle `decodeTimestampFromResumeToken` if we encounter this specific `the resume token was not found.` error which is quite rare

Test: locally + add e2e test

### DIFF
--- a/flow/connectors/mongo/cdc.go
+++ b/flow/connectors/mongo/cdc.go
@@ -131,6 +131,7 @@ func (c *MongoConnector) PullRecords(
 
 	var resumeToken bson.Raw
 	var err error
+
 	if req.LastOffset.Text != "" {
 		// If we have a last offset, we resume from that point
 		c.logger.Info("[mongo] resuming change stream", slog.String("resumeToken", req.LastOffset.Text))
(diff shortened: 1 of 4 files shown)


heavycrystal (commented):
Is there a difference between extracting the time from the resume token to be used for `startAtOperationTime` and just giving the resume token as input to `startAfter`?

serprex (approved):
Good to see chess advice "When you see a good move, look for a better one" applied to this problem

iamKunalGupta:
> persist per-table resume tokens to lastCheckpointText, so we can always resume to a different table's token that exists in the pipeline

This is something that might be added for iceberg as a source eventually

jgao54:
@heavycrystal  

re: giving the resume token as input to startAfter

it won't work, for the same reason that resumeAfter won't work (i.e. can't reconcile with the changestream)

heavycrystal (approved):

PR #3268, opened 2025-07-24
scheduled tasks as one long running activity

serprex:
Adds a flag to api so that it runs the scheduled tasks itself using a couple tickers rather than pushing this through temporal

Eventually this flag will be the default, & the old code path can be removed

Put it in api for now since we'd rather not introduce yet another process, & api process was already chosen for killing/starting scheduled tasks. This is because api pod isn't expected to need horizontal scaling

This greatly cleans up temporal ui being flooded in record-slot-size workflows, & reduce temporal costs

Closes #2380

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -730,9 +730,9 @@ func (a *FlowableActivity) SendWALHeartbeat(ctx context.Context) error {
 
 func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
 	ticker := time.NewTicker(time.Minute)
-	walHeartbeatCounter := 12
+	walHeartbeatCounter := 10
 	for range ticker.C {
-		activity.RecordHeartbeat(ctx, "recording slot size")
+		activity.RecordHeartbeat(ctx, "running")
 		if err := ctx.Err(); err != nil {
 			return err
 		}
@@ -741,7 +741,7 @@ func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
 		}
 		walHeartbeatCounter -= 1
 		if walHeartbeatCounter <= 0 {
-			walHeartbeatCounter = 12
+			walHeartbeatCounter = 10
 			if err := a.SendWALHeartbeat(ctx); err != nil {
 				slog.Error("[scheduled-tasks] SendWALHeartbeat failed", slog.Any("error", err))
 			}
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -728,6 +728,28 @@ func (a *FlowableActivity) SendWALHeartbeat(ctx context.Context) error {
 	return nil
 }
 
+func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
+	ticker := time.NewTicker(time.Minute)
+	walHeartbeatCounter := 12
+	for range ticker.C {
+		activity.RecordHeartbeat(ctx, "recording slot size")
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		if err := a.RecordSlotSizes(ctx); err != nil {
+			slog.Error("[scheduled-tasks] RecordSlotSizes failed", slog.Any("error", err))
+		}
+		walHeartbeatCounter -= 1
+		if walHeartbeatCounter <= 0 {
+			walHeartbeatCounter = 12
+			if err := a.SendWALHeartbeat(ctx); err != nil {
+				slog.Error("[scheduled-tasks] SendWALHeartbeat failed", slog.Any("error", err))
+			}
+		}
+	}
+	return nil
+}
+
 type flowInformation struct {
 	config     *protos.FlowConnectionConfigs
 	workflowID string
--- a/flow/cmd/api.go
+++ b/flow/cmd/api.go
@@ -251,11 +251,7 @@ func APIMain(ctx context.Context, args *APIServerParams) error {
 		TaskQueue: taskQueue,
 	}
 
-	if _, err := flowHandler.temporalClient.ExecuteWorkflow(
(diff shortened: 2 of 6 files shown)


flow/main.go
```diff
@@ -207,11 +211,12 @@ func main() {
...
-						TemporalHostPort:  temporalHostPort,
-						GatewayPort:       uint16(clicmd.Uint("gateway-port")),
-						TemporalNamespace: clicmd.String("temporal-namespace"),
-						EnableOtelMetrics: clicmd.Bool(otelMetricsFlag.Name),
+						Port:                             uint16(clicmd.Uint("port")),
+						TemporalHostPort:                 temporalHostPort,
+						GatewayPort:                      uint16(clicmd.Uint("gateway-port")),
+						TemporalNamespace:                clicmd.String("temporal-namespace"),
+						EnableOtelMetrics:                clicmd.Bool(otelMetricsFlag.Name),
+						RunScheduledTasksWithoutTemporal: clicmd.Bool("run-scheduled-tasks-without-temporal"),
```
serprex: Little messy having flag name repeated here, may clean this up a bit here after #3266 merges

flow/cmd/api.go
```diff
@@ -246,18 +249,56 @@ func APIMain(ctx context.Context, args *APIServerParams) error {
 		return fmt.Errorf("unable to kill existing scheduler flows: %w", err)
 	}
 
-	workflowID := fmt.Sprintf("scheduler-%s", uuid.New())
-	workflowOptions := client.StartWorkflowOptions{
-		ID:        workflowID,
-		TaskQueue: taskQueue,
-	}
+	if args.RunScheduledTasksWithoutTemporal {
+		go func() {
```
serprex: should probably move this logic into `internal` but wanted to get something out that can be reviewed before tidying up on the level of which files code belongs in

flow/workflows/scheduled_flows.go
```diff
@@ -3,61 +3,18 @@ package peerflow
...
-	slotSizeFuture := workflow.ExecuteActivity(ctx, flowable.RecordSlotSizes)
-	return slotSizeFuture.Get(ctx, nil)
-}
-
-// HeartbeatFlowWorkflow sends WAL heartbeats
-func HeartbeatFlowWorkflow(ctx workflow.Context) error {
-	if ctx.Err() != nil {
-		return ctx.Err()
+	slotSizeFuture := workflow.ExecuteActivity(ctx, flowable.ScheduledTasks)
+	if err := slotSizeFuture.Get(ctx, nil); err != nil {
```
ilidemi:
```suggestion
	scheduledTasksFuture := workflow.ExecuteActivity(ctx, flowable.ScheduledTasks)
	if err := scheduledTasksFuture(ctx, nil); err != nil {
```

ilidemi (approved):

flow/activities/flowable.go
```diff
@@ -728,6 +728,28 @@ func (a *FlowableActivity) SendWALHeartbeat(ctx context.Context) error {
...
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		if err := a.RecordSlotSizes(ctx); err != nil {
+			slog.Error("[scheduled-tasks] RecordSlotSizes failed", slog.Any("error", err))
+		}
+		walHeartbeatCounter -= 1
+		if walHeartbeatCounter <= 0 {
+			walHeartbeatCounter = 12
+			if err := a.SendWALHeartbeat(ctx); err != nil {
```
heavycrystal, line 745: do we want to heartbeat every 12 minutes still? seems like we can make this more frequent

flow/activities/flowable.go
```diff
@@ -728,6 +728,28 @@ func (a *FlowableActivity) SendWALHeartbeat(ctx context.Context) error {
...
+		activity.RecordHeartbeat(ctx, "recording slot size")
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		if err := a.RecordSlotSizes(ctx); err != nil {
+			slog.Error("[scheduled-tasks] RecordSlotSizes failed", slog.Any("error", err))
+		}
+		walHeartbeatCounter -= 1
+		if walHeartbeatCounter <= 0 {
+			walHeartbeatCounter = 12
```
heavycrystal: need a activity heartbeat here

flow/activities/flowable.go
serprex: it could be, but I'd like to avoid changing things as much as reasonable with this change. But since we don't have global cron making these fire at exact minute counts on the hour, you're right, I'll change it to 10 so the number seems normal to people

flow/activities/flowable.go
serprex: Heartbeat is run at every tick, heartbeat timeout is long enough that we shouldn't be in a rush to fire it again within loop

flow/cmd/api.go
```diff
@@ -32,11 +32,12 @@ import (
...
-	TemporalNamespace string
-	Port              uint16
-	GatewayPort       uint16
-	EnableOtelMetrics bool
+	TemporalHostPort                 string
+	TemporalNamespace                string
+	Port                             uint16
+	GatewayPort                      uint16
+	EnableOtelMetrics                bool
+	RunScheduledTasksWithoutTemporal bool
```
Amogh-Bharadwaj: I don't see this used -- is it for something in the future?

serprex: oops, this was leftover from when this flag would stop running workflow & run completely without temporal, but various reasons have come together so that we're only going with a long running activity now

serprex:
So it seems the way to go is to introduce a new workflow, & leave the existing workflow, but we only kick off the new workflow, so that soon the old workflow can be removed

PR #3269, opened 2025-07-25
Granular status tracking

ilidemi:
Track snapshot/sync/normalize/slot lag status in a granular way - sync and normalize can fail and recover independently, QRep runs and partitions can fail and recover independently.

Every error in Snapshot and running CDC flow is now reported to flow_errors. Calling alerter logging is unified to the top level activity level rather than being sprinkled in across the code.

Todo:
- [ ] Add the extra lookup fields and indices into flow_errors
- [ ] Double check PG writes are reliable
- [ ] Report the new status and errors in MirrorStatus
- [ ] Double check the lifecycles of status values
- [ ] What will happen when something is erroring out then there's a signal to switch, will there be a blip of stale status in the future
- [ ] Why are we treating ApplicationErrors and replState changed/slot is already active in a special way?
- [ ] Either log internal errors as internal errors and remove _is_internal_error or filter them out and use the field
- [ ] Slot lag threshold should be coming from the individual overrides in catalog
- [ ] Thresholding like IMR does - either move it over or integrate
- [ ] Testing


flow/internal/workflow.go
```diff
@@ -75,6 +76,8 @@ func GetWorkflowStatus(ctx context.Context, pool shared.CatalogPool,
 func UpdateFlowStatusInCatalog(ctx context.Context, pool shared.CatalogPool,
```
serprex, line 76: this can be local activity

flow/e2e/api/api_test.go
```diff
@@ -132,7 +131,7 @@ func (s Suite) checkCatalogTableMapping(
 	}
 
 	var config protos.FlowConnectionConfigs
-	err = proto.Unmarshal(configBytes, &config)
+	err = internal.ProtoUnmarshal(configBytes, &config)
 	if err != nil {
```
serprex:
```suggestion
	if err := internal.ProtoUnmarshal(configBytes, &config); err != nil {
```

PR #3270, opened 2025-07-25
always get status from catalog

serprex:
avoids various issues we've been having where status is in one place but not the other

TODO update tests to use catalog, this was also contributing to not seeing this issue in CI since catalog status was never used even tho it's what's always relied on in practice

### DIFF
--- a/flow/cmd/handler.go
+++ b/flow/cmd/handler.go
@@ -70,10 +70,15 @@ func (h *FlowRequestHandler) createCdcJobEntry(ctx context.Context,
 			req.ConnectionConfigs.DestinationName, dstErr)
 	}
 
+	cfgBytes, err := proto.Marshal(req.ConnectionConfigs)
+	if err != nil {
+		return fmt.Errorf("unable to marshal flow config: %w", err)
+	}
+
 	if _, err := h.pool.Exec(ctx,
-		`INSERT INTO flows (workflow_id, name, source_peer, destination_peer, description,
-		source_table_identifier, destination_table_identifier) VALUES ($1, $2, $3, $4, 'gRPC', '', '')`,
-		workflowID, req.ConnectionConfigs.FlowJobName, sourcePeerID, destinationPeerID,
+		`INSERT INTO flows (workflow_id, name, source_peer, destination_peer, config_proto, status,
+		description, source_table_identifier, destination_table_identifier) VALUES ($1, $2, $3, $4, 'gRPC', '', '', $5)`,
+		workflowID, req.ConnectionConfigs.FlowJobName, sourcePeerID, destinationPeerID, cfgBytes, protos.FlowStatus_STATUS_SETUP,
 	); err != nil {
 		return fmt.Errorf("unable to insert into flows table for flow %s: %w",
 			req.ConnectionConfigs.FlowJobName, err)
@@ -141,11 +146,6 @@ func (h *FlowRequestHandler) CreateCDCFlow(
 		return nil, fmt.Errorf("unable to create flow job entry: %w", err)
 	}
 
-	if err := h.updateFlowConfigInCatalog(ctx, cfg); err != nil {
-		slog.Error("unable to update flow config in catalog", slog.Any("error", err))
-		return nil, fmt.Errorf("unable to update flow config in catalog: %w", err)
-	}
-
 	if _, err := h.temporalClient.ExecuteWorkflow(ctx, workflowOptions, peerflow.CDCFlowWorkflow, cfg, nil); err != nil {
 		slog.Error("unable to start PeerFlow workflow", slog.Any("error", err))
 		return nil, fmt.Errorf("unable to start PeerFlow workflow: %w", err)
@@ -154,13 +154,6 @@ func (h *FlowRequestHandler) CreateCDCFlow(
 	return &protos.CreateCDCFlowResponse{
 		WorkflowId: workflowID,
 	}, nil
-}
-
-func (h *FlowRequestHandler) updateFlowConfigInCatalog(
-	ctx context.Context,
-	cfg *protos.FlowConnectionConfigs,
-) error {
-	return internal.UpdateCDCConfigInCatalog(ctx, h.pool, slog.Default(), cfg)
 }
 
 func (h *FlowRequestHandler) CreateQRepFlow(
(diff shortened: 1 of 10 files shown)


flow/e2e/snowflake/qrep_flow_sf_test.go
```diff
@@ -190,7 +190,7 @@ func (s PeerFlowE2ETestSuiteSF) Test_Complete_QRep_Flow_Avro_SF_Upsert_XMIN() {
 	qrepConfig.WatermarkColumn = "xmin"
 	qrepConfig.SetupWatermarkTableOnDestination = true
 
-	env := e2e.RunXminFlowWorkflow(s.t.Context(), tc, qrepConfig)
+	env := e2e.RunQRepFlowWorkflow(s.t, tc, qrepConfig)
```
ilidemi, line 193: Intended? The test says XMIN

flow/e2e/postgres/qrep_flow_pg_test.go
```diff
@@ -450,7 +450,7 @@ func (s PeerFlowE2ETestSuitePG) TestXminPause() {
 	config.InitialCopyOnly = false
 
 	tc := e2e.NewTemporalClient(s.t)
-	env := e2e.RunXminFlowWorkflow(s.t.Context(), tc, config)
+	env := e2e.RunQRepFlowWorkflow(s.t, tc, config)
```
ilidemi, line 453: xmin test

flow/cmd/handler.go
```diff
@@ -193,10 +196,6 @@ func (h *FlowRequestHandler) CreateQRepFlow(
 		workflowFn = peerflow.QRepFlowWorkflow
 	}
 
-	if req.QrepConfig.SyncedAtColName == "" {
-		cfg.SyncedAtColName = "_PEERDB_SYNCED_AT"
```
ilidemi, line 197: So we don't want this default anymore?

flow/cmd/handler.go
```diff
@@ -156,31 +164,26 @@ func (h *FlowRequestHandler) CreateCDCFlow(
...
+	}
+	cfg.Version = internalVersion
 
 	workflowID := fmt.Sprintf("%s-qrepflow-%s", cfg.FlowJobName, uuid.New())
 	workflowOptions := client.StartWorkflowOptions{
 		ID:                    workflowID,
 		TaskQueue:             h.peerflowTaskQueueID,
 		TypedSearchAttributes: shared.NewSearchAttributes(cfg.FlowJobName),
 	}
-	if req.CreateCatalogEntry {
```
ilidemi, line 178: This one seems to be always set to true at call sites anyway. Should the proto field be removed or we might still want it?

ilidemi (approved):

flow/e2e/snowflake/qrep_flow_sf_test.go
serprex: Yes. API endpoint figures out which workflow to run based on watermark column being xmin

flow/cmd/handler.go
serprex: We don't, as this prevents having no synced at column via api. Instead ui will be explicit

no impact on clickpipes, as it's only cdc

flow/cmd/handler.go
serprex: Trouble is backwards compatibility. Probably safe to remove since it's only request

flow/workflows/cdc_flow.go
```diff
@@ -435,11 +435,10 @@ func CDCFlowWorkflow(
 	}); err != nil {
 		return state, fmt.Errorf("failed to set `%s` query handler: %w", shared.CDCFlowStateQuery, err)
 	}
-	if err := workflow.SetQueryHandler(ctx, shared.FlowStatusQuery, func() (protos.FlowStatus, error) {
+	_ = workflow.SetQueryHandler(ctx, "q-flow-status", func() (protos.FlowStatus, error) {
+		// no longer used, handler kept to avoid nondeterminism
```
Amogh-Bharadwaj, line 438: Is there a way we can eventually remove this from the code?

Amogh-Bharadwaj (approved):
Will need some functional testing

flow/workflows/cdc_flow.go
iamKunalGupta: I'm not sure we need it, I think query handlers are independent of the internal statemachine, but we need to test

iamKunalGupta: Since we still use `state.CurrentFlowStatus`, I think it is worth keeping it around and returning workflowStatus as a new field in GetStatus API calls (at flow API). We can then detect any potential drift if needed.

iamKunalGupta (commented):
Do we fail if we are not able to update status in catalog during pause/resume operations?

flow/workflows/cdc_flow.go
serprex: state should only be updated with method that first updates catalog, update to catalog should be retried until it succeeds, shouldn't ignore state failing to update

flow/workflows/cdc_flow.go
```diff
@@ -70,10 +70,14 @@ func syncStatusToCatalog(ctx workflow.Context, logger log.Logger, status protos.
 		StartToCloseTimeout: 1 * time.Minute,
 	})
 
-	updateFuture := workflow.ExecuteLocalActivity(updateCtx, updateFlowStatusInCatalogActivity,
-		workflow.GetInfo(ctx).WorkflowExecution.ID, status)
-	if err := updateFuture.Get(updateCtx, nil); err != nil {
-		logger.Warn("Failed to update flow status in catalog", slog.Any("error", err), slog.String("flowStatus", status.String()))
+	for {
```
ilidemi: An exponential backoff would be good. If RDS is down for a minute every instance will hammer it full speed. Curious, why not a normal activity? Status updates should be infrequent but durable-durable

serprex: temporal default is that activities retry with 1 second interval & 2x exponential backoff, so this error path is only hit when temporal itself is giving up (probably because of panics)

I don't expect this code to be hit, but have added 1 second sleep

flow/workflows/cdc_flow.go
serprex: the state is accessible from the other query we're still keeping around

PR #3271, opened 2025-07-25
uuid prefix: job must come after uuid

serprex:


heavycrystal (approved):

serprex:
> why?

the point is to diversify partitioning prefix, which flow name does not do

thus:
```
failed to push records: failed to write records to S3: failed to upload file: operation error S3: PutObject, exceeded maximum number of attempts, 3, https response error StatusCode: 503, RequestID: ..., api error SlowDown: A timeout exceeded while waiting to proceed with the request, please reduce your request rate
```

iamKunalGupta:
iirc the full prefix is used to compute partitioning on the AWS, is it different for MinIO?

PR #3281, opened 2025-07-29
Another CDC graph endpoint

Amogh-Bharadwaj:
Currently we have a CDC graph where the time period selection denotes how the data is grouped.
However there is another convention where you'd have the selector options be "Last hour" or "Last day" etc where the tick size would be 1 minute, 1 hour etc respectively.

This PR introduces a new endpoint in flow api which gives these graph data points in the above format, and a field for sum of rows for the given interval of time

### DIFF
--- a/flow/cmd/mirror_status.go
+++ b/flow/cmd/mirror_status.go
@@ -202,60 +202,44 @@ func (h *FlowRequestHandler) cdcFlowStatus(
 	}, nil
 }
 
-type graphParams struct {
-	TickInterval  string
-	NumberOfTicks int
-	TruncUnit     string
-}
-
-func getGraphParams(aggType protos.TimeAggregateType, mode protos.GraphMode) graphParams {
+// returns truncation, tick size, & number of ticks
+func getGraphParams(aggType protos.TimeAggregateType, mode protos.GraphMode) (string, string, int) {
 	switch aggType {
 	case protos.TimeAggregateType_TIME_AGGREGATE_TYPE_FIVE_MIN:
 		if mode == protos.GraphMode_GRAPH_MODE_LAST_X {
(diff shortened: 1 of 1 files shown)


flow/cmd/mirror_status.go
```diff
@@ -234,6 +234,59 @@ func (h *FlowRequestHandler) CDCGraph(ctx context.Context, req *protos.GraphRequ
 	return &protos.GraphResponse{Data: data}, nil
 }
 
+func (h *FlowRequestHandler) CDCGraphForLastXTime(ctx context.Context, req *protos.GraphRequest) (*protos.GraphResponse, error) {
```
Amogh-Bharadwaj: need a better name

serprex: This should be combined with `CDCGraph` where CDCGraph is default behavior while this behavior is based on bool

flow/cmd/mirror_status.go
```diff
@@ -203,6 +203,14 @@ func (h *FlowRequestHandler) cdcFlowStatus(
 }
 
 func (h *FlowRequestHandler) CDCGraph(ctx context.Context, req *protos.GraphRequest) (*protos.GraphResponse, error) {
+	if req.Mode == protos.GraphMode_GRAPH_MODE_LAST_X {
+		return h.CDCGraphForLastXTime(ctx, req)
+	}
+
+	return h.CDCGraphGroupByXTime(ctx, req)
```
Amogh-Bharadwaj: Done like this for backward compatibility

iamKunalGupta:
Are we doing this for ClickPipes or UI in general?

serprex (approved):

PR #3283, opened 2025-07-29
pg: detect when partitioning fails due to timescale compressed tables

serprex:
& fallback to not using ctid

### DIFF
--- a/flow/connectors/postgres/qrep.go
+++ b/flow/connectors/postgres/qrep.go
@@ -3,6 +3,7 @@ package connpostgres
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"strconv"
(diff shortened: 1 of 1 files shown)


flow/connectors/postgres/qrep.go
```diff
@@ -186,6 +188,19 @@ func (c *PostgresConnector) getNumRowsPartitions(
 	}
 
 	if err := rows.Err(); err != nil {
+		var pgErr *pgconn.PgError
+		if config.WatermarkColumn == "ctid" && errors.As(err, &pgErr) &&
+			pgErr.Code == "XX000" && pgErr.Message == "transparent decompression only supports tableoid system column" {
```
heavycrystal: it looks like it can also be "42P10"
also can use `pgerrcode.InternalError` instead

serprex: since we have to do a message check anyways might as well drop code check in that case

heavycrystal: in this case that works

flow/connectors/postgres/qrep.go
```diff
@@ -186,6 +188,19 @@ func (c *PostgresConnector) getNumRowsPartitions(
 	}
 
 	if err := rows.Err(); err != nil {
+		var pgErr *pgconn.PgError
+		if config.WatermarkColumn == "ctid" && errors.As(err, &pgErr) &&
+			pgErr.Message == "transparent decompression only supports tableoid system column" {
+			// timescale, fallback to full table partition
```
heavycrystal, line 194: please add a log here to indicate switch

PR #3285, opened 2025-07-29
Add e2e test bucket to dev config

ilidemi:


serprex:
curious if this is what was giving me 404s in #3225

PR #3288, opened 2025-07-30
ClickHouse: Sanitize initial load parsing error when inserting

Amogh-Bharadwaj:
For ClickHouse's incorrect data error type, we do not want to log the full error message

### DIFF
--- a/flow/connectors/clickhouse/qrep_avro_sync.go
+++ b/flow/connectors/clickhouse/qrep_avro_sync.go
@@ -2,12 +2,14 @@ package connclickhouse
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


flow/connectors/clickhouse/qrep_avro_sync.go
```diff
@@ -356,6 +358,12 @@ func (s *ClickHouseAvroSyncMethod) pushS3DataToClickHouse(
 					slog.Uint64("numParts", numParts),
 					slog.Int("chunkIdx", chunkIdx),
 					slog.Any("error", err))
+
+				var chException *clickhouse.Exception
+				if errors.As(err, &chException) {
```
serprex, line 363: this is too late, we already logged error at start of block

Amogh-Bharadwaj: Ah right

PR #3289, opened 2025-07-30
ch: redact error 117 INCORRECT_DATA

serprex:


iskakaushik (approved):

flow/shared/clickhouse/query_retry.go
```diff
@@ -58,6 +59,13 @@ func Exec(ctx context.Context, logger log.Logger,
 			time.Sleep(time.Second * time.Duration(i*5+1))
 		}
 	}
+	if ex, ok := err.(*clickhouse.Exception); ok {
+		if chproto.Error(ex.Code) == chproto.ErrIncorrectData {
+			// error message often includes json blobs, avoid logging these
+			ex.Message = "REDACTED"
```
Amogh-Bharadwaj, line 65: Do we want to put a generic error message instead?

iskakaushik: i think REDACTED is good!

PR #3291, opened 2025-07-30
refactor(fetched-bytes): metric only looks at mapped tables

iamKunalGupta:


flow/connectors/postgres/cdc.go
```diff
@@ -588,6 +588,7 @@ func PullCdcRecords[Items model.Items](
 				}
 
 				if rec != nil {
+					p.otelManager.Metrics.FetchedBytesCounter.Add(ctx, int64(len(msg.Data)))
```
iskakaushik, line 591: I don't think its filtered to table yet?

iamKunalGupta: rec should be nil only for non-mapped tables I think
`processMessage` calls `processInsertMessage`/`processUpdateMessage`/`processDeleteMessage` which checks and return nil if relId is not found in `srcTableIDNameMapping`

iskakaushik: got it!

iskakaushik (approved):

PR #3292, opened 2025-07-30
pg: add e2e test for json types

jgao54:
Add e2e test for JSON types.

The following types are currently not supported by CH JSON:
- number
- string
- boolean
- array

Null is supported, but we parse it to empty string with `JSONExtractString(_peerdb_data, <col>)`, so it's failing to insert in the same way that string type does. `JSONExtractRaw(_peerdb_data, <col>)` works locally. Will look into a fix for the 'null' case as a follow-up.

### DIFF
--- a/flow/e2e/clickhouse/peer_flow_ch_test.go
+++ b/flow/e2e/clickhouse/peer_flow_ch_test.go
@@ -1207,6 +1207,116 @@ func (s ClickHouseSuite) Test_Types_CH() {
 	e2e.RequireEnvCanceled(s.t, env)
 }
 
+func (s ClickHouseSuite) Test_JSON_CH() {
+	// TODO: fix and enable failing test cases
+	jsonTestCases := []struct {
+		Value      string
+		Desc       string
+		SkipReason string
+		Skip       bool
+	}{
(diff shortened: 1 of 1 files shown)


serprex (approved):

flow/e2e/clickhouse/peer_flow_ch_test.go
```diff
@@ -1207,6 +1207,116 @@ func (s ClickHouseSuite) Test_Types_CH() {
...
+			Value:      `'{"key1": null, "key2": {"nested_key": null}}'`,
+			Skip:       true,
+			SkipReason: "insertion succeeds; but comparison fails because CH JSON removes all null fields but postgres does not",
+		},
+	}
+
+	srcTableName := "test_json"
+	dstTableName := "test_json_dst"
+
+	_, err := s.Conn().Exec(s.t.Context(), fmt.Sprintf(`
```
serprex, line 1283: `s.Conn()` skips test for mysql. Would be good to have this work with mysql too. Would need some if/else on source type since MySQL lacks jsonb (& I think on Maria we don't have a real JSON type)

PR #3293, opened 2025-07-30
[mongo] parallel snapshot for ObjectID key

heavycrystal:
Need to give `WatermarkColumn` as input, otherwise falls back to full table partition.
Only `ObjectID` partitioning is supported, so it is intended for use with `_id` but should be general.

Also added "adjusted partitions" logic to MySQL

TODO testing

### DIFF
--- a/flow/connectors/mongo/qrep.go
+++ b/flow/connectors/mongo/qrep.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"time"
 
+	"github.com/google/uuid"
 	"go.mongodb.org/mongo-driver/v2/bson"
 	"go.mongodb.org/mongo-driver/v2/mongo/options"
 
@@ -18,7 +19,6 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared"
 	shared_mongo "github.com/PeerDB-io/peerdb/flow/shared/mongo"
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
-	"github.com/google/uuid"
 )
 
 func (c *MongoConnector) GetQRepPartitions(
@@ -160,7 +160,6 @@ func (c *MongoConnector) PullQRepRecords(
 	defer shutDown()
 
 	filter := bson.D{}
-	// FullTablePartition is always true until parallel initial load is implemented, see `GetQRepPartitions`
 	if !partition.FullTablePartition {
 		filter, err = toRangeFilter(config.WatermarkColumn, partition.Range)
 		if err != nil {
--- a/flow/connectors/mongo/qrep.go
+++ b/flow/connectors/mongo/qrep.go
@@ -18,6 +18,7 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared"
 	shared_mongo "github.com/PeerDB-io/peerdb/flow/shared/mongo"
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
+	"github.com/google/uuid"
 )
 
 func (c *MongoConnector) GetQRepPartitions(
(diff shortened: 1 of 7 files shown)


flow/connectors/mysql/qrep.go
```diff
@@ -115,30 +115,29 @@ func (c *MySqlConnector) GetQRepPartitions(
 	}
 
 	// Calculate the number of partitions
-	numPartitions := totalRows / numRowsPerPartition
-	if totalRows%numRowsPerPartition != 0 {
-		numPartitions++
-	}
+	adjustedPartitions := shared.AdjustNumPartitions(totalRows, numRowsPerPartition)
+	c.logger.Info("[mysql] partition details",
```
jgao54, line 119: nit: `[mongo]`

flow/connectors/mongo/qrep.go
```diff
@@ -25,14 +25,99 @@ func (c *MongoConnector) GetQRepPartitions(
...
+	// no need to bother with bucketAuto if we have only one partition
+	if adjustedPartitions.AdjustedNumPartitions == 1 {
+		return fullTablePartition, nil
+	}
+
+	// Use bucketAuto to create partitions based on _id field
+	bucketAutoPipeline := []bson.D{
+		{
+			{Key: "$bucketAuto", Value: bson.D{
+				{Key: "groupBy", Value: "$" + config.WatermarkColumn},
```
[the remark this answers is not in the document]
jgao54: thanks copilot. 

can config.WatermarkColumn be passed in from user? if so we should sanitize the column.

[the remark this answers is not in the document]
heavycrystal: For now, no. My idea is a toggle that enables `_id` since that's the only column we can reliably expect to be an ObjectID which is the only thing we can partition by.

will take a look at sanitization though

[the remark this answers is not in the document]
jgao54: in that case a simple `if config.WatermarkColumn != DefaultDocumentKeyColumnName { error }` is sufficient for now

flow/connectors/mysql/qrep.go
```diff
@@ -115,30 +115,29 @@ func (c *MySqlConnector) GetQRepPartitions(
 	}
 
 	// Calculate the number of partitions
-	numPartitions := totalRows / numRowsPerPartition
-	if totalRows%numRowsPerPartition != 0 {
-		numPartitions++
-	}
+	adjustedPartitions := shared.AdjustNumPartitions(totalRows, numRowsPerPartition)
```
jgao54: ooc, is the main difference here limiting `maxPartitions` to 1000?

flow/connectors/utils/partition.go
```diff
@@ -356,8 +341,6 @@ func (p *PartitionHelper) getPartitionForStartAndEnd(start any, end any) (*proto
 		return createTimePartition(v, end.(time.Time)), nil
 	case pgtype.TID:
 		return createTIDPartition(v, end.(pgtype.TID)), nil
-	case bson.ObjectID:
```
jgao54, line 359: do we not want to use partition helper for Mongo for consistency across sources? also thinking if we ever want to partition on other fields other than _id (i.e. with flattened mode in the future),  but not a big deal, we can add later too.

flow/connectors/mysql/qrep.go
heavycrystal: yes, and adjusting numRowsPerPartition accordingly

flow/connectors/utils/partition.go
heavycrystal: `partitionHelper` can be used; it's just a bit cludgy to use it here (`bson.ObjectID` is not the type stored in protobufs and the helper tends to use types to figure out the type of range, which is `string`)

It doesn't give us any advantage here since we don't coalesce partitions together based on ObjectID range (could be done down the line)

PR #3294, opened 2025-07-30
num partitions override

serprex:
certain workloads (both pg & mysql) can have count(*) take over 24 hours to complete,
while getting min/max is still able to be looked up quickly

by providing the number of partitions, instead of the size, we can divide the space without knowing count(*)

### DIFF
--- a/flow/connectors/mysql/qrep.go
+++ b/flow/connectors/mysql/qrep.go
@@ -50,7 +50,7 @@ func (c *MySqlConnector) GetQRepPartitions(
 	config *protos.QRepConfig,
 	last *protos.QRepPartition,
 ) ([]*protos.QRepPartition, error) {
-	if config.WatermarkColumn == "" {
+	if config.WatermarkColumn == "" || config.NumPartitionsOverride == 1 {
 		// if no watermark column is specified, return a single partition
 		return []*protos.QRepPartition{
 			{
--- a/flow/connectors/postgres/qrep.go
+++ b/flow/connectors/postgres/qrep.go
@@ -41,7 +41,7 @@ func (c *PostgresConnector) GetQRepPartitions(
 	config *protos.QRepConfig,
 	last *protos.QRepPartition,
 ) ([]*protos.QRepPartition, error) {
-	if config.WatermarkColumn == "" {
+	if config.WatermarkColumn == "" || config.NumPartitionsOverride == 1 {
 		// if no watermark column is specified, return a single partition
 		return []*protos.QRepPartition{
 			{


flow/connectors/mysql/qrep.go
```diff
@@ -104,37 +109,39 @@ func (c *MySqlConnector) GetQRepPartitions(
...
+			c.logger.Warn("no records to replicate, returning")
+			return make([]*protos.QRepPartition, 0), nil
+		}
 
-	// Calculate the number of partitions
-	numPartitions := totalRows / numRowsPerPartition
-	if totalRows%numRowsPerPartition != 0 {
-		numPartitions++
+		// Calculate the number of partitions
+		numPartitions = totalRows / numRowsPerPartition
```
Amogh-Bharadwaj, line 124: Nit: do we want to protect against division by 0 panic? More likely with API users

flow/connectors/mysql/qrep.go
```diff
@@ -104,37 +109,39 @@ func (c *MySqlConnector) GetQRepPartitions(
...
 
-	watermarkMyType := rs.Fields[1].Type
-	watermarkQKind, err := qkindFromMysql(rs.Fields[1])
+	watermarkMyType := rs.Fields[0].Type
+	watermarkQKind, err := qkindFromMysql(rs.Fields[0])
 	if err != nil {
 		return nil, fmt.Errorf("failed to convert mysql type to qvaluekind: %w", err)
 	}
 
-	c.logger.Info("queried info for partitioning",
```
Amogh-Bharadwaj, line 129: I personally think this log is useful, or is this already captured in another log somewhere?

flow/connectors/postgres/qrep.go
```diff
@@ -98,102 +99,133 @@ func (c *PostgresConnector) getNumRowsPartitions(
...
+		var row pgx.Row
+		var minVal any
+		if last != nil && last.Range != nil {
+			switch lastRange := last.Range.Range.(type) {
+			case *protos.PartitionRange_IntRange:
+				minVal = lastRange.IntRange.End
+			case *protos.PartitionRange_UintRange:
+				minVal = lastRange.UintRange.End
+			case *protos.PartitionRange_TimestampRange:
+				minVal = lastRange.TimestampRange.End.AsTime()
```
Amogh-Bharadwaj, line 211: Do we want to have a default case which errors out ? If tomorrow we add a new partition type

Amogh-Bharadwaj (approved):

PR #3301, opened 2025-07-31
replace local activities with plain activities

serprex:
this small optimization increases complexity around reexecution

moving towards keeping our use of temporal api surface minimal

if it's found that a workflow is spending too much time dealing with overhead of activities,
they should be combined into one long running activity


flow/workflows/cdc_flow.go
```diff
@@ -66,11 +66,11 @@ func NewCDCFlowWorkflowState(ctx workflow.Context, logger log.Logger, cfg *proto
 }
 
 func syncStatusToCatalog(ctx workflow.Context, logger log.Logger, status protos.FlowStatus) {
-	updateCtx := workflow.WithLocalActivityOptions(ctx, workflow.LocalActivityOptions{
+	updateCtx := workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
 		StartToCloseTimeout: 1 * time.Minute,
```
ilidemi:
```suggestion
		StartToCloseTimeout: 4 * 24 * time.Hour,
```

ilidemi:
TestDropCompleted and possible TestEditTablesBeforeResync failures are race conditions between querying the temporal status and the catalog activity now taking a lil longer to execute

PR #3303, opened 2025-08-01
mongo: set allow_nullable_key = 1 only when custom sort keys are used

jgao54:
Kept the `allow_nullable_key = 1` logic relatively simple:
- when using custom ordering ->  allow nullable key
- when using default primary keys from source dbs -> do not allow nullable key

In theory, we could tighten `allow_nullable_key = 1`  to only when custom ordering keys contains columns that are nullable and have nullable enabled, but given the current nullable and enableNullable configuration is relatively nuanced (i.e. schema level vs. column level), doing so would make it harder to reason about when allow_nullable_key are configured and also more maintenance overhead. So I kept the logic here relatively simple -- for now it's fine to allow_nullable_key = 1 even if the destination PKs are non-nullable. And this won't apply to Mongo since we don't allow custom sort keys there.

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -220,7 +220,7 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
 	fmt.Fprintf(&stmtBuilder, " ENGINE = %s", engine)
 
 	if tmEngine != protos.TableEngine_CH_ENGINE_NULL {
-		orderByColumns := getOrderedOrderByColumns(tableMapping, tableSchema.PrimaryKeyColumns, colNameMap)
+		orderByColumns, allowNullablePK := getOrderedOrderByColumns(tableMapping, tableSchema.PrimaryKeyColumns, colNameMap)
 		if sourceSchemaAsDestinationColumn {
 			orderByColumns = append([]string{sourceSchemaColName}, orderByColumns...)
 		}
@@ -238,7 +238,9 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
 			fmt.Fprintf(&stmtBuilder, " PARTITION BY (%s)", partitionByStr)
 		}
 
-		stmtBuilder.WriteString(" SETTINGS allow_nullable_key = 1")
+		if allowNullablePK {
+			stmtBuilder.WriteString(" SETTINGS allow_nullable_key = 1")
+		}
 
 		if c.config.Cluster != "" {
 			fmt.Fprintf(&stmtBuilderDistributed, " ENGINE = Distributed(%s,%s,%s",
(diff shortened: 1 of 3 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -238,7 +238,9 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
 			fmt.Fprintf(&stmtBuilder, " PARTITION BY (%s)", partitionByStr)
 		}
 
-		stmtBuilder.WriteString(" SETTINGS allow_nullable_key = 1")
+		if allowNullablePK {
```
serprex: we used to also skip this when nullables were not enabled, what's the impact of having this setting when it isn't used on nullable columns?

jgao54: The setting only affects validation on table schema creation.

For example, these are all valid:
```

CREATE TABLE users (id Int64 primary key) ENGINE=MergeTree() SETTINGS allow_nullable_key = 1;

CREATE TABLE users (id Nullable(Int64) primary key) ENGINE=MergeTree() SETTINGS allow_nullable_key = 1;

CREATE TABLE users (id Int64 primary key) ENGINE=MergeTree() SETTINGS allow_nullable_key = 0;
```

But this will fail:
```
CREATE TABLE users (id Nullable(Int64) primary key) ENGINE=MergeTree() SETTINGS allow_nullable_key = 0;
```

with:
> DB::Exception: Sorting key contains nullable columns, but merge tree setting `allow_nullable_key` is disabled. 

Whether a Null values gets inserted as a `Null` or a corresponding typed value (i.e. 0 for ints, "" for string, false for bool, etc.) only depends on whether the _column_ is `Nullable` or not. 

I think of this change as more of a "best practice recommendation" for how to  create table in CH; as official doc `allow_nullable_key` is strongly [discouraged](https://clickhouse.com/docs/engines/table-engines/mergetree-family/mergetree#primary-keys-and-indexes-in-queries) and has been a source of issues historically.

serprex: It seems odd when this will only hit user if bug doesn't set setting while key has nullable. We should be warning users not to use nullable keys instead of toggling things in background with zero effect & zero user visibility (outside describe table)

jgao54: you raise a good point about user visibility here, and potential bug where we hit an error because `allow_nullable_key` is not set but column is Nullable, since we don't handle this during validation.

I like the idea of warning users, what do you think about this as next steps:
1. during validation, we check if sort key is nullable, if so, we raise a warning and proceed to create table with `allow_nullable_key = 1` if user persists
2. update the logic for CH table creation to set this setting only if key is _actually_ nullable.

jgao54: Updated PR with 2. 

Currently, when ordering key is selected in the UI, we get a warning:
```
Using ordering keys in ClickHouse that differ from the primary key in Postgres has some caveats. Please read [this doc](https://docs.peerdb.io/mirror/ordering-key-different) carefully.
```
Planning to add allow_nullable_key = 1 as an additional caveat in the doc for peerdb for consistency rather than when validation is clicked.

flow/connectors/clickhouse/normalize.go
```diff
@@ -265,46 +267,82 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
...
+	}
+
+	hasNullableKeys := false
+
+	// Case 1: No custom ordering - use primary keys
+	if len(columnOrderingMap) == 0 {
+		pkeys := make([]string, len(sourcePkeys))
+		for idx, pk := range sourcePkeys {
+			pkeys[idx] = peerdb_clickhouse.QuoteIdentifier(pk)
+			if !hasNullableKeys && isNullableAndNullableEnabled(pk) {
```
jgao54: isNullableAndNullableEnabled should always return false for mysql/pg/mongo as source, but checking this for completion in case we add sources that contain nullable primary columns (i.e. CH -> CH)

flow/connectors/clickhouse/normalize.go
```diff
@@ -232,13 +235,13 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
 			stmtBuilder.WriteString(" ORDER BY tuple()")
 		}
 
-		partitionByColumns := getOrderedPartitionByColumns(tableMapping, colNameMap)
+		partitionByColumns, hsaNullablePartitionKey := getOrderedPartitionByColumns(tableMapping, colNameMap, hasNullableKeyFn)
```
serprex: hsa

PR #3304, opened 2025-08-01
fix(record-slot-sizes): add timeout and record slot information in parallel

iamKunalGupta:


iamKunalGupta:
> should probably cap the parallelism, I recall there being some concerns around too many connections to single DB re making this parallel

The issue is that we need to make sure that the next run starts on time for the rest of the metrics. It might be a good idea to separate the tickers for slot related metrics and other metrics down the line (or now?)

serprex:
We'd need a separate goroutine, since right now this logic is single threaded

iamKunalGupta:
> We'd need a separate goroutine, since right now this logic is single threaded

let me take a stab at it

flow/activities/flowable.go
```diff
@@ -729,25 +730,39 @@ func (a *FlowableActivity) SendWALHeartbeat(ctx context.Context) error {
...
+	go runEveryDurationInfinitely(ctx, "SendWALHeartbeat", 10*time.Minute, a.SendWALHeartbeat)
+	go runEveryDurationInfinitely(ctx, "RecordMetrics", 1*time.Minute, a.RecordMetrics)
+	go runEveryDurationInfinitely(ctx, "RecordSlotSizes", 1*time.Minute, a.RecordSlotSizes)
+	<-ctx.Done()
+	logger := internal.LoggerFromCtx(ctx)
+	logger.Info("Stopping scheduled tasks due to context done", slog.Any("error", ctx.Err()))
+	return nil
+}
+
+func runEveryDurationInfinitely(ctx context.Context, funcName string, duration time.Duration, fn func(context.Context) error) {
```
serprex: use `shared.Interval`

iamKunalGupta: done

serprex (approved):

PR #3310, opened 2025-08-04
remove tables: use selector

serprex:
could be a little more DRY

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -341,6 +341,40 @@ func processTableRemovals(
 	cfg *protos.FlowConnectionConfigs,
 	state *CDCFlowWorkflowState,
 ) error {
+	removeTablesSelector := workflow.NewNamedSelector(ctx, "AddTables")
+	removeTablesSelector.AddReceive(ctx.Done(), func(_ workflow.ReceiveChannel, _ bool) {})
+	flowSignalStateChangeChan := model.FlowSignalStateChange.GetSignalChannel(ctx)
+	flowSignalStateChangeChan.AddToSelector(removeTablesSelector, func(val *protos.FlowStateChangeRequest, _ bool) {
+		if val.RequestedFlowState == protos.FlowStatus_STATUS_TERMINATING {
+			logger.Info("terminating CDCFlow during table removal")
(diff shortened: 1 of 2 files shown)


flow/workflows/cdc_flow.go
```diff
@@ -236,25 +222,50 @@ func processTableAdditions(
...
+
+func processTableAdditions(
+	ctx workflow.Context,
+	logger log.Logger,
+	cfg *protos.FlowConnectionConfigs,
+	state *CDCFlowWorkflowState,
+	mirrorNameSearch temporal.SearchAttributes,
+) error {
+	flowConfigUpdate := state.FlowConfigUpdate
+	if len(flowConfigUpdate.AdditionalTables) == 0 {
```
heavycrystal, line 254: nit: not related to PR, but we already check this before calling the function so could remove one or the other

flow/workflows/cdc_flow.go
```diff
@@ -236,25 +222,50 @@ func processTableAdditions(
...
 			// since we are adding to TableMappings, multiple signals can lead to duplicates
 			// we should ContinueAsNew after the first signal in the selector, but just in case
 			cfg.Resync = true
 			cfg.DoInitialSnapshot = true
 			state.DropFlowInput = &protos.DropFlowInput{
 				// to be filled in just before ContinueAsNew
-				FlowJobName:           "",
-				FlowConnectionConfigs: nil,
+				FlowJobName:           cfg.FlowJobName,
+				FlowConnectionConfigs: cfg,
```
heavycrystal, line 235: If this is safe to do, why was the logic otherwise to begin with?

serprex: still ends up being replaced afterwards, so these values don't matter, but flow name is same either way. tried lifting this here hoping to have more code paths use this function, but they all have their own quirks

flow/workflows/cdc_flow.go
serprex: could, but code calling this avoids an extra sync when neither addition or removal happens, but still don't like trusting caller to have to check for 0 if we remove this check. Just one of those things that happens when you split out functions that are only called once

PR #3318, opened 2025-08-06
fallback to full table partitions for Timescale compressed hypertables

heavycrystal:
The PR adds a new activity, `GetDefaultPartitionKeyForTables,` during Snapshot, which is supposed to return the default "partition" key for each table if not overridden by the user. For Postgres, we check for PG14+ (when TID range scan was added), and if that is there, we also check if the Timescale extension is installed and for compressed hypertables. If any are found, disable the partition key for those.

For Mongo/MySQL it currently returns an empty key for every table. But can be changed.

### DIFF
--- a/flow/connectors/mongo/qrep.go
+++ b/flow/connectors/mongo/qrep.go
@@ -137,10 +137,10 @@ func (c *MongoConnector) GetQRepPartitions(
 
 func (c *MongoConnector) GetParallelLoadKeyForTables(
 	ctx context.Context,
-	input *protos.GetParallelLoadKeyForTablesInput,
-) (*protos.GetParallelLoadKeyForTablesOutput, error) {
-	return &protos.GetParallelLoadKeyForTablesOutput{
-		TableParallelLoadKeyMapping: make(map[string]string),
+	input *protos.GetDefaultPartitionKeyForTablesInput,
+) (*protos.GetDefaultPartitionKeyForTablesOutput, error) {
+	return &protos.GetDefaultPartitionKeyForTablesOutput{
+		TableDefaultPartitionKeyMapping: make(map[string]string),
 	}, nil
 }
 
--- a/flow/connectors/mongo/qrep.go
+++ b/flow/connectors/mongo/qrep.go
@@ -140,7 +140,7 @@ func (c *MongoConnector) GetDefaultPartitionKeyForTables(
 	input *protos.GetDefaultPartitionKeyForTablesInput,
 ) (*protos.GetDefaultPartitionKeyForTablesOutput, error) {
 	return &protos.GetDefaultPartitionKeyForTablesOutput{
-		TableDefaultPartitionKeyMapping: make(map[string]string),
+		TableDefaultPartitionKeyMapping: nil,
 	}, nil
 }
 
--- a/flow/connectors/mongo/qrep.go
+++ b/flow/connectors/mongo/qrep.go
@@ -135,6 +135,15 @@ func (c *MongoConnector) GetQRepPartitions(
 	return partitions, nil
 }
 
+func (c *MongoConnector) GetParallelLoadKeyForTables(
+	ctx context.Context,
+	input *protos.GetParallelLoadKeyForTablesInput,
+) (*protos.GetParallelLoadKeyForTablesOutput, error) {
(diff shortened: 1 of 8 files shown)


flow/connectors/mongo/qrep.go
```diff
@@ -135,6 +135,15 @@ func (c *MongoConnector) GetQRepPartitions(
 	return partitions, nil
 }
 
+func (c *MongoConnector) GetDefaultPartitionKeyForTables(
+	ctx context.Context,
+	input *protos.GetDefaultPartitionKeyForTablesInput,
+) (*protos.GetDefaultPartitionKeyForTablesOutput, error) {
+	return &protos.GetDefaultPartitionKeyForTablesOutput{
+		TableDefaultPartitionKeyMapping: make(map[string]string),
```
serprex: this can be nil. protobuf will encode empty map to nil anyways, & `nil[x]` doesn't error, only an issue if you try adding elements to nil

flow/connectors/postgres/qrep.go
```diff
@@ -69,6 +69,78 @@ func (c *PostgresConnector) GetQRepPartitions(
...
+		var schema, name string
+		if err := r.Scan(&schema, &name); err != nil {
+			return "", err
+		}
+		return strings.ToLower(fmt.Sprintf("%s.%s", schema, name)), nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get compressed hypertables: %w", err)
+	}
+	compressedSet := make(map[string]struct{}, len(compressedSlice))
```
serprex: can skip slice, keep pushing to set with pgx.ForEachRow

flow/connectors/postgres/qrep.go
```diff
@@ -69,6 +69,78 @@ func (c *PostgresConnector) GetQRepPartitions(
...
+	if !hasTimescale {
+		return output, nil
+	}
+
+	// compressed hypertables cannot do ctid scans, so disable for them
+	// NOTE: it appears that the hypercore "TAM" may give us access to ctid scans, but that's to be removed in Timescale 2.22
+	rows, err := c.conn.Query(ctx, `SELECT hypertable_schema, hypertable_name
+		FROM timescaledb_information.chunks
+		WHERE is_compressed='t'
+		GROUP BY hypertable_schema, hypertable_name;`)
```
serprex:
```suggestion
	rows, err := c.conn.Query(ctx, `SELECT DISTINCT hypertable_schema, hypertable_name
		FROM timescaledb_information.chunks
		WHERE is_compressed='t'`)
```

flow/connectors/postgres/qrep.go
```diff
@@ -69,6 +69,78 @@ func (c *PostgresConnector) GetQRepPartitions(
...
+		return nil, fmt.Errorf("failed to get compressed hypertables: %w", err)
+	}
+	compressedSet := make(map[string]struct{}, len(compressedSlice))
+	for _, id := range compressedSlice {
+		compressedSet[id] = struct{}{}
+	}
+
+	for _, tm := range input.TableMappings {
+		if _, found := compressedSet[strings.ToLower(tm.SourceTableIdentifier)]; found {
+			delete(output.TableDefaultPartitionKeyMapping, tm.SourceTableIdentifier)
```
serprex: can't skip creating set & instead try remove each compressed table from TableDefaultPartitionKeyMapping?

flow/connectors/postgres/qrep.go
```diff
@@ -69,6 +69,78 @@ func (c *PostgresConnector) GetQRepPartitions(
...
+
+	output := &protos.GetDefaultPartitionKeyForTablesOutput{
+		TableDefaultPartitionKeyMapping: make(map[string]string, len(input.TableMappings)),
+	}
+
+	pgVersion, err := shared.GetMajorVersion(ctx, c.conn)
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine server version: %w", err)
+	}
+	supportsTidScans := pgVersion >= shared.POSTGRES_14
```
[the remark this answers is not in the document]
heavycrystal: 14 is correct

flow/connectors/postgres/qrep.go
heavycrystal: yea should work

PR #3319, opened 2025-08-07
mongo: handle inf/-inf/NaN in connector

jgao54:
Use `jsonitor` library instead of default encoding/json, add custom extension to handle NaN/Inf/-Inf cases by writing them to `String` type. As pointed out by @ilidemi , JSON CH does not preserve NaN/Infinity/-Infinity:
```
SELECT CAST('{"a": "NaN"}' AS JSON(a Float))
SETTINGS enable_json_type=1; // returns {"a": null}
```

I end up using the default `ConfigCompatibleWithStandardLibrary` because `ConfigFastest` results in precision loss -- we could consider enabling it as an option in the future for perf-sensitive users.

Test:
- add unit tests and e2e tests
- ran flow locally

### DIFF
--- a/flow/connectors/mongo/cdc.go
+++ b/flow/connectors/mongo/cdc.go
@@ -26,10 +26,10 @@ type Namespace struct {
 }
 
 type ChangeEvent struct {
-	DocumentKey   map[string]any `bson:"documentKey,omitempty"`
-	FullDocument  map[string]any `bson:"fullDocument,omitempty"`
 	Ns            Namespace      `bson:"ns"`
 	OperationType string         `bson:"operationType"`
+	DocumentKey   bson.D         `bson:"documentKey,omitempty"`
+	FullDocument  bson.D         `bson:"fullDocument,omitempty"`
 	ClusterTime   bson.Timestamp `bson:"clusterTime"`
 }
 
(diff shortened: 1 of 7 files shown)


flow/connectors/mongo/encode_test.go
```diff
@@ -0,0 +1,309 @@
...
+
+		// Nested bson.D
+		{
+			desc: "nested bson.D",
+			input: bson.D{{Key: "nested", Value: bson.D{
+				{Key: "inner1", Value: "str"},
+				{Key: "inner2", Value: 1},
+				{Key: "inner3", Value: true},
+				{
+					Key: "inner3", Value: bson.D{
```
ilidemi: nit
```suggestion
					Key: "inner4", Value: bson.D{
```

ilidemi (approved):

flow/connectors/mongo/cdc.go
```diff
@@ -26,8 +26,8 @@ type Namespace struct {
 }
 
 type ChangeEvent struct {
-	DocumentKey   map[string]any `bson:"documentKey,omitempty"`
-	FullDocument  map[string]any `bson:"fullDocument,omitempty"`
+	DocumentKey   bson.D         `bson:"documentKey,omitempty"`
+	FullDocument  bson.D         `bson:"fullDocument,omitempty"`
```
jgao54: I changed this to bson.D so cdc's document type is consistent with initial load, otherwise I have to implement redundant logic for map[string]any datatype in the BsonExtension implementation. 

Also realized that bson decoding libary can deserialize to either `bson.D` or `bson.M`. 

Not sure if there's a huge difference in performance between the two (given bson.D is ordered), we can change this to bson.M later if it's faster.

flow/connectors/mongo/encode.go
```diff
@@ -0,0 +1,173 @@
...
+	return nil
+}
+
+func (extension *BsonExtension) DecorateEncoder(typ reflect2.Type, encoder jsoniter.ValEncoder) jsoniter.ValEncoder {
+	if typ.String() == "bson.D" {
+		return &BsonDCodec{}
+	}
+	if typ.String() == "bson.A" {
+		return &BsonACodec{}
+	}
```
serprex:
```suggestion
	if typ.Type1() == reflect.TypeFor[bson.D]() {
		return &BsonDCodec{}
	}
	if typ.Type1() == reflect.TypeFor[bson.A]() {
		return &BsonACodec{}
	}
```
underneath `typ.String()` is `typ.Type1().String()`

Amogh-Bharadwaj (approved):

PR #3321, opened 2025-08-07
feat(kafka): add support for custom certificates

JanPetr:
Hello there!

In our company we use custom certificates to authenticate to our Kafka instances. Unfortunately PeerDB doesn't support uploading custom certificates for Kafka peers.

This PR adds the possibility. The PR reuses the same logic as used for Clickhouse peers.

Thank you for your work, review, feedback & eventual merge ❤️🙏

image

### DIFF
--- a/flow/connectors/kafka/kafka.go
+++ b/flow/connectors/kafka/kafka.go
@@ -76,7 +76,7 @@ func NewKafkaConnector(
 		kgo.WithLogger(kgoLogger(logger)),
 	)
 	if !config.DisableTls {
-		tlsSetting := &tls.Config{MinVersion: tls.VersionTLS13}
+		tlsSetting := &tls.Config{MinVersion: tls.VersionTLS12}
 		if config.Certificate != nil || config.PrivateKey != nil {
 			if config.Certificate == nil || config.PrivateKey == nil {
 				return nil, errors.New("both certificate and private key must be provided if using certificate-based authentication")
--- a/flow/connectors/kafka/kafka.go
+++ b/flow/connectors/kafka/kafka.go
@@ -3,6 +3,7 @@ package connkafka
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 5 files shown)


flow/connectors/kafka/kafka.go
```diff
@@ -75,7 +76,26 @@ func NewKafkaConnector(
...
+			if config.Certificate == nil || config.PrivateKey == nil {
+				return nil, errors.New("both certificate and private key must be provided if using certificate-based authentication")
+			}
+			cert, err := tls.X509KeyPair([]byte(*config.Certificate), []byte(*config.PrivateKey))
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse provided certificate: %w", err)
+			}
+			tlsSetting.Certificates = []tls.Certificate{cert}
+		}
+		if config.RootCa != nil {
```
heavycrystal, line 90: consider setting up the root CAs with the `shared.CreateTlsConfig` helper

flow/connectors/kafka/kafka.go
```diff
@@ -75,7 +76,26 @@ func NewKafkaConnector(
 		kgo.WithLogger(kgoLogger(logger)),
 	)
 	if !config.DisableTls {
-		optionalOpts = append(optionalOpts, kgo.DialTLSConfig(&tls.Config{MinVersion: tls.VersionTLS13}))
+		tlsSetting := &tls.Config{MinVersion: tls.VersionTLS13}
```
heavycrystal: is TLS1.3 too bleeding edge? I'm not too familiar with the Kafka ecosystem but iirc Confluent may not always support TLS 1.3

heavycrystal (commented):
Thanks for your contribution! Just a few minor comments

flow/connectors/kafka/kafka.go
JanPetr: Good point! I lowered the version to `tls.VersionTLS12`

flow/connectors/kafka/kafka.go
JanPetr: I wanted to use it, but I've ran into the issue that for Kafka peer you may use [multiple hosts split by comma](https://github.com/PeerDB-io/peerdb/blob/main/nexus/analyzer/src/lib.rs#L836-L841). And `shared.CreateTlsConfig()` helper is designed to work with single host only.

So I can either refactor the `shared.CreateTlsConfig()` helper method to work with multiple hosts or keep it as it is. ClickHouse connector uses [this approach](https://github.com/PeerDB-io/peerdb/blob/main/flow/connectors/clickhouse/clickhouse.go#L220-L240) as well.

What would you prefer? 🙏  Thank you for your time!

heavycrystal: ah that's fair, I think it would be nice to refactor that function to accept a slice of hosts and then change both kafka and clickhouse to use. Not a blocker for merging though

JanPetr: OK, thanks @heavycrystal! So I'm keeping it as it is now and won't refactor.

PR #3322, opened 2025-08-07
mongo: support inserting `-9223372036854775808` in MongoDB

jgao54:
```
rs0 [direct: primary] test> db.table.insertOne({ 'int': 9223372036854775808 })
{
  acknowledged: true,
  insertedId: ObjectId('68951d213f61d77f9e1703b4')
}
rs0 [direct: primary] test> db.table.find()
[
  {
    _id: ObjectId('68951d213f61d77f9e1703b4'),
    int: 9223372036854776000
  }
]
```
This is because the value got converted to float _in the shell_, result in precision loss, and so it got stored as a float.

Since we want to replicate the exact value from MongoDB, we replicate the "9223372036854775808" value to CH as expected.  In ClickHouse this becomes a Uint64.

The scenario that breaks the pipe is when `-9223372036854775808` is inserted, this becomes `-9223372036854776000` in MongoDB, and then fails during insertion due to parse failure. 

The workaround introduced in this PR is to write it as a string.

Tested e2e locally.


flow/e2e/mongo/mongo_test.go
```diff
@@ -770,6 +772,8 @@ func (s MongoClickhouseSuite) Test_Json_Types() {
 		require.Contains(t, row, `"float32_nan":"NaN"`)
 		require.Contains(t, row, `"float32_pos_inf":"+Inf"`)
 		require.Contains(t, row, `"float32_neg_inf":"-Inf"`)
+		require.Contains(t, row, `"precision_loss_float64_pos":9223372036854776000`)
+		require.Contains(t, row, `"precision_loss_float64_neg":"-9223372036854776000"`)
```
jgao54, line 776: not sure if we want this vs. have both min and max as string

heavycrystal:
I've discussed with @jgao54  and turns out this is a problem in general for values that MongoDB think of as floats but get interpreted as `int64` et al. by ClickHouse when there are no decimal places.

1. If the float fits in integer range, it gets inserted as an integer.
2. If the float does not fit inside integer range, it blocks the pipe.

This PR and associated issue are just a single case of this phenomenon, albeit one that is easy to hit.

Issue #3274, opened 2025-07-28
mongo: numbers less than -Int64.MinValue fail normalize


jgao54:
Current `JSONExtractString` behavior:
- number less than Abs(Int64.MaxValue) get casted to int64
- numbers greater than Int64.MaxValue get casted to uint64
- numbers less than Int64.MinValue will return an error during Normalize (`Cannot parse JSON object here:...`)

Last case is problematic as it will block workflow.

Repro:
```
db.table.insertOne({largeNumRaw: -9223372036854775807})
```
mongo will round this value to `-9223372036854776000`, this will break workflow with the parse error.

jgao54:
broader issue: https://github.com/PeerDB-io/peerdb/issues/3323

PR #3324, opened 2025-08-08
mongo: preserve float type from mongo to ch

jgao54:
- Remove codec + test for float32 handling (Mongo does not have a float32 equivalent so everything coming from changestream is float64)

- With this change, CDC path honors Double from Mongo to ClickHouse. However, QRep path does not. Given the same codec is used for marshaling, the difference in behavior here has to with how we handle data in qrep vs cdc.  

```

07bd249d78c1 :) SELECT JSONExtractString('{"number": 1.0, "number2": 1.1}');

   ┌─JSONExtractSt⋯er2": 1.1}')─┐
1. │ {"number":1,"number2":1.1} │
   └────────────────────────────┘
```
JSONExtractString converts `_full_document` column to JSON during initial load, erasing the explicit decimal and result in an Int64. Haven't thought of a good workaround for this without introducing a staging table similar to CDC.  Ideas welcome.

Update: 
Replace `JSONExtractString` with `CAST(<col>, 'JSON')` during qrep, which avoids this above quirk and correctly interprets floats with `.0` to Float64 in ClickHouse.

This also solves https://github.com/PeerDB-io/peerdb/issues/3274.

### DIFF
--- a/flow/connectors/clickhouse/qrep_avro_sync.go
+++ b/flow/connectors/clickhouse/qrep_avro_sync.go
@@ -289,7 +289,7 @@ func (s *ClickHouseAvroSyncMethod) pushS3DataToClickHouse(
 
 		if (field.Type == types.QValueKindJSON || field.Type == types.QValueKindJSONB) &&
 			qvalue.ShouldUseNativeJSONType(ctx, config.Env, s.ClickHouseConnector.chVersion) {
-			avroColName = fmt.Sprintf("JSONExtractString(%s)", avroColName)
+			avroColName = fmt.Sprintf("CAST(%s, 'JSON')", avroColName)
 		}
 
 		selectedColumnNames = append(selectedColumnNames, avroColName)
--- a/flow/connectors/mongo/codec.go
+++ b/flow/connectors/mongo/codec.go
@@ -3,6 +3,7 @@ package connmongo
 import (
 	"math"
 	"reflect"
+	"strconv"
 	"unsafe"
 
 	jsoniter "github.com/json-iterator/go"
@@ -78,16 +79,6 @@ func (codec *BsonDCodec) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
 		stream.WriteObjectField(elem.Key)
 
 		switch v := elem.Value.(type) {
-		case float32:
-			if math.IsNaN(float64(v)) {
-				stream.WriteString("NaN")
-			} else if math.IsInf(float64(v), 1) {
-				stream.WriteString("+Inf")
-			} else if math.IsInf(float64(v), -1) {
-				stream.WriteString("-Inf")
-			} else {
-				stream.WriteFloat32(v)
-			}
 		case float64:
 			if math.IsNaN(v) {
 				stream.WriteString("NaN")
@@ -96,7 +87,7 @@ func (codec *BsonDCodec) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
 			} else if math.IsInf(v, -1) {
 				stream.WriteString("-Inf")
 			} else {
-				stream.WriteFloat64(v)
+				writeFloat64WithExplicitDecimal(v, stream)
 			}
 		default:
 			// Delegate to json-iterator's encoding system for all other types.
@@ -128,16 +119,6 @@ func (codec *BsonACodec) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
 			stream.WriteMore()
 		}
 		switch v := elem.(type) {
-		case float32:
-			if math.IsNaN(float64(v)) {
-				stream.WriteString("NaN")
-			} else if math.IsInf(float64(v), 1) {
-				stream.WriteString("+Inf")
-			} else if math.IsInf(float64(v), -1) {
-				stream.WriteString("-Inf")
-			} else {
-				stream.WriteFloat32(v)
-			}
 		case float64:
 			if math.IsNaN(v) {
 				stream.WriteString("NaN")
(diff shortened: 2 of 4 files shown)


flow/connectors/mongo/codec.go
```diff
@@ -172,3 +153,12 @@ func CreateExtendedJSONMarshaler() jsoniter.API {
 	config.RegisterExtension(&BsonExtension{})
 	return config
 }
+
+func writeFloat64WithExplicitDecimal(v float64, stream *jsoniter.Stream) {
+	if v == math.Trunc(v) {
+		// use explicit decimal to hint ClickHouse to parse as float
+		stream.WriteRaw(strconv.FormatFloat(v, 'f', 1, 64))
```
ilidemi, line 160: How does this work with numbers like 1e100? Will it keep to the scientific notation or try to spell it out with a .0?

jgao54: Just tested this, it keeps it as 1e100

flow/connectors/clickhouse/qrep_avro_sync.go
```diff
@@ -289,7 +289,7 @@ func (s *ClickHouseAvroSyncMethod) pushS3DataToClickHouse(
 
 		if (field.Type == types.QValueKindJSON || field.Type == types.QValueKindJSONB) &&
 			qvalue.ShouldUseNativeJSONType(ctx, config.Env, s.ClickHouseConnector.chVersion) {
-			avroColName = fmt.Sprintf("JSONExtractString(%s)", avroColName)
```
jgao54, line 292: turns out this was main problem -- `JSONExtractString` removes the `.0` we added, i created an issue [here](https://github.com/ClickHouse/ClickHouse/issues/85267).

CAST to JSON just do the right thing and keeps the `.0`, so the json parser successfully interprets it as float64

flow/e2e/mongo/mongo_test.go
```diff
@@ -760,16 +756,13 @@ func (s MongoClickhouseSuite) Test_Json_Types() {
 		require.Contains(t, row, `"neg_int16":-32768`)
 		require.Contains(t, row, `"neg_int32":-2147483648`)
 		require.Contains(t, row, `"neg_int64":-9223372036854775807`)
-		require.Contains(t, row, `"float32":3.14`)
 		require.Contains(t, row, `"float64":3.14159265359`)
-		require.Contains(t, row, `"neg_float32":-3.14`)
 		require.Contains(t, row, `"neg_float64":-3.14159265359`)
+		require.Contains(t, row, `"max_float64":9223372036854776000`)
+		require.Contains(t, row, `"min_float64":-9223372036854776000`)
```
jgao54: note: this is display issue in clickhouse console, just like javascript

flow/e2e/mongo/mongo_test.go
```diff
@@ -648,19 +648,15 @@ func (s MongoClickhouseSuite) Test_Json_Types() {
 		{Key: "neg_int64", Value: int64(-9223372036854775807)},
 
 		// Floating point types
-		{Key: "float32", Value: float32(3.14)},
 		{Key: "float64", Value: float64(3.14159265359)},
-		{Key: "neg_float32", Value: float32(-3.14)},
 		{Key: "neg_float64", Value: float64(-3.14159265359)},
+		{Key: "max_float64", Value: float64(math.MaxInt64)},
+		{Key: "min_float64", Value: float64(math.MinInt64)},
```
ilidemi:
```suggestion
		{Key: "float64_max_int64", Value: float64(math.MaxInt64)},
		{Key: "float64_min_int64", Value: float64(math.MinInt64)},
```

flow/e2e/mongo/mongo_test.go
ilidemi: This one is coming from `encoding/json.Marshal()`, and Delve debugger does the same :/
Seems like it's just in the imprecise float territory across the implementations, even if the specific value might seem special

ilidemi:
```suggestion
		require.Contains(t, row, `"float64_max_int64":9223372036854776000`)
		require.Contains(t, row, `"float64_min_int64":-9223372036854776000`)
```

flow/connectors/mongo/codec.go
ilidemi: Unit test would be good

flow/e2e/mongo/mongo_test.go
ilidemi: If the goal was to also test that the e2e doesn't crash on outside-of-int64 integers on CH side, maybe add a +-2^65 or something like that (via strconv.ParseFloat())?

serprex (approved):

flow/connectors/mongo/codec.go
jgao54: looking at how mongo handles this:
- it will automatically switch to scientific notation for very small / large values (and vice versa, switching from scientific notation if it can be represented ), regardless of how it's entered. For example:
```
rs0 [direct: primary] test> db.num.insertOne({'num': Double(-1e10)})
{
  acknowledged: true,
  insertedId: ObjectId('6897d0e9d5ded0d0071703c3')
}
rs0 [direct: primary] test> db.num.insertOne({'num': 100000000000000000000000000000000000000000000000000})
{
  acknowledged: true,
  insertedId: ObjectId('6897d0ecd5ded0d0071703c4')
}
rs0 [direct: primary] test> db.num.find()
[
  { _id: ObjectId('6897d0e9d5ded0d0071703c3'), num: -10000000000 },
  { _id: ObjectId('6897d0ecd5ded0d0071703c4'), num: 1e+50 }
]
```

In this case with `1e100`, unit test will actually return `10000000000000000159028911097599180468360808563945281389781327557747838772170381060813469985856815104.0` but e2e it shows up as  `1e100`.

So adding e2e test here to look for `1e100`.

ilidemi: The concern would be that for a specific workload we’re multiplying the size of S3 blob and the staging row by 10x or more. We can say it’s fine because it’s rare but would have to be mindful in either case

ilidemi: Actually, does this only to apply to some special numbers or all big numbers? All big float64s are technically integers afaict

ilidemi: #3327

PR #3325, opened 2025-08-08
Validate mirror: no need to get source schemas for resync

Amogh-Bharadwaj:
We get source schemas for validating destination tables' schemas, which in the case of resync we do not do. Hence no need to fetch schemas from source in the case of resync

### DIFF
--- a/flow/cmd/validate_mirror.go
+++ b/flow/cmd/validate_mirror.go
@@ -12,6 +12,7 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/connectors"
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
 	"github.com/PeerDB-io/peerdb/flow/internal"
+	"github.com/PeerDB-io/peerdb/flow/shared"
 	"github.com/PeerDB-io/peerdb/flow/shared/exceptions"
 )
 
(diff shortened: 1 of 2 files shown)


Amogh-Bharadwaj:
Currently no step in `ValidateMirrorDestination` is relevant for resync. I am open to gating that entire call, but not doing so for now in case we plan to add more checks

flow/cmd/validate_mirror.go
```diff
@@ -96,13 +98,23 @@ func (h *FlowRequestHandler) ValidateCDCMirror(
...
+		if getTableSchemaError != nil {
+			return nil, fmt.Errorf("failed to get source table schema: %w", getTableSchemaError)
+		}
+	} else {
+		// No need to get table schema for resync, as we will create or replace the tables
+		tableSchemaMap = make(map[string]*protos.TableSchema, len(req.ConnectionConfigs.TableMappings))
+		for _, tm := range req.ConnectionConfigs.TableMappings {
+			tableSchemaMap[tm.DestinationTableIdentifier] = &protos.TableSchema{}
+		}
 	}
```
serprex:
```suggestion
	}
```
why does validation need this map filled with blanks?

adding nil handling in connector might address e2e failures

PR #3327, opened 2025-08-11
Mongo: limit max length of marshaled floats

ilidemi:
Numbers like 1e300 (anything > 1e21) technically pass the check of being an integer but scientific notation would be a better hint then spelling out all their digits with a `.0`, also avoiding the size of s3 payload and the raw table getting ballooned.

We're not expecting a lot of such values but just something for a peace of mind.

### DIFF
--- a/flow/connectors/mongo/codec_test.go
+++ b/flow/connectors/mongo/codec_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"math"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


jgao54 (approved):
> also avoiding the size of s3 payload and the raw table getting ballooned

Good point!

flow/connectors/mongo/codec_test.go
```diff
@@ -345,3 +346,32 @@ func TestMarshalId(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, `"6893edbecb1f9508891bbb84"`, string(result))
 }
+
+// Tests that floats of all magnitudes are marshalled into a resonable length and have a signifier
+// that they're not integers
+func TestMarshalFloatLengths(t *testing.T) {
```
jgao54, line 352: nice test

PR #3337, opened 2025-08-12
Allow to pass existing CH tables to validate

ilidemi:
When the pipe is resumed with a table config update, we should validate existing destination tables too as they could change while the pipe was paused.

This change allows the shared package to support that.


ilidemi:
Not doing the existing tables

PR #3341, opened 2025-08-12
clickhouse: PARTITION BY expression

serprex:
I was wrong, making partitioning as inflexible as ORDER BY columns doesn't work

Most easily is that you should pretty much never partition by a datetime without rounding off most of its lower bits

Previous partitioning should be considered deprecated


flow/connectors/clickhouse/normalize.go
```diff
@@ -235,13 +235,17 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
...
-			partitionByStr := strings.Join(partitionByColumns, ",")
-			fmt.Fprintf(&stmtBuilder, " PARTITION BY (%s)", partitionByStr)
+		if tableMapping != nil && tableMapping.PartitionByExpr != "" {
+			fmt.Fprintf(&stmtBuilder, " PARTITION BY (%s)", tableMapping.PartitionByExpr)
+		} else {
+			partitionByColumns := getOrderedPartitionByColumns(tableMapping, colNameMap)
+			if len(partitionByColumns) > 0 {
+				partitionByStr := strings.Join(partitionByColumns, ",")
+				fmt.Fprintf(&stmtBuilder, " PARTITION BY (%s)", partitionByStr)
+			}
```
jgao54: > Previous partitioning should be considered deprecated

with this change, user can still provide columns as partition key right?

if so i think by removing `hasNullablePartitionKey`,  partition key on Nullable column will break `create table` statement since it won't have `allow_nullable_key = 1`.

did you intend to always set `allow_nullable_key = 1` when there is a partition key?

jgao54: Actually even with expression, the same settings apply: https://fiddle.clickhouse.com/fde246da-7404-40a0-80d6-f31b1519b3fe

Could we still use `hasNullablePartitionKey`, but set to true always when an PartitionByExpr is present?

serprex: does nullable key apply to partitioning? when I looked at docs it seemed like it only mattered for order by

jgao54: see the fiddle link i provided above

serprex: yep, seems like it does

bit tempted to make `allow_nullable_key = 1` set unconditionally

PR #3342, opened 2025-08-13
mysql: ignore zerofill on type

serprex:
this is a display hint


ilidemi (approved):

heavycrystal:
looks like you can have `zerofill unsigned` as well, this PR wouldn't handle that
raising followup

PR #3343, opened 2025-08-13
Better flow statuses for renaming and table removal phases 

Amogh-Bharadwaj:
## Why
There are currently two states in the lifecycle of a mirror where the status is `Running` but the mirror is not running/in CDC.

### Table removal

Currently, when a mirror is in the process of removing tables, the status shows up as `Running`, when in reality the mirror is not in a position to accept flow state change signals; if you were to request a table removal during an ongoing table removal, your request would "succeed" as far as the API call goes but the signal is not acted upon by the worker.

A consequence and another pain point is that there is no way to programmatically know if a mirror is finished with its table removal as the state is always Running.

### Renaming phase of resync

The last step of a resync is an activity where we exchange the _resync tables with the originals. Much like table removal, this is a state where the mirror cannot be paused/edited.

### What
- This PR introduces a `Modifying` flow state for table removal.
- Sets the status to `RESYNC` during renaming.

- [ ] Automated and functional testing pending

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -351,6 +351,7 @@ func processTableRemovals(
 	cfg *protos.FlowConnectionConfigs,
 	state *CDCFlowWorkflowState,
 ) error {
+	state.updateStatus(ctx, logger, protos.FlowStatus_STATUS_MODIFYING)
 	removeTablesSelector := workflow.NewNamedSelector(ctx, "RemoveTables")
 	removeTablesSelector.AddReceive(ctx.Done(), func(_ workflow.ReceiveChannel, _ bool) {})
 	flowSignalStateChangeChan := model.FlowSignalStateChange.GetSignalChannel(ctx)
@@ -728,6 +729,7 @@ func CDCFlowWorkflow(
 		}
 
 		if cfg.Resync {
+			state.updateStatus(ctx, logger, protos.FlowStatus_STATUS_RESYNC)
 			renameOpts := &protos.RenameTablesInput{
 				FlowJobName:       cfg.FlowJobName,
 				PeerName:          cfg.DestinationName,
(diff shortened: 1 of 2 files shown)


flow/workflows/cdc_flow.go
```diff
@@ -728,6 +729,7 @@ func CDCFlowWorkflow(
 		}
 
 		if cfg.Resync {
+			state.updateStatus(ctx, logger, protos.FlowStatus_STATUS_RESYNC)
```
heavycrystal, line 728: It would make more sense to mark the whole resync step as this rather than having SNAPSHOT -> RESYNC transition

Amogh-Bharadwaj: True
The implementation for that would be a bit tricky since resync is essentially a fresh new mirror. Perhaps changing the state here to `RENAME` is better?

heavycrystal: why is it tricky? I didn't quite follow
IMO `RENAME` is both a weird status and a leaky abstraction

Amogh-Bharadwaj: > why is it tricky? I didn't quite follow


Because there is no "resync step". Resync is an abstraction over dropping the existing mirror and creating a new one. We would have to add overrides in`cdc_flow.go` to set the status to `RESYNC` everywhere instead of the usual `SETUP`, `SNAPSHOT` etc in the case where `resync` is set to true in the config -- something which I'm not sure we'd want to do as those statuses are important for users to see, and them seeing `RESYNC` for the entire lifecycle of the mirror can be troublesome.

cc @iskakaushik

Amogh-Bharadwaj: > IMO RENAME is both a weird status and a leaky abstraction

I am open to having a different name, but having a dedicated status for that specific activity is what I was thinking.

iamKunalGupta:
Also need to track any new statuses to wait on during Maintenance

iamKunalGupta:
Need to add `STATUS_MODIFYING` in the waitStatuses for maintenance added in https://github.com/PeerDB-io/peerdb/pull/3377

PR #3349, opened 2025-08-14
mongo: enable ShardedCluster + better code sharing for validation

jgao54:
Change Streams API is documented to work out-of-the-box for sharded cluster. This PR updates the shared validation code to:
- check to make sure that the instance is either replica set or sharded cluster (previously, it ran `replSetGetStatus` to check this, we can instead use 'hello' command which is more general)
- add explicit validation for required roles (`readAnyDatabase`, `clusterMonitor`)
- better separation of concern by creating a `commands.go` file for all the db commands, while `validation.go` implements the validation logic -- will follow up with ClickPipes side as well. 

Testing:
- [x]  tested replica set on Atlas
- [x] test replica set on EC2
- [x] test sharded cluster on Atlas
- [x] test sharded clsuter on EC2

### DIFF
--- a/flow/shared/mongo/commands.go
+++ b/flow/shared/mongo/commands.go
@@ -12,7 +12,7 @@ type BuildInfo struct {
 	Version string `bson:"version"`
 }
 
-func GetBuildInfo(ctx context.Context, client *mongo.Client) (*BuildInfo, error) {
+func GetBuildInfo(ctx context.Context, client *mongo.Client) (BuildInfo, error) {
 	return runCommand[BuildInfo](ctx, client, "buildInfo")
 }
 
@@ -21,7 +21,7 @@ type ReplSetGetStatus struct {
 	MyState int    `bson:"myState"`
 }
 
-func GetReplSetGetStatus(ctx context.Context, client *mongo.Client) (*ReplSetGetStatus, error) {
+func GetReplSetGetStatus(ctx context.Context, client *mongo.Client) (ReplSetGetStatus, error) {
 	return runCommand[ReplSetGetStatus](ctx, client, "replSetGetStatus")
 }
 
@@ -38,7 +38,7 @@ type ServerStatus struct {
 	OplogTruncation OplogTruncation `bson:"oplogTruncation"`
 }
 
-func GetServerStatus(ctx context.Context, client *mongo.Client) (*ServerStatus, error) {
+func GetServerStatus(ctx context.Context, client *mongo.Client) (ServerStatus, error) {
 	return runCommand[ServerStatus](ctx, client, "serverStatus")
 }
 
@@ -55,7 +55,7 @@ type Role struct {
 	DB   string `bson:"db"`
 }
 
-func GetConnectionStatus(ctx context.Context, client *mongo.Client) (*ConnectionStatus, error) {
+func GetConnectionStatus(ctx context.Context, client *mongo.Client) (ConnectionStatus, error) {
 	return runCommand[ConnectionStatus](ctx, client, "connectionStatus")
 }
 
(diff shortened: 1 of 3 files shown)


flow/shared/mongo/validation.go
```diff
@@ -2,75 +2,125 @@ package mongo
...
-		return nil, fmt.Errorf("failed to decode ReplSetGetStatus: %w", err)
+
+	topology, err := GetTopologyType(ctx, client)
+	if err != nil {
+		return err
+	}
+	if topology == ReplicaSet {
+		return validateOplogRetention(ctx, client)
+	} else {
+		// TODO: run validation on shard
```
jgao54, line 97: left this out as a follow-up, to keep this PR size small.

flow/shared/mongo/commands.go
```diff
@@ -0,0 +1,84 @@
...
+	})
+	if singleResult.Err() != nil {
+		return nil, fmt.Errorf("'%s' failed: %v", command, singleResult.Err())
+	}
+
+	var result T
+	if err := singleResult.Decode(&result); err != nil {
+		return nil, fmt.Errorf("'%s' failed: %v", command, err)
+	}
+	return &result, nil
```
serprex:
```suggestion
func runCommand[T any](ctx context.Context, client *mongo.Client, command string) (T, error) {
	var result T
	singleResult := client.Database("admin").RunCommand(ctx, bson.D{
		bson.E{Key: command, Value: 1},
	})
	if singleResult.Err() != nil {
		return result, fmt.Errorf("'%s' failed: %v", command, singleResult.Err())
	}

	if err := singleResult.Decode(&result); err != nil {
		return result, fmt.Errorf("'%s' failed: %v", command, err)
	}
	return result, nil
```

flow/shared/mongo/validation.go
```diff
@@ -2,75 +2,125 @@ package mongo
...
+		return nil
+	}
 }
 
-func GetBuildInfo(ctx context.Context, client *mongo.Client) (*BuildInfo, error) {
-	singleResult := client.Database("admin").RunCommand(ctx, bson.D{bson.E{Key: "buildInfo", Value: 1}})
-	if singleResult.Err() != nil {
-		return nil, fmt.Errorf("failed to run 'buildInfo' command: %w", singleResult.Err())
+func ValidateUserRoles(ctx context.Context, client *mongo.Client) error {
+	RequiredRoles := []string{"readAnyDatabase", "clusterMonitor"}
```
serprex: RequiredRoles can move to top level var, & `[...]` to make it an array instead of a slice

flow/shared/mongo/validation.go
```diff
@@ -2,75 +2,125 @@ package mongo
...
+
+	connectionStatus, err := GetConnectionStatus(ctx, client)
+	if err != nil {
+		return err
 	}
-	var info BuildInfo
-	if err := singleResult.Decode(&info); err != nil {
-		return nil, fmt.Errorf("failed to decode BuildInfo: %w", err)
+
+	hasRole := func(roles []Role, targetRole string) bool {
```
serprex: inline into loop, or use `slices.ContainsFunc`

flow/shared/mongo/validation.go
```diff
@@ -2,75 +2,125 @@ package mongo
...
 }
 
-func GetReplSetGetStatus(ctx context.Context, client *mongo.Client) (*ReplSetGetStatus, error) {
-	singleResult := client.Database("admin").RunCommand(ctx, bson.D{
-		bson.E{Key: "replSetGetStatus", Value: 1},
-	})
-	if singleResult.Err() != nil {
-		return nil, fmt.Errorf("failed to run 'replSetGetStatus' command: %w", singleResult.Err())
+func ValidateOplogRetention(ctx context.Context, client *mongo.Client) error {
+	validateOplogRetention := func(instanceCtx context.Context, instanceClient *mongo.Client) error {
```
serprex, line 77: inline into if

flow/shared/mongo/validation.go
```diff
@@ -2,75 +2,125 @@ package mongo
...
+	if cmp, err := CompareServerVersions(buildInfo.Version, MinSupportedVersion); err != nil {
+		return err
+	} else if cmp < 0 {
+		return fmt.Errorf("require minimum mongo version %s", MinSupportedVersion)
+	}
 
-type OplogTruncation struct {
-	OplogMinRetentionHours float64 `bson:"oplogMinRetentionHours"`
-}
+	validateStorageEngine := func(instanceCtx context.Context, instanceClient *mongo.Client) error {
```
serprex, line 34: inline into if

jgao54: ah, this and `validateOplogRetention` was intentional as the function gets reused in https://github.com/PeerDB-io/peerdb/pull/3356

flow/shared/mongo/commands.go
jgao54: I had *T to avoid copying but this is safer 👍

serprex (approved):

PR #3350, opened 2025-08-14
Replace local activities with plain activities, except status in catalog

ilidemi:
Like #3301 but trying to take the status in catalog separately

Plan: look carefully that this won't lead to desyncs then tackle the status in catalog separately


serprex (approved):

ilidemi:
To make sure there won't be a desync:
- There are 3 local activities total
- getQRepOverwriteFullRefreshMode and getPeerType are called from snapshot workflow(s), which we're letting run till the end before initiating the deployment, so no chance of them being in history
- uploadConfigToCatalog is trickier as it's called all around the pause loop. But they only kick in on terminating, resync and flow config update actions, and all of those can only come through the FlowStateChange handler, which is blocked during the maintenance mode. So maintenance mode would cause all CDC flows to go into pause state, then the code gets swapped, then the replay till pause handler doesn't have any config updates, and once the maintenance resumes the flow, the new execution will avoid the lines with the config updates and run till ContinueAsNew. After that code changes are fair game, so we're all good.

PR #3361, opened 2025-08-18
3 max conn, no slot sizing parallelism

serprex:


flow/activities/flowable.go
```diff
@@ -745,9 +742,9 @@ func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
 			logger.Info(name+" completed", slog.Duration("duration", time.Since(now)))
 		}
 	}
-	defer shared.Interval(ctx, 10*time.Minute, wrapWithLog(ctx, "SendWALHeartbeat", a.SendWALHeartbeat))()
-	defer shared.Interval(ctx, 1*time.Minute, wrapWithLog(ctx, "RecordMetrics", a.RecordMetrics))()
-	defer shared.Interval(ctx, 1*time.Minute, wrapWithLog(ctx, "RecordSlotSizes", a.RecordSlotSizes))()
+	defer shared.Interval(ctx, 10*time.Minute, wrapWithLog("SendWALHeartbeat", a.SendWALHeartbeat))()
```
serprex: before we had a ticker so only one of these would run at a time, thus I expect these will still overlap in catalog access

something else I've tended to try is have these sort of scheduled tasks be on coprime intervals, but not sure if that'll make metrics screwy in this case

PR #3369, opened 2025-08-19
replace normalize requests channel

serprex:
tracks changes to atomic.Int64 so normalization always receives latest value,
each update closing a series of channels for synchronization

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -334,12 +334,18 @@ func (a *FlowableActivity) SyncFlow(
 	// Wait on normalizeDone at end to not interrupt final normalize
 	syncDone := make(chan struct{})
 	normRequests := concurrency.NewLastChan()
+	normResponses := concurrency.NewLastChan()
+	normBufferSize, err := internal.PeerDBNormalizeBufferSize(ctx, config.Env)
+	if err != nil {
+		connectors.CloseConnector(ctx, srcConn)
+		return a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
+	}
 
 	group, groupCtx := errgroup.WithContext(ctx)
 	group.Go(func() error {
 		normalizeCtx := internal.WithOperationContext(groupCtx, protos.FlowOperation_FLOW_OPERATION_NORMALIZE)
 		// returning error signals sync to stop, normalize can recover connections without interrupting sync, so never return error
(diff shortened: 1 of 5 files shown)


flow/activities/flowable.go
```diff
@@ -398,6 +396,9 @@ func (a *FlowableActivity) SyncFlow(
 
 	syncState.Store(shared.Ptr("cleanup"))
 	close(syncDone)
+	normRequests.Close()
```
heavycrystal, line 400: no defer for these?

flow/activities/flowable.go
```diff
@@ -381,6 +378,7 @@ func (a *FlowableActivity) SyncFlow(
 			logger.Error("failed to sync records", slog.Any("error", syncErr))
 			syncState.Store(shared.Ptr("cleanup"))
 			close(syncDone)
+			normRequests.Close()
```
heavycrystal, line 381: should we close `normResponses` here too?

flow/activities/flowable_core.go
```diff
@@ -334,10 +337,13 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 
 	if recordBatchSync.NeedsNormalize() {
 		syncState.Store(shared.Ptr("normalizing"))
-		select {
-		case normRequests <- NormalizeBatchRequest{BatchID: res.CurrentSyncBatchID}:
-		case <-ctx.Done():
-			return res, nil
+		normRequests.Update(res.CurrentSyncBatchID)
+		for normResponses.Load() <= res.CurrentSyncBatchID-normBufferSize {
```
heavycrystal: could be worth asserting `normBufferSize` >=0 otherwise this is infinite

flow/activities/flowable.go
serprex: double close is panic & we need this to happen precisely here

heavycrystal (approved):

flow/activities/flowable.go
heavycrystal: could add a variable to LastChan preventing double close

serprex: LastChan handles double close fine, `close(syncDone)` is issue. But want to keep all the close logic coupled

PR #3372, opened 2025-08-20
catalog pool: keep connections at 1 aggressively

serprex:


flow/internal/catalog.go
```diff
@@ -29,8 +29,9 @@ func GetCatalogConnectionPoolFromEnv(ctx context.Context) (shared.CatalogPool, e
 			return shared.CatalogPool{},
 				exceptions.NewCatalogError(fmt.Errorf("unable to parse catalog connection string: %w", err))
 		}
+		config.MinConns = 1
 		config.MaxConns = 3
-		config.MaxConnIdleTime = 5 * time.Minute
+		config.MaxConnIdleTime = time.Second
```
heavycrystal, line 34: Would this cause a lot of connections "churn"? not sure how often we need a connection on aggregate but if an instance has several mirrors it could be 10+ times a minute

serprex: those mirrors will generally have non overlapping catalog access

PR #3375, opened 2025-08-21
Log all progress

ilidemi:
If something is slow, hanging or not hanging, we should be able to tell what it is from logs.

Add progress logs where they're missing, in pull, sync, normalize, qrep/cdc
Standardize cadence to 50000 records (10000 in OCF is a bit spammy)
Consistently log # of records, bytes, channel length (correlated with memory consumed by that pipe/partition), whether we're waiting for commit
Tabulate start/end for long running operations
Unify names of some properties so we can do a column in grafana
Rename CH qrep_avro_sync.go to avro_sync.go as it's actually shared with CDC

### DIFF
--- a/flow/connectors/utils/metered_dialer.go
+++ b/flow/connectors/utils/metered_dialer.go
@@ -6,30 +6,51 @@ import (
 	"sync/atomic"
 )
 
+type Meter struct {
+	lastValue int64
+	bytesRead atomic.Int64
+}
+
+func (m *Meter) Reset() {
+	m.bytesRead.Store(0)
+	m.lastValue = 0
+}
+
+func (m *Meter) TotalBytesRead() int64 {
+	return m.bytesRead.Load()
+}
+
+func (m *Meter) DeltaBytesRead() int64 {
+	current := m.bytesRead.Load()
+	delta := current - m.lastValue
+	m.lastValue = current
+	return delta
+}
+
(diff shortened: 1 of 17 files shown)


flow/connectors/utils/metered_dialer.go
```diff
@@ -6,30 +6,51 @@ import (
...
+}
+
+func (m *Meter) TotalBytesRead() int64 {
+	return m.bytesRead.Load()
+}
+
+func (m *Meter) DeltaBytesRead() int64 {
+	current := m.bytesRead.Load()
+	delta := current - m.lastValue
+	m.lastValue = current
```
serprex: this isn't thread safe, `lastValue` isn't atomic & we could have duplicate reads of `bytesRead.Load`

would prefer to keep single atomic.Int64, but if you wanted running total you could use a 2nd atomic.Int64 which we add the swapped value to:

```go
    delta := m.bytesRead.Swap(0)
    m.totalBytesRead.Add(delta)
    return delta
```

this way we never have two concurrent loads doublebook reading

alternative is to add to two separate ints at the time of original addition (probably better, since that way `TotalBytesRead` returns total read so far, not total reported to metrics so far)

ilidemi: Adding to two ints is genius

serprex (approved):

PR #3379, opened 2025-08-22
Temporal: Set MaxHeartbeatThrottleInterval to 10 seconds

Amogh-Bharadwaj:
Cause for slow pausing was that cancellation requests were being acted upon only after 0.8 times heartbeat timeout had passed.

This PR lowers the interval of pending heartbeats being sent. This is a worker-level setting so affects all our activities. More info here:
https://docs.temporal.io/encyclopedia/detecting-activity-failures#throttling

It might also mean higher network traffic from our worker to Temporal

### DIFF
--- a/flow/cmd/worker.go
+++ b/flow/cmd/worker.go
@@ -92,7 +92,7 @@ func WorkerSetup(ctx context.Context, opts *WorkerSetupOptions) (*WorkerSetupRes
 		OnFatalError: func(err error) {
 			slog.Error("Peerflow Worker failed", slog.Any("error", err))
 		},
-		MaxHeartbeatThrottleInterval: 15 * time.Second,
+		MaxHeartbeatThrottleInterval: 10 * time.Second,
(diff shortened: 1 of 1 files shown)


heavycrystal (approved):

serprex (approved):

flow/cmd/worker.go
```diff
@@ -91,6 +92,7 @@ func WorkerSetup(ctx context.Context, opts *WorkerSetupOptions) (*WorkerSetupRes
 		OnFatalError: func(err error) {
 			slog.Error("Peerflow Worker failed", slog.Any("error", err))
 		},
+		MaxHeartbeatThrottleInterval: 15 * time.Second,
```
serprex:
```suggestion
		MaxHeartbeatThrottleInterval: 10 * time.Second,
```
avoids potentially skipping a beat when we're reporting at 15s interval

PR #3380, opened 2025-08-22
fix connection leak in GetDefaultPartitionKeyForTables

heavycrystal:


ilidemi:
@heavycrystal @Amogh-Bharadwaj is this a similar leak? Every other call to GetByNameAs has a Close()
https://github.com/PeerDB-io/peerdb/blob/a7810adf658ce2e8cb94943fdb2f56f0f1ae71d1/flow/activities/flowable_core.go#L146-L156

PR #3385, opened 2025-08-22
try and add temporal metadata to slog.Logger

heavycrystal:
followup to #3382


ilidemi (approved):

iamKunalGupta:
Can we use activity.GetLogger and convert it to a slog logger?
This way we can reuse loggerFromContext

iamKunalGupta:
ok so slog is one of the implementations of `activity.GetLogger` and we pass a slog Logger when initializing temporal
Might be a good idea to check the type of the logger returned from `LoggerFromCtx` and use it if it is slog else go the usual route of creating everything from scratch

PR #3387, opened 2025-08-22
mongo: error classification

jgao54:
Add error classification for MongoDB. The driver has a convenient `RetryableRead` function, which keeps track of retryable errors. 

This covers two errors we have encountered in the past:
- ShutdownInProgress
- AuthorizeError


heavycrystal (approved):
Can cover `ChangeStreamHistoryLost` as well, should be a separate `ErrorClass` that will notify the user.

PR #3395, opened 2025-08-26
refactor: disable telemetry error sender by default

iamKunalGupta:


flow/internal/config.go
```diff
@@ -193,10 +193,11 @@ func PeerDBMaintenanceModeWaitAlertSeconds() int {
 	return getEnvConvert("PEERDB_MAINTENANCE_MODE_WAIT_ALERT_SECONDS", 600, strconv.Atoi)
 }
 
-func PeerDBTelemetryErrorActionBasedAlertingEnabled() bool {
-	enabled, err := strconv.ParseBool(GetEnvString("PEERDB_TELEMETRY_ERROR_ACTION_BASED_ALERTING_ENABLED", "false"))
+// PEERDB_TELEMETRY_SENDER_SEND_ERROR_ALERTS_ENABLED is whether to send error alerts to the telemetry sender
+func PeerDBTelemetrySenderSendErrorAlertsEnabled() bool {
```
heavycrystal, line 197: why swap the dynconf name? just change the default value?

iamKunalGupta: This is env var, not dynconf, we don't use it anywhere

serprex (approved):

PR #3367, opened 2025-08-19
Optimisation: Do not open destination connection for last offset fetch unless Postgres target

Amogh-Bharadwaj:
Decoupling from https://github.com/PeerDB-io/peerdb/pull/3206 and addressing the [comment](https://github.com/PeerDB-io/peerdb/pull/3206#discussion_r2205157385) there.

This PR makes it so that we do not connect to the destination peer to get the last offset for sync flow.

PR #3396, opened 2025-08-26
Last offset fetch: prevent destination connection by returning early

Amogh-Bharadwaj:
Implementation in https://github.com/PeerDB-io/peerdb/pull/3367 misses the fact that GetConnector which we call under the hood opens a destination connection, and only after we successfully get a connector do we assert that the connector supports T.

This PR clones GetByNameAs and passes in an `expectedType` parameter which is used to error out with unsupported early

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -144,7 +144,8 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	lastOffset, err := func() (model.CdcCheckpoint, error) {
 		// special case pg-pg replication, where offsets are stored on destination instead of catalog
 		if _, isPg := any(srcConn).(*connpostgres.PostgresConnector); isPg {
-			dstPgConn, err := connectors.GetByNameAs[*connpostgres.PostgresConnector](ctx, config.Env, a.CatalogPool, config.DestinationName)
+			dstPgConn, err := connectors.GetByNameAndForPeerAs[*connpostgres.PostgresConnector](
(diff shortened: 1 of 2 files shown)


flow/activities/flowable_core.go
```diff
@@ -144,7 +144,8 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 	lastOffset, err := func() (model.CdcCheckpoint, error) {
 		// special case pg-pg replication, where offsets are stored on destination instead of catalog
 		if _, isPg := any(srcConn).(*connpostgres.PostgresConnector); isPg {
-			dstPgConn, err := connectors.GetByNameAs[*connpostgres.PostgresConnector](ctx, config.Env, a.CatalogPool, config.DestinationName)
+			dstPgConn, err := connectors.GetByNameAndForPeerAs[*connpostgres.PostgresConnector](
```
serprex: Type parameter must always be same with last parameter. Better to create specific (non generic) function for postgres. Or switch over type parameter

serprex (approved):

flow/connectors/core.go
```diff
@@ -485,6 +485,19 @@ func GetByNameAs[T Connector](ctx context.Context, env map[string]string, catalo
 	return GetAs[T](ctx, env, peer)
 }
 
+func GetByNameAndForPeerAs[T Connector](ctx context.Context, env map[string]string, catalogPool shared.CatalogPool, name string, expectedType protos.DBType) (T, error) {
+	peer, err := LoadPeer(ctx, catalogPool, name)
+	if err != nil {
+		var none T
```
serprex: Can put none at top of function, no need to declare twice

PR #3401, opened 2025-08-27
Bigquery source

serprex:


flow/activities/snapshot_activity.go
```diff
@@ -190,3 +190,25 @@ func (a *SnapshotActivity) GetDefaultPartitionKeyForTables(
...
+		return err
+	}
+	uris, err := src.AvroExport(ctx, config)
+	if err != nil {
+		return err
+	}
+
+	// Some changes that could happen here:
+	// - Make URIs a channel so import can start before all tables exported
+	// - Split up activity so exports aren't retried
```
serprex, line 207: bring retries into activity, if activity fails workflow fails, disable workflow level retries

PR #3403, opened 2025-08-27
pg/mysql cdc: report fetched bytes metric with interval

serprex:
@ilidemi noticed 10% of cpu usage in pg/mysql was attributed to metrics

reuse pattern used by qrep where we add to atomic counters & regularly empty those counters

mongo cdc already does this

### DIFF
--- a/flow/connectors/mysql/cdc.go
+++ b/flow/connectors/mysql/cdc.go
@@ -10,6 +10,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-mysql-org/go-mysql/mysql"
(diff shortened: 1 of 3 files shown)


flow/connectors/mysql/cdc.go
```diff
@@ -397,7 +410,7 @@ func (c *MySqlConnector) PullRecords(
 				} else if inTx {
 					c.logger.Info("[mysql] timeout reached, but still in transaction, waiting for inTx false",
 						slog.Uint64("records", uint64(recordCount)),
-						slog.Int("bytes", bytesRead),
+						slog.Int64("bytes", fetchedBytes.Load()),
```
ilidemi:
```suggestion
						slog.Int64("bytes", totalFetchedBytes.Load()),
```

ilidemi (approved):

PR #3404, opened 2025-08-27
catalog pool: don't set MinConns to 1

serprex:
health check only runs once a minute, so pool wasn't aggressively culled

meanwhile MinConns counterbalance keeps cold services lukewarm

### DIFF
--- a/flow/internal/catalog.go
+++ b/flow/internal/catalog.go
@@ -29,9 +29,8 @@ func GetCatalogConnectionPoolFromEnv(ctx context.Context) (shared.CatalogPool, e
 			return shared.CatalogPool{},
 				exceptions.NewCatalogError(fmt.Errorf("unable to parse catalog connection string: %w", err))
 		}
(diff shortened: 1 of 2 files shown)


iamKunalGupta (approved):

flow/internal/catalog.go
```diff
@@ -29,9 +29,8 @@ func GetCatalogConnectionPoolFromEnv(ctx context.Context) (shared.CatalogPool, e
 			return shared.CatalogPool{},
 				exceptions.NewCatalogError(fmt.Errorf("unable to parse catalog connection string: %w", err))
 		}
-		config.MinConns = 1
 		config.MaxConns = 3
-		config.MaxConnIdleTime = time.Second
+		config.MaxConnIdleTime = time.Minute
```
ilidemi: 90 * time.Second so there isn't minute - close - open - minute - close - open

PR #3407, opened 2025-08-28
Fetch configuration from DB

alexstoick:
No longer passing the entire configuration object to the Temporal job as we can hit the 2MB hard limit imposed by Temporal. Instead, pass flowJobName and fetch the config from the DB in all the places required.

There are some additional changes that were required to be made, as some of the jobs are used from multiple contexts - and I had to adapt some of the information passed to ensure that they continue to work as expected.

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -464,7 +464,7 @@ func CDCFlowWorkflow(
 		return state, nil
 	}
 
-	mirrorNameSearch := shared.NewSearchAttributescfg.FlowJobName
+	mirrorNameSearch := shared.NewSearchAttributes(cfg.FlowJobName)
 
 	if state.ActiveSignal == model.PauseSignal {
 		selector := workflow.NewNamedSelector(ctx, "PauseLoop")
--- a/flow/workflows/snapshot_flow.go
+++ b/flow/workflows/snapshot_flow.go
@@ -120,6 +120,11 @@ func (s *SnapshotFlowExecution) cloneTable(
 		slog.String(string(shared.FlowNameKey), flowName),
 		slog.String("snapshotName", snapshotName))
 
+	config, err := internal.FetchConfigFromDB(s.FlowJobName)
+	if err != nil {
+		return fmt.Errorf("unable to fetch config from DB for flow-job-name %s; err : %w", s.FlowJobName, err)
+	}
+
 	srcName := mapping.SourceTableIdentifier
 	dstName := mapping.DestinationTableIdentifier
 	originalRunID := workflow.GetInfo(ctx).OriginalRunID
@@ -281,6 +286,15 @@ func (s *SnapshotFlowExecution) cloneTables(
 	config, err := internal.FetchConfigFromDB(s.FlowJobName)
 	if err != nil {
 		return fmt.Errorf("unable to query flow config from catalog: %w", err)
+	}
+
+	sourcePeerType, err := getPeerType(ctx, config.SourceName)
+	if err != nil {
+		return err
+	}
+	destinationPeerType, err := getPeerType(ctx, config.DestinationName)
+	if err != nil {
+		return err
 	}
 
 	var res *protos.GetDefaultPartitionKeyForTablesOutput


docker-compose.yml
```diff
@@ -142,7 +150,10 @@ services:
 
   flow-worker:
     container_name: flow-worker
-    image: ghcr.io/peerdb-io/flow-worker:stable-v0.32.1
+    build:
+      context: .
+      dockerfile: stacks/flow.Dockerfile
+      target: flow-worker
```
iamKunalGupta: instead of making changes here, can simply use `docker-compose-dev.yaml`

alexstoick: yep, reverted these!

flow/workflows/setup_flow.go
```diff
@@ -121,7 +120,7 @@ func (s *SetupFlowExecution) ensurePullability(
 	ensurePullabilityInput := &protos.EnsurePullabilityBatchInput{
 		PeerName:               config.SourceName,
 		FlowJobName:            s.cdcFlowName,
-		SourceTableIdentifiers: slices.Sorted(maps.Keys(s.tableNameMapping)),
+		SourceTableIdentifiers: []string{},
```
serprex, line 123:
```suggestion
		SourceTableIdentifiers: nil,
```

serprex (approved):

flow/workflows/cdc_flow.go
```diff
@@ -157,34 +137,19 @@ func processCDCFlowConfigUpdate(
...
-	if flowConfigUpdate.SnapshotNumPartitionsOverride > 0 {
-		state.SnapshotNumPartitionsOverride = flowConfigUpdate.SnapshotNumPartitionsOverride
-	}
-	if flowConfigUpdate.SnapshotMaxParallelWorkers > 0 {
-		state.SnapshotMaxParallelWorkers = flowConfigUpdate.SnapshotMaxParallelWorkers
-	}
-	if flowConfigUpdate.SnapshotNumTablesInParallel > 0 {
-		state.SnapshotNumTablesInParallel = flowConfigUpdate.SnapshotNumTablesInParallel
-	}
+	cfg = syncStateToConfigProtoInCatalog(ctx, cfg, state.FlowConfigUpdate)
```
serprex, line 138: this `> 0` behavior was added recently in #3435, this probably regresses

flow/workflows/cdc_flow.go
```diff
@@ -289,22 +248,13 @@ func processTableAdditions(
 			additionalTablesUUID := GetUUID(ctx)
 			childAdditionalTablesCDCFlowID := GetChildWorkflowID("additional-cdc-flow", cfg.FlowJobName, additionalTablesUUID)
 			additionalTablesCfg := proto.CloneOf(cfg)
+			// TODOAS: should the `doInitialSnapshot` just come from cfg?
```
serprex: can remove TODO, this question belongs in another PR

flow/workflows/cdc_flow.go
```diff
@@ -330,13 +281,16 @@ func processTableAdditions(
 	})
 
 	// additional tables should also be resynced, we don't know how much was done so far
-	state.SyncFlowOptions.TableMappings = append(state.SyncFlowOptions.TableMappings, flowConfigUpdate.AdditionalTables...)
+	// state.SyncFlowOptions.TableMappings = append(state.SyncFlowOptions.TableMappings, flowConfigUpdate.AdditionalTables...)
```
serprex: if this was handled above we can remove this

flow/workflows/cdc_flow.go
```diff
@@ -482,12 +431,17 @@ func addCdcPropertiesSignalListener(
 
 func CDCFlowWorkflow(
 	ctx workflow.Context,
-	cfg *protos.FlowConnectionConfigs,
+	flowJobName string,
+	// cfg *protos.FlowConnectionConfigs,
```
serprex: this breaks backwards compatibility. Should still take config & pull name from that to maintain function signature

flow/workflows/cdc_flow.go
```diff
@@ -608,20 +561,17 @@ func CDCFlowWorkflow(
...
 		// a suffix to the table names.
 		if cfg.Resync {
-			for _, mapping := range state.SyncFlowOptions.TableMappings {
-				if mapping.Engine != protos.TableEngine_CH_ENGINE_NULL {
-					mapping.DestinationTableIdentifier += "_resync"
-				}
-			}
-			// because we have renamed the tables.
-			cfg.TableMappings = state.SyncFlowOptions.TableMappings
+			return nil, errors.New("cannot start CDCFlow with Resync enabled, please drop the flow and start again")
```
serprex, line 572: TODO

flow/workflows/cdc_flow.go
alexstoick: The problem with keeing the `protos.FlowConnectionConfig` here, is that in cases where this is too large - it will still fail.

Is the suggestion to pass **both** and then remove the `cfg` from here?

serprex: no, only pass cfg. new code should pass an empty cfg with only FlowJobName present

alexstoick: OK understood, I'll go in this direction then :)

flow/workflows/cdc_flow.go
alexstoick: given that we no longer have or rely on state, this should be a non-issue.

flow/internal/flow_configuration_helpers.go
```diff
@@ -0,0 +1,46 @@
...
+	for _, v := range tableMappings {
+		tblNameMapping[v.SourceTableIdentifier] = v.DestinationTableIdentifier
+	}
+	return tblNameMapping
+}
+
+func FetchConfigFromDB(flowName string) (*protos.FlowConnectionConfigs, error) {
+	var configBytes sql.RawBytes
+	dbCtx := context.Background()
+	pool, _ := GetCatalogConnectionPoolFromEnv(dbCtx)
```
serprex: should take context as param

flow/workflows/setup_flow.go
```diff
@@ -227,9 +224,14 @@ func (s *SetupFlowExecution) setupNormalizedTables(
 // executeSetupFlow executes the setup flow.
 func (s *SetupFlowExecution) executeSetupFlow(
 	ctx workflow.Context,
-	config *protos.FlowConnectionConfigs,
+	flowJobName string,
 ) (*protos.SetupFlowOutput, error) {
 	s.Info("executing setup flow")
+	// gotta fetch the config from the catalog.
+	config, err := internal.FetchConfigFromDB(flowJobName)
```
serprex, line 231: can't call this in workflows

could maybe run as side effect, but then it ends up in history

alexstoick: Why can this not be called in workflows? 🤔

PR #3410, opened 2025-08-29
mongo: add pool error handling

jgao54:


flow/shared/exceptions/mongo.go
```diff
@@ -0,0 +1,6 @@
+package exceptions
+
+type MongoRetryablePoolError interface {
```
serprex: fine to move this to classifier.go

flow/shared/exceptions/mongo.go
```diff
@@ -0,0 +1,6 @@
...
+	error
+	Retryable() bool
```
serprex: unfortunately this will catch all errors that implement this rather genericly named method

fix is to do one of:
1. make this something we'll generally accept, but then don't tag as mongo
2. use reflect to make sure the error we get is from mongo type (`reflect.TypeOf(err).PkgPath()`)

could combine 1/2, using reflect to properly classify type (mongo if mongo package, otherwise OTHER, with code set to name of type)

jgao54: for simplicity, i like 1; for monitoring, i like 2. will do 1/2 as you suggested

PR #3418, opened 2025-09-02
CDC-Only table addition support exposed via Flow API

Amogh-Bharadwaj:
# Why
- Currently, table additions in mirrors without initial load are having their initial loads be done. We have received feedback that this is unideal and comes across as a bug. This lays the foundation for exposing an opt-in/out flag in PeerDB client software for initial load for added tables.
- In multi-tenant source data store setups, when migrating from one PeerDB instance to another, it is often desirable to migrate one tenant at a time, using the same replication slot. 

# What
- Adds a field to the FlowConfigUpdate object (which is the request for table editing) to skip initial load for table additions. This way this PR is backwards-compatible versus `doInitialSnapshotForAddedTables`.
- `doInitialSnapshot` will be set to the opposite of above value for table addition workflow
- Adds an API test for this feature.

I plan to expose this via PeerDB UI as a follow-up

### DIFF
--- a/flow/e2e/api/api_test.go
+++ b/flow/e2e/api/api_test.go
@@ -1273,7 +1273,6 @@ func (s Suite) TestTableAdditionWithoutInitialLoad() {
 	e2e.RequireEmptyDestinationTable(s.ch, "added", cols)
 
 	// cdc should occur for added table still, so insert row into added table to test cdc
-
 	switch s.source.(type) {
 	case *e2e.PostgresSource, *e2e.MySqlSource:
 		require.NoError(s.t, s.source.Exec(s.t.Context(),


flow/workflows/cdc_flow.go
```diff
@@ -281,7 +281,7 @@ func processTableAdditions(
 			additionalTablesUUID := GetUUID(ctx)
 			childAdditionalTablesCDCFlowID := GetChildWorkflowID("additional-cdc-flow", cfg.FlowJobName, additionalTablesUUID)
 			additionalTablesCfg := proto.CloneOf(cfg)
-			additionalTablesCfg.DoInitialSnapshot = true
+			additionalTablesCfg.DoInitialSnapshot = !flowConfigUpdate.SkipInitialSnapshotForTableAdditions
 			additionalTablesCfg.InitialSnapshotOnly = true
```
Amogh-Bharadwaj, line 293: This seems odd and hacky but these settings is what allows us currently to skip initial load and also skip slot creation/snapshot connection. See [here](https://github.com/PeerDB-io/peerdb/pull/3418/files#diff-da30b1dd687ec7ecf48644d30c60eab1f2c0103c739aac30593ad91a35bb91afR366-R378)

flow/workflows/snapshot_flow.go
```diff
@@ -363,7 +363,7 @@ func SnapshotFlowWorkflow(
 	}
 	defer workflow.CompleteSession(sessionCtx)
 
-	if !config.DoInitialSnapshot {
+	if !config.DoInitialSnapshot && !config.InitialSnapshotOnly {
```
Amogh-Bharadwaj, line 366: [Added a validation guard](https://github.com/PeerDB-io/peerdb/pull/3418/files#diff-8579cdc263380818be5a29192c58ea7e4a64f7ce544c24d35483ef5542e416d0R53-R56) so that only table addition can set DoInitialSnapshot to false and InitialSnapshotOnly to true

flow/e2e/clickhouse/clickhouse.go
```diff
@@ -365,6 +365,16 @@ func (s ClickHouseSuite) queryRawTable(conn clickhouse.Conn, table string, cols
 	)
 }
 
+func (s ClickHouseSuite) RunInsertIntoExistingDestinationTable(tableName string, insertCommand string) error {
```
serprex: name could be more generic, since this can run arbitrary queries

flow/workflows/cdc_flow.go
serprex: this works since it maintains backwards compat with protos, but would this've been clearer if DoInitialSnapshot/InitialSnapshotOnly were a single enum?

flow/e2e/api/api_test.go
```diff
@@ -886,3 +886,96 @@ func (s Suite) TestQRep() {
...
+			e2e.AttachSchema(s, "original"),
+		})
+		if err != nil {
+			return false
+		}
+
+		return valid && env.GetFlowStatus(s.t) == protos.FlowStatus_STATUS_RUNNING
+	})
+
+	// No initial load should have happened for added table, so only the pre-existing row should be there
```
serprex: really we want to test:
1. adding row to source beforehand isn't replicated
2. adding row to source afterwards is replicated

flow/e2e/clickhouse/clickhouse.go
Amogh-Bharadwaj: Removed

flow/workflows/cdc_flow.go
Amogh-Bharadwaj: Yup it would actually be clearer

flow/cmd/validate_mirror.go
```diff
@@ -50,6 +50,10 @@ func (h *FlowRequestHandler) ValidateCDCMirror(
 		return nil, errors.New("connection configs is nil")
 	}
 
+	if !req.ConnectionConfigs.DoInitialSnapshot && req.ConnectionConfigs.InitialSnapshotOnly {
+		return nil, errors.New("invalid config: initial_snapshot_only is true but do_initial_snapshot is false")
```
iamKunalGupta:
```suggestion
		return nil, exceptions.NewInvalidArgumentApiError(errors.New("invalid config: initial_snapshot_only is true but do_initial_snapshot is false"))
```

All errors from API now must be wrapped in `exceptions.New.*ApiError`. The test failing below will also let you know

iamKunalGupta: Updated test wording to make more sense https://github.com/PeerDB-io/peerdb/pull/3462

PR #3419, opened 2025-09-02
mysql: try handle strings destined for integers/bools/floats

serprex:
this has been coming up somehow, usually needing us to suggest resync

embrace mysql culture & try handle shoving square pegs in round holes

### DIFF
--- a/flow/connectors/mysql/qvalue_convert.go
+++ b/flow/connectors/mysql/qvalue_convert.go
@@ -3,6 +3,7 @@ package connmysql
 import (
 	"encoding/binary"
 	"fmt"
+	"log/slog"
 	"math"
 	"math/bits"
 	"slices"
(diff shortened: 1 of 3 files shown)


flow/connectors/mysql/qvalue_convert.go
```diff
@@ -492,6 +507,39 @@ func QValueFromMysqlRowEvent(
 				return nil, err
 			}
 			return types.QValueTimestamp{Val: tm.UTC()}, nil
+		case types.QValueKindBoolean:
+			// integer types shouldn't get here, but try work with schema changes
+			return types.QValueBoolean{Val: strings.EqualFold(val, "true") || strings.EqualFold(val, "t") || strings.EqualFold(val, "on") || strings.EqualFold(val, "yes") || strings.EqualFold(val, "1")}, nil
+		case types.QValueKindInt8:
+			v, _ := strconv.ParseInt(val, 10, 8)
```
ilidemi: Up to you, but I think a customer warning would be appropriate for these

serprex: added logging. in future these can be customer facing alongside skew loss warnings

PR #3425, opened 2025-09-04
nullable partition key should fail validation

jgao54:


jgao54:
Discussed with @serprex  -- we want to support cases were column is nullable but does not contain null columns (i.e. DEFAULT NOW()), so this is a bit restricting.

PR #3430, opened 2025-09-05
update flow status metrics when status is updated

jgao54:
Updated PR to emit metrics whenever status is updated. Moved `UpdateFlowStatusInCatalogActivity` to `FlowableActivity` which contains the OtelManager for emitting metrics. This avoids instantiating and passing around OtelManager to all workflows.

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -1416,7 +1416,7 @@ func (a *FlowableActivity) PeerDBFullRefreshOverwriteMode(ctx context.Context, e
 	return internal.PeerDBFullRefreshOverwriteMode(ctx, env)
 }
 
-func (a *FlowableActivity) UpdateFlowStatusInCatalogActivity(
+func (a *FlowableActivity) UpdateFlowStatusInCatalogLocalActivity(
 	ctx context.Context,
 	workflowID string,
 	status protos.FlowStatus,
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -73,7 +73,7 @@ func syncStatusToCatalog(ctx workflow.Context, logger log.Logger, status protos.
 	})
 
 	updateFuture := workflow.ExecuteLocalActivity(updateCtx,
-		flowable.UpdateFlowStatusInCatalogActivity, workflow.GetInfo(ctx).WorkflowExecution.ID, status)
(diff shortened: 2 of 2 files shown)


flow/activities/flowable.go
```diff
@@ -750,13 +750,12 @@ func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
...
-	var allFlows atomic.Pointer[[]flowInformation]
-	defer shared.Interval(ctx, 59*time.Second, func() {
+
+	queryFlowsInfo := func() ([]flowInformation, error) {
 		rows, err := a.CatalogPool.Query(ctx,
 			"SELECT DISTINCT ON (name) name, config_proto, workflow_id, updated_at FROM flows WHERE query_string IS NULL")
 		if err != nil {
-			logger.Error("failed to query all flows", slog.Any("error", err))
-			return
+			return []flowInformation{}, err
```
```suggestion
			return nil, err
```

flow/activities/flowable.go
```diff
@@ -779,8 +778,26 @@ func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
 				updatedAt:  updatedAt,
 			}, nil
 		})
+		return infos, err
+	}
+
+	// final metrics recording before workflow exits (but don't sweat it if it fails)
+	defer func() {
+		infos, _ := queryFlowsInfo()
+		if infos == nil {
```
serprex: should check err, on error slog it & skip this

flow/activities/flowable.go
```diff
@@ -779,8 +778,26 @@ func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
...
+			return
+		}
+		finalCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		_ = a.RecordMetrics(finalCtx, infos)
+		_ = a.RecordSlotSizes(finalCtx, infos)
```
```suggestion
		_ = a.RecordMetrics(finalCtx, infos)
```
don't think slot size info is that urgent

ilidemi (approved):

jgao54:
@serprex DMed context, ended up going with a different approach because flow status is only updated in worker while the scheduled metrics reporting is done in the api server, so my original implementation of emitting metric in api server upon its exit wouldn't help much.

PR #3434, opened 2025-09-05
feat(logging): add request id middleware for logging

iamKunalGupta:


flow/middleware/tracing.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func RequestIdMiddleware() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
+		md, _ := metadata.FromIncomingContext(ctx)
+		requestIds := md.Get(shared.RequestIdKey.String())
+		requestId := uuid.NewString()
+		if requestIds != nil {
+			requestId = requestIds[0]
+		}
```
serprex:
```suggestion
		var requestId string
		if len(requestIds) > 1 {
			requestId = requestIds[0]
		} else {
			requestId = uuid.NewString()
		}
```
avoids panic on empty slice, avoids generating random uuid when requestIds not empty

iamKunalGupta:
```
		if len(requestIds) > 1 {
```
check should be for 0, but yes, will fix

serprex (approved):

PR #3436, opened 2025-09-06
reduce channel sizes to 1024

serprex:
based on https://github.com/PeerDB-io/ab-scale-testing there was no benefit

this optimization was made for a particular usecase before other optimizations were made

these channel sizes become particularly painful when multiple mirrors are spun up


ilidemi:
What were the parallelism configs and the hardware used? We know of one heavy scaling user with 2 pipes * 25 workers and 12x node size.
What was the throughput?
Were smaller row sizes tested? Blog says 230 bytes, I've seen 100 bytes when looking at logs, no evidence that that's the lowest we get in existing workloads.
What about CDC with 50ish pipes like some customers also have?
A single number with any result?

We have such good benchmarking culture in the company ([private repo example](https://c.house/I5rgxEZLnS)), this is very different from that.

PR #3437, opened 2025-09-06
Support very long numbers as strings in PG JSON

ilidemi:
Python transparently makes ints bigints, serializes to JSON as is, PG accepts that JSON, then we and CH can't consume that JSON. Go can produce such JSON too if you put a big.Int in there

The software that hit this is https://github.com/JudgmentLabs/judgeval and we're The leading database for AI

Add a custom decoder so that we replace what we can't parse into float64 with strings. This is backcompatible because only the values that were previously failing would see the new behavior and succeed

### DIFF
--- a/flow/connectors/mongo/codec.go
+++ b/flow/connectors/mongo/codec.go
@@ -16,31 +16,7 @@ import (
 // by converting them to JSON strings. Decoders are implemented to satisfy json-iterator's
 // extension interface requirements, even though decoding is not used by MongoDB Connector.
 type BsonExtension struct {
-	jsoniter.Extension
-}
-
-func (extension *BsonExtension) UpdateStructDescriptor(structDescriptor *jsoniter.StructDescriptor) {
-	// not used (bson documents does not contain struct)
-}
-
-func (extension *BsonExtension) CreateEncoder(typ reflect2.Type) jsoniter.ValEncoder {
-	// use default
-	return nil
-}
-
-func (extension *BsonExtension) CreateDecoder(typ reflect2.Type) jsoniter.ValDecoder {
-	// use default
-	return nil
-}
-
-func (extension *BsonExtension) CreateMapKeyEncoder(typ reflect2.Type) jsoniter.ValEncoder {
-	// use default
-	return nil
-}
-
-func (extension *BsonExtension) CreateMapKeyDecoder(typ reflect2.Type) jsoniter.ValDecoder {
-	// use default
-	return nil
+	jsoniter.DummyExtension
 }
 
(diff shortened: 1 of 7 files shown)


ilidemi:
MySQL is not a concern: `SQL Error [3140] [22001]: Data truncation: Invalid JSON text: "Number too big to be stored in double." at position 9 in value for column 'json_long_num.j'.`

jgao54 (approved):
nice! 🚢

flow/connectors/mongo/codec.go
```diff
@@ -16,31 +16,7 @@ import (
...
-
-func (extension *BsonExtension) CreateMapKeyEncoder(typ reflect2.Type) jsoniter.ValEncoder {
-	// use default
-	return nil
-}
-
-func (extension *BsonExtension) CreateMapKeyDecoder(typ reflect2.Type) jsoniter.ValDecoder {
-	// use default
-	return nil
+	jsoniter.DummyExtension
```
jgao54, line 19: 👍

flow/connectors/postgres/cdc.go
```diff
@@ -306,7 +311,49 @@ func (p *PostgresCDCSource) decodeColumnData(
...
+			}
+		} else if formatCode != pgtype.TextFormatCode {
+			return nil, fmt.Errorf("unknown format code for json: %d", formatCode)
+		}
+
+		if err := p.jsonApi.Unmarshal(jsonData, &parsedData); err != nil {
+			p.logger.Error("[pg_cdc] failed to unmarshal json", slog.Any("error", err))
+			return nil, fmt.Errorf("failed to unmarshal json: %w", err)
+		}
+		return p.parseFieldFromPostgresOID(dataType, typmod, parsedData, customTypeMapping, p.internalVersion)
```
jgao54: nit: using the same approach as how JSONArrayOID/JSONBArrayOID are handled,  could this be simplified to 
```
	if dataType == pgtype.JSONOID || dataType == pgtype.JSONBOID {
		var text pgtype.Text
		if err := p.typeMap.Scan(dataType, formatCode, data, text); err != nil {
			return nil, fmt.Errorf("error decoding binary column data: %w", err)
		}
		if err := p.jsonApi.UnmarshalFromString(text.String, &parsedData); err != nil {
			p.logger.Error("[pg_cdc] failed to unmarshal json", slog.Any("error", err))
			return nil, fmt.Errorf("failed to unmarshal json: %w", err)
		}
		return p.parseFieldFromPostgresOID(dataType, typmod, parsedData, customTypeMapping, p.internalVersion)
```
which avoid needing to know about encodings (albeit it being a very simple one)

PR #3446, opened 2025-09-09
Minor: improve CDC logging

ilidemi:
A neon peer was pulling records for longer than the batch timeout but unclear why
Also their batch time was 3 hours so hard to browse through the logs

### DIFF
--- a/flow/connectors/mongo/cdc.go
+++ b/flow/connectors/mongo/cdc.go
@@ -176,6 +176,7 @@ func (c *MongoConnector) PullRecords(
 	defer changeStream.Close(ctx)
 
 	var recordCount uint32
+	pullStart := time.Now()
 	defer func() {
 		if recordCount == 0 {
 			req.RecordStream.SignalAsEmpty()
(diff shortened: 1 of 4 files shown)


flow/connectors/mongo/cdc.go
```diff
@@ -176,14 +176,16 @@ func (c *MongoConnector) PullRecords(
...
 	defer func() {
 		if recordCount == 0 {
 			req.RecordStream.SignalAsEmpty()
 		}
 		c.logger.Info("[mongo] PullRecords finished streaming",
 			slog.Uint64("records", uint64(recordCount)),
 			slog.Int64("bytes", c.totalBytesRead.Load()),
-			slog.Int("channelLen", req.RecordStream.ChannelLen()))
+			slog.Int("channelLen", req.RecordStream.ChannelLen()),
+			slog.Float64("elapsedMinutes", time.Since(pullStart).Minutes()))
```
serprex, line 188:
```suggestion
			slog.Duration("elapsed", time.Since(pullStart)))
```
can be applied throughout

ilidemi:
```
logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
logger.Info("processed items",
	slog.Int("size", 23),
	slog.Duration("duration", time.Hour+time.Minute),
)
```
`{"time":"2009-11-10T23:00:00Z","level":"INFO","msg":"processed items","size":23,"duration":3660000000000}`
I don't like the 0000000000

ilidemi: In the other batching PR the durations are actually appropriate because in that logger the default duration unit is milliseconds and we want milliseconds for tables. Here we wouldn't set minutes globally

serprex (approved):

PR #3451, opened 2025-09-09
ClickHouse peer: Add flag to disable validation of ClickHouse setup

Amogh-Bharadwaj:
### Why
There are some scenarios where ClickHouse connector objects would be needed to be created and where validation of ClickHouse permissions would be an unnecessary overhead.

### What
This PR adds a flag - `disable_clickhouse_validation` - which is wired in the `validateCheck` method of the connector.
Note that S3 bucket validation is still being performed there.


Amogh-Bharadwaj:
Closed in view of https://github.com/PeerDB-io/peerdb/pull/3452

PR #3453, opened 2025-09-09
Downgrade min required mongo version to 4.4

IvanTakarlikov-st:
I don't see any reason so far to require Mongo higher than version 5.1

### DIFF
--- a/flow/shared/mongo/validation.go
+++ b/flow/shared/mongo/validation.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	MinSupportedVersion    = "5.1.0"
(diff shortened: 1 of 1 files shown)


jgao54 (approved):
Thanks @IvanTakarlikov-st 

Versions below 5.1 have not been extensively tested, but happy to open it up for your use case.  Note that perf optimizations to Change Streams API were made for 5.1+ and older versions may experience slower change stream processing.

IvanTakarlikov-st:
We will test and let you know how it works for us and if it will have issues! 
Waiting to get it merged. 
I see that CI failed, but not sure if it's related to our code changes.

PR #3454, opened 2025-09-09
refactor(flow-api): use proper error codes everywhere

iamKunalGupta:
1. Allows for proper error logging by middleware and allowing for alerting on grpc metrics for internal error status codes
2. Add a test for this for any future modifications which will fail if proper error code is not returned
3. Also fixes logging of request ID in flow-api by replacing `slog\.(Debug|Info|Warn|Error)` -> `slog.$1Context(ctx, `
4. Add test to suggest replacing the generic slog methods with the Context based methods


flow/cmd/validate_mirror.go
```diff
@@ -61,24 +64,25 @@ func (h *FlowRequestHandler) ValidateCDCMirror(
...
+			fmt.Sprintf("failed to validate source connector %s: %v", req.ConnectionConfigs.SourceName, err))
 	}
 
 	dstConn, err := connectors.GetByNameAs[connectors.MirrorDestinationValidationConnector](
 		ctx, req.ConnectionConfigs.Env, h.pool, req.ConnectionConfigs.DestinationName,
 	)
 	if err != nil {
 		if errors.Is(err, errors.ErrUnsupported) {
-			return &protos.ValidateCDCMirrorResponse{}, nil
+			return nil, exceptions.NewUnimplementedApiError("connector is not a supported destination type")
```
iamKunalGupta: @serprex this is causing tests like ES to fail

iamKunalGupta: I am reverting the change specifically for this place, but if we don't care about this, we should remove the check if possible

flow/shared/exceptions/constructor_test.go
```diff
@@ -33,11 +33,14 @@ func TestErrorConstructorsShouldReturnPointers(t *testing.T) {
 						returnType := fn.Type.Results.List[0].Type
 						// If it's not a pointer (*Ident), it's a violation
 						if _, isPointer := returnType.(*ast.StarExpr); !isPointer {
-							pos := pkg.Fset.Position(fn.Pos())
-							assert.Fail(
-								t, "Error constructor should return pointer",
-								"%s:%s should return *%s",
-								pos.Filename, fn.Name.Name, strings.TrimPrefix(fn.Name.Name, "New"))
+							// Although we are ok if the return type is error
+							if ident, isIdent := returnType.(*ast.Ident); !(isIdent && ident.Name == "error") {
```
iamKunalGupta, line 37: @ilidemi added a case for `func New.*Error` returning `error`, hopefully the tests will pass

flow/cmd/handler.go
```diff
@@ -151,12 +153,12 @@ func (h *FlowRequestHandler) CreateCDCFlow(
...
-		slog.Error("unable to create flow job entry", slog.Any("error", err))
-		return nil, fmt.Errorf("unable to create flow job entry: %w", err)
+		slog.ErrorContext(ctx, "unable to create flow job entry", slog.Any("error", err))
+		return nil, exceptions.NewInternalApiError(fmt.Sprintf("unable to create flow job entry: %v", err))
 	}
 
 	if _, err := h.temporalClient.ExecuteWorkflow(ctx, workflowOptions, peerflow.CDCFlowWorkflow, cfg, nil); err != nil {
-		slog.Error("unable to start PeerFlow workflow", slog.Any("error", err))
+		slog.ErrorContext(ctx, "unable to start PeerFlow workflow", slog.Any("error", err))
 		return nil, fmt.Errorf("unable to start PeerFlow workflow: %w", err)
```
ilidemi: Why are some errors converted but not the others? There are a few more instances in this file

iamKunalGupta: I mostly tried to only touch top level functions that are returning response
I might have missed out this one

In some cases (like get tables/publications etc), I did not do it either, but I think maybe an err check with Failed Precondition should be good

flow/shared/exceptions/api.go
```diff
@@ -0,0 +1,93 @@
+package exceptions
```
serprex, line 1: this could include wrapping/formatting so existing code using fmt.Errorf can mostly stay the same

flow/cmd/handler.go
ilidemi: Ah right, didn't notice the public-private distinction. I think for the top level ones it's important to do it consistently so the line is clear.

Helpful Claude:
```
  Files with Unwrapped Errors:
  2. alerts.go: PostAlertConfig method - 1 unwrapped database error
  4. mirror_status.go:
    - ListMirrors method - 1 unwrapped error
    - CDCBatches method - 1 unwrapped error
  5. validate_mirror.go: CheckIfMirrorNameExists method - 1 unwrapped fmt.Errorf
  8. handler.go: CreateCDCFlow method - 2 unwrapped fmt.Errorf returns
```

ilidemi: Also one in FlowStateChange
yeah, this needs to be thorough

flow/cmd/handler.go
```diff
@@ -470,6 +476,9 @@ func (h *FlowRequestHandler) getWorkflowID(ctx context.Context, flowJobName stri
 	q := "SELECT workflow_id FROM flows WHERE name = $1"
 	var workflowID string
 	if err := h.pool.QueryRow(ctx, q, flowJobName).Scan(&workflowID); err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return "", exceptions.NewNotFoundApiError(fmt.Sprintf("flow job %s not found", flowJobName))
```
ilidemi: but then this one breaks the pattern with public/private. Maybe something like ErrNotFound and construct the message in the caller?

ilidemi (commented):
What would be the plan for finishing the rest of the logs where context is applicable? Would be annoying to have it half-done and idk if anyone volunteers just the less-pleasant cleanup

flow/shared/exceptions/api.go
iamKunalGupta: I was thinking the same, but we would end up calling .Error() anyways, but I think we can hide that

flow/middleware/logging.go
```diff
@@ -4,58 +4,62 @@ import (
...
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/reflect/protoregistry"
 
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
 	"github.com/PeerDB-io/peerdb/flow/internal"
 )
 
-const grpcFullServiceName = "peerdb_route.FlowService"
+var GrpcFullServiceName = protos.FlowService_ServiceDesc.ServiceName
```
iamKunalGupta, line 19: @jgao54 I made a small change to refer to the string from the generated grpc code

iamKunalGupta:
> What would be the plan for finishing the rest of the logs where context is applicable? Would be annoying to have it half-done and idk if anyone volunteers just the less-pleasant cleanup

Taken care of these as well

PR #3459, opened 2025-09-10
Allow to connect to Mongo without readConcern: majority

ST-eugenekrivdyuk:
Hello there!

I stumbled upon inability to mirror our MongoDB database due to the following error:
```
[qrep] failed to pull records: failed to query for records: (ReadConcernMajorityNotEnabled) Majority read concern is not enabled.
```

While I agree that it's a good default, it basically makes it impossible to work for certain MongoDB installations. In our case it's PSA (Primary, Secondayr, Arbiter) where setting read concern to majority is not always good.

Hence I propose to add an option to allow disabling majority read concern.

I am very new to the project - so it's likely that I've missed some things - apologies if that's the case 🙏


protos/peers.proto
```diff
@@ -77,11 +77,12 @@ message MongoConfig {
...
-  string tls_host = 5;
-  optional string root_ca = 6 [(peerdb_redacted) = true];
-  ReadPreference read_preference = 7;
-  optional SSHConfig ssh_config = 8;
+  bool require_read_concern_majority = 4;
+  bool disable_tls = 5;
+  string tls_host = 6;
+  optional string root_ca = 7 [(peerdb_redacted) = true];
+  ReadPreference read_preference = 8;
+  optional SSHConfig ssh_config = 9;
```
ST-eugenekrivdyuk: Oops, I should put the new one at the end.

serprex: option should also be made so false will maintain existing behavior in order to be backwards compatible

jgao54:
Thanks for the update @ST-eugenekrivdyuk. Seems like this is a MongoDB changestream limitation unfortunately.  Will require enabling the server side config replication.enableMajorityReadconcern in order to be supported.

ST-eugenekrivdyuk:
Closing as it doesn't work with our setup.

PR #3461, opened 2025-09-11
Fail on invalid snapshot

serprex:
this is made a bit complicated since temporal activities don't pick up on wrapped non recoverable errors

by default temporal workflows don't retry. we should not be using temporal retries with workflows. retryable errors should be handled by having activity retry

dropping failed mirrors wasn't working since it'd try signalling completed workflow, needed small fix in request handler

I don't know why CloseSlotKeepAlive was failing to schedule after restart. Somehow queued to a nonsense UUID task queue name

### DIFF
--- a/flow/workflows/snapshot_flow.go
+++ b/flow/workflows/snapshot_flow.go
@@ -86,8 +86,7 @@ func (s *SnapshotFlowExecution) closeSlotKeepAlive(
 	s.logger.Info("closing slot keep alive for peer flow")
 
 	ctx = workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
-		ScheduleToStartTimeout: time.Minute,
-		StartToCloseTimeout:    10 * time.Minute,
+		StartToCloseTimeout: 10 * time.Minute,
 		RetryPolicy: &temporal.RetryPolicy{
 			InitialInterval: 1 * time.Minute,
 		},
@@ -313,7 +312,9 @@ func (s *SnapshotFlowExecution) cloneTablesWithSlot(
 		return fmt.Errorf("failed to setup replication: %w", err)
 	}
 	defer func() {
-		if err := s.closeSlotKeepAlive(sessionCtx); err != nil {
+		dCtx, cancel := workflow.NewDisconnectedContext(sessionCtx)
+		defer cancel()
+		if err := s.closeSlotKeepAlive(dCtx); err != nil {
 			s.logger.Error("failed to close slot keep alive", slog.Any("error", err))
 		}
 	}()


flow/cmd/handler.go
```diff
@@ -373,7 +373,7 @@ func (h *FlowRequestHandler) FlowStateChange(
 			}
 		case protos.FlowStatus_STATUS_TERMINATING, protos.FlowStatus_STATUS_TERMINATED:
 			if currState != protos.FlowStatus_STATUS_TERMINATED && currState != protos.FlowStatus_STATUS_TERMINATING {
-				if currState == protos.FlowStatus_STATUS_COMPLETED {
+				if currState == protos.FlowStatus_STATUS_COMPLETED || currState == protos.FlowStatus_STATUS_FAILED {
```
iamKunalGupta, line 376: In shutdownFlow we would also need to handle cancellation of missing workflow itself if possible

flow/shared/err_types.go
```diff
@@ -31,3 +36,13 @@ func SkipSendingToIncidentIo(errTags []string) bool {
 }
 
 type QRepWarnings []error
+
+func WrapError(s string, err error) error {
```
iamKunalGupta, line 40:
```suggestion
func WrapAsTemporalError(s string, err error) error {
```

flow/shared/err_types.go
```diff
@@ -1,10 +1,15 @@
...
+	"fmt"
+
+	"go.temporal.io/sdk/temporal"
+)
 
 var (
-	ErrSlotAlreadyExists error = errors.New("slot already exists")
-	ErrTableDoesNotExist error = errors.New("table does not exist")
+	ErrSlotAlreadyExists error = temporal.NewNonRetryableApplicationError("slot already exists", "snapshot", nil)
+	ErrTableDoesNotExist error = temporal.NewNonRetryableApplicationError("table does not exist", "snapshot", nil)
```
iamKunalGupta, line 12: Might be a good idea to re-use `exceptions/application.go` and add `irrecoverable_invalid_snapshot` there

flow/workflows/cdc_flow.go
```diff
@@ -702,11 +702,9 @@ func CDCFlowWorkflow(
...
 		taskQueue := internal.PeerFlowTaskQueueName(shared.SnapshotFlowTaskQueue)
 		childSnapshotFlowOpts := workflow.ChildWorkflowOptions{
-			WorkflowID:        snapshotFlowID,
-			ParentClosePolicy: enums.PARENT_CLOSE_POLICY_REQUEST_CANCEL,
-			RetryPolicy: &temporal.RetryPolicy{
-				MaximumAttempts: 20,
-			},
+			WorkflowID:            snapshotFlowID,
+			ParentClosePolicy:     enums.PARENT_CLOSE_POLICY_REQUEST_CANCEL,
+			RetryPolicy:           &temporal.RetryPolicy{MaximumAttempts: 1},
```
iamKunalGupta, line 707: Would this affect any retryable errors like connection lost with the worker or worker restarts?

flow/workflows/snapshot_flow.go
```diff
@@ -312,9 +313,7 @@ func (s *SnapshotFlowExecution) cloneTablesWithSlot(
 		return fmt.Errorf("failed to setup replication: %w", err)
 	}
 	defer func() {
-		dCtx, cancel := workflow.NewDisconnectedContext(sessionCtx)
-		defer cancel()
-		if err := s.closeSlotKeepAlive(dCtx); err != nil {
+		if err := s.closeSlotKeepAlive(sessionCtx); err != nil {
```
iamKunalGupta: would we handle missing slots, locks etc without affecting failure of wf or is it ok?

flow/shared/err_types.go
serprex: this is designed so it won't necessarily wrap as temporal error unless it's already wrapping a temporal error

flow/workflows/cdc_flow.go
serprex: worker restarts/crashes are invisible to workflow code

flow/workflows/snapshot_flow.go
serprex: I did this while testing, but think it can be reverted for this PR

flow/cmd/handler.go
serprex: `shutdownFlow` works since it spins up new workflow for drop logic, whereas other code path wants to transform existing workflow into drop flow

flow/shared/err_types.go
serprex: that may come with time, but for now going to have the cleaner diff

PR #3466, opened 2025-09-11
report flow status metric at start of drop flow

serprex:
avoids race where drop would happen before RecordMetrics could move flow out of running

followup to #3465

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -1420,7 +1420,7 @@ func (a *FlowableActivity) PeerDBFullRefreshOverwriteMode(ctx context.Context, e
 	return internal.PeerDBFullRefreshOverwriteMode(ctx, env)
 }
 
-func (a *FlowableActivity) ReportStatusMetric(ctx context.Context, flowJobName string, status protos.FlowStatus) error {
+func (a *FlowableActivity) ReportStatusMetric(ctx context.Context, status protos.FlowStatus) error {
 	_, isActive := activeFlowStatuses[status]
 	a.OtelManager.Metrics.FlowStatusGauge.Record(ctx, 1, metric.WithAttributeSet(attribute.NewSet(
 		attribute.String(otel_metrics.FlowStatusKey, status.String()),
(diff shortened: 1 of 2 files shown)


flow/activities/flowable.go
```diff
@@ -1419,3 +1419,12 @@ func (a *FlowableActivity) UpdateCDCConfigInCatalogActivity(ctx context.Context,
 func (a *FlowableActivity) PeerDBFullRefreshOverwriteMode(ctx context.Context, env map[string]string) (bool, error) {
 	return internal.PeerDBFullRefreshOverwriteMode(ctx, env)
 }
+
+func (a *FlowableActivity) ReportStatusMetric(ctx context.Context, flowJobName string, status protos.FlowStatus) error {
+	_, isActive := activeFlowStatuses[status]
+	a.OtelManager.Metrics.FlowStatusGauge.Record(ctx, 1, metric.WithAttributeSet(attribute.NewSet(
```
serprex, line 1425: @iamKunalGupta I don't understand how metrics know which flow this is about

iamKunalGupta: Line 159 of drop_flow.go -> GetFlowContextMetadata builds the context and otel_metrics looks up the context keys

Context propagator ensures that these keys go from wf to acivity

serprex: ah, good that I moved activity invocation below that then instead of right after syncStatusToCatalog

flow/workflows/drop_flow.go
```diff
@@ -161,6 +161,11 @@ func DropFlowWorkflow(ctx workflow.Context, input *protos.DropFlowInput) error {
 		return fmt.Errorf("failed to get flow metadata context: %w", err)
 	}
 
+	_ = workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
```
jgao54, line 164: nit: maybe add a comment that this must run after GetFlowMetadataContext as it's not obvious

jgao54: actually i can do this in a separate PR, as I want to add some comments about the `updateFlowStatusInCatalogActivity` as well :)

PR #3472, opened 2025-09-12
parallelize api tests, move unsafe to new suite

heavycrystal:


heavycrystal:
could be flaky, merging and waiting for data

PR #3474, opened 2025-09-12
fix drop regression when source/destination peer not specified for metadata context

serprex:
regression introduced in #3432

### DIFF
--- a/flow/workflows/drop_flow.go
+++ b/flow/workflows/drop_flow.go
@@ -147,7 +147,7 @@ func DropFlowWorkflow(ctx workflow.Context, input *protos.DropFlowInput) error {
 		slog.String(string(shared.FlowNameKey), input.FlowJobName))
 	contextMetadataInput := &protos.FlowContextMetadataInput{
 		FlowName: input.FlowJobName,
-		Status:   protos.FlowStatus_STATUS_UNKNOWN,
+		Status:   status,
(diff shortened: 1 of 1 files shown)


flow/activities/flowable.go
```diff
@@ -1390,25 +1390,39 @@ func (a *FlowableActivity) GetFlowMetadata(
 	input *protos.FlowContextMetadataInput,
 ) (*protos.FlowContextMetadata, error) {
 	logger := log.With(internal.LoggerFromCtx(ctx), slog.String(string(shared.FlowNameKey), input.FlowName))
-	peerTypes, err := connectors.LoadPeerTypes(ctx, a.CatalogPool, []string{input.SourceName, input.DestinationName})
+	peerNames := make([]string, 0, 2)
+	if input.SourceName != "" {
```
jgao54, line 1394: Is the regression happening because `ReportStatusMetric` is introduced?

if so, we could also optionally not emit `ReportStatusMetric` when `FlowConnectionConfigs == nil` in Dropflow.

It means qrep flow status won't be up to date, but given dropflow does not populate `cdcConfig`, not sure if it would work today anyways.

jgao54: nvm, thought this was related to #3466, but it's actually #3432.  metrics is a separate thing.

PR #3479, opened 2025-09-15
Postgres: Classify replica identity nothing error

Amogh-Bharadwaj:
If a table has replica identity nothing then we currently error out wherever we get table schema. This PR makes sure to emit this error. It will notify users

### DIFF
--- a/flow/shared/err_types.go
+++ b/flow/shared/err_types.go
@@ -8,8 +8,9 @@ import (
 )
 
 var (
-	ErrSlotAlreadyExists error = temporal.NewNonRetryableApplicationError("slot already exists", "snapshot", nil)
-	ErrTableDoesNotExist error = temporal.NewNonRetryableApplicationError("table does not exist", "snapshot", nil)
+	ErrSlotAlreadyExists      error = temporal.NewNonRetryableApplicationError("slot already exists", "snapshot", nil)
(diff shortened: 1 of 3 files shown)


heavycrystal (approved):

flow/shared/err_types.go
```diff
@@ -8,8 +8,9 @@ import (
 )
 
 var (
-	ErrSlotAlreadyExists error = temporal.NewNonRetryableApplicationError("slot already exists", "snapshot", nil)
-	ErrTableDoesNotExist error = temporal.NewNonRetryableApplicationError("table does not exist", "snapshot", nil)
+	ErrSlotAlreadyExists      error = temporal.NewNonRetryableApplicationError("slot already exists", "snapshot", nil)
+	ErrTableDoesNotExist      error = temporal.NewNonRetryableApplicationError("table does not exist", "snapshot", nil)
+	ErrReplicaIdentityNothing error = errors.New("table has replica identity 'n'/NOTHING")
```
iamKunalGupta, line 13: might be a better idea to change this to a struct and add the table name in the alert

Amogh-Bharadwaj: Addressed here: https://github.com/PeerDB-io/peerdb/pull/3481

PR #3483, opened 2025-09-15
refactor(record-metrics): use single query and add more metrics

iamKunalGupta:


flow/activities/flowable.go
```diff
@@ -824,9 +789,105 @@ type flowInformation struct {
...
+	var scannedFlow string
+	tableCount := &protos.CDCTableRowCounts{
+		Counts: &protos.CDCRowCounts{},
+	}
+
+	operationValueMapping := map[string]*int64{
+		otel_metrics.RecordOperationTypeInsert: &tableCount.Counts.InsertsCount,
+		otel_metrics.RecordOperationTypeUpdate: &tableCount.Counts.UpdatesCount,
+		otel_metrics.RecordOperationTypeDelete: &tableCount.Counts.DeletesCount,
+	}
```
```suggestion
	operationValueMapping := map[string]int64{
		otel_metrics.RecordOperationTypeInsert: tableCount.Counts.InsertsCount,
		otel_metrics.RecordOperationTypeUpdate: tableCount.Counts.UpdatesCount,
		otel_metrics.RecordOperationTypeDelete: tableCount.Counts.DeletesCount,
	}
```
really tho this should be `[3]struct{op string; count int64}`

flow/activities/flowable.go
```diff
@@ -824,9 +789,105 @@ type flowInformation struct {
...
+					attribute.String(otel_metrics.RecordOperationTypeKey, op),
+				)))
+			}
+		} else {
+			logger.Error("Flow not found for metrics",
+				slog.String("flow", scannedFlow), slog.String("tableName", tableCount.TableName))
+		}
+		return nil
+	})
+	if err != nil {
```
```suggestion
	if _, err := pgx.ForEachRow(rows, []any{
		&tableCount.TableName,
		&tableCount.Counts.InsertsCount,
		&tableCount.Counts.UpdatesCount,
		&tableCount.Counts.DeletesCount,
		&scannedFlow,
	}, func() error {
		if flowData, ok := flowsMap[scannedFlow]; ok {
			eCtx := context.WithValue(ctx, internal.FlowMetadataKey, flowData.toFlowContextMetadata())
			for op, valuePtr := range operationValueMapping {
				a.OtelManager.Metrics.RecordsSyncedPerTableGauge.Record(eCtx, *valuePtr, metric.WithAttributeSet(attribute.NewSet(
					attribute.String(otel_metrics.FlowNameKey, scannedFlow),
					attribute.String(otel_metrics.DestinationTableNameKey, tableCount.TableName),
					attribute.String(otel_metrics.RecordOperationTypeKey, op),
				)))
			}
		} else {
			logger.Error("Flow not found for metrics",
				slog.String("flow", scannedFlow), slog.String("tableName", tableCount.TableName))
		}
		return nil
	}); err != nil {
```

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -112,19 +115,29 @@ func UpdateEndTimeForCDCBatch(
...
+	tableNameRowsMapping map[string]*model.RecordTypeCounts,
+	otelManager *otel_metrics.OtelManager,
 ) error {
 	insertBatchTablesTx, err := pool.Begin(ctx)
 	if err != nil {
 		return fmt.Errorf("error while beginning transaction for inserting statistics: %w", err)
 	}
 	defer shared.RollbackTx(insertBatchTablesTx, internal.LoggerFromCtx(ctx))
-
+	tableNameOperations := make(map[string]map[string]int32, len(tableNameRowsMapping))
```
serprex: Insert/Update/Delete doesn't need to be map, can be `[3]struct{op string;count int32}`

flow/activities/flowable.go
iamKunalGupta: shouldn't it be `[3]struct{op string; count *int64}` else `operationValueMapping` will just have dereferenced value from before the scan?

serprex: if you move away from map you can get reference to field for scan *(go just has funny semantics about not giving away pointers to objects in maps so that maps can move their elements around without breaking memory safety)*. I see why you went with pointers here now with map tho

iamKunalGupta: yeah, I took a look at the full code now, I don't need the proto anymore

PR #3493, opened 2025-09-17
Idempotent CDC flow creation

ilidemi:
In ClickPipes the flow name is always unique. Add a new endpoint that makes sure retrying with the same flow name always returns the same result, keying on the CDCFlowWorkflow id.

Concurrency is tested with toxiproxy, I am open to removing it if that's too much


flow/cmd/handler.go
```diff
@@ -154,7 +168,65 @@ func (h *FlowRequestHandler) CreateCDCFlow(
...
+	if err != nil {
+		return nil, exceptions.NewInternalApiError(fmt.Errorf("failed to get internal version: %w", err))
+	}
+	cfg.Version = internalVersion
+
+	if cfg.Resync {
+		return nil, exceptions.NewInvalidArgumentApiError(errors.New("resync is not supported in the managed API"))
+	}
+
+	workflowID := cfg.FlowJobName + "-peerflow"
```
iamKunalGupta: are we making wf IDs deterministic without the uuid suffix?

ilidemi: FlowJobName contains pipe id which is a primary db key on the caller side, so 1 pipe = 1 workflow is what we want. Any UUID generation on PeerDB side would break this

flow/cmd/handler.go
```diff
@@ -78,11 +79,24 @@ func (h *FlowRequestHandler) createCdcJobEntry(ctx context.Context,
...
 	}
 
-	if _, err := h.pool.Exec(ctx,
-		`INSERT INTO flows (workflow_id, name, source_peer, destination_peer, config_proto, status,
-		description, source_table_identifier, destination_table_identifier) VALUES ($1,$2,$3,$4,$5,$6,'gRPC','','')`,
-		workflowID, req.ConnectionConfigs.FlowJobName, sourcePeerID, destinationPeerID, cfgBytes, protos.FlowStatus_STATUS_SETUP,
-	); err != nil {
+	if idempotent {
+		_, err = h.pool.Exec(ctx,
+			`INSERT INTO flows (workflow_id, name, source_peer, destination_peer, config_proto, status,
```
serprex: you can't instead keep this as one code path & check if error is conflict?

also, can't use `ON CONFLICT DO NOTHING`?

ilidemi: Thing is, name is not a key at all for flows, so no conflict. Idk how would it impact OSS users if we change that but there has been some duplication in the managed instances already.

flow/cmd/handler.go
iamKunalGupta: One issue with having a new API is that we need to wait for all users to be up to date before switching the API. Instead if we add a new parameter, it would allow us to switch out of the box

iamKunalGupta: > FlowJobName contains pipe id which is a primary db key on the caller side, so 1 pipe = 1 workflow is what we want. Any UUID generation on PeerDB side would break this

For managed call yes, it makes sense, I meant to say for non-managed creation we still use uuids

ilidemi: Might rename the existing endpoint then so the managed caller doesn't have to be deployed after

flow/cmd/handler.go
ilidemi: Made unique

flow/cmd/handler.go
ilidemi: One endpoint with a boolean is good

serprex (approved):

PR #3494, opened 2025-09-17
[mysql] add SSH keepalive

heavycrystal:


flow/connectors/utils/ssh.go
```diff
@@ -79,21 +84,63 @@ func NewSSHTunnel(
...
+		ticker := time.NewTicker(time.Duration(keepaliveIntervalSeconds) * time.Second)
+		defer ticker.Stop()
+		logger := internal.LoggerFromCtx(ctx)
+		for {
+			select {
+			case <-ticker.C:
+				_, _, err := tunnel.Client.SendRequest("keepalive@openssh.com", true, nil)
+				if err != nil {
+					logger.Error("Failed to send keep alive", slog.Any("error", err))
+					// don't close it, otherwise a select can read from channel forever until we error out
```
serprex: selecting a closed channel immediately returns

heavycrystal: just clarified comment, I don't want this behaviour as it'll keep trying to close MySQL connection while the rest of the system catches up

flow/connectors/utils/ssh.go
```diff
@@ -79,21 +84,63 @@ func NewSSHTunnel(
...
 	if tunnel.Client != nil {
+		close(tunnel.closeKeepaliveChan)
 		return tunnel.Client.Close()
 	}
 	return nil
 }
+
+// returns a channel that will receive a value if the SSH keepalive fails
+// or nil if no SSH tunnel is configured
+func (tunnel *SSHTunnel) GetKeepaliveChan(ctx context.Context) chan struct{} {
```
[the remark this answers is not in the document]
heavycrystal: shouldn't happen as one connection owns one tunnel currently, can be changed later

flow/connectors/utils/ssh.go
```diff
@@ -79,21 +84,63 @@ func NewSSHTunnel(
...
+		logger := internal.LoggerFromCtx(ctx)
+		for {
+			select {
+			case <-ticker.C:
+				_, _, err := tunnel.Client.SendRequest("keepalive@openssh.com", true, nil)
+				if err != nil {
+					logger.Error("Failed to send keep alive", slog.Any("error", err))
+					// don't close it, otherwise a select can read from channel forever until we error out
+					// relying on it being closed by .Close()
+					tunnel.keepaliveChan <- struct{}{}
```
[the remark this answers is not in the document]
heavycrystal: true, but we want a goroutine to be waiting to catch the keepalive failure so working as intended

flow/connectors/utils/ssh.go
serprex: you could close it, then nil out field in tunnel, so later calls to `GetKeepaliveChan` returns nil, then in code check that if function returns nil connection is down

flow/connectors/utils/ssh.go
```diff
@@ -79,21 +84,63 @@ func NewSSHTunnel(
...
+		return SSHTunnel{Client: client, closeKeepaliveChan: make(chan struct{})}, nil
 	}
 
 	return SSHTunnel{}, nil
 }
 
-func (tunnel SSHTunnel) Close() error {
+func (tunnel *SSHTunnel) Close() error {
 	if tunnel.Client != nil {
+		close(tunnel.closeKeepaliveChan)
```
serprex: panics when Close called twice

not an issue if switching to closing channel then setting field to nil

flow/connectors/utils/ssh.go
```diff
@@ -79,21 +84,66 @@ func NewSSHTunnel(
...
+	tunnel.keepaliveChan = make(chan struct{})
+
+	go func() {
+		ticker := time.NewTicker(time.Duration(keepaliveIntervalSeconds) * time.Second)
+		defer ticker.Stop()
+		logger := internal.LoggerFromCtx(ctx)
+		for {
+			select {
+			case <-ticker.C:
+				_, _, err := tunnel.Client.SendRequest("keepalive@openssh.com", true, nil)
```
ilidemi: What happens when the receiver is a black hole? I don't think we're setting a net.Conn deadline (different from Dialer deadline) explicitly anywhere so this might hang indefinitely (our [reference](https://github.com/golang/go/issues/21478#issue-250696693) extends the deadline on every request). Alternatively, can do time.After concurrently.

heavycrystal: Unlike the reference, we don't have a handle on the underlying `net.Conn` and having one is non-trivial because the MySQL connector can create multiple connections with the dialer we provide.

Can revisit this if we see the failure mode in practice

serprex: You'd have a time.After goroutine that closes connection if request is blocked too long

serprex (approved):

flow/connectors/mysql/ssh_keepalive_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// Test initial connection works
+	err = connector.ConnectionActive(ctx)
+	require.NoError(t, err, "Initial connection should work")
+
+	return connector, sshProxy
+}
+
+func TestMySQLSSHKeepaliveWithToxiproxy(t *testing.T) {
+	if testing.Short() {
```
ilidemi: We're not really using .Short() elsewhere, is this intentional?

flow/connectors/mysql/ssh_keepalive_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+	ctx := t.Context()
+
+	// Connect to Toxiproxy - NewClient automatically adds http:// if not present
+	toxiproxyClient := toxiproxy.NewClient(toxiproxyHost + ":" + toxiproxyAPIPort)
+
+	// Test if toxiproxy is available
+	_, err := toxiproxyClient.Version()
+	if err != nil {
+		t.Skipf("Toxiproxy not available: %v", err)
```
ilidemi: When would we want it to skip the test vs help us detect when something broke?

ilidemi: If we expect version check failures to fail the test, the check and the client creation could be in a helper that always returns a good client

flow/connectors/mysql/ssh_keepalive_test.go
```diff
@@ -0,0 +1,251 @@
...
+		expectedErrors := []string{"connection reset", "broken pipe", "EOF", "connection closed", "use of closed network connection"}
+		for _, expected := range expectedErrors {
+			if assert.Contains(t, errorMsg, expected) {
+				hasExpectedError = true
+				break
+			}
+		}
+		assert.True(t, hasExpectedError, "Error should indicate immediate connection failure, got: %s", errorMsg)
+
+	case <-time.After(15 * time.Second):
```
ilidemi: Injecting the keepalive interval could help speed up the tests and make it clearer how it relates to test thresholds, especially if we tweak the constant in the future

PR #3500, opened 2025-09-18
Classify pg_inherits query failure

Amogh-Bharadwaj:
To obtain parent-child table mapping in Postgres CDC we query this catalog table in Postgres - `pg_inherits` - when this fails, we are not classifying the error.
This PR classifies it


flow/connectors/postgres/cdc.go
```diff
@@ -1254,7 +1254,7 @@ func (p *PostgresCDCSource) checkIfUnknownTableInherits(ctx context.Context,
 			if errors.Is(err, pgx.ErrNoRows) {
 				return relID, nil
 			}
-			return 0, fmt.Errorf("failed to query pg_inherits: %w", err)
+			return 0, exceptions.NewPostgresDependencyError(fmt.Errorf("failed to query pg_inherits: %w", err))
```
iamKunalGupta: Can you check if there is an underlying connectionc closed in go? Might be simpler

iamKunalGupta: I think we can do all 4 of these: https://github.com/jackc/pgx/blob/master/pgconn/pgconn.go#L733-L740

flow/alerting/classifier.go
```diff
@@ -261,6 +261,17 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
...
 	}
 
+	// Reference:
+	// https://github.dev/jackc/pgx/blob/master/pgconn/pgconn.go#L733-L740
+	if strings.Contains(err.Error(), "conn closed") ||
+		strings.Contains(err.Error(), "conn uninitialized") ||
+		strings.Contains(err.Error(), "conn busy") {
+		return ErrorNotifyConnectivity, ErrorInfo{
+			Source: ErrorSourceNet,
+			Code:   "UNKNOWN",
```
heavycrystal, line 275: why not recoverable?

iamKunalGupta: we anyways send it on the third retry, we cannot really do anything in case it does not recover

Amogh-Bharadwaj: I thought errRecoverable is what has the 3rd retry logic, will revert in that case

flow/alerting/classifier.go
```diff
@@ -261,6 +261,17 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
+	// Reference:
+	// https://github.dev/jackc/pgx/blob/master/pgconn/pgconn.go#L733-L740
```
serprex:
```suggestion
	// https://github.com/jackc/pgx/blob/61d3c965ad442cc14d6b0e39e0ab3821f3684c03/pgconn/pgconn.go#L731
```
1. link doesn't need to grant oauth to github.dev
2. link uses hash so line reference doesn't drift

flow/alerting/classifier.go
```diff
@@ -261,6 +261,17 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
...
+	if strings.Contains(err.Error(), "conn closed") ||
+		strings.Contains(err.Error(), "conn uninitialized") ||
+		strings.Contains(err.Error(), "conn busy") {
```
serprex: these aren't really net errors, so much as status of pgconn object. eg `conn busy` is the result of lock being called while connection object already locked

seems like an error in our logic somewhere racing on usage of pgconn

Amogh-Bharadwaj: I agree with conn busy -- the source code also says that indicates an app error so I can remove that. The error we've seen is conn closed

Amogh-Bharadwaj: Changed to only classify conn closed

serprex (approved):

PR #3503, opened 2025-09-21
Classify: MySQL multi-source GTID conflict

Amogh-Bharadwaj:
Error encountered was:
```
ERROR 1943 (HY000): GTID 0-4294967295-115345509 and 0-1-115345510 conflict (duplicate domain id 0)
```

Relevant material:
https://mariadb.com/docs/server/ha-and-performance/standard-replication/gtid#use-with-multi-source-replication-and-other-multi-primary-setups

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -472,9 +472,7 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		case 1146: // ER_NO_SUCH_TABLE
 			return ErrorNotifySourceTableMissing, myErrorInfo
 		case 1943:
-			if myErr.State == "HY000" && strings.Contains(myErr.Message, "duplicate domain id") {
-				return ErrorNotifyBadGTIDSetup, myErrorInfo
-			}
+			return ErrorNotifyBadGTIDSetup, myErrorInfo
 		default:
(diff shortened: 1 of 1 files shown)


flow/alerting/classifier.go
```diff
@@ -468,6 +471,10 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 			return ErrorOther, myErrorInfo
 		case 1146: // ER_NO_SUCH_TABLE
 			return ErrorNotifySourceTableMissing, myErrorInfo
+		case 1943:
```
serprex, line 474: https://mariadb.com/docs/server/reference/error-codes/mariadb-error-codes-1900-to-1999/e1943

No need to check for HY000 & error message

flow/alerting/classifier.go
```diff
@@ -468,6 +471,10 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
...
+			if myErr.State == "HY000" && strings.Contains(myErr.Message, "duplicate domain id") {
+				return ErrorNotifyBadGTIDSetup, myErrorInfo
+			}
```
serprex:
```suggestion
			return ErrorNotifyBadGTIDSetup, myErrorInfo
```

PR #3512, opened 2025-09-23
Enforce API errors via the type system

ilidemi:
Sadly there doesn't seem to be a way to completely avoid AST code but at least it is simpler and the consumer side is as pleasant as it can be


stacks/flow.Dockerfile
```diff
@@ -17,6 +17,8 @@ RUN rm -f go.work*
 # build the binary from flow folder
 WORKDIR /root/flow
 ENV CGO_ENABLED=1
+# Generate the typed handler wrapper
+RUN go run cmd/gen-grpc-wrapper/main.go
```
serprex: not `go generate`?

PR #3513, opened 2025-09-23
fix(classifier): invalid snapshot

iamKunalGupta:


flow/alerting/classifier_test.go
```diff
@@ -533,10 +533,17 @@ func TestTemporalKnownErrorsShouldBeCorrectlyClassified(t *testing.T) {
...
+			errInfo: ErrorInfo{
+				Source: ErrorSourcePostgres,
+				Code:   "SNAPSHOT_DOES_NOT_EXIST",
+			},
+		},
 	} {
 		t.Run(code.String(), func(t *testing.T) {
 			errorClass, errInfo := GetErrorClass(t.Context(), temporal.NewNonRetryableApplicationError(
-				"irrecoverable error",
+				"irrecoveragble error",
```
iskakaushik: typo

serprex:
```suggestion
				"irrecoverable error",
```

flow/shared/err_types.go
```diff
@@ -5,11 +5,13 @@ import (
...
 	"go.temporal.io/sdk/temporal"
+
+	"github.com/PeerDB-io/peerdb/flow/shared/exceptions"
 )
 
 var (
-	ErrSlotAlreadyExists error = temporal.NewNonRetryableApplicationError("slot already exists", "snapshot", nil)
-	ErrTableDoesNotExist error = temporal.NewNonRetryableApplicationError("table does not exist", "snapshot", nil)
+	ErrSlotAlreadyExists error = temporal.NewNonRetryableApplicationError("slot already exists", exceptions.ApplicationErrorTypeIrrecoverableSlotMissing.String(), nil)
+	ErrTableDoesNotExist error = temporal.NewNonRetryableApplicationError("table does not exist", exceptions.ApplicationErrorTypeIrrecoverablePublicationMissing.String(), nil)
```
iamKunalGupta: weird, I'll fix this

serprex: it's probably trying to blend in with nonretryable error types

iamKunalGupta: we had just set non recoverable types to `snapshot` earlier
I'm thinking if this can cause non determinism? it does not change the state machine tho

iamKunalGupta: let me add new codes for these and add them to non recoverable

PR #3523, opened 2025-09-25
[postgres] classify slot read failing due to inactive sync standby

heavycrystal:
Still figuring out exactly when this can happen, but classify regardless

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -438,7 +438,7 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 			}
 
 			if strings.Contains(pgErr.Message,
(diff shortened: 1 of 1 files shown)


flow/alerting/classifier.go
```diff
@@ -437,6 +437,11 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 				return ErrorNotifySlotInvalid, pgErrorInfo
 			}
 
+			if strings.Contains(pgErr.Message,
+				"specified in parameter \"synchronized_standby_slots\" does not have active_pid") {
```
iamKunalGupta: can use backticks to prevent quote escaping

iamKunalGupta:
```suggestion
				`specified in parameter "synchronized_standby_slots" does not have active_pid`) {
```

PR #3545, opened 2025-10-01
fix(ch): validatePeer increase timeout, CREATE OR REPLACE

heavycrystal:
There are instances (either overloaded instances or instances waking up in Cloud) where a service may take longer than 15s to respond to our validation checks, adding increased timeline for ClickHouse.

Also instead of creating validation tables repeatedly, `CREATE OR REPLACE` a table with a fixed suffix (currently peer attributes)

### DIFF
--- a/flow/cmd/validate_peer.go
+++ b/flow/cmd/validate_peer.go
@@ -14,8 +14,6 @@ func (h *FlowRequestHandler) ValidatePeer(
 	ctx context.Context,
 	req *protos.ValidatePeerRequest,
 ) (*protos.ValidatePeerResponse, APIError) {
-	ctx, cancelCtx := context.WithTimeout(ctx, 15*time.Second)
-	defer cancelCtx()
 	if req.Peer == nil {
 		return &protos.ValidatePeerResponse{
 			Status:  protos.ValidatePeerStatus_INVALID,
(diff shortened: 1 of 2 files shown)


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -166,7 +168,14 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
 	if err := ValidateClickHouseHost(ctx, c.Config.Host, allowedDomains); err != nil {
 		return err
 	}
-	validateDummyTableName := "peerdb_validation_" + shared.RandomString(4)
+
+	hasher := sha256.New()
```
serprex: fine to use [fnv64a](https://pkg.go.dev/hash/fnv)

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -166,7 +168,14 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
...
+	hasher.Write([]byte(c.Config.Host))
+	hasher.Write([]byte(c.Config.Database))
+	hasher.Write([]byte(c.Config.User))
```
serprex: use io.WriteString

flow/cmd/validate_peer.go
```diff
@@ -30,6 +28,14 @@ func (h *FlowRequestHandler) ValidatePeer(
 		}, NewInvalidArgumentApiError(errors.New("no peer name provided"))
 	}
 
+	validatePeerDeadline := 15 * time.Second
```
Amogh-Bharadwaj, line 31: Maybe a comment explaining why 15s for this 1 min for that  would be good

Amogh-Bharadwaj (approved):

PR #3547, opened 2025-10-02
Move shared validations into their own module

ilidemi:
This allows to import just the validation code with minimal dependencies and avoid tracking unrelated dependency updates in PeerDB.

Connector-specific code remaining in shared/ was kind of small so moved it to other places. CH type conversion was supposed to be shared with other services in the future but it pulls in QValue so we can decide to move all of it to shared-light if we want to expose QValue or to move just type conversions and give it its own small data contracts.


flow/connectors/clickhouse/type_conversion.go
```diff
@@ -2,13 +2,69 @@ package connclickhouse
 
 import (
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
-	"github.com/PeerDB-io/peerdb/flow/shared/clickhouse"
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
 
+/*
```
jgao54, line 8: Thanks, just for context, originally this was moved to shared because of [this PR](https://github.com/ClickHouse/clickpipes-platform/pull/4260/files#diff-ba27dcf24cfc8a7d9c9f94eddaa369d85303e5f96b7e8a09f1782ba77d38b24aR408) which never got merged/prioritized.

good idea to keep it internal for now and we can move to shared later when this feature is actually prioritized.

flow/connectors/utils/partition.go
```diff
@@ -16,6 +16,12 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared"
 )
 
+// QRep partition constants
+const (
+	MySQLFullTablePartitionID = "mysql-full-table-partition-id"
```
serprex: these can be merged, no need to put mysql/mongo in id

serprex:
if `shared-light` is split out from the rest of `flow` it can live beside `flow` rather than inside of it, in terms of directory structure

`shared-light` feels like it could easily become outdated. Hard to come up with good name tho. `peer-support`?

flow/shared/mysql/validation.go
```diff
@@ -64,12 +62,12 @@ func CheckMySQL5BinlogSettings(conn *client.Conn, logger log.Logger) error {
 	}
 	row := rs.Values[0]
 
-	binlogFormat := shared.UnsafeFastReadOnlyBytesToString(row[0].AsString())
```
serprex, line 67: these functions get copied around a lot, in go-mysql it's under [zeroalloc](https://github.com/go-mysql-org/go-mysql/blob/master/utils/zeroalloc.go) & every large go package ends up rolling copies of these

for validation it's fine to skip this optimization

PR #3556, opened 2025-10-04
drop completed mirror: just delete flow entry

serprex:
bypasses mess of shutdowFlow logic, in particular doesn't try connect to source/destination, avoiding issues when those are unavailable

### DIFF
--- a/flow/cmd/handler.go
+++ b/flow/cmd/handler.go
@@ -211,6 +211,36 @@ func (h *FlowRequestHandler) CreateQRepFlow(
 	}, nil
 }
 
+func (h *FlowRequestHandler) dropFlow(
+	ctx context.Context,
+	flowJobName string,
+	deleteStats bool,
+) error {
+	logs := slog.Group("shutdown-log", slog.String(string(shared.FlowNameKey), flowJobName))
+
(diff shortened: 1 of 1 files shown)


flow/cmd/handler.go
```diff
@@ -211,6 +211,36 @@ func (h *FlowRequestHandler) CreateQRepFlow(
...
+		ID:                    dropFlowWorkflowID,
+		TaskQueue:             h.peerflowTaskQueueID,
+		TypedSearchAttributes: shared.NewSearchAttributes(flowJobName),
+	}
+
+	if dropFlowHandle, err := h.temporalClient.ExecuteWorkflow(ctx, workflowOptions, peerflow.DropFlowWorkflow, &protos.DropFlowInput{
+		FlowJobName:           flowJobName,
+		DropFlowStats:         deleteStats,
+		FlowConnectionConfigs: nil,
+		SkipDestinationDrop:   true,
```
ilidemi, line 297: A bit confusing to have SkipDestinationDrop but not SkipSourceDrop. Could do very explicit with both true, or let FlowConnectionConfigs nil do the job

ilidemi (approved):

flow/cmd/handler.go
serprex: May want to still pass FlowConnectionConfigs so that `DropFlowStats` gets handled

PR #3558, opened 2025-10-06
BigQuery source

jkaflik:


flow/activities/flowable.go
```diff
@@ -118,7 +118,7 @@ func (a *FlowableActivity) EnsurePullability(
 	config *protos.EnsurePullabilityBatchInput,
 ) (*protos.EnsurePullabilityBatchOutput, error) {
 	ctx = context.WithValue(ctx, shared.FlowNameKey, config.FlowJobName)
-	srcConn, err := connectors.GetByNameAs[connectors.CDCPullConnector](ctx, nil, a.CatalogPool, config.PeerName)
+	srcConn, err := connectors.GetByNameAs[connectors.CDCPullConnectorCore](ctx, nil, a.CatalogPool, config.PeerName) // todo: I assume this was mistake before to expect CDCPullConnector
```
serprex: Yep

flow/activities/flowable.go
```diff
@@ -561,36 +561,95 @@ func (a *FlowableActivity) ReplicateQRepPartitions(ctx context.Context,
...
+		case connectors.QRepPullConnector:
+			destConn, ok := qRepSyncCoreConn.(connectors.QRepSyncConnector)
+			if !ok {
+				err = fmt.Errorf("source connector is QRepPullConnector but destination connector is not QRepSyncConnector, got %T", qRepSyncCoreConn)
+				break
+			}
+
+			err = qRecordReplication(srcConn, destConn, partition)
+		case connectors.QRepPullObjectsConnector:
+			// todo: should we bring Lua support?
```
serprex: No, especially since with BQ data never hits peerdb

flow/activities/flowable.go
```diff
@@ -561,36 +561,95 @@ func (a *FlowableActivity) ReplicateQRepPartitions(ctx context.Context,
...
+			}
+		}
+
+		return replicateQRepPartition(ctx, a, qRepPullConn, destConn, config, partition, runUUID, stream, outstream,
+			connectors.QRepPullConnector.PullQRepRecords,
+			connectors.QRepSyncConnector.SyncQRepRecords,
+		)
+	}
+
+	// todo: figure out if we are allowed to reuse the same connection for multiple partitions
```
serprex: I don't understand, but generally yes if code is within same activity. If it isn't then you need to use session in workflow so group of activities run on same activity worker & can share state between them on the FlowableActivity

flow/connectors/bigquery/avro.go
```diff
@@ -0,0 +1,73 @@
...
+		return 0, fmt.Errorf("failed to create OCF decoder: %w", err)
+	}
+
+	var totalRows uint64
+	for decoder.HasNext() {
+		select {
+		case <-ctx.Done():
+			return totalRows, ctx.Err()
+		default:
+		}
```
serprex:
```suggestion
		if err := ctx.Err(); err != nil {
			return totalRows, err
		}
```

flow/activities/flowable.go
```diff
@@ -562,36 +562,104 @@ func (a *FlowableActivity) ReplicateQRepPartitions(ctx context.Context,
...
+			}
+		}
+
+		return replicateQRepPartition(ctx, a, qRepPullConn, destConn, config, partition, runUUID, stream, outstream,
+			connectors.QRepPullConnector.PullQRepRecords,
+			connectors.QRepSyncConnector.SyncQRepRecords,
+		)
+	}
+
+	for _, partition := range partitions.Partitions {
```
jkaflik: could we have subfunctions per replication method?

flow/connectors/bigquery/bigquery.go
```diff
@@ -110,46 +135,9 @@ func NewBigQueryConnector(ctx context.Context, config *protos.BigqueryConfig) (*
...
-
-	// Drop the table
-	deleteErr := newTable.Delete(ctx)
-	if deleteErr != nil {
-		errs = append(errs, fmt.Errorf("unable to delete table :%w. ", deleteErr))
-	}
-
-	if len(errs) > 0 {
-		return errors.Join(errs...)
+	if _, err := c.client.DatasetInProject(c.projectID, c.datasetID).Metadata(ctx); err != nil {
```
jkaflik, line 140: move the create/query/delete validation to mirror creation validation

flow/activities/flowable.go
serprex: I don't follow, what do you mean by subfunctions?

flow/connectors/bigquery/qvalue_convert.go
```diff
@@ -133,6 +133,9 @@ func BigQueryTypeToQValueKind(fieldSchema *bigquery.FieldSchema) types.QValueKin
 		return types.QValueKindGeography
 	case bigquery.JSONFieldType:
 		return types.QValueKindJSON
+	case bigquery.RecordFieldType:
+		// RECORD/STRUCT types are represented as String in ClickHouse
+		return types.QValueKindString
```
serprex, line 138: Should this be KindJSON? currently we map JSON to string, but have an option to map JSON to CH's JSON type

jkaflik: I think it makes sense. I will validate this.

PR #3568, opened 2025-10-07
drop flow: use same workflow id as peerflow

serprex:
alternative to #3549, fixes drop flow failing to update status in catalog when workflow id doesn't match cdc workflow id

### DIFF
--- a/flow/cmd/handler.go
+++ b/flow/cmd/handler.go
@@ -283,7 +283,7 @@ func (h *FlowRequestHandler) dropFlow(
 		}
 	}
 
-	dropFlowWorkflowID := fmt.Sprintf("%s-dropflow-%s", flowJobName, uuid.New())
+	dropFlowWorkflowID := flowJobName + "-peerflow"
 	workflowOptions := client.StartWorkflowOptions{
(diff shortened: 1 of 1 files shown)


flow/cmd/handler.go
```diff
@@ -283,7 +283,7 @@ func (h *FlowRequestHandler) dropFlow(
 		}
 	}
 
-	dropFlowWorkflowID := fmt.Sprintf("%s-dropflow-%s", flowJobName, uuid.New())
+	dropFlowWorkflowID := flowJobName + "-peerflow"
 	workflowOptions := client.StartWorkflowOptions{
```
serprex, line 287: setting conflict policy to reuse, & removing uuid, could work in #3549 but with this PR the stupid scenario where drop flow is created while cdc flow still running makes me hesitant to have it return a handle to a running cdc flow

heavycrystal:
I think you may need to specify a https://docs.temporal.io/workflow-execution/workflowid-runid#workflow-id-reuse-policy and https://docs.temporal.io/workflow-execution/workflowid-runid#workflow-id-conflict-policy too?

PR #3572, opened 2025-10-08
handle infinite time

serprex:
Postgres supports infinite time values, we coalesce these to null, this is incompatible with non nullable avro schema

In CH, QRep & CDC exhibit different behavior as inserting NULL into column inserts default value, where default for `Date32` is `1970-01-01`, while inserting `time.Time{}` into avro file gives minimum `Date32`, `1900-01-01`

Unifying this would be through tracking nullability in CDC & filling JSON with non-null values. Or accepting that avro schema & destination schema don't belong together, & discarding any notion of non-null in avro schema

### DIFF
--- a/flow/e2e/clickhouse_test.go
+++ b/flow/e2e/clickhouse_test.go
@@ -1317,7 +1317,7 @@ func (s ClickHouseSuite) Test_InfiniteTimestamp() {
 
 	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(`
 		CREATE TABLE IF NOT EXISTS %s (
-			id SERIAL PRIMARY KEY,
+			id INT PRIMARY KEY,
 			t_null TIMESTAMP NULL,
 			t_notnull TIMESTAMP NOT NULL,
 			d_null DATE NULL,
@@ -1327,8 +1327,8 @@ func (s ClickHouseSuite) Test_InfiniteTimestamp() {
 		);
 	`, srcFullName)))
 
-	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(`INSERT INTO %s (t_null,t_notnull,d_null,d_notnull,n_null,n_notnull)
-		VALUES ('infinity'::timestamp,'infinity'::timestamp,'infinity'::date,'infinity'::date,'infinity'::numeric,'infinity'::numeric)`,
+	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(`INSERT INTO %s (id,t_null,t_notnull,d_null,d_notnull,n_null,n_notnull)
+		VALUES (1,'infinity'::timestamp,'infinity'::timestamp,'infinity'::date,'infinity'::date,'infinity'::numeric,'infinity'::numeric)`,
 		srcFullName)))
 
 	connectionGen := FlowConnectionGenerationConfig{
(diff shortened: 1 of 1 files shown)


flow/connectors/postgres/qvalue_convert.go
```diff
@@ -391,9 +391,29 @@ func convertToArray[T any](kind types.QValueKind, value any) ([]T, error) {
...
+	typmod int32,
+	value any,
+	customTypeMapping map[uint32]shared.CustomDataType,
+	version uint32,
+) (types.QValue, error) {
+	return c.parseFieldFromPostgresOIDWithQValueKind(
+		oid,
+		typmod,
+		c.postgresOIDToQValueKind(oid, customTypeMapping, version),
+		true,
```
serprex: this matches previous behavior, but is incorrect: we should be maintaining correct null info during CDC

flow/connectors/postgres/qvalue_convert.go
jgao54: when it's not a valid decimal, should this also be checking for nullable?

flow/model/qvalue/avro_converter.go
```diff
@@ -311,8 +315,11 @@ func (c *QValueAvroConverter) processGoTime(t time.Duration) any {
 }
 
 func (c *QValueAvroConverter) processGoTimestampTZ(t time.Time) any {
-	if DisallowedTimestamp(c.TargetDWH, t, c.logger) {
-		return nil
+	if newVal, disallowed := DisallowedTimestamp(c.TargetDWH, t, c.logger); disallowed {
+		if c.Nullable {
+			return nil
+		}
```
jgao54: nit: since `DisallowedTimestamp` is a helper for BQ, maybe just move this check to inside the helper?

flow/connectors/postgres/qrep_query_executor.go
```diff
@@ -409,7 +412,25 @@ func (qe *QRepQueryExecutor) mapRowToQRecord(
...
-		tmp, err := qe.parseFieldFromPostgresOID(fd.DataTypeOID, fd.TypeModifier, values[i], qe.customTypeMapping, qe.version)
+		var nullable bool
+		if i < len(schema.Fields) && schema.Fields[i].Name == fd.Name {
+			nullable = schema.Fields[i].Nullable
+		} else {
+			for _, field := range schema.Fields {
+				if field.Name == fd.Name {
+					nullable = field.Nullable
+					break
+				}
```
jgao54: looks like mapRowToRecord is in the hot path, could/should we aim to do some preprocessing before `for rows.Next() {...}` loop is entered to make this O(1)

ilidemi: Newbie question, when can the schema fields have different order from fd fields? This is qrep so no schema changes in between rows

serprex: I don't expect it, but schema skew would be the easiest way. I need to check how things work out with column exclusion too

flow/connectors/postgres/qvalue_convert.go
serprex: Yes

flow/e2e/clickhouse_test.go
```diff
@@ -1306,6 +1306,52 @@ func (s ClickHouseSuite) Test_Types_CH() {
...
+	}
+
+	srcTableName := "test_infinite_time"
+	srcFullName := s.attachSchemaSuffix(srcTableName)
+	dstTableName := "test_infinite_time"
+
+	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(`
+		CREATE TABLE IF NOT EXISTS %s (
+			id SERIAL PRIMARY KEY,
+			t_null TIMESTAMP NULL,
```
ilidemi, line 1321: 2 tests down, 3 (4 with numeric) to go 👍

serprex: timestamptz, time, timetz? `infinity` isn't valid for time/timetz

ilidemi: Sorry I meant geography/geometry

serprex: geography/geometry case is hard to hit, & is only snowflake, snowflake tends to have everything nullable anyways, so not concerned about that case. Would be good to know how Snowflake handles empty string for geometry

PR #3573, opened 2025-10-08
Do not use slow query from `INFORMATION_SCHEMA`

alexstoick:
When the `INFORMATION_SCHEMA` table baloons it will lead to very slow query execution. In our case - the queries end up taking ~5mins, which drastically slow down the ingest.

The proposal in this PR is to swap to using a `SHOW COLUMNS IN TABLE ...` and then querying the result from that. There is a Snowflake documentation link which highlights this problem, and solution as well, https://community.snowflake.com/s/article/Select-the-list-of-columns-in-the-table-without-using-information-schema

There is an alternative to continue using the `INFORMATION_SCHEMA` but have a cache for all the tables in the sync, I however, think that will be trickier to implement.

### DIFF
--- a/flow/connectors/snowflake/qrep.go
+++ b/flow/connectors/snowflake/qrep.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/snowflakedb/gosnowflake"
 
 	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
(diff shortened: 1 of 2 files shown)


flow/connectors/snowflake/qrep.go
```diff
@@ -182,27 +183,44 @@ func (c *SnowflakeConnector) getColsFromTable(ctx context.Context, tableName str
 		return nil, fmt.Errorf("failed to parse table name: %w", err)
 	}
 
-	rows, err := c.QueryContext(
-		ctx,
-		getTableSchemaSQL,
+	query := fmt.Sprintf(`SHOW COLUMNS IN %s.%s`,
 		strings.ToUpper(schemaTable.Schema),
```
serprex: Ideally escape Schema/Table

alexstoick: Updated so now this gets escaped, as well as the `queryID` in the following query 👍

PR #3574, opened 2025-10-08
Ensure PeerDB UI preserves existing default delete behaviour or soft-delete

Amogh-Bharadwaj:
https://github.com/PeerDB-io/peerdb/pull/2008 + the fact that PeerDB UI sets a softDeleteColName value of `_PEERDB_IS_DELETED` by default in its request to Flow API means that new ClickHouse mirrors created via PeerDB UI will have hard delete turned on by default and not off as intended.

This PR fixes this.

Also since the existing soft delete column name is `_peerdb_is_deleted` and not `_PEERDB_IS_DELETED` the PR fixes this for PeerDB UI as well so that hard delete will not make the is_deleted column uppercase

### DIFF
--- a/ui/app/mirrors/create/helpers/cdc.ts
+++ b/ui/app/mirrors/create/helpers/cdc.ts
@@ -173,7 +173,8 @@ export const cdcSettings: MirrorSetting[] = [
         (curr: CDCConfig): CDCConfig => ({
           ...curr,
           softDeleteColName: (value as boolean)
-            ? curr.softDeleteColName || blankCDCSetting.softDeleteColName.toLowerCase()
+            ? curr.softDeleteColName ||
+              blankCDCSetting.softDeleteColName.toLowerCase()
             : '',
         })
       ),


ui/app/mirrors/create/helpers/common.ts
```diff
@@ -20,6 +20,8 @@ export interface MirrorSetting {
   command?: string;
 }
 
+export const SOFT_DELETE_COLUMN_NAME = '_PEERDB_IS_DELETED';
```
jgao54, line 23: question: instead of `blankCDCSetting.softDeleteColName.toLowerCase()`, could we just make sure the constant for the column name here is lower case? i.e. `export const SOFT_DELETE_COLUMN_NAME = '_peerdb_is_deleted'`, or is there a reason this is in upper case?

Amogh-Bharadwaj: Unfortunately we can't because for Snowflake and BigQuery (which that const also applies to), the default is intended to be upper-case :)
The hard delete option is only visible for ClickHouse

PR #3585, opened 2025-10-10
Add `gocache` to speed up local Docker builds

alexstoick:
Caching link from Go documentation:
https://pkg.go.dev/cmd/go#hdr-Build_and_test_caching

Locally, when re-running `docker build` commands this take the build from ~260-290s down to 20-26s.

### DIFF
--- a/stacks/flow.Dockerfile
+++ b/stacks/flow.Dockerfile
@@ -20,7 +20,8 @@ WORKDIR /root/flow
 ENV CGO_ENABLED=1
 # Generate the typed handler wrapper
 RUN go generate
-RUN go build -o /root/peer-flow
+ENV GOCACHE=/root/.cache/go-build
+RUN --mount=type=cache,target="/root/.cache/go-build" go build -o /root/peer-flow
 
 FROM alpine:3.22@sha256:4bcff63911fcb4448bd4fdacec207030997caf25e9bea4045fa6c8c44de311d1 AS flow-base
 ENV TZ=UTC


stacks/flow.Dockerfile
```diff
@@ -20,7 +20,8 @@ WORKDIR /root/flow
 ENV CGO_ENABLED=1
 # Generate the typed handler wrapper
 RUN go generate
-RUN go build -o /root/peer-flow
+ENV GOCACHE=/root/.cache/go-build
```
heavycrystal, line 23: Is it possible to scope this to local development and not CI?

alexstoick: @heavycrystal - the options I see are:
- having a different `dockerfile` that gets referenced in the `docker-compose-dev.yml`
- having a `buildarg` that you can pass for local development and then having a conditional like:
`RUN if [[ -z "$gocache" ]] ; then `go build -cache...` ; else `go build...` fi`

Let me know which of those, or if you prefer a different approach!

jgao54: thinking more about this, i think this change should be fine. It is pretty standard practice to mount build cache to images. Go should invalidate the cache entries and rebuild when there is a hash mismatch (the cache key is arch and os aware). And the cache mount gets unmounted after build so no image bloat. (CI and local development should also be isolated from one another since they are on different machines)

heavycrystal: It appears that persisting the cache in CI requires additional setup, as mentioned [here](https://docs.docker.com/build/ci/github-actions/cache/#github-cache). We can land this for now.

PR #3586, opened 2025-10-10
Flow API: Add nil check in create CDC request

Amogh-Bharadwaj:
If you send a create mirror request via flow api and you do not wrap your request in `connection_configs`, flow-api will run into a panic. This adds a nil check for this pitfall


flow/cmd/handler.go
```diff
@@ -136,6 +136,10 @@ func (h *FlowRequestHandler) CreateCDCFlow(
 	ctx context.Context, req *protos.CreateCDCFlowRequest,
 ) (*protos.CreateCDCFlowResponse, APIError) {
 	cfg := req.ConnectionConfigs
+	if cfg == nil {
+		// todo(@iamKunalGupta): add a panic recovery handler with metric and alert for it
```
heavycrystal, line 140: I think this is better as a standalone issue so can be assigned

PR #3589, opened 2025-10-10
Proposal: Activity to fetch config from the DB

alexstoick:
This uses a new helper in `internal.FetchConfigFromDB` to fetch a fully hydrated `protos.FlowConnectionConfigs`. When passing this config to Temporal we strip the `tableMappings` array element which can cause it to go over the 2MB limit.

This contains a subset of the changes which were originally proposed in: https://github.com/PeerDB-io/peerdb/pull/3407

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -282,7 +282,7 @@ func processTableAdditions(
 	}
 
 	addTablesSelector.AddFuture(alterPublicationAddAdditionalTablesFuture, func(f workflow.Future) {
-		addTablesFlowErr = f.Get(alterPublicationAddAdditionalTablesCtx, f)
+		addTablesFlowErr = f.Get(alterPublicationAddAdditionalTablesCtx, nil)
 		if addTablesFlowErr == nil {
 			logger.Info("additional tables added to publication")
 			additionalTablesUUID := GetUUID(ctx)
@@ -296,6 +296,7 @@ func processTableAdditions(
 			)
 			if err != nil {
 				addTablesFlowErr = fmt.Errorf("failed to update flow config table mappings for additional tables: %w", err)
+				return
 			}
 			tableMappingAdditionVersions = tableMappingVersions
 			additionalTablesCfg.TableMappingVersion = tableMappingVersions.PartialTableMappingVersion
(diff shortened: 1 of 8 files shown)


nexus/catalog/migrations/V49__table_mappings_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS table_mappings (
+    flow_name varchar(255) not null,Expand commentComment on line R2ResolvedCode has comments. Press enter to view.
+    version int32 not null default 1,
```
ilidemi: bigint, in case someone spams the API endpoint for updating tables so we don't even have to think through overflows

nexus/catalog/migrations/V49__table_mappings_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS table_mappings (
+    flow_name varchar(255) not null,Expand commentComment on line R2ResolvedCode has comments. Press enter to view.
```
ilidemi: copypaste quirk?

flow/workflows/cdc_flow.go
```diff
@@ -291,7 +305,11 @@ func processTableAdditions(
 			additionalTablesCfg := proto.CloneOf(cfg)
 			additionalTablesCfg.DoInitialSnapshot = !flowConfigUpdate.SkipInitialSnapshotForTableAdditions
 			additionalTablesCfg.InitialSnapshotOnly = true
-			additionalTablesCfg.TableMappings = flowConfigUpdate.AdditionalTables
+			// TODO: thought - maybe we need to pass additionalTables and then in the
+			// CDCWorkflow we persist them to the DB and send an incremented `tableMappingVersion`
+			// to the new workflow?
```
ilidemi: Allowing up to ~10K additional tables at a time seems like a generous restriction. Although now it does make sense to have API explicitly turn down config update requests that Temporal won't take

ilidemi (commented):
Options argument - moving SrcTableIdNameMapping to DB too makes sense to me, and use a version instead (or -1/nil). Rest of fields change with config/state separation, and although config is not fully immutable, merging the two would require more thinking (if it's a good idea) and would be too much scope for this PR. If it's easy enough and useful to do, we can do it separately.

Changing activity signatures - the only fully supported upgrade route for PeerDB is to pause CDC mirrors (and wait for setup/snapshot/drop to complete), then upgrade, then resume. It is nice when a surgical change is possible for running snapshots but we're not shooting for that even with smaller changes. We can just treat them as functions (besides the codepath from the start of CDCFlowWorkflow to the pause loop) and not worry about the rest. Adding an argument is good.

flow/cmd/validate_mirror.go
```diff
@@ -65,7 +65,15 @@ func (h *FlowRequestHandler) validateCDCMirrorImpl(
 			errors.New("invalid config: initial_snapshot_only is true but do_initial_snapshot is false"))
 	}
 
-	for _, tm := range connectionConfigs.TableMappings {
+	// fetch connection configs from DB
+	slog.Info("!!!!! YO", slog.Any("connectionConfigs", connectionConfigs), slog.Any("tableMappingVersion", connectionConfigs.TableMappingVersion))
```
serprex: debug logging should be cleaned up

PR #3591, opened 2025-10-11
mongo error clean up

jgao54:
`changestream.Err()` calls [`wrapErrors`](https://github.com/mongodb/mongo-go-driver/blob/1e7cea0fc96b08059915a3d1b0903be2c7c1ed0c/mongo/change_stream.go#L569-L579) which [_attempts_ to wraps errors in one of CommandError, MarshalError, MongocryptError, etc.](https://github.com/mongodb/mongo-go-driver/blob/1e7cea0fc96b08059915a3d1b0903be2c7c1ed0c/mongo/errors.go#L90-L106), but errors can fall thru if it doesn't match one of the known error targets, and the original error is returned. This PR updates driver.Error matching with the wrapped errors. 

There's also been a few iterations of this error handling such that this section no longer makes sense (comment says we retry, but code has since been changed to notify, but notify to all retryable errors seems too aggressive):
```
		// some error codes are defined to be retryable by the driver, so we retry them
		var retryableError RetryableError
		if errors.As(mongoErr, &retryableError) && retryableError.Retryable() {
			return ErrorNotifyConnectivity, mongoErrorInfo
		}
```

So instead notify on the errors we've seen so far: ShutdownInProgress (which has code=91). Would be better if we are more explicit about which one to retry and which one to notify, as they come up in practice.

Finally, fixes https://linear.app/clickhouse/issue/DBI-111/classify-connection-pool-error-with-mongodb-cluster by making it retryable when "connection reset by peers" is detected.

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -617,6 +617,12 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 			return ErrorRetryRecoverable, mongoErrorInfo
 		}
 
+		// this often happens on Mongo Atlas as part of maintenance, and should recover, but we notify if exceed default threshold
+		// (ShutdownInProgress code should be 91, but we have observed 0 in the past, so string match to be safe)
+		if mongoCmdErr.HasErrorMessage("(ShutdownInProgress) The server is in quiesce mode and will shut down") {
+			return ErrorNotifyConnectivity, mongoErrorInfo
+		}
+
 		// This should recover, but we notify if exceed default threshold
 		if mongoCmdErr.HasErrorLabel(driver.TransientTransactionError) {
 			return ErrorNotifyConnectivity, mongoErrorInfo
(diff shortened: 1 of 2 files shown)


flow/alerting/classifier.go
```diff
@@ -610,41 +591,52 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
...
-			}
+		if mongoCmdErr.HasErrorMessage("connection reset by peer") {
+			return ErrorRetryRecoverable, mongoErrorInfo
 		}
 
-		// some error codes are defined to be retryable by the driver, so we retry them
-		var retryableError RetryableError
-		if errors.As(mongoErr, &retryableError) && retryableError.Retryable() {
+		// This should recover, but we notify if exceed default threshold
+		if mongoCmdErr.HasErrorLabel(driver.TransientTransactionError) {
```
jgao54, line 627: a bit inclined to remove this and fallback to using code matching as much as we can.

flow/alerting/classifier.go
```diff
@@ -92,11 +93,6 @@ type ErrorClass struct {
 	action ErrorAction
 }
 
-type RetryableError interface {
```
jgao54, line 95: removing this -- what Mongo driver deems retryable, we might instead want to notify user

PR #3602, opened 2025-10-14
Classify invalid parameter value for synchronized_standby_slots

Amogh-Bharadwaj:
Encountered an error like:
```
failed to get partitions from source:
 failed to query for total rows: 
ERROR: invalid value for parameter "synchronized_standby_slots": "azure_standby_f1ffd344ad1e, azure_standby_bbb297c256bd" (SQLSTATE 22023)
```

More information here:
https://postgresqlco.nf/doc/en/param/synchronized_standby_slots/

[Relevant source code](https://github.com/postgres/postgres/blob/5f3808646f67260d5dc0d95f2602886da5cf5c52/src/backend/replication/slot.c#L2753)

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -518,11 +518,6 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 				return ErrorRetryRecoverable, pgErrorInfo
 			}
 
-			if strings.Contains(pgErr.Message,
-				`specified in parameter "synchronized_standby_slots" does not have active_pid`) {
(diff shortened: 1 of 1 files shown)


flow/alerting/classifier.go
```diff
@@ -515,6 +518,11 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 				return ErrorRetryRecoverable, pgErrorInfo
 			}
 
+			if strings.Contains(pgErr.Message,
```
heavycrystal: isn't this already present a few lines above?

PR #3605, opened 2025-10-15
Decouple FlowConnectionConfigs from API

ilidemi:
In #3589 we want to stop using a field in a way that's not error-prone in the rest of the code (so, removing) but also doesn't break API backcompat (so, not removing).

Trying to split FlowConnectionConfigs into two contracts that are made sure to be in sync with a little codegen. Generated code has small TODOs that would be removed by #3589 if we go this route.

Validated that renaming the type of a workflow arg works well with history replay, only the payload is important.

Curious what the team thinks of this approach, versus having a nullable field that has to be rehydrated in every activity.

### DIFF
--- a/flow/.golangci.yml
+++ b/flow/.golangci.yml
@@ -60,6 +60,8 @@ linters:
       settings:
         hugeParam:
           sizeThreshold: 1024
+        ruleguard:
+          rules: cmd/codegen/ruleguard.rules.go
     gosec:
       excludes:
         - G115
--- a/flow/cmd/codegen/ruleguard.rules.go
+++ b/flow/cmd/codegen/ruleguard.rules.go
@@ -1,0 +1,19 @@
+//go:build ruleguard
+
+package gorules
+
+import (
+	"github.com/quasilyte/go-ruleguard/dsl"
+)
+
+// Enforce FlowConnectionConfigsCore usage outside cmd/ and e2e/
+func flowConfigUsage(m dsl.Matcher) {
+	m.Match(
+		"*protos.FlowConnectionConfigs",     // pointer type
+		"protos.FlowConnectionConfigs",      // non-pointer type
+		"protos.FlowConnectionConfigs{$*_}", // struct literal
+	).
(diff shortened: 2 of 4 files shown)


serprex (approved):

jgao54 (approved):

flow/cmd/codegen/flow_config_converter.go
```diff
@@ -0,0 +1,197 @@
+package main
```
jgao54, line 1: Some questions/comments:

1. for my understanding, what does removing a new field looks like going forward with this change? iiuc it would require:
a. remove the field from FlowConnectionConfigsCore
b. updating the codegen template as needed
c. update FlowConnectionConfigsToCore call site to pass in new args as needed

Last step can be a bit tedious, but given it's not frequent we remove fields, should be fine.

2.  One thing i do worry a little is it's possible to accidentally use the wrong message in the proto file, i.e. need to make sure that requests are using  `FlowConnectionConfigs` but internal ones  like `DropFlowInput` are using `FlowConnectionConfigsCore`. Or is there a simpler way to make sense of this?

3. One last nit-pick is that may be a bit confusing is that FlowConnectionConfigs also has a bunch of removed/reserved fields from the past, so without context and just from reading comments at the proto files, it is a little confusing why we have removed from from both. 

Overall the approach makes sense. Just want to bring these things up in case there is a better solution; or if there's something i misunderstood here please lmk!

ilidemi: 1. Yup, that's the process. Right now there's one that has to be removed if we want to support the use case, in general we stick to keeping the functionality around, and if there is a critical mass of changes at some point we'd do API versioning. Current change is too small to require versioning, and the codegen would fit right in with contract conversions when/if that comes, only the helper signature would potentially need to change.

2. Added a linter enforcing the FlowConnectionConfigs to stick to cmd/ and e2e/. The infrastructure around a small solution is growing but I guess with API versioning we'd also have contracts we don't want to be used outside API.

3. I don't know if they're technically needed (as Temporal is I think using JSON which doesn't know about reserved) but it helps to have the two in lockstep, just for easy visual comparison. I'd introduce new fields to Core with immediately reserving that number in API.

PR #3607, opened 2025-10-15
Spew full structure for all flow_errors

ilidemi:
Chasing DB errors is getting ridiculous lately, adding spew logging that looks like this:
```
(*utils.connectionError)(0x14000e024c0)({
  message: (string) (len=25) "connection pool exhausted",
  wrapped: (*connpostgres.queryExecutionError)(0x14000e024a0)({
    query: (string) (len=33) "SELECT * FROM users WHERE id = $1",
    wrapped: (*errors.errorString)(0x140008136b0)({
      s: (string) (len=15) "network timeout"
    })
  })
})
```

### DIFF
--- a/flow/alerting/alerting.go
+++ b/flow/alerting/alerting.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"log/slog"
 	"net"
-	"reflect"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


flow/alerting/alerting.go
```diff
@@ -456,7 +463,10 @@ func (a *Alerter) logFlowErrorInternal(
 	logger := internal.LoggerFromCtx(ctx)
 	inErrWithStack := fmt.Sprintf("%+v", inErr)
 	errError := inErr.Error()
-	loggerFunc(errError, slog.String("stack", inErrWithStack))
+	loggerFunc(errError,
+		slog.String("stack", inErrWithStack),
+		slog.String("type", reflect.TypeOf(inErr).String()),
+		slog.String("spew", errSpew.Sdump(inErr)))
```
jgao54, line 467: good idea 👍

flow/alerting/alerting.go
```diff
@@ -8,11 +8,13 @@ import (
 	"io"
 	"log/slog"
 	"net"
+	"reflect"
 	"slices"
 	"strings"
 	"time"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/davecgh/go-spew/spew"
```
serprex, line 16: go.mod should be updated, as this will no longer be an indirect dependency

PR #3610, opened 2025-10-16
Classify slot creation attempt on Postgres reader in multi-az

Amogh-Bharadwaj:
Encountered this Postgres error:
```
ERROR: Creating logical replication slot peerflow_slot_mirror_1cd7f87b__d143__4cea__a247__a2acc5f5b746 is not supported on the Multi-AZ DB cluster reader node. Create the replication slot from the writer node instead. (SQLSTATE XX000)
```

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -197,6 +197,7 @@ var (
 	ErrorNotifyPostgresSlotMemalloc = ErrorClass{
 		Class: "NOTIFY_POSTGRES_SLOT_MEMALLOC", action: NotifyUser,
 	}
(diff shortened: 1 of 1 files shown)


flow/alerting/classifier.go
```diff
@@ -497,6 +500,9 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 				return ErrorNotifyPostgresSlotMemalloc, pgErrorInfo
 			}
 
+			if strings.Contains(pgErr.Message, "Create the replication slot from the writer node instead") {
```
heavycrystal, line 536: nit: comment about being RDS specific please

heavycrystal (approved):

PR #3612, opened 2025-10-16
[nexus] base64 decode SSH private key before use

heavycrystal:
Also fixing INSERT statement for `create_qrep_flow_job_entry`, it had more placeholders than columns post https://github.com/PeerDB-io/peerdb/pull/3493

### DIFF
--- a/nexus/postgres-connection/Cargo.toml
+++ b/nexus/postgres-connection/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2024"
 
 [dependencies]
 anyhow = "1"
+base64 = "0.22"
 futures-util = { version = "0.3", default-features = false, features = ["io"] }
 pt = { path = "../pt" }
 rustls = { version = "0.23", default-features = false, features = ["aws-lc-rs"] }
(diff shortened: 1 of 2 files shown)


nexus/postgres-connection/src/lib.rs
```diff
@@ -94,7 +95,21 @@ pub async fn create_tunnel(
...
-        session.userauth_pubkey_memory(&ssh_config.user, None, &ssh_config.private_key, None)?;
+        let private_key_bytes = base64::engine::general_purpose::STANDARD
+            .decode(&ssh_config.private_key)
+            .map_err(|e| {
+                io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!("Failed to decode private key: {e}"),
+                )
+            })?;
+        let private_key = String::from_utf8(private_key_bytes).map_err(|e| {
```
serprex: you can avoid copy using `str::from_utf8`

nexus/postgres-connection/Cargo.toml
```diff
@@ -18,3 +18,4 @@ tokio-util = { version = "0.7", features = ["compat"] }
 tokio-stream = "0.1"
 tracing.workspace = true
 urlencoding = "2"
+base64 = "0.22"
```
serprex: be alphabetical

PR #3613, opened 2025-10-16
pg qrep: avro schema nullable by default

serprex:
despite our efforts, nulled values still frequently leak through. put behind setting


ilidemi (commented):
Logic LGTM. Please add test with and without the flag

PR #3614, opened 2025-10-17
make settings versioning more robust

jgao54:
- Make sure `type_json_skip_duplicated_paths` is also conditionally set only if clickhouse version supports it.
- Create a setting_util.go file to centralize settings definition and minimum versioning, to make setting generation more robust. This has been an issue in the past, and we'll likely need to continue to add more settings later.
- Separately, added some comments for how PeerDB Internal Version System works

We also have a number of settings that we set at the client level [here](https://github.com/PeerDB-io/peerdb/blob/a6d0527bb3443dac36097e0a47affc286c8eab23/flow/connectors/clickhouse/clickhouse.go#L243-L260). Not sure if client automatically set settings based on server version, but since we don't check ch version there explicitly in code, that seems to be the case. The caveats of these settings is that they are not backwards compatible, since this is on the peer creation flow and not mirror creation flow, so internal version is not currently available if needed.

The change in this PR only affects query-level settings. It may be possible that some of these query-level SETTINGs could instead be client-level (i.e. `type_json_skip_duplicated_paths`), but query-level settings seem to be more flexible with backwards compatibility support.

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -264,7 +264,7 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
 		}
 
 		if allowNullableKey {
-			stmtBuilder.WriteString(NewChSettingsString(chVersion, CHSettingEntry{SettingAllowNullableKey, "1"}))
+			stmtBuilder.WriteString(NewCHSettingsString(chVersion, CHSettingEntry{SettingAllowNullableKey, "1"}))
 		}
 
 		if c.Config.Cluster != "" {
--- a/flow/connectors/clickhouse/setting_util.go
+++ b/flow/connectors/clickhouse/setting_util.go
@@ -47,8 +47,8 @@ func GetMinVersion(name CHSetting) (chproto.Version, bool) {
 	return chproto.Version{}, false
 }
 
-// NewChSettingsString is a one-liner method to generate an immutable settings string
-func NewChSettingsString(version *chproto.Version, settings ...CHSettingEntry) string {
+// NewCHSettingsString is a one-liner method to generate an immutable settings string
+func NewCHSettingsString(version *chproto.Version, settings ...CHSettingEntry) string {
 	sg := NewCHSettings(version)
 	for _, setting := range settings {
 		sg.Add(setting.key, setting.val)
--- a/flow/connectors/clickhouse/setting_util_test.go
+++ b/flow/connectors/clickhouse/setting_util_test.go
@@ -48,12 +48,12 @@ func TestCHSettingsVersionFiltering(t *testing.T) {
 	// settings not having minimum version should be included
 	_, minVersionExist := GetMinVersion(SettingThrowOnMaxPartitionsPerInsertBlock)
 	require.False(t, minVersionExist)
-	result = NewChSettingsString(&chproto.Version{Major: 25, Minor: 8, Patch: 0},
+	result = NewCHSettingsString(&chproto.Version{Major: 25, Minor: 8, Patch: 0},
 		CHSettingEntry{SettingThrowOnMaxPartitionsPerInsertBlock, "0"})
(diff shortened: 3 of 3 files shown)


flow/shared/constants.go
```diff
@@ -15,10 +15,30 @@ const (
...
+// ## How to add a new version:
+//  1. Add a new constant below with a descriptive name and a comment explaining the change
+//  2. Add version checks in the relevant connector code (e.g., if version >= InternalVersion_MyFeature)
+//  3. Add e2e test to verify both old and new versions work correctly (old mirrors keep old behavior, new mirrors get new behavior)
+
 const (
 	InternalVersion_First uint32 = iota
+	// Postgres: vector types ("vector", "halfvec", "sparsevec") replicated as float arrays instead of string
 	InternalVersion_PgVectorAsFloatArray
+	// MongoDB: rename `fullDocument` column to `doc`
```
ilidemi:
```suggestion
	// MongoDB: rename `_full_document` column to `doc`
```

flow/shared/constants.go
```diff
@@ -15,10 +15,30 @@ const (
...
 	IntervalVersion_MongoDBFullDocumentColumnToDoc
+	// All: setting json_type_escape_dots_in_keys = true when inserting JSON column to ClickHouse (only impact MongoDB today)
```
ilidemi:
```suggestion
	// All: setting json_type_escape_dots_in_keys = true when inserting JSON column to ClickHouse (only impacts MongoDB today)
```

flow/connectors/clickhouse/setting_util.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func GetMinVersion(name CHSetting) (chproto.Version, bool) {
+	if minVersion, exists := settingMinVersions[name]; exists {
+		return minVersion, true
+	}
+	return chproto.Version{}, false
+}
+
+type SettingGenerator struct {
```
ilidemi:
```suggestion
type CHSettings struct {
```

flow/connectors/clickhouse/setting_util.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func NewSettingGenerator(version *chproto.Version) *SettingGenerator {
+	return &SettingGenerator{
+		settings:  make(map[CHSetting]string),
+		chVersion: version,
+	}
+}
+
+func (sg *SettingGenerator) AddSetting(name CHSetting, value string) *SettingGenerator {
```
ilidemi:
```suggestion
func (sg *SettingGenerator) Add(name CHSetting, value string) *SettingGenerator {
```

flow/connectors/clickhouse/setting_util.go
```diff
@@ -0,0 +1,91 @@
...
+	sg.settings[name] = value
+	return sg
+}
+
+// ToString generates settings string ' SETTINGS <key1> = <val1>, <key2> = <val2>, ...';
+// If ClickHouse version is set in the SettingGenerator, settings that do not meet CH version
+// requirement will be filtered out. Otherwise, all settings are included.
+func (sg *SettingGenerator) ToString() string {
```
ilidemi:
```suggestion
func (sg *SettingGenerator) String() string {
```

flow/connectors/clickhouse/normalize.go
```diff
@@ -264,7 +264,9 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
 		}
 
 		if allowNullableKey {
-			stmtBuilder.WriteString(" SETTINGS allow_nullable_key = 1")
+			settingGenerator := NewSettingGenerator(chVersion)
+			settingGenerator.AddSetting(SettingAllowNullableKey, "1")
+			stmtBuilder.WriteString(settingGenerator.ToString())
```
ilidemi: Can this one have a one-liner API?

ilidemi (approved):

PR #3622, opened 2025-10-22
Add time limit for get variant to be 10s

iskakaushik:
During DropFlow we have get flow metadata, there is a 1 min timeout. But the failure can take longer.

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -1747,7 +1747,9 @@ func (a *FlowableActivity) GetFlowMetadata(
 		// Use a short timeout for optional variant detection to avoid consuming entire activity timeout
 		variantCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
 		defer cancel()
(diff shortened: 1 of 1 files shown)


flow/activities/flowable.go
```diff
@@ -1744,8 +1744,11 @@ func (a *FlowableActivity) GetFlowMetadata(
 
 	// Detect source database variant
 	if input.SourceName != "" {
-		if srcConn, err := connectors.GetByNameAs[connectors.DatabaseVariantConnector](ctx, nil, a.CatalogPool, input.SourceName); err == nil {
-			if variant, variantErr := srcConn.GetDatabaseVariant(ctx); variantErr == nil {
+		// Use a short timeout for optional variant detection to avoid consuming entire activity timeout
+		variantCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+		defer cancel()
+		if srcConn, err := connectors.GetByNameAs[connectors.DatabaseVariantConnector](variantCtx, nil, a.CatalogPool, input.SourceName); err == nil {
```
```suggestion
		if srcConn, err := connectors.GetByNameAs[connectors.DatabaseVariantConnector](
		    variantCtx, nil, a.CatalogPool, input.SourceName,
		); err == nil {
```

PR #3623, opened 2025-10-22
simplify

jgao54:


flow/connectors/clickhouse/setting_util.go
```diff
@@ -48,11 +48,9 @@ func GetMinVersion(name CHSetting) (chproto.Version, bool) {
 }
 
 // NewCHSettingsString is a one-liner method to generate an immutable settings string
-func NewCHSettingsString(version *chproto.Version, settings ...CHSettingEntry) string {
+func NewCHSettingsString(version *chproto.Version, setting CHSettingEntry) string {
```
ilidemi: Does CHSettingEntry need to be here?

PR #3624, opened 2025-10-22
add support for CTID bucketing with snapshotNumPartitionsOverride

jgao54:
PeerDB supports parallel snapshotting to optimize initial load time. To do this there are two ways today:

1) we compute the total row count in a table, bucket the data evenly by watermark column -- this is the default behavior and provides even distribution of data for parallel initial load. However calculating total row count can be slow on large tables.

2) with `SnapshotNumPartitionsOverride` enabled, rather than calculating num partitions,  we fetch the min / max values of watermark column, and increment the column evenly to get the partition ranges. In this case we can't guarantee even distribution of data across partitions, but can speed up initial snapshot on large table by bypassing calculating total row count.

This change handles the case for 2) when watermark column is not explicitly passed in and defaults to `ctid`, which currently is a no-op. For append-only tables, we expect even distribution of data, so this change should result in a pure performance win. For updatable tables, this may result in uneven distribution of data partitioning, but this is already the case with approach 2). 

Thank you @alon-zeltser-cyera for the contribution.

Separate note: `SnapshotNumPartitionsOverride` was introduced to support use case where num partitions is explicitly provided. There's no reason that it has to be tied to the two initial snapshot bucketing approaches, so we may want to evaluate decoupling the two concept later on if we want to provide this feature more widely.

TODO: 
- [x]  Add e2e test
- [x]  Run test against a large table

### DIFF
--- a/flow/e2e/clickhouse_test.go
+++ b/flow/e2e/clickhouse_test.go
@@ -2887,7 +2887,7 @@ func (s ClickHouseSuite) Test_Partition_By_CTID_With_Num_Partitions_Override() {
 	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(`
 			INSERT INTO %s (name, age, email) VALUES ('user_%d', %d, 'user_%d@example.com')
 		`, srcFullName, numRows+1, 25, numRows+1)))
-	EnvWaitForCount(env, s, "wait on cdc", dstTableName, "id", numRows-deletedRows)
+	EnvWaitForCount(env, s, "wait on cdc", dstTableName, "id", numRows-deletedRows+1)
 
 	rows, err := s.Conn().Query(s.t.Context(),
 		`SELECT partition_start, partition_end FROM peerdb_stats.qrep_partitions WHERE parent_mirror_name = $1


flow/connectors/postgres/qrep.go
```diff
@@ -295,6 +302,88 @@ func (c *PostgresConnector) getNumRowsPartitions(
...
+	if last != nil && last.Range != nil {
+		if lr, ok := last.Range.Range.(*protos.PartitionRange_TidRange); ok {
+			resume := tid{b: lr.TidRange.End.BlockNumber, o: lr.TidRange.End.OffsetNumber}
+			v := inc(resume)
+			resumeFrom = &v
+		}
+	}
+
+	partitionHelper := utils.NewPartitionHelper(c.logger)
+	for i := int64(0); i < numPartitions; i++ {
```
serprex:
```suggestion
	for i := range numPartitions {
```

flow/connectors/postgres/qrep.go
```diff
@@ -295,6 +302,88 @@ func (c *PostgresConnector) getNumRowsPartitions(
...
+		end := tid{b: endBlockExclusive - 1, o: 0xFFFF}
+		start := tid{b: startBlock, o: 0}
+		if resumeFrom != nil && cmp(end, *resumeFrom) < 0 {
+			continue
+		}
+		if resumeFrom != nil && cmp(start, *resumeFrom) < 0 && cmp(*resumeFrom, end) <= 0 {
+			start = *resumeFrom
+		}
+
+		if err := partitionHelper.AddPartition(pgtype.TID{BlockNumber: start.b, OffsetNumber: uint16(start.o), Valid: true}, pgtype.TID{BlockNumber: end.b, OffsetNumber: uint16(end.o), Valid: true}); err != nil {
```
serprex:
```suggestion
		if err := partitionHelper.AddPartition(
		    pgtype.TID{BlockNumber: start.b, OffsetNumber: uint16(start.o), Valid: true},
		    pgtype.TID{BlockNumber: end.b, OffsetNumber: uint16(end.o), Valid: true},
		); err != nil {
```

flow/connectors/postgres/qrep.go
```diff
@@ -295,6 +302,91 @@ func (c *PostgresConnector) getNumRowsPartitions(
...
+	blocksQuery := "SELECT (pg_relation_size(to_regclass($1)) / current_setting('block_size')::int)::bigint"
+	var totalBlocks pgtype.Int8
+	if err := tx.QueryRow(ctx, blocksQuery, parsedWatermarkTable.String()).Scan(&totalBlocks); err != nil {
+		return nil, fmt.Errorf("failed to get relation blocks: %w", err)
+	}
+	if totalBlocks.Int64 <= 0 {
+		return nil, nil
+	}
+
+	type tid struct {
```
serprex: instead of using new `tid` type, use `pgtype.TID`. Instead of using `*pgtype.TID` have the nil case represented with `Valid: false`

flow/connectors/postgres/qrep.go
```diff
@@ -295,6 +302,91 @@ func (c *PostgresConnector) getNumRowsPartitions(
...
+		b uint32
+		o uint32
+	}
+	cmp := func(a tid, b tid) int {
```
serprex: should mirror https://github.com/PeerDB-io/peerdb/blob/c020dd474098a73aeb2c093eb505c2666b0dcc45/flow/connectors/utils/partition.go#L224 using `cmp.Compare`

flow/connectors/postgres/qrep.go
```diff
@@ -264,6 +264,13 @@ func (c *PostgresConnector) getNumRowsPartitions(
 
 		return partitionHelper.GetPartitions(), nil
 	} else {
+		// Special handling for CTID when user wants a fixed number of partitions:
+		// build approximate physical (block-based) TID partitions without sorting or MIN/MAX(ctid).
+		if strings.EqualFold(config.WatermarkColumn, "ctid") {
```
serprex:
```suggestion
		if config.WatermarkColumn == "ctid" {
```
only lowercase ctid is special

flow/connectors/postgres/qrep.go
```diff
@@ -295,6 +312,85 @@ func (c *PostgresConnector) getNumRowsPartitions(
...
+	tx pgx.Tx,
+	parsedWatermarkTable utils.SchemaTable,
+	numPartitions int64,
+	last *protos.QRepPartition,
+) ([]*protos.QRepPartition, error) {
+	if numPartitions == 0 {
+		return nil, fmt.Errorf("numPartitions must be greater than 0")
+	}
+
+	blocksQuery := "SELECT (pg_relation_size(to_regclass($1)) / current_setting('block_size')::int)::bigint"
```
ilidemi, line 327: Any reason to not go for min(ctid), max(ctid)? Like with the example from the test the bottom half will be dead and min..max will actually cover the full range
(in this case the logic would move into a case under PartitionHelper.AddPartitionsWithRange)

jgao54: `ctid` is not indexed so calculating min/max would require a sequential scan, so this approach bypasses this for performance reasons. While it's less precise than `select min(ctid), max(ctid) from <table>` (i.e. it includes deletes), it's much faster.

ilidemi: Riiight, when the rows are actually recognized as dead that's VACUUM FULL and ctids start from 0 anyway

Issue #3546, opened 2025-10-01
[BUG] Error when syncing a NULL JSONB column from Postgres to ClickHouse


aldofunes:
# Settings

I have enabled both these settings:

- `PEERDB_CLICKHOUSE_ENABLE_JSON: true`
- `PEERDB_NULLABLE: true`

# Context

I'm using the docker tag `stable-v0.34.2`

Here is a sample table DDL

```sql
create table meta
(
    id         uuid primary key
    metadata   jsonb,
    created_at timestamp with time zone not null default CURRENT_TIMESTAMP
);
```

The columns are created correctly in CH:

```sql
create table default.publi_meta
(
    id                 UUID,
    created_at         DateTime64(6),
    metadata           Nullable(String),
    _peerdb_synced_at  DateTime64(9) default now64(),
    _peerdb_is_deleted Int8,
    _peerdb_version    Int64
)
    engine = ReplacingMergeTree(_peerdb_version)
        PRIMARY KEY id
        ORDER BY id
        SETTINGS index_granularity = 8192;
```

# Problem

However, when the initial sync is triggered, PeerDB seems to attempt to cast the empty string to a JSON because I get these errors in the logs

```
failed to sync records: QRepSync Error: code: 349, message: Cannot convert NULL value to non-Nullable type: while executing 'FUNCTION CAST(__table1.meta_9 :: 8, 'JSON'_String :: 10) -> CAST(__table1.meta_9, 'JSON'_String) JSON : 7'
```

aldofunes:
I was browsing the commit history and noticed this one: https://github.com/PeerDB-io/peerdb/commit/52591d5c1d426994ff0bed24bc15cbff533012d0, which looks like it intends to address this issue. Am I missing something?

PR #3626, opened 2025-10-24
Support Nullable(JSON) in ClickHouse

aldofunes:
This is my attempt at addressing issue #3546.

I managed to make it work on my setup, which pulls NULL JSONB columns in a Postgres db into Nullabl(JSON) columns in ClickHouse

### DIFF
--- a/flow/connectors/clickhouse/normalize_query.go
+++ b/flow/connectors/clickhouse/normalize_query.go
@@ -246,19 +246,25 @@ func (t *NormalizeQueryGenerator) BuildQuery(ctx context.Context) (string, error
 					peerdb_clickhouse.QuoteIdentifier(dstColName),
 				)
 			}
-		case "JSON", "Nullable(JSON)":
-			fmt.Fprintf(&projection,
-				"JSONExtractString(_peerdb_data, %s)::JSON AS %s,",
-				peerdb_clickhouse.QuoteLiteral(colName),
-				peerdb_clickhouse.QuoteIdentifier(dstColName),
-			)
-			if t.enablePrimaryUpdate {
-				fmt.Fprintf(&projectionUpdate,
-					"JSONExtractString(_peerdb_match_data, %s)::JSON AS %s,",
+			case "JSON", "Nullable(JSON)":
+				stringType := strings.Replace(clickHouseType, "JSON", "String", 1)
+
+				fmt.Fprintf(&projection,
+					"JSONExtract(_peerdb_data, %s, '%s')::%s AS %s,",
 					peerdb_clickhouse.QuoteLiteral(colName),
+					stringType,
+					clickHouseType,
 					peerdb_clickhouse.QuoteIdentifier(dstColName),
 				)
-			}
+				if t.enablePrimaryUpdate {
+					fmt.Fprintf(&projectionUpdate,
+						"JSONExtract(_peerdb_match_data, %s, '%s')::%s AS %s,",
+						peerdb_clickhouse.QuoteLiteral(colName),
+						stringType,
+						clickHouseType,
+						peerdb_clickhouse.QuoteIdentifier(dstColName),
+					)
+				}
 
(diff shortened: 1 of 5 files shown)


aldofunes:
@jgao54, you are totally right, how did I miss that?

I found that in the clone step, the avro files have a different schema from the CDC step.

Here's an example:

`java -jar avro-tools-1.12.1.jar getschema cdc_file.avro`
```json
{
  "type" : "record",
  "name" : "_peerdb_raw_aggregators_basculin_dwh",
  "fields" : [ {
    "name" : "_peerdb_uid",
    "type" : "string"
  }, {
    "name" : "_peerdb_timestamp",
    "type" : "long"
  }, {
    "name" : "_peerdb_destination_table_name",
    "type" : "string"
  }, {
    "name" : "_peerdb_data",
    "type" : "string"
  }, {
    "name" : "_peerdb_record_type",
    "type" : [ "null", "long" ]
  }, {
    "name" : "_peerdb_match_data",
    "type" : [ "null", "string" ]
  }, {
    "name" : "_peerdb_batch_id",
    "type" : [ "null", "long" ]
  }, {
    "name" : "_peerdb_unchanged_toast_columns",
    "type" : [ "null", "string" ]
  } ]
}
```

`java -jar avro-tools-1.12.1.jar getschema clone_file.avro`
```json
{
  "type" : "record",
  "name" : "aggregators_basculin_activity_views",
  "fields" : [ {
    "name" : "user_id_0",
    "type" : {
      "type" : "string",
      "logicalType" : "uuid"
    }
  }, {
    "name" : "operation_id_1",
    "type" : {
      "type" : "string",
      "logicalType" : "uuid"
    }
  }, {
    "name" : "data_2",
    "type" : [ "null", "string" ]
  }, {
    "name" : "created_at_3",
    "type" : {
      "type" : "long",
      "logicalType" : "timestamp-micros"
    }
  }, {
    "name" : "updated_at_4",
    "type" : {
      "type" : "long",
      "logicalType" : "timestamp-micros"
    }
  } ]
}
```

In other words, the CDC step is wrapping the record in an envelope, while the Clone step is dumping the record by itself.

For the life of me, I could not get the tests to run on my machine, but I was able to figure out that in the CDC step, it needs to first extract the `jsonb` column to a `Nullable(String)` and then cast that to a `Nullable(JSON)`, using `JSONExtractString()` does not work because `null` values get converted to an empty string (instead of `null`) and hence, the cast to `Nullable(JSON)` fails.

I basically had to split the handling of both `JSON` and `Nullable(JSON)` as separate steps. I have zero experience with golang, so please forgive me if my approach is too naive.

flow/connectors/clickhouse/normalize_query.go
```diff
@@ -193,7 +193,7 @@ func (t *NormalizeQueryGenerator) BuildQuery(ctx context.Context) (string, error
 					peerdb_clickhouse.QuoteIdentifier(dstColName),
 				)
 			}
-		case "JSON", "Nullable(JSON)":
+		case "JSON":
```
jgao54: Can be combined into:

```suggestion
		case "JSON", "Nullable(JSON)":
			stringType := strings.Replace(clickHouseType, "JSON", "String", 1)
			fmt.Fprintf(&projection,
				"JSONExtract(_peerdb_data, %s, '%s')::%s AS %s,",
				peerdb_clickhouse.QuoteLiteral(colName),
				stringType,
				clickHouseType,
				peerdb_clickhouse.QuoteIdentifier(dstColName),
			)
			if t.enablePrimaryUpdate {
				fmt.Fprintf(&projectionUpdate,
					"JSONExtract(_peerdb_match_data, %s, '%s')::%s AS %s,",
					peerdb_clickhouse.QuoteLiteral(colName),
					stringType,
					clickHouseType,
					peerdb_clickhouse.QuoteIdentifier(dstColName),
				)
			}

```

PR #3635, opened 2025-10-30
skip unsupported validations for documentdb

jgao54:
- DocumentDB does not provide infos like oplog retention and storage engine, so skip validations for them.
- Remove explicitly setting `SetFullDocumentBeforeChange` . this is `off` by default for MongoDB so setting this is redundant. For DocumentDB variant, setting this field is not supported by the server (`BSON field '$changeStream.fullDocumentBeforeChange' is an unknown field.`). The more complicated solution is to look at host domain to try to determine if it's documentdb, and optionally set this setting if it's not; the simpler solution here is just not try to set it at all. 
 
Notes:
- DocumentDB has a `change_stream_log_retention_duration` config that needs to be changed via AWS ; will provide separate documentation for this. There's no good way to validate this from a MongoDB client at the moment. 
- See [DocumentDB version support](https://docs.aws.amazon.com/documentdb/latest/developerguide/docdb-version-support-dates.html). 3.6 and 4.0 requires change stream to be opened from primary node, which is not ideal. Since we can't easily test it, will scope to latest version (5.0) for now. In MongoDB we currently support 4.4+, so no change needed here. 

Testing:
- [x] validate changes unblocks peer creation (via ssh tunnel to a documentdb instance) and mirror creation (validation works)
- [x] test both initial snapshot and cdc with a diverse set of data types to make sure replication works

### DIFF
--- a/flow/connectors/mongo/cdc.go
+++ b/flow/connectors/mongo/cdc.go
@@ -78,8 +78,7 @@ func (c *MongoConnector) GetTableSchema(
 func (c *MongoConnector) SetupReplication(ctx context.Context, input *protos.SetupReplicationInput) (model.SetupReplicationResult, error) {
 	changeStreamOpts := options.ChangeStream().
 		SetComment("PeerDB changeStream").
-		SetFullDocument(options.UpdateLookup).
-		SetFullDocumentBeforeChange(options.Off)
+		SetFullDocument(options.UpdateLookup)
 
 	pipeline, err := createPipeline(nil)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


flow/pkg/mongo/validation.go
```diff
@@ -15,6 +16,8 @@ const (
 
 	ReplicaSet     = "ReplicaSet"
 	ShardedCluster = "ShardedCluster"
+
+	DocumentDBDomain = "docdb.amazonaws.com"
```
ilidemi, line 20: With private link the hostnames are going to be weird/mangled, seems like the storage engine check is hard then?

jgao54: ss.Host uses domain fetched from DocumentDB server-side so  it will still be the actual private DNS that customer has.

ilidemi (approved):

PR #3640, opened 2025-11-01
improve normalize logic readability

jgao54:
The clickhouse normalize logic/metrics uses a mix of startBatchId/normBatchId, endBatchId/syncBatchId. This makes the code more confusing than it needs to be:
- `startBatchID` metric attribute can be misleading because StartBatchID should be inclusive according to `NormalizeResponse`, but it is used to represent last normalized batch ID -- changed metrics to always use `lastNormBatchID` instead.
- `syncBatchID` metric attribute can be misleading because of the GroupNormalize setting, since the `endBatchID` is determined by `lastNormBatchId + batchGroup`, and not necessarily `req.SyncBatchID` -- changed it to always use `endBatchID` instead.
- when executing per-table normalize queries, per-table metric attribute can be misleading, since individual queries can have different starting batchID (i.e. if previous iteration failed half way), so make sure the query-level metric attribute is used.
- removing a footgun where `NormalizeQueryGenerator` was created twice, once to generate query and another partially created to pass query info; this avoids accidentally using unset attributes.

Previously considered standardize on [startBatchID, endBatchID] throughout, but keeping (lastNormBatchID, endBatchID] to avoid introducing new concept.

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -409,31 +409,29 @@ func (c *ClickHouseConnector) NormalizeRecords(
 	ctx context.Context,
 	req *model.NormalizeRecordsRequest,
 ) (model.NormalizeResponse, error) {
-	normBatchID, err := c.GetLastNormalizeBatchID(ctx, req.FlowJobName)
+	lastNormBatchID, err := c.GetLastNormalizeBatchID(ctx, req.FlowJobName)
 	if err != nil {
 		c.logger.Error("[clickhouse] error while getting last sync and normalize batch id", slog.Any("error", err))
 		return model.NormalizeResponse{}, err
 	}
 
 	// normalize has caught up with sync, chill until more records are loaded.
-	if normBatchID >= req.SyncBatchID {
+	if lastNormBatchID >= req.SyncBatchID {
 		return model.NormalizeResponse{
-			StartBatchID: normBatchID,
+			StartBatchID: lastNormBatchID,
 			EndBatchID:   req.SyncBatchID,
 		}, nil
 	}
 
-	endBatchID := req.SyncBatchID
 	groupBatches, err := internal.PeerDBGroupNormalize(ctx, req.Env)
-	if err != nil {
+	if err != nil || groupBatches <= 0 {
 		c.logger.Error("failed to lookup PEERDB_GROUP_NORMALIZE, only normalizing 1 batch")
 		groupBatches = 1
 	}
-	if groupBatches > 0 {
-		endBatchID = min(endBatchID, normBatchID+groupBatches)
-	}
 
-	if err := c.copyAvroStagesToDestination(ctx, req.FlowJobName, normBatchID, endBatchID, req.Env, req.Version); err != nil {
+	endBatchID := min(req.SyncBatchID, lastNormBatchID+groupBatches)
+
+	if err := c.copyAvroStagesToDestination(ctx, req.FlowJobName, lastNormBatchID, endBatchID, req.Env, req.Version); err != nil {
 		return model.NormalizeResponse{}, fmt.Errorf("failed to copy avro stages to destination: %w", err)
 	}
 
@@ -441,7 +439,7 @@ func (c *ClickHouseConnector) NormalizeRecords(
 		ctx,
 		req.FlowJobName,
 		endBatchID,
-		normBatchID,
+		lastNormBatchID,
 		req.TableNameSchemaMapping,
 	)
 	if err != nil {
@@ -463,9 +461,10 @@ func (c *ClickHouseConnector) NormalizeRecords(
 	if err != nil {
 		return model.NormalizeResponse{}, err
 	}
+	// parallelize normalization up to the number of destination tables
 	parallelNormalize = min(max(parallelNormalize, 1), len(destinationTableNames))
-	c.logger.Info("[clickhouse-cdc] inserting batch...",
-		slog.Int64("startBatchID", normBatchID),
+	c.logger.Info("[clickhouse-cdc] inserting batches...",
+		slog.Int64("lastNormBatchID", lastNormBatchID),
 		slog.Int64("endBatchID", endBatchID),
 		slog.Int("connections", parallelNormalize))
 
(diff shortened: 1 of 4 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -409,39 +409,39 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
 	}
+	if lastNormalizeBatchId > req.SyncBatchID {
+		return model.NormalizeResponse{}, fmt.Errorf(
+			"[clickhouse] unexpected: syncBatchId (%d) is behind lastNormalizeBatchId (%d)", req.SyncBatchID, lastNormalizeBatchId)
+	}
 
-	endBatchID := req.SyncBatchID
 	groupBatches, err := internal.PeerDBGroupNormalize(ctx, req.Env)
-	if err != nil {
+	if err != nil || groupBatches <= 0 {
```
ilidemi, line 427: 0 wouldn't be an error as it's a default?

flow/connectors/clickhouse/normalize.go
```diff
@@ -409,39 +409,39 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
 		return model.NormalizeResponse{
-			StartBatchID: normBatchID,
-			EndBatchID:   req.SyncBatchID,
+			StartBatchID: lastNormalizeBatchId,
+			EndBatchID:   lastNormalizeBatchId,
 		}, nil
 	}
+	if lastNormalizeBatchId > req.SyncBatchID {
+		return model.NormalizeResponse{}, fmt.Errorf(
+			"[clickhouse] unexpected: syncBatchId (%d) is behind lastNormalizeBatchId (%d)", req.SyncBatchID, lastNormalizeBatchId)
```
ilidemi: Is there a way to confirm from logs that it's for sure not happening before making it an error? The `>=` is a pattern generally protecting from overruns in for loops

flow/connectors/clickhouse/normalize.go
```diff
@@ -526,38 +527,39 @@ func (c *ClickHouseConnector) NormalizeRecords(
 				if insertIntoSelectQuery.Part == numParts-1 {
 					c.logger.Info("[clickhouse] set last normalized batch id for table",
 						slog.String("table", insertIntoSelectQuery.TableName),
-						slog.Int64("syncBatchID", endBatchID),
-						slog.Int64("lastNormalizedBatchID", normBatchID),
+						slog.Int64("endBatchId", insertIntoSelectQuery.endBatchId),
+						slog.Int64("startBatchId", insertIntoSelectQuery.startBatchId),
 						slog.Int("parallelWorker", i))
-					err := c.SetLastNormalizedBatchIDForTable(ctx, req.FlowJobName, insertIntoSelectQuery.TableName, endBatchID)
+					err := c.SetLastNormalizedBatchIDForTable(ctx, req.FlowJobName, insertIntoSelectQuery.TableName, insertIntoSelectQuery.endBatchId)
```
ilidemi: Seems like this could run before a previous one runs in parallel? Great that we're reviewing the code like this

flow/connectors/clickhouse/normalize.go
```diff
@@ -610,30 +612,30 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
 		c.logger.Error("[clickhouse] error while updating normalize batch id",
-			slog.Int64("batchID", endBatchID), slog.Any("error", err))
+			slog.Int64("batchID", endBatchIdInclusive), slog.Any("error", err))
 		return model.NormalizeResponse{}, err
 	}
 
 	return model.NormalizeResponse{
-		StartBatchID: normBatchID + 1,
-		EndBatchID:   endBatchID,
+		StartBatchID: startBatchIdInclusive,
```
ilidemi: I kind of like the two/three-pointer model more than introducing new concepts to keep track of. Is it the opposite for you?

flow/connectors/clickhouse/normalize.go
```diff
@@ -507,17 +508,17 @@ func (c *ClickHouseConnector) NormalizeRecords(
 
 			for insertIntoSelectQuery := range queries {
 				c.logger.Info("executing INSERT command to ClickHouse table",
-					slog.Int64("syncBatchID", endBatchID),
-					slog.Int64("normalizeBatchID", normBatchID),
+					slog.Int64("endBatchId", insertIntoSelectQuery.endBatchId),
```
ilidemi: Shame on the one who added these logs (me)

flow/connectors/clickhouse/normalize.go
jgao54: agree with this, kept this model

flow/connectors/clickhouse/normalize.go
jgao54: default is 1

flow/connectors/clickhouse/normalize.go
jgao54: decide to leave this as is for now.

was treating this more like an assertion (if we get >, it means something wrong with the implementation, i.e. normBatchId somehow got ahead of syncBatchId) -- but  for now keeping current code since it's beyond scope of this PR, in case i have overlooked something.

flow/connectors/clickhouse/normalize.go
jgao54: The queries are parallelized by table, so processing out-of-order across tables is acceptable

flow/connectors/clickhouse/normalize.go
```diff
@@ -474,19 +473,18 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
 	}
+	numParts = max(numParts, 1)
 
 	// This is for cases where currently normalizing can take a looooong time
 	// there is no other indication of progress, so we log every 5 minutes.
 	periodicLogger := shared.Interval(ctx, 5*time.Minute, func() {
-		c.logger.Info("[clickhouse-cdc] inserting batch...",
-			slog.Int64("startBatchID", normBatchID),
+		c.logger.Warn("[clickhouse-cdc] inserting batches...",
+			slog.Int64("lastNormBatchID", lastNormBatchID),
```
jgao54, line 482: `startBatchID` is misleading here because we define startBatchID to be inclusive

```
	return model.NormalizeResponse{
		StartBatchID: normBatchID + 1,
		EndBatchID:   endBatchID,
	}, nil
```

flow/connectors/clickhouse/normalize.go
jgao54: actually you are right, forgot we have the query partitioning dimension as well.

So i think the following race condition bug is possible when numPart > 1:
- we have table t1 and t2,  numParts=2,  parallelNormalize=2, lastNormBatchId=122, endBatchId=123
- this generate queries [t1p1, t1p2, t2p1, t2p2]
- t1p1 and t1p2 runs in parallel, t1p1 fails, but t1p2 succeeds and `SetLastNormalizedBatchIDForTable` update t1's batchId to 123
- `NormalizeRecords` return an error
- on retry, `GetLastNormalizedBatchIDForTable` fetches batch 123 for t1, assumes completes even though t1p1 never got replicated

out of scope for this PR but needs a bug fix.

iskakaushik (approved):
LGTM

ilidemi:
Looks awesome

PR #3644, opened 2025-11-01
Speed up dev nexus build

ilidemi:
Build in debug mode in dev, use build caching.

Benchmarked with three scenarios. In each you start with a local checkout that was built before, then pull/code the change and see how much does the rebuild take.

**Scenario 1: all-up dependency updates** (#3632 + #3638)
Baseline: 3m33s
With debug: 1m48s
With debug+caching: 1m39s

**Scenario 2: Rust+Go dependency updates** (#3632)
Baseline: 3m18s
With debug: 1m34s
With debug+caching: 1m19s

**Scenario 3: a change in protos + a small fix in Rust usage code**
Baseline: 1m7s
With debug: 24s
With debug+caching: 16s

### DIFF
--- a/stacks/peerdb-server.Dockerfile
+++ b/stacks/peerdb-server.Dockerfile
@@ -11,7 +11,7 @@ RUN cargo chef prepare --recipe-path recipe.json
 
 FROM chef AS builder
 ENV OPENSSL_STATIC=1
-ARG CARGO_BUILD_FLAGS="--release"
+ARG BUILD_MODE="release"
 RUN apk add --no-cache build-base pkgconfig curl unzip openssl-dev openssl-libs-static
 WORKDIR /root/nexus
 COPY scripts /root/scripts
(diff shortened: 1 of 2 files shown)


stacks/peerdb-server.Dockerfile
```diff
@@ -11,17 +11,27 @@ RUN cargo chef prepare --recipe-path recipe.json
...
 COPY protos /root/protos
 WORKDIR /root/nexus
-RUN cargo build --release --bin peerdb-server
+# Build the actual binary with cache mounts
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/usr/local/cargo/git \
+    --mount=type=cache,target=/root/nexus/target \
+    cargo build ${CARGO_BUILD_FLAGS} --bin peerdb-server && \
+    mkdir -p /root/target && \
+    cp target/*/peerdb-server /root/target/
```
jgao54: nit: `target/release/peerdb-server` to be safe (or specify mount cache to point to release folder only)

given that we are mounting everything in target (i.e. `--mount=type=cache,target=/root/nexus/target`), could accidentally pick up non-release builds, and introduce ambiguity.

jgao54 (approved):
small nit, lgtm!🔥

stacks/peerdb-server.Dockerfile
ilidemi: was hoping to avoid conditionals that docker doesn't do well but a really good point

PR #3663, opened 2025-11-05
bug fixes for NormalizeRecords

jgao54:
- use errCtx in goroutine
- close channel on error
- move logging so it's reporting for every query

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -540,14 +540,14 @@ func (c *ClickHouseConnector) NormalizeRecords(
 	}
 
 	// wrap query generation logic in a function to ensure queriesCh always closes once
-	response, err := func() (model.NormalizeResponse, error) {
+	if err := func() error {
 		defer close(queriesCh)
 
 		for _, tbl := range destinationTableNames {
 			lastNormBatchIDForTable, err := c.GetLastNormalizedBatchIDForTable(ctx, req.FlowJobName, tbl)
 			if err != nil {
 				c.logger.Error("[clickhouse] error while getting last synced batch id for table", "table", tbl, slog.Any("error", err))
-				return model.NormalizeResponse{}, err
+				return err
 			}
 			c.logger.Info("[clickhouse] last normalized batch id for table",
 				"table", tbl, "lastNormBatchID", lastNormBatchIDForTable, "endBatchID", endBatchID)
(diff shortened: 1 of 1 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -524,22 +524,24 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
+					slog.Int("parallelWorker", i))
+			}
 			return nil
 		})
 	}
 
 	for _, tbl := range destinationTableNames {
 		lastNormBatchIDForTable, err := c.GetLastNormalizedBatchIDForTable(ctx, req.FlowJobName, tbl)
 		if err != nil {
+			close(queries)
```
ilidemi: This change is good, but holy the existing pattern seems error-prone. One way to make it simpler could be to wrap the whole loop in `func(){}()` with `defer close(queries)` at the beginning

jgao54: good idea

flow/connectors/clickhouse/normalize.go
```diff
@@ -524,91 +525,93 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
-	for _, tbl := range destinationTableNames {
-		lastNormBatchIDForTable, err := c.GetLastNormalizedBatchIDForTable(ctx, req.FlowJobName, tbl)
-		if err != nil {
-			c.logger.Error("[clickhouse] error while getting last synced batch id for table", "table", tbl, slog.Any("error", err))
-			return model.NormalizeResponse{}, err
-		}
-		c.logger.Info("[clickhouse] last normalized batch id for table",
-			"table", tbl, "lastNormBatchID", lastNormBatchIDForTable, "endBatchID", endBatchID)
+	// wrap query generation logic in a function to ensure queriesCh always closes once
+	response, err := func() (model.NormalizeResponse, error) {
```
ilidemi: The error is not handled

flow/connectors/clickhouse/normalize.go
```diff
@@ -524,91 +525,93 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
+					slog.Any("error", errCtx.Err()),
+					slog.Any("cause", context.Cause(errCtx)))
+				return model.NormalizeResponse{}, context.Cause(errCtx)
+			}
 		}
-	}
-	close(queries)
+		return model.NormalizeResponse{
+			StartBatchID: lastNormBatchID + 1,
+			EndBatchID:   endBatchID,
```
ilidemi: lastNormBatchID and endBatchID are not modified in the helper function so it doesn't need to be involved with the response return?

PR #3669, opened 2025-11-10
mongo: measure event size instead of metered dialer

jgao54:
For Postgres/MySQL, we were manually tracking bytes processed for cdc, but for MongoDB we were using metered dialer for cdc. This PR updates MongoDB to use the same mechanism as Postgres/MySQL for CDC -- we explicitly measure and track cdc bytes as well as overall change stream bytes. Since Mongo provides server-side change stream filtering, these two  metrics are represented by the same `deltaBytesProcessed` variable.

#### Testing
Tested locally before and after this change to make sure that bytes reported only reflect when CDC events are processed. 

#### Follow-up
Prior to this PR, the metric was reflecting application traffic in raw bytes. And after some investigation, it turns out change stream was generate 400 bytes of traffic every seconds (which equates to ~35MB / day), mainly due to change stream's internal mechanism where it calls `GetMore` every second to look for new records.

I looked into configuring the `GetMore` frequency with [`setMaxAwaitTime`](https://www.mongodb.com/docs/manual/reference/method/cursor.maxAwaitTimeMS/#mongodb-method-cursor.maxAwaitTimeMS) to reduce network traffic, however this introduces a different error: `MaxAwaitTime must be less than the operation timeout`.  This requires more investigations as I was not able to determine how "operation timeout" is set.  Note the CommandMonitor was added in this PR to support further debugging.

To avoids confusion on data ingestion volume, use the change stream event size for now.

### DIFF
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -168,8 +168,7 @@ func parseAsClientOptions(config *protos.MongoConfig, meteredDialer utils.Metere
 		clientOptions.SetTLSConfig(tlsConfig)
 	}
 
-	// For debugging only: set env var MONGO_COMMAND_MONITOR to "true" to log all change stream commands
-	if os.Getenv("MONGO_COMMAND_MONITOR") == "true" {
+	if level, ok := os.LookupEnv("PEERDB_LOG_LEVEL"); ok && level == "DEBUG" {
 		clientOptions.SetMonitor(NewCommandMonitor(logger))
 	}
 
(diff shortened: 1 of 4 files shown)


flow/connectors/mongo/mongo.go
```diff
@@ -167,6 +168,11 @@ func parseAsClientOptions(config *protos.MongoConfig, meteredDialer utils.Metere
 		clientOptions.SetTLSConfig(tlsConfig)
 	}
 
+	// For debugging only: set env var MONGO_COMMAND_MONITOR to "true" to log all change stream commands
+	if os.Getenv("MONGO_COMMAND_MONITOR") == "true" {
```
ilidemi: There's also [PEERDB_LOG_LEVEL](https://github.com/PeerDB-io/peerdb/pull/3647) that we're consuming when printing PG messages in debug only

jgao54: ah good call, will use this instead

PR #3685, opened 2025-11-12
Add more WAL metrics

ilidemi:
Adding a bunch of WAL metrics in hopes of better seeing the impact of deployment on a busy instance.

`pg_stat_activity.wait_event_type/wait_event` for our backend: these are really the source of truth on when the wal sender becomes idle (`wait_event_type=Client, wait_event= WalSenderWaitForWAL`) because there's nothing to send (even if the current lsn is ahead of what was sent to us because of an unfinished transaction). However, it also can change many times a second and polling once every few minutes is pretty sparse. Putting it in in case it's still useful.

Commit LSN on the receiver - the latest we've received from Postgres so far

`pg_stat_replication.sent_lsn` - we don't always have permission to read this one but saw some discrepancies with commit LSN so keeping as optional for debugging purposes

Current LSN - to diff with commit LSN

Deltas between PG-side LSNs, just for convenience

`pg_stat_replication_slots.spill_txns/spill_count/spill_bytes` (PG16+), `logical_decoding_work_mem` (PG13+) - to see what the PG is doing after reconnect and before it sent anything to us, and be able to diagnose bad setting

---

Misc quality of life metrics:

`pg_replication_slots.safe_wal_size` (PG13+) - monitor danger zone when we're behind or there's a burst

`pg_replication_slots.active/wal_status`, `pg_stat_activity.state`

---
- [x] actually test this

### DIFF
--- a/flow/otel_metrics/otel_manager.go
+++ b/flow/otel_metrics/otel_manager.go
@@ -363,7 +363,7 @@ func (om *OtelManager) setupMetrics(ctx context.Context) error {
 
 	if om.Metrics.LogicalDecodingWorkMemGauge, err = om.GetOrInitInt64Gauge(BuildMetricName(LogicalDecodingWorkMemGaugeName),
 		metric.WithUnit("MiBy"),
-		metric.WithDescription("Current logical_decoding_work_mem setting in MB (with pendingRestart attribute)"),
+		metric.WithDescription("Current logical_decoding_work_mem setting in MB"),
 	); err != nil {
 		return err
 	}


flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -281,15 +281,36 @@ func AppendSlotSizeInfo(
...
+		slotInfo.Active,
+		slotInfo.WaitEventType,
+		slotInfo.WaitEvent,
+		slotInfo.BackendState,
+		slotInfo.StatsReset,
+		slotInfo.SpillTxns,
+		slotInfo.SpillCount,
+		slotInfo.SpillBytes,
+		slotInfo.LogicalDecodingWorkMemMb,
+		slotInfo.LogicalDecodingWorkMemPendingRestart,
```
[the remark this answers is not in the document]
ilidemi: thanks robot

nexus/catalog/migrations/V49__enhanced_slot_metrics.sql
```diff
@@ -0,0 +1,18 @@
+-- Add new columns for WAL metrics
+ALTER TABLE peerdb_stats.peer_slot_size
+ADD COLUMN IF NOT EXISTS sent_lsn TEXT,
```
iamKunalGupta: do we need this or are otel metrics enough?

ilidemi: we do not

flow/connectors/postgres/client.go
```diff
@@ -278,52 +278,180 @@ func (c *PostgresConnector) checkSlotAndPublication(ctx context.Context, slot st
...
+	}
+
+	ldwMBSelect := "NULL::bigint"
+	ldwPendingSelect := "NULL::boolean"
+	if pgversion >= shared.POSTGRES_13 {
+		ldwMBSelect = `(
+			SELECT (pg_size_bytes(setting || COALESCE(unit,'')) / 1024 / 1024)::bigint
+			FROM pg_settings WHERE name='logical_decoding_work_mem'
+		)`
+		ldwPendingSelect = `(
```
heavycrystal: `setting` will display the current value of `logical_decoding_work_mem`, based on my testing. So, isn't it enough to emit this? Why is emitting `pending_restart` important? 

Also due to `logical_decoding_work_mem` being `PGC_USERSET`, it can be set liberally. Therefore I'm not sure if it can even be `pending_restart`

flow/connectors/postgres/client.go
```diff
@@ -278,52 +278,180 @@ func (c *PostgresConnector) checkSlotAndPublication(ctx context.Context, slot st
...
+			round((cw.current_lsn - psr.sent_lsn) / 1024 / 1024),
+			psa.wait_event_type,
+			psa.wait_event,
+			psa.state,
+			%s, -- logical_decoding_work_mem megabytes
+			%s, -- logical_decoding_work_mem pending_restart
+			%s  -- stats
+		FROM current_wal cw,
+			pg_control_checkpoint() as pcc,
+			(pg_replication_slots as prs
```
heavycrystal, line 353: With the permissions we currently document, `pg_stat_replication` is not very usable (`sent_lsn` is NULL). You'll need to grant `pg_monitor`. If most of our current users don't have the permission, I'd recommend not emitting the metrics unless that's addressed
```
postgres=> SELECT * FROM pg_stat_replication;
-[ RECORD 1 ]----+------------------------------
pid              | 1028
usesysid         | 18295
usename          | clickpipes_user
application_name | peerdb
client_addr      | 192.168.107.1
client_hostname  |
client_port      | 20007
backend_start    | 2025-11-17 10:24:13.323278+00
backend_xmin     |
state            |
sent_lsn         |
write_lsn        |
flush_lsn        |
replay_lsn       |
write_lag        |
flush_lag        |
replay_lag       |
sync_priority    |
sync_state       |
reply_time       |
```

flow/connectors/postgres/postgres.go
```diff
@@ -1394,51 +1394,93 @@ func (c *PostgresConnector) HandleSlotInfo(
...
+	slotMetricGauges.LogicalDecodingWorkMemGauge.Record(ctx, slotInfo.LogicalDecodingWorkMemMb,
+		metric.WithAttributeSet(attribute.NewSet(
+			attribute.String(otel_metrics.FlowNameKey, alertKeys.FlowName),
+			attribute.String(otel_metrics.PeerNameKey, alertKeys.PeerName),
+			attribute.String(otel_metrics.SlotNameKey, alertKeys.SlotName),
+			attribute.Bool(otel_metrics.PendingRestartKey, slotInfo.LogicalDecodingWorkMemPendingRestart),
+		)))
+
+	// PG 16+ statistics gauges
+	slotMetricGauges.StatsResetGauge.Record(ctx, slotInfo.StatsReset, attributeSet)
```
heavycrystal: These should just not be emitted if Postgres doesn't support it. 0 `spilltxns` constantly would seem great unless someone looks at version too

flow/connectors/postgres/client.go
ilidemi: Good catch. I noticed that sent_lsn was different from received commit LSN by 70MB in testing, will keep as optional for debugging purposes - otherwise could be hard to compare received LSN against others

flow/connectors/postgres/client.go
ilidemi: Indeed it cannot, good catch

PR #3687, opened 2025-11-12
Improve initial load partitioning logic readability in PG

jgao54:
We have expanded our partitioning logic over time to the point that getNumRowsPartitions has gotten quite long, so seems like a good time to break it into more digestible pieces.

I also noticed we were only committing the transaction for the default path, but not for the alternative path with numPartitionsOverride, which doesn't have an impact on correctness because of defer shared.RollbackTx(), but was confusing nonetheless. Moved the "commit" logic to GetQRepPartitions.

The refactor introduces a light function interface to implement partitioning logic. This will make testing/exposing different strategies easier later on. (i.e. if/when we do plan to expose CTID partitioning).

Separately, I think we can use estimated table count instead of SELECT count(*) for NTileBucketPartitioning, which @serprex proposed before. The implementation doesn't need a precise row count. So this should help with improving performance of this count query. Leaving that for a follow-up PR.

Test:
- local smoke testing to ensure it works as expected

### DIFF
--- a/flow/connectors/postgres/qrep_partition.go
+++ b/flow/connectors/postgres/qrep_partition.go
@@ -32,10 +32,6 @@ type PartitioningFunc func(context.Context, PartitionParams) ([]*protos.QRepPart
 // equal-sized partitions based on row count. It uses the NTILE window function to assign rows
 // to buckets and ensures more balanced row distribution across partitions.
 func NTileBucketPartitioningFunc(ctx context.Context, pp PartitionParams) ([]*protos.QRepPartition, error) {
-	if pp.numPartitions <= 1 {
-		return nil, errors.New("expect numPartitions to be greater than 1")
-	}
-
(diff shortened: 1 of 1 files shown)


flow/connectors/postgres/qrep.go
```diff
@@ -68,13 +65,21 @@ func (c *PostgresConnector) GetQRepPartitions(
...
 		return nil, fmt.Errorf("failed to set transaction snapshot: %w", err)
 	}
 
-	return c.getNumRowsPartitions(ctx, getPartitionsTx, config, last)
+	partitions, err := c.getPartitions(ctx, getPartitionsTx, config, last)
+
+	// commit transaction
+	if err := getPartitionsTx.Commit(ctx); err != nil {
+		return nil, fmt.Errorf("failed to commit transaction: %w", err)
+	}
```
jgao54, line 80: also open to just remove this altogether -- since this transaction is used for read-only purpose -- having `defer shared.RollbackTx()` above is sufficient.

PR #3691, opened 2025-11-12
Support TIME64 data type for ClickHouse replication

### LINKED ISSUE #3468: Support replicating TIME64 data type in ClickHouse
With [Time64](https://clickhouse.com/docs/sql-reference/data-types/time64), we can now support this data type in Postgres/MySQL.

Need to bump internal version for backward-wards compatibility.

NehharShah:
Adds support for ClickHouse `TIME64(3)` data type when replicating TIME columns from PostgreSQL and MySQL sources.

**Changes:**
- Bump internal version to `InternalVersion_ClickHouseTime64` for backward compatibility
- Map MySQL `TIME` type to `QValueKindTime` (instead of `QValueKindTimestamp`) for new mirrors
- Use `Time64(3)` for ClickHouse 25.6+ when internal version >= `InternalVersion_ClickHouseTime64`
- Fall back to `DateTime64(6)` for older ClickHouse versions or existing mirrors
- Update normalization queries to handle `Time64(3)` columns

**Backward Compatibility:**
- Existing mirrors continue using `DateTime64(6)` 
- New mirrors with ClickHouse >= 25.6 use `Time64(3)`
- Older ClickHouse versions automatically fall back to `DateTime64(6)`

Fixes #3468

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -234,7 +234,9 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 		defer dstClose(ctx)
 
 		syncState.Store(shared.Ptr("updating schema"))
-		if err := dstConn.ReplayTableSchemaDeltas(ctx, config.Env, flowName, options.TableMappings, recordBatchSync.SchemaDeltas, config.Version); err != nil {
+		if err := dstConn.ReplayTableSchemaDeltas(
+			ctx, config.Env, flowName, options.TableMappings, recordBatchSync.SchemaDeltas, config.Version,
+		); err != nil {
 			return nil, fmt.Errorf("failed to sync schema: %w", err)
 		}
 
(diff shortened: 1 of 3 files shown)


serprex (approved):

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -113,6 +114,44 @@ func (s ClickHouseSuite) Test_MySQL_Time() {
...
+
+	// Check if ClickHouse version >= 25.6
+	versionParts := strings.Split(chVersion, ".")
+	if len(versionParts) >= 2 {
+		major, _ := strconv.Atoi(versionParts[0])
+		minor, _ := strconv.Atoi(versionParts[1])
+
+		// If ClickHouse >= 25.6 and using latest internal version, should use Time64(3)
+		if major > 25 || (major == 25 && minor >= 6) {
+			if flowConnConfig.Version >= shared.InternalVersion_ClickHouseTime64 {
```
jgao54: nit: we currently don't run CI against different versions of PeerDB/ClickHouse, could you add an explicit test case for an older version to verify backwards compatibility by setting something like `flowConnConfig.Version = shared.InternalVersion_First
`

jgao54 (approved):
small nit, lgtm 🚀

PR #3694, opened 2025-11-13
feat: configurable decimal precision/scale for unbounded numeric

### LINKED ISSUE #3336: Enhancement request: Be able to specify the cusom numeric precision and scale for ClickPipes
### Current behaviour

- `PeerDBClickHouseScale` is hard-coded as `38`. https://github.com/PeerDB-io/peerdb/blob/cffcf454cca0f64f72685674a496abd679fa9622/flow/shared/datatypes/numeric.go#L7
- When one tries to get a Postgres `unbounded  numeric` column into ClickHouse via ClickPipe, it is converted into `Decimal(78, 38)` and a user has no control over it.
- This may be a problem when ingesting large numbers from `unbounded  numeric` Postgres columns.

### Suggested behaviour

- Add an option to configure precision and scale for the `unbounded  numeric` Postgres columns at the ClickPipe creation.

Raviikumar001:
closes #3336 

Summary

- Add optional precision/scale overrides for unbounded Postgres NUMERIC → ClickHouse Decimal mapping.
- Proto: introduce FlowConnectionConfigs fields `clickhouse_numeric_default_precision` and `clickhouse_numeric_default_scale`.
- Config: support env-based overrides
- `PEERDB_CLICKHOUSE_NUMERIC_DEFAULT_PRECISION` and `PEERDB_CLICKHOUSE_NUMERIC_DEFAULT_SCALE`
- Validation: enforce ranges and dependency (precision 1–76, scale 0–precision; scale requires precision).
- Wiring: thread overrides through normalize, CDC schema deltas, Avro schema/value conversion, and query generation.
- Model: update qvalue functions (`DetermineNumericSettingForDWH, ToDWHColumnType, GetNumericDestinationType`) to accept overrides; bounded NUMERIC unchanged.
- UI: fix CDCConfig to use camelCase fields (
[clickhouseNumericDefaultPrecision], [clickhouseNumericDefaultScale]) and update handlers.
-Rust: update FlowConnectionConfigs initializer to include the new optional fields.
- Tests: add unit tests for override behavior; add ClickHouse e2e test validating Decimal(60,10) mapping for unbounded NUMERIC.

### DIFF
--- a/flow/cmd/validate_mirror.go
+++ b/flow/cmd/validate_mirror.go
@@ -65,7 +65,6 @@ func (h *FlowRequestHandler) validateCDCMirrorImpl(
 			errors.New("invalid config: initial_snapshot_only is true but do_initial_snapshot is false"))
 	}
 
-	// Validate ClickHouse numeric settings if provided
 	var chDefPrecision, chDefScale int32
 	if connectionConfigs.ClickhouseNumericDefaultPrecision != nil {
 		chDefPrecision = *connectionConfigs.ClickhouseNumericDefaultPrecision
@@ -74,30 +73,25 @@ func (h *FlowRequestHandler) validateCDCMirrorImpl(
 		chDefScale = *connectionConfigs.ClickhouseNumericDefaultScale
 	}
 	if chDefPrecision > 0 || chDefScale > 0 {
-		// Load destination peer to check if it's ClickHouse
 		dstPeer, err := connectors.LoadPeer(ctx, h.pool, connectionConfigs.DestinationName)
 		if err != nil {
 			return nil, NewFailedPreconditionApiError(fmt.Errorf("failed to load destination peer: %w", err))
 		}
 
-		// Only validate ClickHouse numeric settings for ClickHouse destinations
 		if dstPeer.Type == protos.DBType_CLICKHOUSE {
 			precision := chDefPrecision
 			scale := chDefScale
 
-			// If scale is provided but precision is not, reject early with a clear error
 			if precision == 0 && scale > 0 {
 				return nil, NewInvalidArgumentApiError(
 					errors.New("clickhouse numeric precision must be set when providing a custom scale"))
 			}
 
-			// Validate precision: 1-76
 			if precision < 1 || precision > 76 {
 				return nil, NewInvalidArgumentApiError(
 					fmt.Errorf("clickhouse numeric precision must be between 1 and 76, got %d", precision))
 			}
 
-			// Validate scale: 0 to precision
 			if scale < 0 || scale > precision {
 				return nil, NewInvalidArgumentApiError(
 					fmt.Errorf("clickhouse numeric scale must be between 0 and %d (precision), got %d", precision, scale))
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -202,7 +202,6 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
 
 			if clickHouseType == "" {
 				var err error
-				// Fetch optional ClickHouse numeric default overrides from env
 				var chDefPrecision, chDefScale int32
 				if p, errP := internal.PeerDBClickHouseNumericDefaultPrecision(ctx, config.Env); errP == nil {
 					chDefPrecision = p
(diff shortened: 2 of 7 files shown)


Raviikumar001:
> @Raviikumar001 appreciate the initiative and contribution here.
> 
> Configurable decimal precision/scale for unbounded numeric is a great feature, however it's important that it is implemented at the column-level rather than via global environment variable, in order to accommodate different columns with different precision/scale requirement. (historically we have used env var to store data-type state, and each time it had bit us back and we'd regret going down that path).
> 
> In the codebase we already have the concept of "SupportedDestinationTypes". To support configurable decimal precision/scale, that would be the better entry point to introduce this feature (i.e. introduce a `NumericToNumericSchemaConversion` and `NumericToNumericValueConversion`). Although the current design for the conversion system also lacks e2e design, so i am not 100% sure how much of a lift it is to preserve precision/scale e2e without digging into the entire write path.
> 
> TL;DR is this piece is probably going to be a bigger project. If you have ideas on improving the architecture to make type conversion more extensible for scale/precision, would love to hear them.

Thanks for you're response, really appreciate it, i'll have to think on this, i'll share my thoughts on you're query soon!

Raviikumar001:
> @Raviikumar001 appreciate the initiative and contribution here.
> 
> Configurable decimal precision/scale for unbounded numeric is a great feature, however it's important that it is implemented at the column-level rather than via global environment variable, in order to accommodate different columns with different precision/scale requirement. (historically we have used env var to store data-type state, and each time it had bit us back and we'd regret going down that path).
> 
> In the codebase we already have the concept of "SupportedDestinationTypes". To support configurable decimal precision/scale, that would be the better entry point to introduce this feature (i.e. introduce a `NumericToNumericSchemaConversion` and `NumericToNumericValueConversion`). Although the current design for the conversion system also lacks e2e design, so i am not 100% sure how much of a lift it is to preserve precision/scale e2e without digging into the entire write path.
> 
> TL;DR is this piece is probably going to be a bigger project. If you have ideas on improving the architecture to make type conversion more extensible for scale/precision, would love to hear them.

@jgao54 
Thanks for the detailed feedback — I agree that env-level precision/scale isn’t the right abstraction and that this belongs in the column-level schema and conversion layer. To address this properly, I’m planning to split the work into two PRs to keep the scope reviewable and reduce risk.

Apologies for the lengthy comment 🙇‍♂️

PR 1 — Foundation (column-level spec + ClickHouse support)

Core changes:

- Introduce a per-column `DestinationTypeSpec` (`decimal { precision, scale, applyToUnboundedOnly }`).
- Add `NumericToNumericSchemaConversion` + `NumericToNumericValueConversion` for ClickHouse.
- Persist the selected decimal spec in the catalog; all DDL, CDC schema updates, Avro/Arrow schema generation, and value conversion use this single source of truth.
- Precedence:
(1) column-level spec
(2) bounded NUMERIC(p,s) typmod
(3) legacy env defaults (env kept only as a fallback and marked deprecated)

Tests:

- Unit tests for precedence handling, invalid ranges, bounded vs unbounded behavior.
- One ClickHouse end-to-end case with different overrides across multiple columns.

Migration:

- No breaking changes; existing mirrors continue using env defaults until column-level specs are added.

PR 2 — Expansion (UI, additional destinations, policies, deprecation)

Enhancements:

- UI for configuring per-column Decimal precision/scale with validation.
- Add converters for BigQuery (NUMERIC/BIGNUMERIC) and Snowflake.
- Introduce optional rounding/overflow policy (default remains “fail”). 
- Deprecation path: warnings when env fallback is used + option to disable env fallback entirely.

Testing:

- Broader cross-destination conversion matrix.
- CDC schema delta tests involving decimal overrides.
- Rounding and overflow behavior tests.

Splitting this into two PRs keeps the core architectural change (single, column-scoped decimal spec + schema/value unification) focused and reviewable, while the second PR handles UI and multi-destination extensions once the foundation is solid.

Please let me know you're thoughts on this — happy to adjust. Though some testing cannot done on my end(from my laptop) as mentioned above in testing section,  i'd really apppreciate you thoughts on my suggestions.  
I'll be really greateful if i'll get the chance to be able to do these contributions. Thanks!!

jgao54:
@Raviikumar001 thanks for providing your feedback, the general direction does sound right to me. Unfortunately, at the moment given our team's limited capacity, frankly we would not be able to support architectural changes at this scale, mostly due to the state changes this requires across the board (temporal/db/avro). 

It's probably not the best time to introduce this feature; but I expect this to be something we'll revisit sometimes next year.

PR #3695, opened 2025-11-13
Add support for Index / function / triggers for postgres to postgres migrations

yogesh1801:


yogesh1801:
@iskakaushik 

1. For Trigger migration we need to migrate all dependencies also? for now i have enabled function migration so that triggers can work,  other dependencies can be custom types

2. Should PeerDB support for indexes, functions, and triggers, like it does for table columns?"

PR #3697, opened 2025-11-15
Adding support of TIME64 datatype

jgao54:
Postgres TIME supports range:  [00:00:00, 24:00:00], with microsecond precision
MySQL TIME supports range: [-838:59:59.000000, 838:59:59.999999], with microsecond precision
ClickHouse Time64 supports range: [-999:59:59.000000, 999:59:59.999999], with nanosecond precision

This implementation for Time64 allow ClickHouse to fully support Postgres and MySQL Time ranges.

One subtle, but IMO useful improvement from this PR (aside from adding Time64 support) that's worth calling out:
we used to have Postgres convert to QValueTime while MySQL convert to QValueTimestamp prior to Time64 was introduced; meaning Posgres handles conversion to DateTime64 in the destination connector, while MySQL handles the conversion in the source connector. With this PR, conversion is consistently handled in the destination connector. So we always convert Time -> QValueTime, and then handle backwards-compatibility by optionally converting QValueTime -> DateTime64 if on an older internal version or if on an older version of ClickHouse that does not support Time64.

I've also evaluated bringing consistency by aligning Postgres to MySQL, where we map to QValueTimestamp on older internal version. This is a messier approach, since we can't convert to Time64 unless it is also supported by ClickHouse, the source connector would need to know about destination connector's version (later on discussed with @ilidemi and realized we already perform destination-aware transformations for some other data types, but realized this in the _qvalue to avro_ conversion layer, not in the source connector logic that mysql uses. it looks like mysql time handling in source connector is the only edge case here so it's good to get that one aligned to QValueTime). 

Test: added e2e test; tested the upgrade path locally to verify that after the upgrade, existing pipes will continue to use datetime64 instead of time64.

### DIFF
--- a/flow/connectors/clickhouse/normalize_query.go
+++ b/flow/connectors/clickhouse/normalize_query.go
@@ -161,15 +161,18 @@ func (t *NormalizeQueryGenerator) BuildQuery(ctx context.Context) (string, error
 		case "DateTime64(6)", "Nullable(DateTime64(6))":
 			// Handle legacy path where TIME is stored as DateTime64 (before Time64 support)
 			if colType == types.QValueKindTime || colType == types.QValueKindTimeTZ {
-				fmt.Fprintf(&projection,
-					"toDateTime64(toTime64OrNull(JSONExtractString(_peerdb_data, %s), 6), 6) AS %s,",
-					peerdb_clickhouse.QuoteLiteral(colName),
+				minVersion, exists := clickhouse.GetMinVersion(clickhouse.SettingEnableTimeTime64Type)
+				time64Supported := exists && chproto.CheckMinVersion(minVersion, *t.chVersion)
+
+				fmt.Fprintf(&projection, "%s AS %s,",
+					extendedTimeToDateTime(fmt.Sprintf("JSONExtractString(_peerdb_data, %s)",
+						peerdb_clickhouse.QuoteLiteral(colName)), time64Supported),
 					peerdb_clickhouse.QuoteIdentifier(dstColName),
 				)
 				if t.enablePrimaryUpdate {
-					fmt.Fprintf(&projectionUpdate,
-						"toDateTime64(toTime64OrNull(JSONExtractString(_peerdb_match_data, %s), 6), 6) AS %s,",
-						peerdb_clickhouse.QuoteLiteral(colName),
+					fmt.Fprintf(&projectionUpdate, "%s AS %s,",
+						extendedTimeToDateTime(fmt.Sprintf("JSONExtractString(_peerdb_match_data, %s)",
+							peerdb_clickhouse.QuoteLiteral(colName)), time64Supported),
 						peerdb_clickhouse.QuoteIdentifier(dstColName),
 					)
 				}
@@ -329,4 +332,21 @@ func (t *NormalizeQueryGenerator) BuildQuery(ctx context.Context) (string, error
 	t.Query = insertIntoSelectQuery
 
 	return t.Query, nil
+}
+
+func extendedTimeToDateTime(jsonExtractExpr string, time64Supported bool) string {
+	if time64Supported {
+		return fmt.Sprintf("toDateTime64(toTime64OrNull(%s, 6), 6)", jsonExtractExpr)
+	}
+
+	// fallback to manual string parsing for older versions of clickhouse instances
+	// where `toTime64OrNull` is not supported.
+	return fmt.Sprintf(`if(length(%[1]s) > 0,
+		toDateTime64(
+			(if(startsWith(%[1]s, '-'), -1, 1)) *
+			(toInt64(splitByChar(':', if(startsWith(%[1]s, '-'), substring(%[1]s, 2), %[1]s))[1]) * 3600 +
+			 toInt64(splitByChar(':', if(startsWith(%[1]s, '-'), substring(%[1]s, 2), %[1]s))[2]) * 60 +
+			 toFloat64(splitByChar(':', if(startsWith(%[1]s, '-'), substring(%[1]s, 2), %[1]s))[3]))
+		, 6),
+		NULL)`, jsonExtractExpr)
 }
--- a/flow/connectors/clickhouse/normalize_query_test.go
+++ b/flow/connectors/clickhouse/normalize_query_test.go
@@ -1,0 +1,55 @@
+package connclickhouse
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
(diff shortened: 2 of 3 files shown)


flow/connectors/mysql/qvalue_convert.go
jgao54: I think it's safe to delete this actually (i.e. dead code), since we no longer map mysql time -> QValueKindTimestamp.
I kept it here out of precaution seeing the notes on deprecation we used to have below, but i can't think of a path where this would actually hit as it's only used in the paths during initial snapshot. 
so will probably remove.

flow/e2e/clickhouse.go
```diff
@@ -71,7 +71,7 @@ func (s ClickHouseSuite) Peer() *protos.Peer {
 			Config: &protos.Peer_ClickhouseConfig{
 				ClickhouseConfig: &protos.ClickhouseConfig{
 					Host:       "localhost",
-					Port:       9001,
+					Port:       9000,
```
jgao54: oops, this was unintended change, will switch back

flow/internal/clickhouse/settings.go
ilidemi, line 1: `/internal` would be a better place for this - `/pkg` is only for things that are being used from the outside

flow/model/qvalue/equals.go
```diff
@@ -188,7 +188,11 @@ func compareGoTime(value1, value2 any) bool {
 		t2 = tm.Sub(time.Unix(0, 0).UTC())
 	}
 
-	return ok1 && ok2 && t1 == t2
+	// Workaround for Go driver behavior where precision is lost when
+	// scanning to time.Duration (QValue Equal is only used for e2e testing)
+	// https://github.com/ClickHouse/clickhouse-go/issues/1757
```
ilidemi, line 193: Do we want to nudge at least a check-in to main and use that? Or minor enough to wait for next ver

jgao54: wait for next ver seems fine, this is only used for e2e testing so not impacting prod.
also i want to get to the root of the precision issue and get this upstream issue resolved, but don't want to block Time64 rollout. 

will nudge on a release cut once both issues are resolved.

flow/connectors/mysql/qvalue_convert.go
jgao54: this code is only used during qrep where we generate schema from source db; and since we always translate mysql time type to QValueKindTime, confirmed that this is safe to remove with this pr.

ilidemi: The previous one was needed because it didn't have versioning system we use now, so the qkind stored in the catalog schemas was used as a discriminator set at pipe creation. Here the version determines all paths in the same way so the code is indeed dead.

flow/model/qvalue/kind.go
```diff
@@ -116,3 +121,12 @@ func ShouldUseNativeJSONType(ctx context.Context, env map[string]string, chVersi
...
 	return isJsonSupported && isJsonEnabled
 }
+
+func ShouldUseTime64Type(chVersion *chproto.Version, internalVersion uint32) bool {
+	if chVersion == nil {
+		return false
+	}
+	// check clickhouse minimum supported version, and peerdb internal version for backwards-compatibility
+	minSupportedVersion, exists := chinternal.GetMinVersion(chinternal.SettingEnableTimeTime64Type)
+	return exists && chproto.CheckMinVersion(minSupportedVersion, *chVersion) && internalVersion >= shared.InternalVersion_ClickHouseTime64
```
ilidemi: What will happen when the pipe/mirror is created with InternalVersion_ClickHouseTime64 on 25.5 then CH updates? Seems like the queries would change and we'd want them to stay the same.

ilidemi: As a solution, it would seem ok to me to gate the mirror version on CH version from the get go and make the precondition for the new behavior updating CH + recreating the mirror. Checked, none of the CHC customers would be affected.

jgao54: > What will happen when the pipe/mirror is created with InternalVersion_ClickHouseTime64 on 25.5 then CH updates? 

if CH update at that time the destination table is already created with column as DateTime64, the CDC logic checks for destination type and use the correct routing based on the destination schema. In other words, `ToDWHColumnType` is only used during table creation. The time-formatted string is flexible and can be interpreted as either datetime or time depending on destination type during CDC.

ilidemi: I'm not seeing any places where DestinationType is set across the codebase. Also looked through prod catalogs, 0 are set in the protos there. Are you sure it doesn't go to `ToDWHColumnType`?

jgao54: discussed offline, valid concern and good catch on this edge case! I missed the use of `ToDWHColumnType` in the CDC path (i.e. `BuildQuery`)

flow/connectors/clickhouse/normalize_query.go
```diff
@@ -314,3 +333,20 @@ func (t *NormalizeQueryGenerator) BuildQuery(ctx context.Context) (string, error
...
+	// fallback to manual string parsing for older versions of clickhouse instances
+	// where `toTime64OrNull` is not supported.
+	return fmt.Sprintf(`if(length(%[1]s) > 0,
+		toDateTime64(
+			(if(startsWith(%[1]s, '-'), -1, 1)) *
+			(toInt64(splitByChar(':', if(startsWith(%[1]s, '-'), substring(%[1]s, 2), %[1]s))[1]) * 3600 +
+			 toInt64(splitByChar(':', if(startsWith(%[1]s, '-'), substring(%[1]s, 2), %[1]s))[2]) * 60 +
+			 toFloat64(splitByChar(':', if(startsWith(%[1]s, '-'), substring(%[1]s, 2), %[1]s))[3]))
+		, 6),
+		NULL)`, jsonExtractExpr)
```
jgao54, line 351: Discovered that `toDateTime64(toTime64OrNull(%s, 6), 6)` would not work universally here, so fallback to manual parsing. Not super ideal, but haven't thought of a good workaround.

This path should only apply to those on clickhouse version < 25.6, which we could eventually deprecate after some time.

ilidemi (approved):

flow/connectors/clickhouse/normalize_query.go
ilidemi: Approach makes sense 👍 
The expression took some time to wrap the head around, maybe add more comments as a follow up?

jgao54: will do

PR #3700, opened 2025-11-16
Better metrics for CDC store

ilidemi:


flow/connectors/postgres/cdc.go
```diff
@@ -705,6 +705,19 @@ func PullCdcRecords[Items model.Items](
...
 						isFullReplica := req.TableNameSchemaMapping[tableName].IsReplicaIdentityFull
 						if isFullReplica {
+							var colCount int64
+							if recItems, ok := any(r.NewItems).(model.RecordItems); ok {
+								colCount = int64(len(recItems.ColToVal))
+							}
+							p.otelManager.Metrics.ToastValuesCounter.Add(ctx, colCount,
+								metric.WithAttributeSet(attribute.NewSet(
+									attribute.String("table", tableName),
+									attribute.String("backfilled", "replica_identity_full"))))
```
Amogh-Bharadwaj: Further below, "backfilled" seems to have boolean values but here it's "replica_identity_full", what does "backfilled" exactly represent?

flow/connectors/postgres/cdc.go
```diff
@@ -714,23 +727,61 @@ func PullCdcRecords[Items model.Items](
...
+
+							// Report metrics
+							remainingUnchangedToast := len(r.UnchangedToastColumns)
+							if backfilledCount > 0 {
+								p.otelManager.Metrics.ToastValuesCounter.Add(ctx, int64(backfilledCount),
+									metric.WithAttributeSet(attribute.NewSet(
+										attribute.String("table", tableName),
+										attribute.String("backfilled", "true"))))
+							}
+							if remainingUnchangedToast > 0 {
```
Amogh-Bharadwaj: Can we move:
```Go
remainingUnchangedToast := len(r.UnchangedToastColumns)
```
to line 758 for easy reading?

flow/connectors/postgres/cdc.go
```diff
@@ -714,23 +727,61 @@ func PullCdcRecords[Items model.Items](
...
+							switch {
+							case unchangedToastCount == 0:
+								backfillStatus = "no_toast"
+							case backfilledCount == unchangedToastCount:
+								backfillStatus = "fully"
+							case backfilledCount > 0:
+								backfillStatus = "partially"
+							default:
+								backfillStatus = "none"
 							}
```
jgao54, line 800: nit: could we represent these states with `const` and add a comment beside each status? (for instance, does "none" represent "not started"?)

flow/connectors/postgres/cdc.go
ilidemi: Backfilling is not applicable here because of replica identity full but the rows/values should be reported for accurate percentages. Added a comment and `na_` prefix

flow/connectors/postgres/cdc.go
ilidemi: Added comments. The constants would be single use on the very next line so not sure it's worth the extra lines of code somewhere at the top

flow/connectors/postgres/cdc.go
```diff
@@ -749,40 +776,30 @@ func PullCdcRecords[Items model.Items](
...
+								rowBackfillStatus = "partially" // Some values were backfilled
 							default:
-								backfillStatus = "none" // Couldn't backfill anything
+								rowBackfillStatus = "none" // Couldn't backfill anything
 							}
-							p.otelManager.Metrics.ToastRowsCounter.Add(ctx, 1,
-								metric.WithAttributeSet(attribute.NewSet(
-									attribute.String("table", tableName),
-									attribute.String("backfill_status", backfillStatus))))
+							toastValueMetrics[toastMetricKey{tableName, rowBackfillStatus}]++
```
jgao54: `toastRowMetrics`

PR #3704, opened 2025-11-17
Store and migrate Postgres table OIDs to catalog 

Amogh-Bharadwaj:
This PR ensures that setup flow stores Postgres table OID in catalog. Once this is rolled out and the migration is complete, `ensurePullability` can be merged away as `GetTableSchema` makes the exact same calls as it.

### DIFF
--- a/flow/e2e/flow_status_test.go
+++ b/flow/e2e/flow_status_test.go
@@ -164,4 +164,6 @@ func (s APITestSuite) TestFlowStatusUpdate() {
 	require.Equal(s.t, protos.FlowStatus_STATUS_PAUSED, updates[3].NewStatus)
 	require.Equal(s.t, protos.FlowStatus_STATUS_PAUSED, updates[4].OldStatus)
 	require.Equal(s.t, protos.FlowStatus_STATUS_RUNNING, updates[4].NewStatus)
+	env.Cancel(s.t.Context())
+	RequireEnvCanceled(s.t, env)
 }


flow/activities/flowable.go
```diff
@@ -1813,3 +1813,43 @@ func (a *FlowableActivity) ReportStatusMetric(ctx context.Context, status protos
...
+	shutdown := heartbeatRoutine(ctx, func() string {
+		return "migrating oids to table schema"
+	})
+	defer shutdown()
+	logger := internal.LoggerFromCtx(ctx)
+	total := len(oidToTableNameMapping)
+	remaining := total
+	for oid, tableName := range oidToTableNameMapping {
+		var destinationTableIdentifier string
+		for _, tm := range tableMappings {
```
ilidemi, line 1853: O(N^2)

flow/activities/flowable.go
```diff
@@ -1813,3 +1813,43 @@ func (a *FlowableActivity) ReportStatusMetric(ctx context.Context, status protos
...
+			if tm.SourceTableIdentifier == tableName {
+				destinationTableIdentifier = tm.DestinationTableIdentifier
+				break
+			}
+		}
+		err := internal.UpdateTableOIDInTableSchemaInCatalog(ctx, a.CatalogPool, logger, flowName, destinationTableIdentifier, oid)
```
ilidemi: Can we do the updates in a batch? If a customer has 1000 tables (or someone in OSS has 10K), no reason to make it 2000 round-trip queries. Load all in 1 query, update, insert as a batch

flow/workflows/cdc_flow.go
```diff
@@ -41,6 +41,9 @@ type CDCFlowWorkflowState struct {
 	SnapshotNumPartitionsOverride uint32
 	SnapshotMaxParallelWorkers    uint32
 	SnapshotNumTablesInParallel   uint32
+
+	// Mark if OID migration is completed
+	OidMigrationCompleted bool
```
ilidemi: This seems like it would stay in Temporal forever as state gets copied. As this is catalog schema related, maybe create a table `flow_migrations (flow_name text, migration_name text, competed boolean)` and check/update within the activity? We will have more migrations in the future, doesn't hurt to have infra

flow/activities/flowable.go
```diff
@@ -1826,3 +1826,36 @@ func (a *FlowableActivity) ReportStatusMetric(ctx context.Context, status protos
...
+	oidToTableNameMapping map[uint32]string,
+	tableMappings []*protos.TableMapping,
+) error {
+	shutdown := heartbeatRoutine(ctx, func() string {
+		return "migrating oids to table schema"
+	})
+	defer shutdown()
+
+	logger := internal.LoggerFromCtx(ctx)
+	destinationTableOidMap, err := internal.GetDestinationTableOidMap(tableMappings, oidToTableNameMapping)
```
ilidemi: This can be inline, it's a small activity

flow/e2e/api_test.go
```diff
@@ -1114,6 +1132,41 @@ func (s APITestSuite) TestTotalRowsSyncedByMirror() {
 	RequireEqualTables(s.ch, "table1", cols)
 	RequireEqualTables(s.ch, "table2", cols)
 
+	// Check Postgres table OIDs in catalog
+	if pgconn, ok := s.source.Connector().(*connpostgres.PostgresConnector); ok {
```
ilidemi: Let's make a separate test

flow/internal/postgres.go
```diff
@@ -72,3 +72,59 @@ func LoadTableSchemaFromCatalog(
...
+		slog.Int("numTables", len(tableOIDs)))
+
+	tx, err := pool.Pool.Begin(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	defer shared.RollbackTx(tx, logger)
+
+	for destinationTableName, tableOID := range tableOIDs {
+		tableSchema, err := LoadTableSchemaFromCatalog(ctx, pool, flowName, destinationTableName)
```
ilidemi: Sorry I meant load all at once (1 query), process, insert with `pgx.Batch` (many queries one round trip)

flow/internal/postgres.go
```diff
@@ -72,3 +73,130 @@ func LoadTableSchemaFromCatalog(
...
+	logger log.Logger,
+	flowName string,
+	tableOIDs map[string]uint32, // map[destinationTableName]tableOID
+) error {
+	logger.Info("updating table OIDs in catalog",
+		slog.String("flowName", flowName),
+		slog.Int("numTables", len(tableOIDs)))
+
+	if len(tableOIDs) == 0 {
+		return nil
```
ilidemi, line 87: nit: log this happened, will always be the case for mysql/mongo pipes

flow/internal/postgres.go
```diff
@@ -72,3 +73,130 @@ func LoadTableSchemaFromCatalog(
...
+			slog.String("tableName", tableName),
+			slog.Uint64("tableOID", uint64(tableOID)))
+	}
+
+	results := tx.SendBatch(ctx, batch)
+	defer results.Close()
+
+	for i := 0; i < len(tableOIDs); i++ {
+		_, err := results.Exec()
+		if err != nil {
```
ilidemi:
```suggestion
		if _, err := results.Exec(); err != nil {
```

flow/activities/flowable.go
```diff
@@ -1826,3 +1826,57 @@ func (a *FlowableActivity) ReportStatusMetric(ctx context.Context, status protos
...
+	oidToTableNameMapping map[uint32]string,
+	tableMappings []*protos.TableMapping,
+) error {
+	shutdown := heartbeatRoutine(ctx, func() string {
+		return "migrating oids to table schema"
+	})
+	defer shutdown()
+
+	logger := internal.LoggerFromCtx(ctx)
+	migrationName := "migrate_postgres_table_oids_v1"
```
ilidemi: Use the shared constant?

flow/e2e/api_test.go
```diff
@@ -1163,6 +1207,91 @@ func (s APITestSuite) TestTotalRowsSyncedByMirror() {
...
+	EnvWaitFor(s.t, env, 3*time.Minute, "wait for initial load to finish", func() bool {
+		return env.GetFlowStatus(s.t) == protos.FlowStatus_STATUS_RUNNING
+	})
+
+	// Test initial load
+	RequireEqualTables(s.ch, "table1", cols)
+	RequireEqualTables(s.ch, "table2", cols)
+
+	// Check Postgres table OIDs in catalog
+	var table1OID, table2OID uint32
```
ilidemi, line 1249: nit: assert they're not zero

flow/connectors/postgres/postgres_schema_delta_test.go
```diff
@@ -71,6 +71,8 @@ func (s PostgresSchemaDeltaTestSuite) TestSimpleAddColumn() {
 
 	output, err := s.connector.GetTableSchema(s.t.Context(), nil, shared.InternalVersion_Latest, protos.TypeSystem_Q,
 		[]*protos.TableMapping{{SourceTableIdentifier: tableName}})
+	require.NotEqual(s.t, 0, output[tableName].TableOid)
+	output[tableName].TableOid = 0 // zero out TableOid for comparison
```
ilidemi, line 75: Wrong test file?

Amogh-Bharadwaj: During schema changes (in CDC) also we  do `SetupTableSchema` - which does `GetTableSchema` and pushes to table_schema_mapping.
That's why this test file is testing the table schemas returned by GetTableSchema. It was erroring out in CI because the `expectedTableSchema` in these tests didn't have `tableOid` specified.

There isn't a way for us to know ahead of time what the OID of the table created by this test would be and GetTableSchema queries Postgres anyway so doing that in the test as well would be redundant I guess

ilidemi: Ohhh we're fixing the tests here, got it. I was also thinking of asserting a non-zero value in api test as it theoretically could be comparing 0 to 0 https://github.com/PeerDB-io/peerdb/pull/3704#discussion_r2548912014

Amogh-Bharadwaj: Yup I pushed that

ilidemi: my bad looked at wrong commits

PR #3706, opened 2025-11-17
Add rekickoff gate to setup and snapshot flows

Amogh-Bharadwaj:
Adds a field to FlowConnectionConfigs and FlowConnectionConfigsCore - `rekickoff` - a boolean field which when set to true:
- SetupFlow will only get the table OIDs via ensurePullability and return the OID mapping
- SnapshotFlow exits immediately and does not run its activities


ilidemi (commented):
I thought we decided to not go through setup/snapshot? If the migration already happened, API has access to all the mappings and can start the cdc flow in exact state like it was

PR #3707, opened 2025-11-17
[snowflake] Feature Flag for skipping Avro compression algorithm

alexstoick:
We are observing a return of the Snowflake errors: `COPY INTO command: 100079 (22000): Invalid data encountered during decompression for file: 'not_file',compression type used: 'ZSTD', cause: 'Data corruption detected'`

These were previously resolved by using `AUTO_COMPRESS=false`; however it seems that in the meantime further changes make the `zst` compression applied to Avro files to no longer be in compliance with Snowflake, leading to failures to decode the files.

This change introduces a feature flag that allows for the skipping of the `zstd` compression on Avro files, making the files that land in Snowflake fully uncompressed.

---

**NB:**
We are successfully running v0.30.9 in production w/o this error happening, however on updating to v0.34.8 the error returns. I have attempted to find out where between v0.30.9 and v0.34.8 there was the breaking change which led to this issue returning however I was not successful and ran out of allocated time - so I'm suggesting this alternative which would allow us to move to a more recent version of `peerdb`.

### DIFF
--- a/flow/connectors/snowflake/qrep_avro_sync.go
+++ b/flow/connectors/snowflake/qrep_avro_sync.go
@@ -153,9 +153,17 @@ func (s *SnowflakeAvroSyncHandler) writeToAvroFile(
 	flowJobName string,
 ) (utils.AvroFile, error) {
 	if s.config.StagingPath == "" {
-		ocfWriter := utils.NewPeerDBOCFWriter(stream, avroSchema, ocf.ZStandard, protos.DBType_SNOWFLAKE)
+		compression := ocf.ZStandard
+		skipCompression, err := internal.PeerDBSnowflakeSkipCompression(ctx, s.config.Env)
+		if err != nil {
+			s.logger.Warn("Failed to load PEERDB_SNOWFLAKE_SKIP_COMPRESSION, proceeding without", slog.Any("error", err))
+		} else if skipCompression {
+			compression = ocf.Null
+		}
+
+		ocfWriter := utils.NewPeerDBOCFWriter(stream, avroSchema, compression, protos.DBType_SNOWFLAKE)
 		tmpDir := fmt.Sprintf("%s/peerdb-avro-%s", os.TempDir(), flowJobName)
-		err := os.MkdirAll(tmpDir, os.ModePerm)
+		err = os.MkdirAll(tmpDir, os.ModePerm)
 		if err != nil {
 			return utils.AvroFile{}, fmt.Errorf("failed to create temp dir: %w", err)
 		}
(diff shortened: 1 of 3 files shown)


flow/connectors/snowflake/qrep_avro_sync.go
```diff
@@ -153,9 +153,17 @@ func (s *SnowflakeAvroSyncHandler) writeToAvroFile(
...
 ) (utils.AvroFile, error) {
 	if s.config.StagingPath == "" {
-		ocfWriter := utils.NewPeerDBOCFWriter(stream, avroSchema, ocf.ZStandard, protos.DBType_SNOWFLAKE)
+		compression := ocf.ZStandard
+		skipCompression, err := internal.PeerDBSnowflakeSkipCompression(ctx, s.config.Env)
+		if err != nil {
+			s.logger.Warn("Failed to load PEERDB_SNOWFLAKE_SKIP_COMPRESSION, proceeding without", slog.Any("error", err))
+		} else if skipCompression {
+			compression = ocf.Null
+		}
```
jgao54, line 162: change looks good, but i think you can just re-use the existing variable:
```
    compression := ocf.ZStandard
	autoCompression, err := internal.PeerDBSnowflakeAutoCompress(ctx, s.config.Env)
	if err != nil {
		s.logger.Warn("Failed to load PEERDB_SNOWFLAKE_AUTO_COMPRESS, proceeding without", slog.Any("error", err))
	}
    if !autoCompression {
        compression = ocf.Null
    }        
```
this avoid inconsistent setting combinations (i.e PEERDB_SNOWFLAKE_AUTO_COMPRESS = false and PEERDB_SNOWFLAKE_SKIP_COMPRESSION = false)

alexstoick: Thanks for the comment! 

Funnily enough - the settings apply in different ways. The AUTO_COMPRESS stops Snowflake from zipping the input file.
It's perfectly fine to have:
AUTO_COMPRESS = false
SKIP_COMPRESSION = true

Arguably, skip compression could be named slightly better - happy to take a suggestion!

let me know if the above makes sense!

jgao54: got it, makes sense 👍  

Looking at snowflake doc [here](https://docs.snowflake.com/en/sql-reference/sql/put):
```
Specifies whether Snowflake uses gzip to compress files during upload:

TRUE: Snowflake compresses the files (if they are not already compressed).

FALSE: Snowflake doesn’t compress the files.

This option does not support other compression types. To use a different compression type, compress the file separately before executing the PUT command. Then, identify the compression type using the SOURCE_COMPRESSION option.
```
so it looks like in order to support zstd-compressed avro file for snowflake, it would require passing in `SOURCE_COMPRESSION`, but that's a separate change.

PR #3708, opened 2025-11-17
feat: added support for index and trigger migration

pandeykushagra51:
added support for trigger and index migration during snapshot and polling to periodically fetch trigger and index updates.

Implementation Details:

1. Implemented trigger and schema migration where following operation will be done:
   a. During initial migration (snapshot), the triggers and function will be fetched from source pg table and applied to dest table.
   b. There will be periodically scan of triggers on src at interval of 1 min and the updated triggers will be applied to destination table.

Triggers are created disabled on the destination to prevent double execution. If a source trigger increments a counter on UPDATE, the change replicates to the destination. If the destination trigger is enabled, it runs again and increments the counter twice, causing incorrect values. Since data changes come from replication, not local operations, enabling triggers would duplicate side effects. Keeping them disabled ensures the destination matches the source without extra trigger logic. Enable them only if the destination needs its own separate trigger behavior.

The plan is to make polling of trigger and index make it configurable through UI.
Polling was choosen as updating trigger and index with some delay will not break any think (as data is already written immediately so updating trigger with delay doesn't impact.)

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -1813,3 +1813,663 @@ func (a *FlowableActivity) ReportStatusMetric(ctx context.Context, status protos
 	)))
 	return nil
 }
+
+// MigrateTriggersAfterSnapshot migrates triggers from source to destination after snapshot completion.
+// It queries triggers from source and applies them to destination in a disabled state.
+// Retries up to 3 times with exponential backoff on failure.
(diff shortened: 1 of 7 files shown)


pandeykushagra51:
link to explaination:
https://drive.google.com/file/d/1SboAhRFDRpXAfMDjzmNMc45YMkmPMIiy/view?usp=sharing

PR #3715, opened 2025-11-18
track uncompressed bytes for mongodb connector

jgao54:
MongoDB Avro file has a very high compression ratio. This can result in OOM issues during file upload to ClickHouse during initial snapshot.  The proposed solution compute the uncompressed bytes of QRecords based on Avro encoding. It reuses the `PEERDB_S3_BYTES_PER_AVRO_FILE` variable to determine limit, and then based on connector source type, determines whether to compute the compressed bytes (which is conveniently computed by the WatchWriter and how we've tracked this historically) or uncompressed bytes (in the case for MongoDB Connector, where we manually compute bytes as we process QValues). If we want a different default, we could also introduce a new env variable instead, e.g. `PEERDB_S3_UNCOMPRESSED_BYTES_PER_AVRO_FILE`.  Keeping it to initial snapshot for now since that's where OOM issue is happening. If we want to extend this to CDC in the future, the computation would also be different due to the schema of the staging table.

Add unit tests
Run manual smoke test

### DIFF
--- a/flow/model/qrecord_avro_size_test.go
+++ b/flow/model/qrecord_avro_size_test.go
@@ -42,24 +42,24 @@ func TestMongoDBAvroSizeComputation(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
+			// Create temporary file
+			tmpfile, err := os.CreateTemp(t.TempDir(), fmt.Sprintf("test_avro_size_%s.avro", tc.name))
+			require.NoError(t, err)
+			defer os.Remove(tmpfile.Name())
+			defer tmpfile.Close()
+
 			schema := types.QRecordSchema{
 				Fields: []types.QField{
 					{Name: "_id", Type: types.QValueKindString, Nullable: false},
 					{Name: "doc", Type: types.QValueKindJSON, Nullable: false},
 				},
 			}
-			tmpFile := fmt.Sprintf("/tmp/test_avro_size_%s_%d.avro", tc.name, time.Now().Unix())
-			defer func() {
-				if err := os.Remove(tmpFile); err != nil && !os.IsNotExist(err) {
-					t.Logf("Warning: failed to remove temp file %s: %v", tmpFile, err)
-				}
-			}()
-			avroSchema, computedSize := writeAvroFileCompressed(t, schema, tc.numRecords, tmpFile)
+			avroSchema, computedSize := writeAvroFileCompressed(t, schema, tc.numRecords, tmpfile.Name())
 
 			t.Logf("Total computed size for %d records: %d bytes (%.2f MiB)",
 				tc.numRecords, computedSize, float64(computedSize)/(1024*1024))
 
-			actualSize := getActualUncompressedSize(t, tmpFile, avroSchema, tc.numRecords)
+			actualSize := getActualUncompressedSize(t, tmpfile.Name(), avroSchema, tc.numRecords)
 			t.Logf("Actual uncompressed Avro size for %d records: %d bytes (%.2f MiB)",
 				tc.numRecords, actualSize, float64(actualSize)/(1024*1024))
 
@@ -67,15 +67,15 @@ func TestMongoDBAvroSizeComputation(t *testing.T) {
 			diffPercent := float64(diff) / float64(actualSize) * 100
 			t.Logf("Difference: %d bytes (%.2f%%)", diff, diffPercent)
 
-			lowerBound := float64(actualSize) * 0.9
-			upperBound := float64(actualSize) * 1.1
+			lowerBound := float64(actualSize) * 0.95
+			upperBound := float64(actualSize) * 1.05
 			assert.GreaterOrEqual(t, float64(computedSize), lowerBound,
-				"Estimated size should be within 10% of actual (lower bound)")
+				"Estimated size should be within 5% of actual (lower bound)")
 			assert.LessOrEqual(t, float64(computedSize), upperBound,
-				"Estimated size should be within 10% of actual (upper bound)")
+				"Estimated size should be within 5% of actual (upper bound)")
 
 			// for completion, print out the compression ratio
-			fileInfo, err := os.Stat(tmpFile)
+			fileInfo, err := os.Stat(tmpfile.Name())
 			require.NoError(t, err)
 			compressedSize := fileInfo.Size()
 			compressionRatio := float64(actualSize) / float64(compressedSize)
@@ -178,19 +178,20 @@ func writeAvroFileCompressed(
 			types.QValueJSON{Val: randomizedDocContent},
 		}
 
-		avroMap, err := avroConverter.Convert(
+		avroMap, size, err := avroConverter.Convert(
 			context.Background(),
 			env,
 			record,
 			nil,
 			nil,
 			internal.BinaryFormatRaw,
+			true,
 		)
 		require.NoError(t, err)
 		err = encoder.Encode(avroMap)
 		require.NoError(t, err)
 
-		bytes += avroConverter.ComputeSize(record)
+		bytes += size
 	}
 
 	err = encoder.Flush()
--- a/flow/model/qvalue/avro_converter.go
+++ b/flow/model/qvalue/avro_converter.go
@@ -188,9 +188,16 @@ func QValueToAvro(
 	value types.QValue, field *types.QField, targetDWH protos.DBType, logger log.Logger,
 	unboundedNumericAsString bool, stat *NumericStat,
 	binaryFormat internal.BinaryFormat,
-) (any, error) {
+	calcSize bool,
+) (any, int64, error) {
+	// we only use union tag for nullable types
+	nullableSize := int64(0)
+	if calcSize && field.Nullable {
+		nullableSize = 1
+	}
+
 	if value.Value() == nil {
-		return nil, nil
+		return nil, nullableSize, nil
 	}
 
 	c := QValueAvroConverter{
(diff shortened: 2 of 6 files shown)


flow/model/qrecord_avro_size_test.go
```diff
@@ -0,0 +1,269 @@
...
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/internal"
+	"github.com/PeerDB-io/peerdb/flow/shared/types"
+)
+
+// This test verifies that avro size computation is sufficiently accurate for MongoDB documents
+// by comparing them against actual uncompressed Avro file sizes. MongoDB CDC connector
+// produces records with _id (string) and doc (JSON). We will need additional test cases
+// for broader schema coverage when we are introducing uncompressed bytes for other databases.
+func TestMongoDBAvroSizeComputation(t *testing.T) {
```
jgao54, line 24: Claude helped a lot here :)

flow/model/qvalue/avro_converter.go
```diff
@@ -679,3 +680,220 @@ func (ns *NumericStat) CollectWarnings(warnings *shared.QRepWarnings) {
...
+			size += arrTerminationByteSize
+		}
+	case types.QValueArrayFloat64:
+		count := int64(len(v.Val))
+		size = varIntSize(count) + (count * 8)
+		if count > 0 {
+			size += arrTerminationByteSize
+		}
+	case types.QValueArrayInt16:
+		size += varIntSize(int64(len(v.Val)))
```
ilidemi: why are some sizes `=` and some `+=`?

flow/model/qvalue/avro_converter_test.go
```diff
@@ -63,3 +70,234 @@ func TestColumnNameAvroFieldConvert(t *testing.T) {
...
+		{
+			name:         "array string small",
+			qvalue:       types.QValueArrayString{Val: []string{"a", "b"}},
+			expectedSize: 6, // varint(2) + (varint(1)+1) + (varint(1)+1) + varint(0)
+		},
+	}
+
+	for _, tt := range tests {
+		for _, nullable := range []bool{false, true} {
+			t.Run(tt.name, func(t *testing.T) {
```
ilidemi: Test names should include nullable, or just do two computes and two asserts

flow/model/qvalue/avro_converter_test.go
```diff
@@ -63,3 +70,234 @@ func TestColumnNameAvroFieldConvert(t *testing.T) {
...
+		{8192, 3},
+		{-1, 1},
+		{-64, 1},
+		{-65, 2},
+		{1000000000, 5},
+		{-1000000000, 5},
+	}
+
+	for _, tt := range tests {
+		t.Run(string(rune(tt.value)), func(t *testing.T) {
```
ilidemi:
```suggestion
		t.Run(fmt.Sprint(tt.value)), func(t *testing.T) {
```
Random runes can be unprintable

protos/flow.proto
```diff
@@ -398,6 +398,8 @@ message QRepConfig {
 
   repeated ColumnSetting columns = 27;
   uint32 version = 28;
+
+  peerdb_peers.DBType source_type = 30;
```
ilidemi, line 403:
```suggestion
  peerdb_peers.DBType source_type = 29;
```

flow/model/qvalue/avro_converter.go
```diff
@@ -679,3 +680,220 @@ func (ns *NumericStat) CollectWarnings(warnings *shared.QRepWarnings) {
...
+
+	if value.Value() == nil {
+		return unionTagSize
+	}
+
+	// Avro encodes arrays with a zero-byte terminator
+	arrTerminationByteSize := int64(1)
+
+	size := int64(0)
+	switch v := value.(type) {
```
ilidemi: Seems a little bad that the switch branches are duplicated when the size is tied to the value. How about:
```go
func QValueToAvro(..., calcSize bool) (any, int64, error) {
    switch v := value.(type) {
    case types.QValueFloat64:
        return c.processNullableUnion(float64(v.Val)), constSize(8, calcSize), nil
    case types.QValueString:
        return c.processNullableUnion(v.Value()), stringSize(v.Value(), calcSize), nil
    ...
}

func constSize(n int64, calcSize bool) int64 {
    if !calcSize {
        return 0
    }
    return n
}

func stringSize(s string, calcSize bool) int64 {
    if !calcSize {
        return 0
    }
    return calcVarIntSize(int64(len(s))) + int64(len(s))
}
```

flow/model/qrecord_avro_size_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			schema := types.QRecordSchema{
+				Fields: []types.QField{
+					{Name: "_id", Type: types.QValueKindString, Nullable: false},
+					{Name: "doc", Type: types.QValueKindJSON, Nullable: false},
+				},
+			}
+			tmpFile := fmt.Sprintf("/tmp/test_avro_size_%s_%d.avro", tc.name, time.Now().Unix())
```
ilidemi: TestWriteRecordsToAvroFileHappyPath has a pattern for temp files

flow/model/qrecord_avro_size_test.go
```diff
@@ -0,0 +1,269 @@
...
+			name:       "large_file",
+			numRecords: 500_000,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			schema := types.QRecordSchema{
+				Fields: []types.QField{
+					{Name: "_id", Type: types.QValueKindString, Nullable: false},
```
ilidemi, line 53: CDC path/raw table? 👉 👈 
(ok if not)

protos/flow.proto
jgao54: apparently i can't count 🤦‍♀️

jgao54: oh nvm, 29 is used up at line 375

flow/model/qvalue/avro_converter.go
jgao54: i actually started with this, and then switched to the current way for readability (since size can be 0), but the switch branches being a bit wasteful is a good point.

flow/model/qrecord_avro_size_test.go
jgao54: currently we are not tracking bytes for CDC path so i didn't pursue it, if we decide to can implement it as a follow-up

ilidemi: Didn't connect the dots around pushDataToS3 rename, sounds good 👍

flow/model/qvalue/avro_converter.go
```diff
@@ -205,103 +212,126 @@ func QValueToAvro(
...
+		return c.processNullableUnion(int64(v.Val)), varIntSize(int64(v.Val), nullableSize, calcSize), nil
 	case types.QValueUInt256:
-		return c.processUInt256(v.Val), nil
+		return c.processUInt256(v.Val), constSize(32, nullableSize, calcSize), nil
 	case types.QValueBoolean:
-		return c.processNullableUnion(v.Val), nil
+		return c.processNullableUnion(v.Val), constSize(1, nullableSize, calcSize), nil
 	case types.QValueNumeric:
-		return c.processNumeric(v.Val), nil
+		return c.processNumeric(v.Val), stringSize(v.Val.String(), nullableSize, calcSize), nil
```
ilidemi, line 276: If the implementation is not planned, please make a task to push the calculation to processNumeric() and a TODO for when flattened mode comes with Decimal128 or if we decide to use it for PG. Enabling this for PG today would mean double perf hit on long numerics (which is the heaviest type to process).

flow/model/qvalue/avro_converter.go
```diff
@@ -205,103 +212,126 @@ func QValueToAvro(
 	switch v := value.(type) {
 	case types.QValueInvalid:
 		// we will attempt to convert invalid to a string
-		return c.processNullableUnion(v.Val), nil
+		return c.processNullableUnion(v.Val), stringSize(v.Val, nullableSize, calcSize), nil
```
ilidemi, line 215: I think a good goal here is to do code golf on the extra characters in the switch statement. How about collapsing nullableSize and calcSize into something like `sizeFlag`/`sizeOpt` (`Skip`, `Nullable`, `Plain`) at the cost of little packing on top and unpacking in the utils?

flow/model/qvalue/avro_converter.go
```diff
@@ -205,103 +212,126 @@ func QValueToAvro(
...
-		return c.processArrayFloat32(v.Val), nil
+		return c.processArrayFloat32(v.Val), fixedElementArraySize(int64(len(v.Val)), 4, nullableSize, calcSize), nil
 	case types.QValueArrayFloat64:
-		return c.processArrayFloat64(v.Val), nil
+		return c.processArrayFloat64(v.Val), fixedElementArraySize(int64(len(v.Val)), 8, nullableSize, calcSize), nil
 	case types.QValueArrayInt16:
-		return c.processArrayInt16(v.Val), nil
+		return c.processArrayInt16(v.Val), variableElementArraySize(v.Val, func(elem int16) int64 {
+			return varIntSize(int64(elem), 0, calcSize)
+		}, nullableSize, calcSize), nil
```
ilidemi, line 294:
```suggestion
		return c.processArrayInt16(v.Val), varIntArraySize(v.Val, nullableSize, calcSize), nil
```

flow/model/qvalue/avro_converter.go
```diff
@@ -205,103 +212,126 @@ func QValueToAvro(
...
 	case types.QValueArrayInt64:
-		return c.processArrayInt64(v.Val), nil
+		return c.processArrayInt64(v.Val), variableElementArraySize(v.Val, func(elem int64) int64 {
+			return varIntSize(elem, 0, calcSize)
+		}, nullableSize, calcSize), nil
 	case types.QValueArrayString:
-		return c.processArrayString(v.Val), nil
+		return c.processArrayString(v.Val), variableElementArraySize(v.Val, func(elem string) int64 {
+			return stringSize(elem, 0, calcSize)
+		}, nullableSize, calcSize), nil
```
ilidemi, line 306:
```suggestion
		return c.processArrayString(v.Val), varStringArraySize(v.Val, nullableSize, calcSize), nil
```

flow/model/qvalue/avro_converter.go
```diff
@@ -205,103 +212,126 @@ func QValueToAvro(
...
 	case types.QValueArrayUUID:
-		return c.processArrayUUID(v.Val), nil
+		return c.processArrayUUID(v.Val), variableElementArraySize(v.Val, func(elem uuid.UUID) int64 {
+			return stringSize(elem.String(), 0, calcSize)
+		}, nullableSize, calcSize), nil
 	case types.QValueArrayNumeric:
-		return c.processArrayNumeric(v.Val), nil
+		return c.processArrayNumeric(v.Val), variableElementArraySize(v.Val, func(elem decimal.Decimal) int64 {
+			return stringSize(elem.String(), 0, calcSize)
+		}, nullableSize, calcSize), nil
```
ilidemi, line 332: For these, let's add a TODO and a `stringerArraySize[T fmt.Stringer](vals []T, ...)`, or push the calculation down (same decision as for QValueNumeric)

flow/model/qvalue/avro_converter.go
```diff
@@ -681,219 +711,56 @@ func (ns *NumericStat) CollectWarnings(warnings *shared.QRepWarnings) {
...
+		return nullableSize + 1
 	}
+	// Avro uses Variable-Length Zig-Zag encoding
 	encoded := uint64((n << 1) ^ (n >> 63))
 	bitLen := bits.Len64(encoded)
-	return int64((bitLen + 6) / 7)
+	return nullableSize + int64((bitLen+6)/7)
+}
+
+func fixedElementArraySize(count int64, elemSize int64, nullableSize int64, calcSize bool) int64 {
```
ilidemi, line 741:
```suggestion
func fixedArraySize(count int, elemSize int64, nullableSize int64, calcSize bool) int64 {
```

flow/model/qvalue/avro_converter.go
```diff
@@ -205,103 +212,126 @@ func QValueToAvro(
...
+		return c.processUInt256(v.Val), constSize(32, nullableSize, calcSize), nil
 	case types.QValueBoolean:
-		return c.processNullableUnion(v.Val), nil
+		return c.processNullableUnion(v.Val), constSize(1, nullableSize, calcSize), nil
 	case types.QValueNumeric:
-		return c.processNumeric(v.Val), nil
+		return c.processNumeric(v.Val), stringSize(v.Val.String(), nullableSize, calcSize), nil
 	case types.QValueBytes:
-		return c.processBytes(v.Val), nil
+		return c.processBytes(v.Val), stringSize(string(v.Val), nullableSize, calcSize), nil
```
ilidemi, line 278: `string(v.Val)` would mean a reallocation, including for large payloads. This should be computed in processBytes as`shared.UnsafeFastReadOnlyBytesToString()`

ilidemi (approved):

jgao54:
Totally supportive of all the changes in https://github.com/PeerDB-io/peerdb/compare/support-avro-uncompressed...support-avro-uncompressed-ilia2. Going to merge this as is -- to avoid merge conflicts -- and create a separate PR with your improvements -- i'll double-check the comments here and make sure they are addressed in the next PR.

PR #3717, opened 2025-11-19
fix(mysql): only error on minimal binlog_row_metadata if supported

heavycrystal:
`binlog_row_metadata` was added in MySQL 8.0.1, fix in line with existing conditional validation
Also MariaDB introduced `binlog_row_metadata` in 10.5.0, validation also needed a fix for this
Also try to fix the original panic

### DIFF
--- a/flow/connectors/mysql/ssh_keepalive_test.go
+++ b/flow/connectors/mysql/ssh_keepalive_test.go
@@ -2,6 +2,7 @@ package connmysql
 
 import (
 	"context"
+	"os"
 	"strconv"
 	"strings"
 	"testing"
@@ -61,7 +62,9 @@ func setupMySQLConnectorWithSSH(ctx context.Context, t *testing.T,
 
 func TestMySQLSSHKeepaliveWithToxiproxy(t *testing.T) {
 	// Skip if running MariaDB instead of MySQL
-	t.Skip("Skipping SSH keepalive test for TODO")
+	if os.Getenv("CI_MYSQL_VERSION") == "maria" {
+		t.Skip("Skipping SSH keepalive test for MariaDB")
+	}
 
 	ctx := t.Context()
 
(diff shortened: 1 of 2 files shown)


flow/connectors/mysql/cdc.go
```diff
@@ -316,6 +316,15 @@ func (c *MySqlConnector) PullRecords(
...
 	}
 
+	binlogRowMetadataSupported := true
+	cmp, err := c.CompareServerVersion(ctx, "8.0.1")
+	if err != nil {
+		return fmt.Errorf("failed to get server version: %w", err)
+	}
+	if cmp < 0 {
+		binlogRowMetadataSupported = false
+	}
```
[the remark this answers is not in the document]
ilidemi: 👍

flow/connectors/mysql/cdc.go
```diff
@@ -489,16 +500,18 @@ func (c *MySqlConnector) PullRecords(
...
-				e := exceptions.NewMySQLUnsupportedBinlogRowMetadataError(string(ev.Table.Schema), string(ev.Table.Table))
-				c.logger.Error(e.Error())
-				return e
-			}
 			sourceTableName := string(ev.Table.Schema) + "." + string(ev.Table.Table) // TODO this is fragile
 			destinationTableName := req.TableNameMapping[sourceTableName].Name
 			exclusion := req.TableNameMapping[sourceTableName].Exclude
 			schema := req.TableNameSchemaMapping[destinationTableName]
 			if schema != nil {
+				// The issue is global, but only error if we see a table in the pipe
```
jgao54: ooc, how does `schema  == nil`  happen, assuming it's a valid scenario? 

(or if this is a defensive check, in which case we should return an error)

flow/connectors/mysql/cdc.go
```diff
@@ -489,16 +500,18 @@ func (c *MySqlConnector) PullRecords(
...
+				// Otherwise users could be confused
+				if binlogRowMetadataSupported && len(ev.Table.ColumnName) == 0 && len(ev.Table.ColumnType) > 0 {
```
jgao54: nit: my bad for introducing this but the len(columnType) check here is not necessary, can be simplified to:
```suggestion
				if binlogRowMetadataSupported && ev.Table.ColumnName ==  nil {
```

PR #3722, opened 2025-11-20
fix(bigquery): fix GCS bucket listing in PullQRepObjects

jkaflik:
I have found two issues when processed BigQuery peer consisted of more than one partition:
- bucket listing prefix was not suffixed with table name
- missing skip on empty objects return in iterator - apparently that's how GCP API works

Interesting, GCS API is not deterministic here and not always listing issue can be reproduced this way. This PR makes it less error-prone, especially processing multiple tables.

Added a test case to process multiple partitions.

### DIFF
--- a/flow/e2e/bigquery_source_test.go
+++ b/flow/e2e/bigquery_source_test.go
@@ -445,6 +445,52 @@ func (s BigQueryClickhouseSuite) Test_Trips_Flow() {
 	env.Cancel(t.Context())
 }
 
+func (s BigQueryClickhouseSuite) Test_Trips_Flow_Small_Partitions() {
+	t := s.T()
+
+	source := s.Source().(*bigQuerySource)
+	srcTable := "trips_1k"
+	dstTable := "trips_1k_dst_small_partitions"
+
+	t.Logf("ClickHouse database: %s", s.Peer().Config.(*protos.Peer_ClickhouseConfig).ClickhouseConfig.Database)
+
+	count, err := source.helper.countRowsWithDataset(t.Context(), source.config.DatasetId, srcTable, "")
+	require.NoError(t, err, "should be able to count rows in source table")
+	require.Positive(t, count, "source table should have data")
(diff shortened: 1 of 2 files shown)


flow/e2e/bigquery_source_test.go
```diff
@@ -445,6 +445,52 @@ func (s BigQueryClickhouseSuite) Test_Trips_Flow() {
...
+				SourceTableIdentifier:      fmt.Sprintf("%s.%s", source.config.DatasetId, srcTable),
+				DestinationTableIdentifier: s.DestinationTable(dstTable),
+			},
+		},
+		Destination: s.Peer().Name,
+	}
+	flowConnConfig := connectionGen.GenerateFlowConnectionConfigs(s)
+	flowConnConfig.DoInitialSnapshot = true
+	flowConnConfig.InitialSnapshotOnly = true
+	flowConnConfig.SnapshotStagingPath = stagingTestBucket
```
jgao54, line 475: ooc, is this correct to interpret that we are running e2e test with a real gcs bucket?  asking to get a rough sense of whether we need to think about cost as we increase test frequencies. Currently *i think* all our e2e tests for db pipes uses minio

ilidemi: We run TestPeerFlowE2ETestSuiteGCS in CI, and the costs from these look negligible so far

ilidemi: Also BQ itself has costs, although that has been even lower so far

ilidemi (approved):

flow/e2e/bigquery_source_test.go
jgao54: It looks like there's [this project](https://github.com/goccy/bigquery-emulator) that could help with emulating BQ, just an FYI if we decide to mock it later on. 

Minio doesn't support gcs and i didn't see a good alternative out there either.

PR #3735, opened 2025-11-24
Cancel table addition feature

Amogh-Bharadwaj:
## Overview
This PR introduces a new feature -- cancelling ongoing table additions of CDC mirrors. Specifically, if a mirror has been paused and is adding tables but is stuck in the `Setup` or `Snapshot` phases, it can be brought back to CDC aka `Running`, and any tables as part of the addition request which finished snapshot will be part of the restored CDC.

## Why
Suppose you were to pause your CDC mirror and add 5 tables. Table addition of those 5 tables involve setup of those 5 tables where they are created on target, and then snapshot aka initial load. The underlying mirror is paused during this process to maintain data consistency. This can be problematic for a few reasons:
- Your destination tables there are growing stale.
- In the case of Postgres sources, your replication slot is growing and eating up your storage. Also, if a network interruption was to occur, the snapshot is invalid and you have no choice but to resync/create a new mirror.
- In the case of MySQL and MongoDB sources, your replication logs are getting closer to expiry.

We have seen cases where users add very large tables which take time and the above issues are more pronounced, making them wish they could cancel the table addition and have CDC heal their pipeline.

## What

Exposed via an API endpoint, this PR introduces a new CancelTableAdditionFlow temporal workflow. At a high-level, the workflow steps are as follows:
- Takes in as input the mirror name and a list of currently replicating tables.
- Checks the user-facing `qrep_runs` table to get a list of tables which have completed snapshot for this mirror.
- Cleans up the incomplete tables in `qrep_runs` 
- Obtains source table OIDs from catalog, owing to https://github.com/PeerDB-io/peerdb/pull/3704
- Uses the information obtained above to construct the required input to a new CDCFlowWorkflow with the same name as the originally running CDC flow. We construct the input in. a way where this new CDC flow skips setup and snapshot and directly goes to `Running`.

An end-to-end test has been added, and more are pending to be written.

---
_edit_

**NOTE**: Future migrations should also include an alerter call if there's an error so that it pages us quickly if something happens - #3769

### DIFF
--- a/flow/e2e/clickhouse_mv.go
+++ b/flow/e2e/clickhouse_mv.go
@@ -10,11 +10,10 @@ import (
 
 // ClickHouseMVManager manages materialized views for ClickHouse testing
 type ClickHouseMVManager struct {
-	config        *protos.ClickhouseConfig
-	tableName     string
-	mvName        string
-	suffix        string
-	isSourceMongo bool
+	config    *protos.ClickhouseConfig
+	tableName string
+	mvName    string
+	suffix    string
 }
 
 // NewClickHouseMVManager creates a new ClickHouseMVManager instance
@@ -22,14 +21,12 @@ func newClickHouseMVManager(
 	config *protos.ClickhouseConfig,
 	tableName string,
 	suffix string,
-	isSourceMongo bool,
 ) *ClickHouseMVManager {
 	return &ClickHouseMVManager{
-		config:        config,
-		tableName:     tableName,
-		mvName:        fmt.Sprintf("%s_mv_%s", tableName, suffix),
-		suffix:        suffix,
-		isSourceMongo: isSourceMongo,
+		config:    config,
+		tableName: tableName,
+		mvName:    fmt.Sprintf("%s_mv_%s", tableName, suffix),
+		suffix:    suffix,
 	}
 }
 
@@ -52,18 +49,6 @@ func (m *ClickHouseMVManager) CreateBadMV(ctx context.Context) error {
         )
         ENGINE = MergeTree()
         ORDER BY id`, targetName)
-	if m.isSourceMongo {
-		// Mongo mirrors always have fixed schema in ClickHouse:
-		// _id and doc fields
-		targetTableSQL = fmt.Sprintf(`
-		CREATE TABLE IF NOT EXISTS %s (
-			_id   String,
-			doc UInt32
-		)
-		ENGINE = MergeTree()
-		ORDER BY _id`, targetName)
-	}
-
 	err = ch.Exec(ctx, targetTableSQL)
 	if err != nil {
 		return fmt.Errorf("failed to create MV target table: %w", err)
@@ -75,19 +60,9 @@ func (m *ClickHouseMVManager) CreateBadMV(ctx context.Context) error {
         CREATE MATERIALIZED VIEW IF NOT EXISTS %s TO %s
         AS
         SELECT
-            id,
-            toUInt32(val) AS val
+            1 as id,
+			throwIf(1, 'Intentional MV failure') as val
         FROM %s`, m.mvName, targetName, m.tableName)
-	if m.isSourceMongo {
-		mvSQL = fmt.Sprintf(`
-    CREATE MATERIALIZED VIEW IF NOT EXISTS %s TO %s
-    AS
-    SELECT
-        _id,
-        throwIf(1, 'Intentional MV failure') AS doc
-    FROM %s`, m.mvName, targetName, m.tableName)
-	}
-
 	err = ch.Exec(ctx, mvSQL)
 	if err != nil {
 		return fmt.Errorf("failed to create materialized view: %w", err)
@@ -114,11 +89,9 @@ func (m *ClickHouseMVManager) DropBadMV(ctx context.Context) error {
 }
 
 func (s ClickHouseSuite) NewMVManager(tableName string, suffix string) *ClickHouseMVManager {
-	_, isSourceMongo := s.source.(*MongoSource)
 	return newClickHouseMVManager(
 		s.Peer().GetClickhouseConfig(),
 		tableName,
 		suffix,
-		isSourceMongo,
 	)
 }


flow/cmd/handler.go
```diff
@@ -69,47 +57,21 @@ func (h *FlowRequestHandler) cdcJobEntryExists(ctx context.Context, flowJobName
 func (h *FlowRequestHandler) createCdcJobEntry(ctx context.Context,
```
ilidemi, line 69: Can be inlined now

flow/cmd/worker.go
```diff
@@ -115,6 +115,12 @@ func WorkerSetup(ctx context.Context, opts *WorkerSetupOptions) (*WorkerSetupRes
 		TemporalClient: c,
 	})
 
+	w.RegisterActivity(&activities.CancelTableAdditionActivity{
+		CatalogPool:    conn,
+		Alerter:        alerting.NewAlerter(ctx, conn, otelManager),
+		OtelManager:    otelManager,
+		TemporalClient: c,
+	})
```
ilidemi, line 123:
```suggestion
	})

```

flow/shared/catalog.go
ilidemi, line 1: maybe catalog.go?

flow/cmd/cancel_table_addition.go
```diff
@@ -0,0 +1,56 @@
...
+
+	workflowID := fmt.Sprintf("table-addition-cancellation-%s-%s", req.FlowJobName, req.IdempotencyKey)
+
+	// Start the cancel table addition workflow
+	workflowOptions := client.StartWorkflowOptions{
+		ID:                       workflowID,
+		TaskQueue:                string(shared.PeerFlowTaskQueue),
+		TypedSearchAttributes:    shared.NewSearchAttributes(req.FlowJobName),
+		WorkflowIDConflictPolicy: tEnums.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
+		WorkflowIDReusePolicy:    tEnums.WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE,
```
ilidemi:
```suggestion
		WorkflowIDReusePolicy:    tEnums.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE,
```
if the retry comes after the workflow already ran, it shouldn't run again

flow/cmd/cancel_table_addition.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, NewInternalApiError(fmt.Errorf("failed to start cancel table addition workflow: %w", err))
+	}
+
+	slog.InfoContext(ctx, "Started cancel table addition workflow",
+		slog.String("flowJobName", req.FlowJobName),
+		slog.String("workflowID", workflowID),
+		slog.String("runID", run.GetRunID()))
+
+	return &protos.CancelTableAdditionOutput{
```
ilidemi: The caller probably needs the result, let's wait for it

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,221 @@
...
+		RetryPolicy: &temporal.RetryPolicy{
+			InitialInterval:    time.Second * 5,
+			BackoffCoefficient: 2.0,
+			MaximumInterval:    time.Minute * 1,
+			MaximumAttempts:    5,
+		},
+	}
+	getFlowConfigCtx := workflow.WithActivityOptions(ctx, getFlowConfigOptions)
+
+	var flowConfig *protos.FlowConnectionConfigs
```
ilidemi: The one that's returned from catalog is actually `FlowConnectionConfigsCore`. When the table mappings are not in it, there would be an activity getting a bunch of ids to large values in catalog, merging them and returning another id

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,221 @@
...
+	err = workflow.ExecuteActivity(cleanupCurrentParentMirrorCtx, cancelTableAddition.CleanupCurrentParentMirror, flowJobName).Get(ctx, nil)
+	if err != nil {
+		logger.Error("Failed to cleanup current parent mirror", "error", err)
+		return err
+	}
+
+	logger.Info("Successfully cleaned up current parent mirror workflow", "flowName", flowJobName)
+
+	createCdcJobEntryOptions := workflow.ActivityOptions{
+		StartToCloseTimeout: time.Minute * 3,
```
ilidemi:
```suggestion
		StartToCloseTimeout: time.Minute * 5,
```
splitting hairs tbh but the other PG reads are 5-15

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,221 @@
...
+		logger.Error("Failed to create CDC job entry", "error", err)
+		return err
+	}
+
+	logger.Info("Starting CDC flow with updated configuration and OID mappings",
+		"flowName", flowJobName, "snapshottedTables", len(snapshottedTables))
+
+	childWorkflowOptions := workflow.ChildWorkflowOptions{
+		WorkflowID:            childWorkflowID,
+		ParentClosePolicy:     enums.PARENT_CLOSE_POLICY_ABANDON,
```
ilidemi: Cool feature 👍 
If this was a QRepFlowWorkflow we'd be in trouble with updating status in catalog but this also comes with extra pointer in the UI that's actually good

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,243 @@
...
+		WHERE flow_name = $1 AND table_name NOT IN (SELECT unnest($2::text[]))`, flowJobName, completedDestinationTables)
+
+	if err != nil {
+		return fmt.Errorf("failed to cleanup incomplete tables in table_schema_mapping for flow %s: %w", flowJobName, err)
+	}
+
+	return nil
+}
+
+func (a *CancelTableAdditionActivity) GetTableOIDsFromCatalog(
```
ilidemi, line 97: nit: move one up to read in the same order as the workflow

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,243 @@
...
+		return "cleaning up qrep stats for incomplete tables"
+	})
+	defer shutdown()
+
+	completedSourceTables := make([]string, 0, len(completedTables))
+	for _, tm := range completedTables {
+		completedSourceTables = append(completedSourceTables, tm.SourceTableIdentifier)
+	}
+	completedDestinationTables := make([]string, 0, len(completedTables))
+	for _, tm := range completedTables {
```
ilidemi, line 148: nit: merge with the above one

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,243 @@
...
+		completedDestinationTables = append(completedDestinationTables, tm.DestinationTableIdentifier)
+	}
+	_, err := a.CatalogPool.Exec(ctx, `
+		DELETE FROM peerdb_stats.qrep_runs
+		WHERE parent_mirror_name = $1 AND consolidate_complete = false
```
ilidemi: We have to commit to the list of tables we consider complete at this point. Either terminate the CDC workflow before getting here or (my preference) occasionally clean up some `consolidate_complete = true`. Otherwise the tables are ghost-snapshotted in the UI

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,243 @@
...
+	TemporalClient client.Client
+	OtelManager    *otel_metrics.OtelManager
+}
+
+func (a *CancelTableAdditionActivity) GetCompletedTablesInQrepRuns(ctx context.Context, flowJobName string) ([]string, error) {
+	shutdown := heartbeatRoutine(ctx, func() string {
+		return "fetching completed tables from qrep_runs"
+	})
+	defer shutdown()
+	rows, err := a.CatalogPool.Query(ctx, "SELECT source_table FROM peerdb_stats.qrep_runs WHERE parent_mirror_name = $1 AND consolidate_complete = true", flowJobName)
```
ilidemi: Question, what happens when a table was added before, completed snapshot, then removed, and now added and stuck? Will there be a consolidate_complete=true for it?

flow/activities/cancel_table_addition_activity.go
ilidemi: Same in query below

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (a *CancelTableAdditionActivity) CleanupCurrentParentMirror(ctx context.Context, flowJobName string) error {
+	shutdown := heartbeatRoutine(ctx, func() string {
+		return "cleaning up current parent mirror and flow from catalog"
+	})
+	defer shutdown()
+	workflowID := shared.GetWorkflowID(flowJobName)
+	err := a.TemporalClient.TerminateWorkflow(ctx, workflowID, "", "Canceling due to table addition cancellation")
+	if err != nil {
```
ilidemi: Ignore a NotFound error - would happen during retry

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,243 @@
...
+
+	if flowStatus != protos.FlowStatus_STATUS_RUNNING {
+		return fmt.Errorf("expected flow %s to be in RUNNING state, but found %s", flowJobName, flowStatus.String())
+	}
+
+	internal.LoggerFromCtx(ctx).Info("New mirror flow is in RUNNING state", slog.String("flowName", flowJobName))
+	return nil
+}
+
+func (a *CancelTableAdditionActivity) CreateCdcJobEntry(
```
ilidemi: nit: move one up for reading flow

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,243 @@
...
+	ctx context.Context,
+	connectionConfigs *protos.FlowConnectionConfigsCore,
+	workflowID string,
+	idempotent bool,
+) error {
+	err := shared.CreateCdcJobEntry(ctx, a.CatalogPool, connectionConfigs, workflowID, idempotent)
```
ilidemi:
```suggestion
    // idempotent=false so that in an outlandish case when the flow entry is somehow there we have a chance to manually clean it up because its configs are wrong
	err := shared.CreateCdcJobEntry(ctx, a.CatalogPool, connectionConfigs, workflowID, false)
```

flow/shared/catalog.go
```diff
@@ -0,0 +1,63 @@
...
+
+	cfgBytes, err := proto.Marshal(connectionConfigs)
+	if err != nil {
+		return fmt.Errorf("unable to marshal flow config: %w", err)
+	}
+
+	if _, err = pool.Exec(ctx,
+		`INSERT INTO flows (workflow_id, name, source_peer, destination_peer, config_proto, status, description)
+        VALUES ($1,$2,$3,$4,$5,$6,'gRPC')`,
+		workflowID, connectionConfigs.FlowJobName, sourcePeerID, destinationPeerID, cfgBytes, protos.FlowStatus_STATUS_SETUP,
```
ilidemi: Status could be `RUNNING` now

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		logger.Error("Failed to confirm new mirror is running", "error", err)
+		return err
+	}
+
+	logger.Info("Cancel table addition flow completed successfully - CDC workflow started and running",
+		"flowName", flowJobName,
+		"finalTableCount", len(snapshottedTables))
+
+	return nil
```
ilidemi: completed tables can go in the output now

flow/cmd/cancel_table_addition.go
```diff
@@ -0,0 +1,56 @@
...
+	ctx context.Context,
+	req *protos.CancelTableAdditionInput,
+) (*protos.CancelTableAdditionOutput, APIError) {
+
+	flowStatus, err := internal.GetWorkflowStatus(ctx, h.pool, shared.GetWorkflowID(req.FlowJobName))
+	if err != nil {
+		return nil, NewInternalApiError(
+			fmt.Errorf("failed to get workflow status for flow %s: %w", req.FlowJobName, err))
+	}
+	if flowStatus != protos.FlowStatus_STATUS_SETUP &&
```
ilidemi, line 28: Can this really be canceled at any point during setup and snapshot? It's almost a UI button at this point if that's the case, but would need a detailed description of how that's safe to trigger at every step to actually believe it.

flow/activities/cancel_table_addition_activity.go
ilidemi: Also in the query below, didn't we just delete the thing we're joining with

flow/workflows/cancel_table_addition_flow.go
Amogh-Bharadwaj: When would table mappings not be in it?

ilidemi: If the work continues in #3589 or we later get a strong requirement in that direction. Basically if you have 10-100K mappings you can't put them in Temporal, but the API contract also can't change, so FlowConnectionConfigs stay the API input and FlowConnectionConfigsCore replace the big fields with small pointers to PG data. All of this just for general interest, adapting this feature for the new contracts would come as part of that future change.

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (a *CancelTableAdditionActivity) CleanupCurrentParentMirror(ctx context.Context, flowJobName string) error {
+	shutdown := heartbeatRoutine(ctx, func() string {
+		return "cleaning up current parent mirror and flow from catalog"
+	})
+	defer shutdown()
+	workflowID := shared.GetWorkflowID(flowJobName)
+	err := a.TemporalClient.TerminateWorkflow(ctx, workflowID, "", "Canceling due to table addition cancellation")
+	if err != nil {
+		if err.Error() != "workflow execution already completed" {
```
ilidemi: serviceerror.NotFound? It's mentioned in the docstring

flow/activities/cancel_table_addition_activity.go
```diff
@@ -128,26 +89,74 @@ func (a *CancelTableAdditionActivity) GetTableOIDsFromCatalog(
...
 
-	// Extract table OIDs from schemas, mapping OID to source table identifer
+	// Extract table OIDs from schemas, mapping OID to source table identifier
 	tableOIDs := make(map[uint32]string)
 	for _, tm := range tableMappings {
 		schema, exists := tableSchemas[tm.DestinationTableIdentifier]
 		if !exists {
 			return nil, fmt.Errorf("table schema not found in catalog for table %s in flow %s", tm.DestinationTableIdentifier, flowJobName)
 		}
-		if schema.TableOid == 0 {
```
ilidemi: Decided to not check?

Amogh-Bharadwaj: Yup accounting for MySQL and Mongo

flow/internal/workflow.go
```diff
@@ -33,6 +33,27 @@ func GetWorkflowStatus(ctx context.Context, pool shared.CatalogPool,
...
+			slog.WarnContext(ctx, "workflowId not found in catalog, will raise an error",
+				slog.String("flowName", flowName))
+			return protos.FlowStatus_STATUS_UNKNOWN, fmt.Errorf("workflowId not found in catalog: %w", err)
+		}
+		slog.ErrorContext(ctx, "failed to get status for flow from catalog",
+			slog.Any("error", err),
+			slog.String("flowName", flowName))
+		return flowStatus, fmt.Errorf("failed ot get status for flow from catalog: %w", err)
+	} else if flowStatus == protos.FlowStatus_STATUS_UNKNOWN {
+		slog.WarnContext(ctx, "flow status from catalog is unknown", slog.String("floflowNamewID", flowName))
```
ilidemi:
```suggestion
			slog.WarnContext(ctx, "flowName not found in catalog, will raise an error",
				slog.String("flowName", flowName))
			return protos.FlowStatus_STATUS_UNKNOWN, fmt.Errorf("flowName not found in catalog: %w", err)
		}
		slog.ErrorContext(ctx, "failed to get status for flow from catalog",
			slog.Any("error", err),
			slog.String("flowName", flowName))
		return flowStatus, fmt.Errorf("failed to get status for flow from catalog: %w", err)
	} else if flowStatus == protos.FlowStatus_STATUS_UNKNOWN {
		slog.WarnContext(ctx, "flow status from catalog is unknown", slog.String("flowName", flowName))
```

protos/flow.proto
```diff
@@ -631,3 +631,8 @@ message GetDefaultPartitionKeyForTablesInput {
 message GetDefaultPartitionKeyForTablesOutput {
   map<string, string> table_default_partition_key_mapping = 1;
 }
+
+message GetFlowConfigAndWorkflowIdFromCatalogOutput {
+  FlowConnectionConfigsCore flow_connection_configs = 1;
+  string original_workflow_id = 2;
```
ilidemi:
```suggestion
  string workflow_id = 2;
```
We aren't ever changing the workflow id so distinguishing the original would be confusing

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,284 @@
...
+
+	var removedTables []*protos.TableMapping
+	for _, mapping := range flowConfig.TableMappings {
+		if !currentlyReplicatingTableSet[mapping.SourceTableIdentifier] {
+			removedTables = append(removedTables, mapping)
+		}
+	}
+
+	if len(removedTables) > 0 {
+		logger.Error("Detected removed tables during table addition cancellation",
```
ilidemi: This is not an error yet at this point, can be moved into the `if` right below

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,284 @@
...
+			"flowName", flowJobName,
+			"removedTables", removedTables)
+		if !input.AssumeTableRemovalWillNotHappen {
+			return nil, fmt.Errorf("cannot cancel table addition because the following tables were removed during the operation: %v; "+
+				"please set assume_table_removal_will_not_happen to true to override and proceed with cancellation",
```
ilidemi:
```suggestion
			return nil, fmt.Errorf("cannot cancel table addition because the following tables are expected to be removed during the operation and cleanup for that is not implemented: %v; "+
				"if you're certain this won't happen, set assume_table_removal_will_not_happen to true to override and proceed with cancellation",
```

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,467 @@
...
+	ctx context.Context,
+	flowJobName string,
+	workflowId string,
+) ([]string, error) {
+	shutdown := heartbeatRoutine(ctx, func() string {
+		return "fetching completed tables from qrep_runs"
+	})
+	defer shutdown()
+
+	originalRunId, err := a.getRunIDOfLatestRunningPeerFlow(ctx, workflowId)
```
ilidemi:
```suggestion
	peerflowRunId, err := a.getRunIDOfLatestRunningPeerFlow(ctx, workflowId)
```
Original would be the very first one which is not what it is or what we want

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,467 @@
...
+
+type CancelTableAdditionActivity struct {
+	CatalogPool    shared.CatalogPool
+	Alerter        *alerting.Alerter
+	TemporalClient client.Client
+	OtelManager    *otel_metrics.OtelManager
+}
+
+/* GetCompletedTablesInQrepRunsForTableAddition gets the list of source tables whose latest QRep run is completed */
+func (a *CancelTableAdditionActivity) GetCompletedTablesInQrepRunsForTableAddition(
```
ilidemi:
```suggestion
func (a *CancelTableAdditionActivity) GetCompletedTablesFromQrepRuns
```

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,284 @@
...
+				removedTables)
+		}
+		logger.Warn("Proceeding with cancellation despite removed tables as per override flag",
+			"flowName", flowJobName,
+			"removedTables", removedTables)
+	}
+
+	// Get snapshotted tables from qrep_runs
+	getSnapshottedCompletedTablesOptions := workflow.ActivityOptions{
```
ilidemi:
```suggestion
	getCompletedTablesFromQrepRunsOptions := workflow.ActivityOptions{
```

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,467 @@
...
+			slog.String("flowName", flowJobName))
+		return []string{}, nil
+	}
+
+	runIds := make([]string, 0, len(listResp.Executions))
+	for _, execution := range listResp.Executions {
+		runIds = append(runIds, execution.Execution.RunId)
+	}
+
+	rows, err := a.CatalogPool.Query(ctx, `
```
ilidemi, line 75: With the RootRunId filtering can just do simple query with 3 filters, with the way consolidate_complete is set I don't think there's any chance of inconsistencies within a qrep run if the value in catalog is true

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,284 @@
...
+
+	logger.Info("Retrieved completed tables", "flowName", flowJobName, "completedCount", len(snapshottedSourceTables))
+
+	snapshottedTableSet := make(map[string]bool)
+	for _, tableName := range snapshottedSourceTables {
+		snapshottedTableSet[tableName] = true
+	}
+
+	finalListOfTables := make([]*protos.TableMapping, 0)
+	// final list of tables = tables in catalog + tables in this table addition that have completed snapshotting
```
ilidemi:
```suggestion
	// final list of tables = tables in catalog (still before addition/removal) + tables in this table addition that have completed snapshotting
```

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,284 @@
...
+
+	logger.Info("Retrieved PostgreSQL table OIDs", "flowName", flowJobName, "oidCount", len(tableOIDs))
+
+	// update table mappings for upcoming request
+	flowConfig.TableMappings = finalListOfTables
+	flowConfig.DoInitialSnapshot = false
+
+	state := cdc_state.NewCDCFlowWorkflowState(ctx, logger, flowConfig)
+	// update table OIDs for upcoming request
+	if len(tableOIDs) > 0 {
```
ilidemi: At this point they can't be empty, right?

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		logger.Error("Failed to get PostgreSQL table OIDs", "error", err)
+		return nil, err
+	}
+
+	logger.Info("Retrieved PostgreSQL table OIDs", "flowName", flowJobName, "oidCount", len(tableOIDs))
+
+	// update table mappings for upcoming request
+	flowConfig.TableMappings = finalListOfTables
+	flowConfig.DoInitialSnapshot = false
```
ilidemi: DoInitialSnapshot should stay as is, status=RUNNING is skipping this and we don't want to introduce discrepancies further on

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,284 @@
...
+		cancelTableAddition.CleanupCurrentParentMirror, flowJobName, originalWorkflowId).Get(ctx, nil)
+	if err != nil {
+		logger.Error("Failed to cleanup current parent mirror", "error", err)
+		return nil, err
+	}
+	logger.Info("Successfully cleaned up current parent mirror workflow", "flowName", flowJobName)
+
+	logger.Info("Starting CDC flow with updated configuration and OID mappings",
+		"flowName", flowJobName, "snapshottedTables", len(finalListOfTables))
+	// Start new CDC flow as a regular workflow via activity
```
ilidemi:
```suggestion
	// Start new CDC flow as a regular (not child) workflow via activity
```

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,467 @@
...
+
+	slog.InfoContext(ctx, "Successfully started new CDC workflow",
+		slog.String("flowName", flowConfig.FlowJobName),
+		slog.String("workflowID", workflowID),
+		slog.String("runID", run.GetRunID()))
+
+	return nil
+}
+
+func (a *CancelTableAdditionActivity) getRunIDOfLatestRunningPeerFlow(ctx context.Context, workflowId string) (string, error) {
```
ilidemi, line 444:
```suggestion
func (a *CancelTableAdditionActivity) getRunIDOfRunningPeerFlow(ctx context.Context, workflowId string) (string, error) {
```

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,467 @@
...
+}
+
+func (a *CancelTableAdditionActivity) CleanupCurrentParentMirror(ctx context.Context, flowJobName string, workflowId string) error {
+	shutdown := heartbeatRoutine(ctx, func() string {
+		return "cleaning up current parent mirror and flow from catalog"
+	})
+	defer shutdown()
+
+	// Describe to get latest run
+	originalRunId, err := a.getRunIDOfLatestRunningPeerFlow(ctx, workflowId)
```
ilidemi:
```suggestion
	rootRunId, err := a.getRunIDOfLatestRunningPeerFlow(ctx, workflowId)
```

flow/activities/cancel_table_addition_activity.go
```diff
@@ -0,0 +1,467 @@
...
+
+	workflowOptions := client.StartWorkflowOptions{
+		ID:                       workflowID,
+		TaskQueue:                string(shared.PeerFlowTaskQueue),
+		TypedSearchAttributes:    shared.NewSearchAttributes(flowConfig.FlowJobName),
+		WorkflowIDConflictPolicy: tEnums.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING, // idempotent behavior
+		WorkflowIDReusePolicy:    tEnums.WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE, // allow reuse for retries
+	}
+
+	run, err := a.TemporalClient.ExecuteWorkflow(ctx, workflowOptions, "CDCFlowWorkflow", flowConfig, state)
```
ilidemi, line 431: In a perfect world it would be possible to get compiler help with `"CDCFlowWorkflow"` (and `'QRepFlowWorkflow'` above) but renaming workflows is a rare and risky enough event that we'll trust the person doing that to do find and replace. Just expressing thoughts, no action needed

flow/connectors/postgres/postgres.go
```diff
@@ -943,6 +943,66 @@ func (c *PostgresConnector) GetSelectedColumns(
...
+	}
+
+	rows, err := c.conn.Query(ctx, getTablesSQL, args...)
+	if err != nil {
+		return nil, fmt.Errorf("error getting tables from publication %s: %w", publicationName, err)
+	}
+	defer rows.Close()
+
+	var tables []*utils.SchemaTable
+	for rows.Next() {
```
ilidemi: CollectRows please

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,284 @@
...
+		HeartbeatTimeout:    2 * time.Minute,
+		RetryPolicy: &temporal.RetryPolicy{
+			InitialInterval:    time.Second * 10,
+			BackoffCoefficient: 2.0,
+			MaximumInterval:    time.Minute * 5,
+		},
+	}
+	cleanupIncompleteTablesCtx := workflow.WithActivityOptions(ctx, cleanupIncompleteTablesOptions)
+
+	err = workflow.ExecuteActivity(cleanupIncompleteTablesCtx, cancelTableAddition.CleanupIncompleteTablesInStats,
```
ilidemi, line 265: ~~Just remembered, there's also `CleanupQRepFlow` to drop the stage. I think we can return the deleted flow_names from table cleanup and change the args of `dstConn.CleanupQRepFlow` to `stagingPath, flowJobName string` and should be good to do an activity~~
Ok to not do for now to get this merged in - TTL is taking care of these anyway

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+				fmt.Sprintf("expected no table_schema_mapping entry for table %s, got %d",
+					destinationTableName, actualCount))
+		}
+	}
+}
+
+func (s APITestSuite) testCancelTableAddition(
+	assumeTableRemovalWillNotHappen bool, withRemoval bool) {
+	var cols string
+	switch s.source.(type) {
```
ilidemi, line 319: Please use loops, like `range []string{"t1", "t2", "t3", "t4", "t5", "t6"}`

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+
+func (s APITestSuite) checkQrepRuns(
+	flowJobName string,
+	expectedTables []includedTable,
+) {
+	for _, tableEntry := range expectedTables {
+		var actualCount int
+		queryErr := s.pg.PostgresConnector.Conn().QueryRow(s.t.Context(),
+			`SELECT COUNT(*) FROM peerdb_stats.qrep_runs 
+            WHERE parent_mirror_name=$1 AND source_table = $2 AND consolidate_complete = true`,
```
ilidemi: Probably need to query all tables by parent mirror name, make sure the list matches and `consolidate_complete` is all set

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+	s.checkQrepRuns(
+		flowConnConfig.FlowJobName,
+		[]includedTable{
+			{tableName: AttachSchema(s, "t1"), entries: 1},
+			{tableName: AttachSchema(s, "t2"), entries: 1},
+			{tableName: AttachSchema(s, "t3"), entries: 1},
+			{tableName: AttachSchema(s, "t4"), entries: 1},
+		},
+	)
+	s.checkQrepRuns(
```
ilidemi: `checkQrepPartitions`?

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+
+func (s APITestSuite) checkTableSchemaMapping(
+	flowJobName string,
+	expectedTables []includedTable,
+) {
+	for _, tableEntry := range expectedTables {
+		destinationTableName := tableEntry.tableName
+		var actualCount int
+		queryErr := s.pg.PostgresConnector.Conn().QueryRow(s.t.Context(),
+			`SELECT COUNT(*) FROM table_schema_mapping
```
ilidemi: Need to query by flow name and compare sets, what if there are leftovers

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+}
+
+func (s APITestSuite) checkQrepPartitions(
+	flowJobName string,
+	expectedTables []includedTable,
+) {
+	for _, tableEntry := range expectedTables {
+		var actualCount int
+		queryErr := s.pg.PostgresConnector.Conn().QueryRow(s.t.Context(),
+			`SELECT COUNT(*) FROM peerdb_stats.qrep_partitions qp
```
ilidemi: Let's query by `qrep_runs.parent_mirror_name`, then compare table lists and no need to check `consolidate_complete` the second time

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+func (s APITestSuite) checkPublicationTables(
+	publicationName string,
+	includedTables []*utils.SchemaTable,
+	cancelledTables []*utils.SchemaTable,
+) {
+
+	pubTables, err := s.pg.PostgresConnector.GetTablesFromPublication(
```
ilidemi, line 150: If we're getting all tables, can just take `includedTables` as an argument, would also account for strays

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+	}
+
+	// drop target table
+	err = s.ch.DropTable("t5")
+	require.NoError(s.t, err)
+	err = s.ch.DropTable("t6")
+	require.NoError(s.t, err)
+
+	// insert a row into all original tables
+	switch s.source.(type) {
```
ilidemi: Loops please

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+	default:
+		require.Fail(s.t, fmt.Sprintf("unknown source type %T", s.source))
+	}
+
+	EnvWaitForEqualTables(env, s.ch, "cdc after cancellation t1", "t1", cols)
+	EnvWaitForEqualTables(env, s.ch, "cdc after cancellation t2", "t2", cols)
+	EnvWaitForEqualTables(env, s.ch, "cdc after cancellation t3", "t3", cols)
+	EnvWaitForEqualTables(env, s.ch, "cdc after cancellation t4", "t4", cols)
+
+	s.checkMetadataLastSyncStateValues(env, flowConnConfig, "batch id check after cdc", 1, 1)
```
ilidemi, line 577:
```suggestion
	// Batch finished without erroring on t5 and t6 that don't have a target table anymore
	s.checkMetadataLastSyncStateValues(env, flowConnConfig, "batch id check after cdc", 1, 1)
```

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+	s.testCancelTableAddition(false, true)
+}
+
+func (s APITestSuite) TestCancelTableAddition_NoRemovalAssumedWithRemoval() {
+	s.testCancelTableAddition(true, true)
+}
+
+func (s APITestSuite) TestCancelTableAdditionRemoveAddRemove() {
+	var cols string
+	switch s.source.(type) {
```
ilidemi, line 606: Would be a liiittle shorter with loops

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+
+func (s APITestSuite) TestCancelTableAddition_WithRemoval() {
+	s.testCancelTableAddition(false, true)
+}
+
+func (s APITestSuite) TestCancelTableAddition_NoRemovalAssumedWithRemoval() {
+	s.testCancelTableAddition(true, true)
+}
+
+func (s APITestSuite) TestCancelTableAdditionRemoveAddRemove() {
```
ilidemi, line 597:
```suggestion
// Tests that table addition cancellation doesn't get confused by the canceled table having a previous initial load and a previous successful addition then removal
func (s APITestSuite) TestCancelTableAdditionRemoveAddRemove() {
```

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+	}
+	// sort and compare
+	require.ElementsMatch(s.t, expectedTables, outputSourceTables,
+		"expected tables after cancellation to match")
+
+	s.checkQrepRuns(
+		flowConnConfig.FlowJobName,
+		[]includedTable{
+			{tableName: AttachSchema(s, "t1"), entries: 1},
+			{tableName: AttachSchema(s, "t2"), entries: 2},
```
ilidemi, line 707:
```suggestion
			{tableName: AttachSchema(s, "t2"), entries: 2}, // initial snapshot and first add but not the second add
```

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+		},
+	)
+	s.checkQrepPartitions(
+		flowConnConfig.FlowJobName,
+		[]includedTable{
+			{tableName: AttachSchema(s, "t1"), entries: 1},
+			{tableName: AttachSchema(s, "t2"), entries: 2},
```
ilidemi:
```suggestion
			{tableName: AttachSchema(s, "t2"), entries: 2}, // initial snapshot and first add but not the second add
```

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+	require.NoError(s.t, err)
+
+	// insert a row into all original tables
+	switch s.source.(type) {
+	case *PostgresSource, *MySqlSource:
+		require.NoError(s.t, s.source.Exec(s.t.Context(),
+			fmt.Sprintf("INSERT INTO %s(id, val) values (2,'second')", AttachSchema(s, "t1"))))
+		require.NoError(s.t, s.source.Exec(s.t.Context(),
+			fmt.Sprintf("INSERT INTO %s(id, val) values (2,'second')", AttachSchema(s, "t2"))))
+	case *MongoSource:
```
ilidemi: nit: mongo could use a loop

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+
+	s.checkMetadataLastSyncStateValues(env, flowConnConfig, "batch id check after cdc", 1, 1)
+
+	env.Cancel(s.t.Context())
+	RequireEnvCanceled(s.t, env)
+}
+
+func (s APITestSuite) TestCancelTableAdditionDuringSetupFlow() {
+	var cols string
+	switch s.source.(type) {
```
ilidemi, line 998: Please add loops

flow/e2e/cancel_table_addition_test.go
```diff
@@ -0,0 +1,910 @@
...
+	require.NotNil(s.t, response)
+	tc := NewTemporalClient(s.t)
+	env, err := GetPeerflow(s.t.Context(), s.pg.PostgresConnector.Conn(), tc, flowConnConfig.FlowJobName)
+	require.NoError(s.t, err)
+	SetupCDCFlowStatusQuery(s.t, env, flowConnConfig)
+	EnvWaitFor(s.t, env, 3*time.Minute, "wait for initial load to finish", func() bool {
+		return env.GetFlowStatus(s.t) == protos.FlowStatus_STATUS_RUNNING
+	})
+	RequireEqualTables(s.ch, "original", cols)
+	// add table
```
ilidemi:
```suggestion

```
This one is confusing, it's on the same level as the further ones but is a parent

ilidemi (commented):
Tests look great so far!

flow/workflows/cancel_table_addition_flow.go
Amogh-Bharadwaj: Removing it since the content is covered in the subsequent logs

flow/e2e/cancel_table_addition_test.go
Amogh-Bharadwaj: Wouldn't we not want to filter by includedTables if we were checking for strays? (You can't have 2 entries with same (schemaname, tablename) in pg_publication_tables)

flow/workflows/cancel_table_addition_flow.go
Amogh-Bharadwaj: In the case of mysql and mongo they can be

flow/workflows/cancel_table_addition_flow.go
```diff
@@ -0,0 +1,280 @@
...
+	var removedTables []*protos.TableMapping
+	for _, mapping := range flowConfig.TableMappings {
+		if !currentlyReplicatingTableSet[mapping.SourceTableIdentifier] {
+			removedTables = append(removedTables, mapping)
+		}
+	}
+
+	if len(removedTables) > 0 {
+		if !input.AssumeTableRemovalWillNotHappen {
+			return nil, fmt.Errorf("cannot cancel table addition because the following tables are expected to be removed during the operation and cleanup for that is not implemented: %v; "+
```
ilidemi: The caller workflow would need to handle this error and error out without retrying forever. Maybe introduce a well-known error, match it on the API side and return a PreconditionFailed?

flow/workflows/cancel_table_addition_flow.go
ilidemi: Looking at `GetTableOIDsFromCatalog`, the only places it returns nil are error branches, and oids from table mappings are assigned unconditionally, so it would be a map of zeros for them

flow/activities/cancel_table_addition_activity.go
ilidemi: bump

flow/e2e/cancel_table_addition_test.go
ilidemi: Thinking just the includedTables loop and len(includedTables) == len(pubTables) would do it, by strays I mean something in the publication that's not in cancelledTables

flow/e2e/api_test.go
```diff
@@ -95,8 +95,8 @@ func (s APITestSuite) checkMetadataLastSyncStateValues(
 	env WorkflowRun,
 	flowConnConfig *protos.FlowConnectionConfigs,
 	reason string,
-	expectedSyncBatchId int64,
-	expectedNormalizeBatchId int64,
+	expectedSyncBatchId int64, //nolint:unparam
+	expectedNormalizeBatchId int64, //nolint:unparam
```
ilidemi, line 99: Are these needed?

Amogh-Bharadwaj: Currently linter is complaining about these 2 always being set as 1 in all the call sites, but I could easily see this being used in future tests where have multiple CDC batches (I think we already do)

ilidemi: Ahh thought they got confused for completely unused at some point

PR #3739, opened 2025-11-25
support Avro uncompressed - improvements

jgao54:
Follow-up to https://github.com/PeerDB-io/peerdb/pull/3715

Courtesy of @ilidemi

### DIFF
--- a/flow/model/qvalue/avro_converter.go
+++ b/flow/model/qvalue/avro_converter.go
@@ -812,6 +812,9 @@ func fixedArraySize(count int, elemSize int64, so sizeOpt) int64 {
 }
 
 func varIntArraySize[T ~int16 | ~int32 | ~int64](vals []T, so sizeOpt) int64 {
+	if so == sizeSkip {
+		return 0
+	}
 	totalElemSize := int64(0)
 	for _, elem := range vals {
 		totalElemSize += varIntSize(int64(elem), sizePlain)
@@ -820,6 +823,9 @@ func varIntArraySize[T ~int16 | ~int32 | ~int64](vals []T, so sizeOpt) int64 {
 }
 
 func stringArraySize(vals []string, so sizeOpt) int64 {
+	if so == sizeSkip {
+		return 0
+	}
 	totalElemSize := int64(0)
 	for _, elem := range vals {
 		totalElemSize += stringSize(elem, sizePlain)


flow/model/qvalue/avro_converter.go
```diff
@@ -711,56 +780,63 @@ func (ns *NumericStat) CollectWarnings(warnings *shared.QRepWarnings) {
...
-	}
-	return size
+func fixedArraySize(count int, elemSize int64, so sizeOpt) int64 {
+	return arraySize(count, int64(count)*elemSize, so)
 }
 
-func variableElementArraySize[T any](vals []T, elemSizeFn func(T) int64, nullableSize int64, calcSize bool) int64 {
-	if !calcSize {
-		return 0
+func varIntArraySize[T ~int16 | ~int32 | ~int64](vals []T, so sizeOpt) int64 {
```
jgao54, line 814: TIL: this is awesome -- was not aware that generic supports union type 🔥

flow/model/qvalue/avro_converter.go
```diff
@@ -711,56 +780,69 @@ func (ns *NumericStat) CollectWarnings(warnings *shared.QRepWarnings) {
...
+		totalElemSize += varIntSize(int64(elem), sizePlain)
 	}
-	return size
+	return arraySize(len(vals), totalElemSize, so)
 }
 
-func variableElementArraySize[T any](vals []T, elemSizeFn func(T) int64, nullableSize int64, calcSize bool) int64 {
-	if !calcSize {
+func stringArraySize(vals []string, so sizeOpt) int64 {
+	if so == sizeSkip {
```
jgao54, line 826: added this check for `varIntArraySize` and `stringArraySize`, given they are called directly by some data types, this avoids the need to loop thru arrays unnecessarily when size is skipped.

flow/model/qvalue/avro_converter.go
```diff
@@ -520,38 +572,50 @@ func (c *QValueAvroConverter) processArrayBoolean(arrayData []bool) any {
...
 		transformedTimeArr := make([]string, 0, len(arrayDate))
+		totalElemSize := int64(0)
 		for _, t := range arrayDate {
-			transformedTimeArr = append(transformedTimeArr, t.Format("2006-01-02"))
+			s := t.Format("2006-01-02")
+			transformedTimeArr = append(transformedTimeArr, s)
+			totalElemSize += stringSize(s, sizePlain)
 		}
-		return transformedTimeArr
+		return transformedTimeArr, arraySize(len(arrayDate), totalElemSize, so)
```
jgao54, line 609: the original reason I was hesitant to introduce size calculation inside the `process*` functions was because current implementation of size calculation is optional and specific to mongo.

Just noting that this PR introduces a tiny overhead of also computing the size (and then discarding it) when it's `sizeSkip == true` , which is fine because addition is cheap, and because it would be beneficial to support uncompressed bytes in PG/MySQL as well, but just want to point this out.

flow/model/qvalue/avro_converter_test.go
```diff
@@ -76,417 +78,349 @@ func TestColumnNameAvroFieldConvert(t *testing.T) {
...
+	t *testing.T, ctx context.Context, qv types.QValue, field *types.QField, unboundedNumericAsString bool,
+) (any, int64) {
+	t.Helper()
+	avroVal, computedSize, err := QValueToAvro(ctx, qv, field, protos.DBType_CLICKHOUSE,
+		log.NewStructuredLogger(nil), unboundedNumericAsString, nil, internal.BinaryFormatRaw, true)
+	require.NoError(t, err)
+	return avroVal, computedSize
+}
+
+func avroEncodedSize(t *testing.T, schema avro.Schema, value any) int64 {
```
jgao54, line 421: much better than hand-rolling the math

flow/model/qvalue/avro_converter.go
ilidemi: Opus is better at Go than I am 👼

ilidemi (approved):

PR #3742, opened 2025-11-26
better error handling for getPartitions

jgao54:
make sure `getPartitions` error gets surfaced right away to make log less confusing.

### DIFF
--- a/flow/connectors/postgres/qrep.go
+++ b/flow/connectors/postgres/qrep.go
@@ -73,13 +73,16 @@ func (c *PostgresConnector) GetQRepPartitions(
 	}
 
 	partitions, err := c.getPartitions(ctx, getPartitionsTx, config, last)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get partitions: %w", err)
+	}
 
 	// commit transaction
 	if err := getPartitionsTx.Commit(ctx); err != nil {
 		return nil, fmt.Errorf("failed to commit transaction: %w", err)
 	}
 
-	return partitions, err
+	return partitions, nil
 }
 
 func (c *PostgresConnector) GetDefaultPartitionKeyForTables(


flow/connectors/postgres/qrep.go
```diff
@@ -73,13 +73,16 @@ func (c *PostgresConnector) GetQRepPartitions(
...
 
 	partitions, err := c.getPartitions(ctx, getPartitionsTx, config, last)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get partitions: %w", err)
+	}
 
 	// commit transaction
 	if err := getPartitionsTx.Commit(ctx); err != nil {
 		return nil, fmt.Errorf("failed to commit transaction: %w", err)
 	}
```
jgao54, line 83: should we just remove this commit for read-only transactions, is there a reason this was introduced to begin with? cc @serprex / @heavycrystal  if you have context

serprex: you need to commit in order to close the transaction

jgao54: on line 69, we always rollback on function exit already

serprex: rollback's okay, just aesthetically displeasing

serprex (approved):

flow/connectors/postgres/qrep.go
jgao54: got it, was thinking that since we don't need to commit, this feels like it introduces a potential point of failure unnecessarily if the `getPartitions` above it succeeds. However, if the read-only transaction fails to commit, it seems like it could be an indicator of some other problem that's worth knowing. 

going to leave this in for now, but if we do see an error in the future that was a false alarm, we can evaluate removing it.

Just wanted to make sure there wasn't some other important reason I am not aware of w.r.t. to this commit.

PR #3746, opened 2025-11-27
make json the default env for mongo e2e tests

jgao54:
Forgot to do this in the past led to longer debugging session than needed.

### DIFF
--- a/flow/e2e/mongo_test.go
+++ b/flow/e2e/mongo_test.go
@@ -16,6 +16,7 @@ import (
 	"go.mongodb.org/mongo-driver/v2/mongo"
 	"go.mongodb.org/mongo-driver/v2/mongo/options"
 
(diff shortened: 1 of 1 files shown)


flow/e2e/mongo_test.go
```diff
@@ -581,36 +588,38 @@ func (s MongoClickhouseSuite) Test_Enable_Json() {
 	}
 	flowConnConfig := connectionGen.GenerateFlowConnectionConfigs(s)
 	flowConnConfig.DoInitialSnapshot = true
-	flowConnConfig.Env = map[string]string{"PEERDB_CLICKHOUSE_ENABLE_JSON": "true"}
+	flowConnConfig.Env = map[string]string{"PEERDB_CLICKHOUSE_ENABLE_JSON": "false"}
```
ilidemi: Should this one be setting the value rather than the full map?

PR #3749, opened 2025-11-27
Query replication mirrors UI: change upsert key columns input to textfield

Amogh-Bharadwaj:
In a query where they are having aliases for columns (i.e column A in source maps to column B in target), the upsert key columns for upsert qrep mode need to column B (i,e the destination columns).
Currently we list source columns in the UI, this PR changes this to allow typing whatever columns are needed

### DIFF
--- a/ui/app/mirrors/create/qrep/upsertcols.tsx
+++ b/ui/app/mirrors/create/qrep/upsertcols.tsx
@@ -9,14 +9,12 @@ import {
 import { MirrorSetting } from '../helpers/common';
 
 interface UpsertColsProps {
-  columns: { value: string; label: string }[];
   loading: boolean;
   setter: Dispatch<SetStateAction<QRepConfig>>;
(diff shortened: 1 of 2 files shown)


ui/app/mirrors/create/qrep/upsertcols.tsx
```diff
@@ -1,96 +1,93 @@
...
-    }
-    const uniqueColsArr = Array.from(uniqueColumnsSet);
-    setting.stateHandler(uniqueColsArr, setter);
+    setting.stateHandler(columnsArray, setter);
   };
 
   useEffect(() => {
-    const uniqueColsArr = Array.from(uniqueColumnsSet);
+    // Parse comma-separated values and update the config
+    const columnsArray = upsertColumnsText
```
serprex, line 38: can memo this just to not have to repeat the expression so much

PR #3750, opened 2025-11-28
pg qrep: config for avro schema nullable, debug logging

ilidemi:
Current situation:
1. `failed to sync records: failed to write records to S3: failed to write records to OCF writer: failed to write record to OCF: some_column_100: avro: *avro.null is unsupported for Avro long` occurs once a month for one customer
2. The column in question is a nullable integer, all values in it are null, it is also inherited from a parent table
3. Parent table has same name but different schema
4. Parent and child attnums are different, so possibly column deletions were involved, and also child seems to have become inherited later on.
5. I was unable to reproduce the error with that information. It is possible the schema has changed since, so it would be nice to capture the exact data when we have it.

This change is a spiritual successor of #3613, but defaults to strict behavior (which has been working fine for everyone else) and puts the lax one under a config. When lax is enabled, it collects all the inputs that go into deciding whether a column would be nullable under strict behavior, then some extra about table inheritance, and logs it later if any mismatch with strict was detected. Tested that the new logic runs and logs if the code is adjusted to under-do nullable, but as-is the test doesn't do much as the issue is not cleanly reproducible yet.

Also adding a generic code notification metric that's easy to emit from anywhere, will set up a non-paging alert on it once this goes in.

The plan is to enable the setting just for one service, leave it to bake for another month or two, then check back when the notification fires. After the issue is sorted out, all the null tracking can be removed.

### DIFF
--- a/flow/internal/dynamicconf.go
+++ b/flow/internal/dynamicconf.go
@@ -120,7 +120,7 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		TargetForSetting: protos.DynconfTarget_ALL,
 	},
 	{
-		Name:             "PEERDB_NULLABLE_LAX",
+		Name:             "PEERDB_AVRO_NULLABLE_LAX",
 		Description:      "Make all columns nullable in initial load Avro schema (disables strict nullable checking)",
 		DefaultValue:     "false",
 		ValueType:        protos.DynconfValueType_BOOL,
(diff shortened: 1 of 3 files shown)


flow/model/conversion_avro.go
```diff
@@ -161,3 +173,54 @@ func ConstructColumnNameAvroFieldMap(fields []types.QField) map[string]string {
...
+		return nil // Not in lax mode
+	}
+	return &NullMismatchTracker{
+		schemaDebug:    schemaDebug,
+		mismatchedCols: make([]bool, len(schemaDebug.StrictNullable)),
+	}
+}
+
+func (t *NullMismatchTracker) RecordNull(fieldIdx int) {
+	if fieldIdx < len(t.schemaDebug.StrictNullable) && !t.schemaDebug.StrictNullable[fieldIdx] {
```
jgao54, line 196: does this need a nil check for t.schemaDebug

jgao54: oh nvm, i see that it's checked in the constructor 👍

flow/internal/dynamicconf.go
```diff
@@ -119,6 +119,14 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_NEW_MIRROR,
 		TargetForSetting: protos.DynconfTarget_ALL,
 	},
+	{
+		Name:             "PEERDB_NULLABLE_LAX",
```
jgao54: iiuc this only applies to AVRO, to not be confused with `PEERDB_NULLABLE` above, which propogates nullability to ClickHouse, maybe `PEERDB_AVRO_SCHEMA_NULLABLE_LAX`?

PR #3752, opened 2025-12-01
improve schema validations gaps

jgao54:
- `ParseSchemaTable` should fail if it schema or table is empty string
- Add missing source identifier check Mongo in `ValidateMirrorSource`
- Add missing validation for checking that source collections exist in database
- `ValidateMirrorDestination` should check `DestinationTableIdentifier` is not empty string

Testing:
- added api test to ensure this gets caught
- updated e2e tests to ensure tables are always created so tests pass validations too during cdc flow creation
- ran local smoke tests

### DIFF
--- a/flow/connectors/mongo/validate.go
+++ b/flow/connectors/mongo/validate.go
@@ -2,9 +2,7 @@ package connmongo
 
 import (
 	"context"
-	"fmt"
 
-	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
 	shared_mongo "github.com/PeerDB-io/peerdb/flow/pkg/mongo"
 )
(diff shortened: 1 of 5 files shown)


flow/connectors/mongo/validate.go
```diff
@@ -28,5 +30,12 @@ func (c *MongoConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.F
 		return err
 	}
 
+	for _, tableMapping := range cfg.TableMappings {
```
ilidemi: This should be a listing tables validation in `pkg` to align with the rest of the ClickPipes sources, so we can reuse it in preflight too

flow/pkg/mongo/validation.go
```diff
@@ -107,6 +107,31 @@ func ValidateOplogRetention(ctx context.Context, client *mongo.Client) error {
 	}
 }
 
+func ValidateCollections(ctx context.Context, client *mongo.Client, sourceTableIdentifiers []string) error {
+	databaseCollectionsMapping := make(map[string][]string)
+
+	for _, sourceTableIdentifier := range sourceTableIdentifiers {
+		database, collection, hasDot := strings.Cut(sourceTableIdentifier, ".")
```
ilidemi, line 114: Can move `SchameTable`  and `ParseSchemaTable` from `flow/connectors/utils/identifiers.go` to `pkg` and have `ValidateCollections` take `[]SchemaTable` because in the other codebase we have them as separate strings. `ParseSchemaTable` wouldn't be needed here exactly but would be good to align with #5131 over there (`utils.go`)

jgao54: agree that having a shared data structure to represent this is useful (and avoid code duplication). although it expands the scope of this PR a bit more than i'd like (given how `ParseSchemaTable` and `QuoteIdentifier` are used everywhere in our codebase today).

In addition, if we plan to use a shared data structure to represent this, i'd propose something more generic (vs SchemaTable, which is more postgres-specific), maybe something like:
```
type QualifiedTable struct {
	Namespace string // schema in PG, database in MySQL/Mongo
	Table     string // table or collection
}
```
I can do a bigger lift as a follow-up PR with this goal in mind, and then migrate `ValidateCollections` to use it.

PR #3756, opened 2025-12-02
MySQL: support gh-ost migrations

ilidemi:
[gh-ost](https://github.com/github/gh-ost) is a popular MySQL schema migration tool. It does migrations for a table `A` via creating a new table `A_gho`, syncing everything into it via binlog (just like us), then doing an [atomic rename](https://github.com/github/gh-ost/issues/82) `A to A_old, A_gho to A`. For us it means exactly nothing table identifier-wise as we use names, but we'll miss the column adds that would come out of it.

This change replaces silent drop of unknown columns with constructing a table schema delta from the TABLE_MAP_EVENT that we get with every ROWS_EVENT.

### DIFF
--- a/flow/connectors/mysql/cdc.go
+++ b/flow/connectors/mysql/cdc.go
@@ -26,6 +26,7 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/internal"
 	"github.com/PeerDB-io/peerdb/flow/model"
 	"github.com/PeerDB-io/peerdb/flow/otel_metrics"
+	"github.com/PeerDB-io/peerdb/flow/pkg/common"
 	mysql_validation "github.com/PeerDB-io/peerdb/flow/pkg/mysql"
 	"github.com/PeerDB-io/peerdb/flow/shared"
 	"github.com/PeerDB-io/peerdb/flow/shared/datatypes"
@@ -60,7 +61,7 @@ func (c *MySqlConnector) getTableSchemaForTable(
 	tm *protos.TableMapping,
 	system protos.TypeSystem,
 ) (*protos.TableSchema, error) {
-	schemaTable, err := utils.ParseSchemaTable(tm.SourceTableIdentifier)
+	qualifiedTable, err := common.ParseTableIdentifier(tm.SourceTableIdentifier)
 	if err != nil {
 		return nil, err
 	}
@@ -72,7 +73,7 @@ func (c *MySqlConnector) getTableSchemaForTable(
 
 	rs, err := c.Execute(ctx, fmt.Sprintf(`select column_name, column_type, column_key, is_nullable, numeric_precision, numeric_scale
 		from information_schema.columns where table_schema = '%s' and table_name = '%s' order by ordinal_position`,
-		mysql.Escape(schemaTable.Schema), mysql.Escape(schemaTable.Table)))
+		mysql.Escape(qualifiedTable.Namespace), mysql.Escape(qualifiedTable.Table)))
 	if err != nil {
 		return nil, err
 	}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
           json: ${{ secrets.SNOWFLAKE_GH_CI_PKEY }}
           dir: "nexus/server/tests/assets/"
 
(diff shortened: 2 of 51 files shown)


flow/connectors/mysql/qvalue_convert.go
```diff
@@ -24,9 +24,8 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
 
-func qkindFromMysql(field *mysql.Field) (types.QValueKind, error) {
-	unsigned := (field.Flag & mysql.UNSIGNED_FLAG) != 0
-	switch field.Type {
+func qkindFromMysqlType(mytype byte, unsigned bool, charset uint16) (types.QValueKind, error) {
```
jgao54, line 27: ooc why switched this signature? (was ptr, now require data copy) not a lot of data but is in hot path (i.e. per-column)

flow/connectors/mysql/cdc.go
```diff
@@ -770,3 +772,112 @@ func (c *MySqlConnector) processAlterTableQuery(ctx context.Context, catalogPool
...
+	sourceTableName string,
+	destinationTableName string,
+	schema *protos.TableSchema,
+	exclusion map[string]struct{},
+) ([]*protos.FieldDescription, error) {
+	newFds := make([]*protos.FieldDescription, len(tableMap.ColumnName))
+
+	// Build a set of existing column names for quick lookup
+	existingCols := make(map[string]*protos.FieldDescription, len(schema.Columns))
+	for _, col := range schema.Columns {
```
jgao54, line 793: nit: could we shave off this for-loop because this would run per event and could be expensive for wide tables? 

maybe cache the data with 
`existingColsCache := make(map[string]map[string]*protos.FieldDescription) // pass this into processTableMapEventSchema
`

and then in processTableMapEventSchema:
```
if existingCols, exists := existingColsCache[sourceTableName]; !exists {
    existingCols = make(map[string]*protos.FieldDescription, len(schema.Columns))
    for _, col := range schema.Columns {
        existingCols[col.Name] = col
    }
    existingColsCache[sourceTableName] = existingCols
}
```
And then every time a new col is detected, we can update the cache, when we append to schema.columns

flow/connectors/mysql/qvalue_convert.go
jgao54: lol just realized ptr is 8 byte and this is 4 byte and avoids a lookup

flow/connectors/mysql/cdc.go
jgao54: i guess we already have `TableNameSchemaMapping` for all tables,  so we could just build the initial cache upfront instead of lazily -- either works.

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -662,7 +663,204 @@ func (s ClickHouseSuite) Test_MySQL_Schema_Changes() {
 		[]*protos.TableMapping{{SourceTableIdentifier: dstTableName}})
 	EnvNoError(t, env, err)
 	EnvTrue(t, env, CompareTableSchemas(expectedTableSchema, output[dstTableName]))
-	EnvEqualTablesWithNames(env, s, srcTable, dstTable, "id,c1,coalesce(`addedColumn`,0) `addedColumn`")
+
+	env.Cancel(t.Context())
+	RequireEnvCanceled(t, env)
+}
+
+func (s ClickHouseSuite) Test_MySQL_GhOst_Schema_Changes() {
```
jgao54, line 671: awesome that you have an e2e test that validates peerdb behavior under schema migration without introducing ghost dependency

flow/connectors/mysql/qvalue_convert.go
ilidemi: Function arguments are generally passed in registers, the copying concern really only applies to big structs. The switch is because the QRep path and CDC path have different starting contracts to call it with so have to deconstruct

flow/e2e/clickhouse_mysql_test.go
ilidemi: Claude is a good tester

flow/connectors/mysql/cdc.go
ilidemi: Cache invalidation doesn't sound fun, and also this change brings it from O(N^2) to O(N). Happy to take if it shows up in profiles but would like to get this in before the rollout

PR #3758, opened 2025-12-02
Refactor SchemaTable -> QualifiedTable (and move to pkg)

jgao54:
- `ParseSchemaTable` -> `ParseTableIdentifier`
- `SchemaTable` -> `QualifiedTable`
- move `QualifiedTable`, `ParseTableIdentifier`, and `QuoteIdentifier` to `pkg` so they can be re-used (instead of re-implemented 3 times)
- update `ValidateCollection` to take in `[]*common.QualifiedTable`

### DIFF
--- a/flow/pkg/common/identifiers.go
+++ b/flow/pkg/common/identifiers.go
@@ -33,7 +33,7 @@ func ParseTableIdentifier(tableIdentifier string) (*QualifiedTable, error) {
 //
 //	tblname := "my_table"
 //	data := "my_data"
-//	quoted := pq.QuoteIdentifier(tblname)
+//	quoted := common.QuoteIdentifier(tblname)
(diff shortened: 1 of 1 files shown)


flow/pkg/common/identifiers.go
```diff
@@ -0,0 +1,48 @@
...
+
+	return &QualifiedTable{ns, table}, nil
+}
+
+// QuoteIdentifier quotes an "identifier" (e.g. a table or a column name) to be
+// used as part of an SQL statement.  For example:
+//
+//	tblname := "my_table"
+//	data := "my_data"
+//	quoted := pq.QuoteIdentifier(tblname)
```
ilidemi:
```suggestion
//	quoted := common.QuoteIdentifier(tblname)
```

ilidemi (approved):
Beautiful

PR #3765, opened 2025-12-04
Test mysql 9.5 (gtid) and 8.4 (filepos)

jgao54:
Turns out getting mysql 5.7 to play well with our CI was not a trivial feat.

5.7 requires passing in `--log_bin` to explicitly enable binary logging, but there is no way to pass this into the image as an argument with github-action service configs (`options` are docker args, not application args).  And there's no environment variables to pass it in either. So first downside is we can't run it as a service.

Then tried to do something similar to Maria where we just spin up the container with `docker run` but was blocked by networking issues.  `mysql` no longer works for ssh_keepalive_tests for some reason, but even `localhost` did not work for them. For the other e2e tests, was seeing 'connection refused'  errors due to ipv6 not supported by default on mysql 5,  tried passing in `--bind-address="::"` but still getting the same error so i suspect it has to do with docker network connectivity issues that i am not understanding. 

Fallback to 8.4 for now (with some small refactoring around test setups) until i am motivated to look into this again.

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         runner: [ubuntu-latest-16-cores]
-        db-version: [{pg: 16, mysql: 'mysql-gtid', mongo: '6.0'}, {pg: 17, mysql: 'mysql-pos', mongo: '7.0'}, {pg: 18, mysql: 'maria', mongo: '8.0'}]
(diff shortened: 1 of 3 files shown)


.github/workflows/flow.yml
```diff
@@ -501,15 +501,3 @@ jobs:
         uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-
-  finish:
```
jgao54, line 505: didn't get around to figuring out a good workaround for this github action limitation so just going to remove this for now

jgao54:
Testing one more thing [here](https://github.com/PeerDB-io/peerdb/pull/3767)

PR #3770, opened 2025-12-06
Shadow deploy applySchemaDeltasV2 with validations

jgao54:
This pull request introduces a more robust approach for applying schema deltas to the catalog. It's a follow-up to https://github.com/PeerDB-io/peerdb/pull/3768 (and addresses the race condition described in the PR description), with e2e test coverage, and with the new approach shadow-applied for validation before it's fully enabled, given this PR introduces some non-trivial changes db state that is hard to debug/rollback.

Changes:
- Moved previous `applySchemaDeltas` logic to `applySchemaDeltasV1`, and introduce a separate `applySchemaDeltasV2`. Note that `applySchemaDeltasV2` applies change in-memory, and works in conjunction with a `ReadModifyWriteTableSchemasToCatalog` method to support transaction with ReadModifyWrite patterns.
- Added a feature flag (`PEERDB_APPLY_SCHEMA_DELTA_TO_CATALOG`) and logic to choose between the legacy (v1) and new (v2) approaches for applying schema deltas to the catalog. For now the FF is **disabled** by default, which means the old approach continue to be used as source of truth. 
- The new v2 method is shadow-applied. This is done by adding temporary validation utilities to compare the results of v1 and v2 schema delta approaches, logging discrepancies and reporting metrics for monitoring. 
- Improved end-to-end tests to cover a race condition where columns added without subsequent DML operations could be lost in the destination schema, verifying catalog correctness after each sync.

### DIFF
--- a/flow/e2e/generic_test.go
+++ b/flow/e2e/generic_test.go
@@ -689,6 +689,16 @@ func (s Generic) Test_Schema_Change_Lost_Column_Bug() {
 func (s Generic) Test_Schema_Change_Drop_Consecutive_Columns() {
 	t := s.T()
 
+	_, isPostgres := s.Source().(*PostgresSource)
+	mySource, isMySQL := s.Source().(*MySqlSource)
+	if isMySQL {
+		cmp, err := mySource.CompareServerVersion(t.Context(), mysql_validation.MySQLMinVersionForBinlogRowMetadata)
+		require.NoError(t, err)
+		if cmp < 0 {
+			t.Skip("skipping test because drop column is not supported")
+		}
+	}
+
 	srcTable := "test_ddl_drop_column"
 	dstTable := "test_ddl_drop_column_dst"
 	srcTableName := AttachSchema(s, srcTable)
@@ -718,9 +728,6 @@ func (s Generic) Test_Schema_Change_Drop_Consecutive_Columns() {
 		`ALTER TABLE %s DROP COLUMN col_to_drop_second`, srcTableName)))
 
 	EnvWaitForEqualTablesWithNames(env, s, "wait for first row synced", srcTable, dstTable, "id")
-
-	_, isPostgres := s.Source().(*PostgresSource)
-	_, isMySQL := s.Source().(*MySqlSource)
 
 	// for integer, postgres reports typmod -1 while mysql reports precision-based typmod (precision=10, scale=0)
 	var integerTypmod int32


jgao54:
CI Failure is unrelated to this PR: https://github.com/PeerDB-io/peerdb/pull/3784 , but has to do with the latest ClickHouse version. looking into that separately.

flow/internal/postgres.go
```diff
@@ -74,6 +74,74 @@ func LoadTableSchemaFromCatalog(
...
+		return fmt.Errorf("failed to close batch results: %w", err)
+	}
+
+	if err := tx.Commit(ctx); err != nil {
+		return fmt.Errorf("failed to commit transaction: %w", err)
+	}
+	return nil
+}
+
 func UpdateTableOIDsInTableSchemaInCatalog(
```
jgao54, line 156: note: we can apply this function to use ReadModifyWriteTableSchemasToCatalog as well later.

flow/activities/flowable_core.go
```diff
@@ -93,6 +151,57 @@ func (a *FlowableActivity) applySchemaDeltas(
...
+	schemaDeltas []*protos.TableSchemaDelta,
+) (map[string]*protos.TableSchema, error) {
+	logger := internal.LoggerFromCtx(ctx)
+
+	schemasCopy := make(map[string]*protos.TableSchema, len(schemasInCatalog))
+	for tableName, schema := range schemasInCatalog {
+		if schema == nil {
+			return nil, fmt.Errorf("failed to deep copy table schema from catalog: table %s has nil schema", tableName)
+		}
+		schemasCopy[tableName] = proto.Clone(schema).(*protos.TableSchema)
```
serprex: use `proto.CloneOf` which is generic so won't need cast

flow/activities/flowable_core.go
```diff
@@ -93,6 +151,57 @@ func (a *FlowableActivity) applySchemaDeltas(
...
+	for _, schemaDelta := range schemaDeltas {
+		ddlTableNames = append(ddlTableNames, schemaDelta.DstTableName)
+	}
+
+	if len(schemasCopy) != len(ddlTableNames) {
+		matchedTables := make([]string, 0, len(schemasCopy))
+		for name := range schemasCopy {
+			matchedTables = append(matchedTables, name)
+		}
+		logger.Warn(fmt.Sprintf("not all tables are found, expected: %v, actual, %v", ddlTableNames, matchedTables))
```
serprex:
```suggestion
		logger.Warn(fmt.Sprintf("not all tables are found, expected: %v, actual: %v", ddlTableNames, matchedTables))
```
but really use slog.Any

flow/activities/flowable_core.go
```diff
@@ -93,6 +151,57 @@ func (a *FlowableActivity) applySchemaDeltas(
...
+	}
+
+	for _, schemaDelta := range schemaDeltas {
+		schema := schemasCopy[schemaDelta.DstTableName]
+		columnNames := make(map[string]bool, len(schema.GetColumns()))
```
serprex:
```suggestion
		columnNames := make(map[string]struct{}, len(schema.GetColumns()))
```

flow/activities/flowable_core.go
```diff
@@ -67,6 +68,64 @@ func (a *FlowableActivity) applySchemaDeltas(
 	config *protos.FlowConnectionConfigsCore,
 	options *protos.SyncFlowOptions,
 	schemaDeltas []*protos.TableSchemaDelta,
+) error {
+	logger := internal.LoggerFromCtx(ctx)
+
+	ddlTableNames := make([]string, 0, len(schemaDeltas))
```
Amogh-Bharadwaj: Maybe `dstTableNamesInDeltas` ?

flow/activities/flowable_core.go
```diff
@@ -93,6 +152,59 @@ func (a *FlowableActivity) applySchemaDeltas(
...
+
+	schemasCopy := make(map[string]*protos.TableSchema, len(schemasInCatalog))
+	for tableName, schema := range schemasInCatalog {
+		if schema == nil {
+			return nil, fmt.Errorf("failed to deep copy table schema from catalog: table %s has nil schema", tableName)
+		}
+		schemasCopy[tableName] = proto.CloneOf(schema)
+	}
+
+	ddlTableNames := make([]string, 0, len(schemaDeltas))
```
Amogh-Bharadwaj: Maybe `dstTableNamesInDeltas` ?

flow/activities/flowable_core.go
```diff
@@ -93,6 +152,59 @@ func (a *FlowableActivity) applySchemaDeltas(
...
+// schemaDeltas that are not in the existing mapping. This function returns
+// a copy of schemasInCatalog with schemaDeltas applied.
+func applySchemaDeltaV2(
+	ctx context.Context,
+	schemasInCatalog map[string]*protos.TableSchema,
+	schemaDeltas []*protos.TableSchemaDelta,
+) (map[string]*protos.TableSchema, error) {
+	logger := internal.LoggerFromCtx(ctx)
+
+	schemasCopy := make(map[string]*protos.TableSchema, len(schemasInCatalog))
```
Amogh-Bharadwaj:
```suggestion
	schemasInCatalogCopy := make(map[string]*protos.TableSchema, len(schemasInCatalog))
```

flow/activities/flowable_core.go
```diff
@@ -93,6 +152,59 @@ func (a *FlowableActivity) applySchemaDeltas(
...
+		}
+		schemasCopy[tableName] = proto.CloneOf(schema)
+	}
+
+	ddlTableNames := make([]string, 0, len(schemaDeltas))
+	for _, schemaDelta := range schemaDeltas {
+		ddlTableNames = append(ddlTableNames, schemaDelta.DstTableName)
+	}
+
+	if len(schemasCopy) != len(ddlTableNames) {
```
Amogh-Bharadwaj: Here you're comparing the lengths of:
1. Number of table schemas in catalog for the mirror throughout its lifetime
2. Number of unique destination tables in schema changes in the current sync batch

and warning if the two aren't equal. Why should they be equal?

Amogh-Bharadwaj: Did you mean to warn for a non-empty list of tables which are in the schema deltas but not in catalog ?

flow/activities/flowable_core.go
```diff
@@ -93,6 +152,59 @@ func (a *FlowableActivity) applySchemaDeltas(
...
+	}
+
+	for _, schemaDelta := range schemaDeltas {
+		schema := schemasCopy[schemaDelta.DstTableName]
+		columnNames := make(map[string]struct{}, len(schema.GetColumns()))
+		for _, col := range schema.GetColumns() {
+			columnNames[col.Name] = struct{}{}
+		}
+		for _, newCol := range schemaDelta.GetAddedColumns() {
+			// only add columns that don't already exists
```
Amogh-Bharadwaj:
```suggestion
			// only add columns that don't already exist
```

flow/activities/flowable_core.go
```diff
@@ -93,6 +152,59 @@ func (a *FlowableActivity) applySchemaDeltas(
...
+	for _, schemaDelta := range schemaDeltas {
+		ddlTableNames = append(ddlTableNames, schemaDelta.DstTableName)
+	}
+
+	if len(schemasCopy) != len(ddlTableNames) {
+		matchedTables := make([]string, 0, len(schemasCopy))
+		for name := range schemasCopy {
+			matchedTables = append(matchedTables, name)
+		}
+		logger.Warn("not all tables are found",
```
Amogh-Bharadwaj:
```suggestion
		logger.Warn("not all tables in deltas are found in catalog",
```

flow/e2e/generic_test.go
```diff
@@ -568,6 +570,122 @@ func (s Generic) Test_Schema_Changes_Cutoff_Bug() {
...
+// The scenario:
+//  1. CDC mirror is running with a small batch size
+//  2. ALTER TABLE adds good_column + INSERT (relation message sent for good_column)
+//  3. ALTER TABLE adds lost_column (NO subsequent DML, so no relation message yet)
+//  4. PeerDB syncs: adds good_column to destination, then applySchemaDelta updates catalog
+//     with latest source db schema (which includes lost_column)
+//  5. Next INSERT triggers relation message, adds lost_column to schemaDeltas
+//  6. PeerDB compares against catalog (which has lost_column) -> no delta detected
+//  7. Error: lost_column doesn't exist on destination but we try to insert data for it
+func (s Generic) Test_Schema_Change_Lost_Column_Bug() {
```
Amogh-Bharadwaj, line 584: Nice! Curious if this test does indeed fail with v1 (to validate our RCA)

jgao54: @Amogh-Bharadwaj  yep, was able to confirm that it does fail :)

flow/activities/flowable_core.go
jgao54: ah, i introduced this prior to the readModifyWrite method, so this warning should probably go in there instead to be more obvious. 

`LoadTableSchemasFromCatalog` takes in a list of tableNames, and returns schema for only the tables specified in the input, since the query is `"SELECT table_name, table_schema FROM table_schema_mapping WHERE flow_name = $1 AND table_name = ANY($2)",`

so, adding a warning if a specified table from schemaDelta is not in `table_schema_mapping`. this shouldn't happen because the logic to process relation messages filters irrelevant DDLs out, add a warning to help with debugging if that's not the case.

will move this logic to `ReadModifyWriteTableSchemasToCatalog`

Amogh-Bharadwaj: Ah that makes a lot of sense, my bad 🤦

flow/e2e/generic_test.go
```diff
@@ -686,6 +686,111 @@ func (s Generic) Test_Schema_Change_Lost_Column_Bug() {
...
+	EnvTrue(t, env, RequireEqualTableSchemas(t, expectedCatalogSchema, catalogSchema))
+
+	// Add a column to validate previously drop columns are handled correctly after schema deltas are applied
+	EnvNoError(t, env, s.Source().Exec(t.Context(), fmt.Sprintf(
+		`ALTER TABLE %s ADD COLUMN col_to_add text`, srcTableName)))
+	EnvNoError(t, env, s.Source().Exec(t.Context(), fmt.Sprintf(
+		`INSERT INTO %s (id, col_to_add) VALUES (2, 'add2')`, srcTableName)))
+	EnvWaitForEqualTablesWithNames(env, s, "wait for second row synced", srcTable, dstTable, "id, col_to_add")
+
+	expectedCatalogSchema.Columns = append(expectedCatalogSchema.Columns, &protos.FieldDescription{
```
jgao54, line 784: note that if PEERDB_APPLY_SCHEMA_DELTA_TO_CATALOG is set to false, `expectedCatalogSchema` after ADD COLUMN here would _not_ contain `col_to_drop_first` and `col_to_drop_second` columns (because of schema fetched from source)

PR #3778, opened 2025-12-09
feat(bigquery): verbose into flow info on GCS snapshot

jkaflik:
The idea is to give an insight on ongoing snapshot GCS export. For large datasets, it can take few minutes.

### DIFF
--- a/flow/connectors/bigquery/qrep_object_pull.go
+++ b/flow/connectors/bigquery/qrep_object_pull.go
@@ -341,7 +341,11 @@ func (c *BigQueryConnector) ExportTxSnapshot(
 			var apiErr *googleapi.Error
 			if errors.As(err, &apiErr) {
 				if apiErr.Code == 403 {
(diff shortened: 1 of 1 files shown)


flow/connectors/bigquery/qrep_object_pull.go
```diff
@@ -339,14 +339,16 @@ func (c *BigQueryConnector) ExportTxSnapshot(
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to start export job for table %s: %w", tm.SourceTableIdentifier, err)
 		}
-		jobs = append(jobs, job)
+		jobs[tm.SourceTableIdentifier] = job
```
heavycrystal, line 354: Is there any problem with running too many extract jobs at once? Should it be tied to parallelism

jkaflik: In theory it could be a problem, but in reality at the moment I don't see exporting dozens could be problematic in terms of parallelism. Only limitation is exceeding quota, which is a problem regardless of parallelism.

PR #3782, opened 2025-12-10
BigQuery performance improvements and bugfixes

jkaflik:
Two major improvements:
- Use Parquet instead of Avro for better performance on the ClickHouse cluster
  - In order to support JSON column type export, the `EXPORT DATA` query statement is built directly, instead of the SDK API. 
  - For Geo types we have to cast into string first
- Achieve higher batching with [a globs](https://clickhouse.com/docs/sql-reference/table-functions/url#globs-in-url) (multiple files in a single table function call)
- `totalRows` are not return in BigQuery source connector as it's non-accurrate informaton. Synced total rows should be taken into account.

Plus several bugfixes, mostly for handling data types like Array of Record.

### DIFF
--- a/flow/connectors/bigquery/utils.go
+++ b/flow/connectors/bigquery/utils.go
@@ -7,7 +7,7 @@ import (
 func quotedIdentifier(value string) string {
 	var result strings.Builder
 	result.WriteByte('`')
-	for i := 0; i < len(value); i++ {
+	for i := range len(value) {
 		ch := value[i]
 		if ch == '`' {
 			result.WriteString("\\`")


flow/connectors/clickhouse/object_sync_test.go
```diff
@@ -0,0 +1,224 @@
...
+		stream := model.NewQObjectStream(10)
+
+		batchCh := collectAndBatchObjects(ctx, stream, 10000, longInterval)
+
+		stream.Objects <- &model.Object{URL: "http://example.com/1.avro", Size: 100}
+
+		cancel()
+
+		select {
+		case _, _ = <-batchCh: //nolint:staticcheck
```
ilidemi:
```suggestion
		case <-batchCh:
```

flow/connectors/bigquery/avro_test.go
```diff
@@ -1,18 +0,0 @@
...
-
-import (
-	"os"
-	"testing"
-
-	"github.com/stretchr/testify/require"
-)
-
-func TestAvroTotalRows(t *testing.T) {
-	f, err := os.Open("../../e2e/test_data/1k.avro")
```
ilidemi, line 11: Does 1k.avro need to be removed?

flow/connectors/bigquery/data_format_utils.go
```diff
@@ -0,0 +1,163 @@
...
+const (
+	// minParquetFooterSize is the minimum footer size: 4 bytes for footer length + 4 bytes for magic
+	minParquetFooterSize = 8
+)
+
+// gcsObjectSeeker implements io.ReadSeeker for a GCS object using range reads.
+// Designed for a quick reading specific parts of a Parquet file. Like file header.
+// Not efficient for many small reads/seeks.
+type gcsObjectSeeker struct {
+	ctx context.Context //nolint:containedctx // context for GCS operations
```
ilidemi, line 21:
```suggestion
	ctx context.Context //nolint:containedctx // to fit the io.ReadSeeker signatures
```

flow/connectors/bigquery/data_format_utils.go
```diff
@@ -0,0 +1,163 @@
...
+
+	object   *storage.ObjectHandle
+	fileSize int64
+	offset   int64
+}
+
+func (r *gcsObjectSeeker) Read(p []byte) (n int, err error) { //nolint:nonamedreturns // standard Read signature
```
ilidemi:
```suggestion
func (r *gcsObjectSeeker) Read(p []byte) (int, error) {
```

flow/connectors/bigquery/data_format_utils.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	r := &gcsObjectSeeker{
+		ctx:      ctx,
+		object:   object,
+		fileSize: attrs.Size,
+		offset:   0,
+	}
+
+	n, err := readParquetTotalRows(r, attrs.Size)
```
ilidemi: Can we use arrow-go/v18/parquet/file.NewParquetReader().NumRows() here? I think it does the same, and even allows to get away with just supporting SeekEnd and not tracking offsets if we want to go extra short

flow/connectors/bigquery/data_format_utils.go
```diff
@@ -0,0 +1,163 @@
...
+		return 0, fmt.Errorf("failed to parse parquet file metadata: %w", err)
+	}
+
+	return fileMetaData.FileMetaData.NumRows, nil
+}
+
+func parquetObjectAverageRowSize(ctx context.Context, logger log.Logger, object *storage.ObjectHandle) uint64 {
+	attrs, err := object.Attrs(ctx)
+	if err != nil {
+		logger.Error("failed to get object attrs for parquet object, using default compressed row size",
```
ilidemi, line 102: Log the size used in these too, in case we change it and there are different bits across services

flow/connectors/bigquery/qrep_object_pull.go
```diff
@@ -366,6 +368,58 @@ func (c *BigQueryConnector) ExportTxSnapshot(
...
+	exportSQL := fmt.Sprintf(`EXPORT DATA OPTIONS(
+			uri='%s',
+			format='PARQUET',
+			compression='GZIP',
+			overwrite=true
+		) AS
+		SELECT %s FROM %s`,
+		uri,
+		strings.Join(columnSelects, ", "),
+		sourceTableIdentifier,
```
ilidemi: Needs to be quoted, possibly along with introducing a QuoteIdentifier helper and aligning a few usages

flow/connectors/clickhouse/object_sync_test.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func TestCollectAndBatchObjects(t *testing.T) {
+	ctx := context.Background()
+	longInterval := time.Hour
+
+	t.Run("single object", func(t *testing.T) {
+		stream := model.NewQObjectStream(1)
+		go func() {
+			stream.Objects <- &model.Object{URL: "http://example.com/1.avro", Size: 100}
```
ilidemi: `.parquet`?

flow/connectors/clickhouse/object_sync.go
```diff
@@ -31,64 +112,74 @@ func (c *ClickHouseConnector) SyncQRepObjects(
...
+		slog.String("format", string(format)))
 
 	schema, err := stream.Schema()
 	if err != nil {
 		return 0, nil, fmt.Errorf("failed to get schema from stream: %w", err)
 	}
 
+	// For ClickHouse records sync we have Avro file size limits.
+	// We reuse the same value as a good heuristic for batching inserts.
+	maxBatchSize, err := internal.PeerDBS3BytesPerAvroFile(ctx, config.Env)
```
ilidemi, line 128: Warning, if parquets are compressed and the data is repetitive (e.g. jsons), batching by compressed size may still OOM ClickHouse. With Mongo we've seen 0.5-1% compression ratios and had to [switch](https://github.com/PeerDB-io/peerdb/pull/3715) to tracking uncompressed size the best we can. Ok to see first if that actually happens but just FYI

flow/connectors/bigquery/qrep_object_pull.go
ilidemi, line 77: The header situation is a bit mind-bending. They're queued along with each object, but are actually a reference to the same map, and then object sync assumes they're the same within batch? Wonder if there's a way to have one object (per snapshot) that everyone calls and it's always up to date

flow/connectors/clickhouse/object_sync.go
```diff
@@ -31,64 +112,74 @@ func (c *ClickHouseConnector) SyncQRepObjects(
...
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to get batch size config: %w", err)
+	}
+
+	batches := collectAndBatchObjects(ctx, stream, maxBatchSize, defaultBatchFlushInterval)
```
ilidemi, line 133: Would need to give object_sync a more thorough pass, but is there a way to inline collectAndBatchObjects and have one for loop that sometimes cuts the batch and sometimes continues? Channels into channels are harder to wrap one's head around

flow/connectors/bigquery/data_format_utils.go
jkaflik: could you elaborate?

flow/connectors/bigquery/data_format_utils.go
jkaflik: It's something I missed when investigating parquet reader... not sure why.
Indeed, it's much simpler to just re-use existing code. Changing.

flow/connectors/bigquery/qrep_object_pull.go
jkaflik: Like a callback function to retrieve headers? I think it could make sense.

The idea behind this to be unique per object was to keep `object sync` mechanism generic. 
I would leave it as a potentical refactor item.

flow/connectors/bigquery/data_format_utils.go
ilidemi: To have the comment explain "why" instead of "what". On any random struct the linter would be correct because the context should be passed to functions, but in this case the functions have to conform to a standard interface without context, so it's justified. A comment prevents the reader from having to figure that out by rereading code

flow/connectors/bigquery/utils.go
```diff
@@ -1 +1,16 @@
 package connbigquery
+
+import (
+	"fmt"
+	"strings"
+)
+
+func quoteIdentifier(sourceTableIdentifier string) string {
+	parts := strings.Split(sourceTableIdentifier, ".")
```
ilidemi: QuoteIdentifier in CH and Postgres utils is assuming a simple db/table/column identifier, without splitting on periods. Can we follow the suit here and pass `dsTable.dataset` and `dsTable.table` separately?

flow/connectors/bigquery/utils.go
```diff
@@ -1 +1,16 @@
...
+	for i, part := range parts {
+		// Escape backticks by doubling them
+		escapedPart := strings.ReplaceAll(part, "`", "``")
```
ilidemi: BQ escaping is backslashes https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/lexical#quoted_identifiers

flow/connectors/bigquery/qrep_object_pull.go
ilidemi: We only have one implementation so it's ok to change expectations to fit BigQuery. If that changes for another object source, we can revisit

good bot claude https://github.com/PeerDB-io/peerdb/compare/customer-kuba-bq-improvements...header-provider?expand=1

flow/connectors/clickhouse/object_sync.go
ilidemi: Experimented with this approach and the one currently implemented is actually more testable. Go gets convoluted sometimes

flow/connectors/clickhouse/object_sync.go
```diff
@@ -162,18 +255,34 @@ func (c *ClickHouseConnector) insertFromURLWithMapping(
...
+		expr.WriteString(peerdb_clickhouse.QuoteLiteral(batch.objects[0].URL))
+	} else {
+		// Multiple URLs: use brace expansion {url1,url2,url3}
+		var urlList strings.Builder
+		urlList.WriteString("{")
+		for i, obj := range batch.objects {
+			if i > 0 {
+				urlList.WriteString(",")
+			}
+			urlList.WriteString(obj.URL)
```
ilidemi, line 276: Add a defensive check that URL doesn't contain `,{}`? Based on the current implementation it should be ok (bucket name restrictions + url.PathEscape) but that needed tracking down

ilidemi (commented):
Looking good, only a few comments

flow/connectors/bigquery/qrep_object_pull.go
jkaflik: Let's follow up in seperate PR if we want to.

jkaflik: https://github.com/PeerDB-io/peerdb/issues/3828

flow/connectors/bigquery/data_format_utils.go
jkaflik: My comment explains on "why".

PR #3786, opened 2025-12-12
support multiple ch versions in ci

jgao54:
Test against multiple ClickHouse versions to improve robustness, now that we are pinning CH versions.


jgao54:
updated to use more abstract names to avoid having to update 'required github checks' every time we bump version.

and bringing latest back as one of the matrix. reason: we've caught a few bugs in ClickHouse with using the latest release.  This might be a good way for us to raise issue early. And having different versions will make it more obvious when the failure is version-specific.

PR #3793, opened 2025-12-16
support specifying x509 auth mechanism for mongodb

jgao54:
support MONGODB-X509 (TLS) authentication. Only checks to make sure username/password are not specified. previous check `HasAuthParameters` will block if authMechanism is specified in the connection string, but that should be allowed.

### DIFF
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -8,7 +8,6 @@ import (
 	"log/slog"
 	"net"
 	"os"
-	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


serprex (dismissed):

flow/connectors/mongo/mongo.go
```diff
@@ -126,7 +127,9 @@ func parseAsClientOptions(config *protos.MongoConfig, meteredDialer utils.Metere
 		return nil, fmt.Errorf("error parsing uri: %w", err)
 	}
 
-	if connStr.HasAuthParameters() {
+	if connStr.Username != "" || connStr.Password != "" ||
+		// handle the edge case where '@' or ':@' in uri but no username/password provided
+		strings.Contains(connStr.Original[len(connStr.Scheme+"://"):len(connStr.Scheme+"://")+2], "@") {
```
```go
if connStr.UsernameSet {
```

jgao54: ah, missed that. great!

PR #3797, opened 2025-12-18
mysql and clickhouse classification

jgao54:
MySQL classifications (plus a missing `io.ErrUnexpectedEOF` in clickhouse.

The mysql ones are based on call stack , i.e.:
```
"stack":"writeAuthHandshake: tls: first record does not look like a TLS handshake
github.com/pingcap/errors.Trace
	/go/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20250523034308-74f78ae071ee/juju_adaptor.go:15
github.com/go-mysql-org/go-mysql/client.(*Conn).handshake
	/go/pkg/mod/github.com/go-mysql-org/go-mysql@v1.13.0/client/conn.go:204
github.com/go-mysql-org/go-mysql/client.ConnectWithDialer
	/go/pkg/mod/github.com/go-mysql-org/go-mysql@v1.13.0/client/conn.go:163
github.com/go-mysql-org/go-mysql/replication.(*BinlogSyncer).newConnection
	/go/pkg/mod/github.com/go-mysql-org/go-mysql@v1.13.0/replication/binlogsyncer.go:1004
github.com/go-mysql-org/go-mysql/replication.(*BinlogSyncer).registerSlave
	/go/pkg/mod/github.com/go-mysql-org/go-mysql@v1.13.0/replication/binlogsyncer.go:280
github.com/go-mysql-org/go-mysql/replication.(*BinlogSyncer).prepare
	/go/pkg/mod/github.com/go-mysql-org/go-mysql@v1.13.0/replication/binlogsyncer.go:407
github.com/go-mysql-org/go-mysql/replication.(*BinlogSyncer).StartSyncGTID
	/go/pkg/mod/github.com/go-mysql-org/go-mysql@v1.13.0/replication/binlogsyncer.go:484
github.com/PeerDB-io/peerdb/flow/connectors/mysql.(*MySqlConnector).startCdcStreamingGtid
	/root/flow/connectors/mysql/cdc.go:288
github.com/PeerDB-io/peerdb/flow/connectors/mysql.(*MySqlConnector).startStreaming
	/root/flow/connectors/mysql/cdc.go:261
github.com/PeerDB-io/peerdb/flow/connectors/mysql.(*MySqlConnector).PullRecords
```
So far we've only see this error with gtid-based streaming, but adding an error handling around filepos-based streaming just in case.

### DIFF
--- a/flow/shared/exceptions/mysql.go
+++ b/flow/shared/exceptions/mysql.go
@@ -2,9 +2,9 @@ package exceptions
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
-	"strings"
 )
 
 type MySQLIncompatibleColumnTypeError struct {
@@ -46,30 +46,30 @@ func (e *MySQLUnsupportedBinlogRowMetadataError) Error() string {
 		e.SchemaName, e.TableName)
 }
 
-type MySQLStreamingTransientError struct {
+type MySQLStreamingError struct {
 	error
+	Retryable bool
 }
 
-func (e *MySQLStreamingTransientError) Error() string {
-	return "MySQL streaming transient error: " + e.error.Error()
-}
-
-func (e *MySQLStreamingTransientError) Unwrap() error {
-	return e.error
-}
-
-func AsMySQLStreamingTransientError(err error) *MySQLStreamingTransientError {
-	if err == nil {
-		return nil
-	}
-
+func NewMySQLStreamingError(err error) *MySQLStreamingError {
 	if errors.Is(err, context.DeadlineExceeded) {
-		return &MySQLStreamingTransientError{err}
+		return &MySQLStreamingError{err, true}
 	}
 
-	if strings.Contains(err.Error(), "first record does not look like a TLS handshake") {
-		return &MySQLStreamingTransientError{err}
+	var recordHeaderError tls.RecordHeaderError
+	if errors.As(err, &recordHeaderError) {
+		if recordHeaderError.Msg == "first record does not look like a TLS handshake" {
+			return &MySQLStreamingError{err, true}
+		}
 	}
 
-	return nil
+	return &MySQLStreamingError{err, false}
+}
+
+func (e *MySQLStreamingError) Error() string {
+	return "MySQL streaming error: " + e.error.Error()
+}
+
+func (e *MySQLStreamingError) Unwrap() error {
+	return e.error
 }
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -909,11 +909,18 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
-	var mysqlStreamingTransientError *exceptions.MySQLStreamingTransientError
-	if errors.As(err, &mysqlStreamingTransientError) {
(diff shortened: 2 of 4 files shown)


flow/shared/exceptions/mysql.go
```diff
@@ -42,3 +43,30 @@ func (e *MySQLUnsupportedBinlogRowMetadataError) Error() string {
...
+
+func AsMySQLStreamingTransientError(err error) *MySQLStreamingTransientError {
+	if err == nil {
+		return nil
+	}
+	msg := err.Error()
+	switch {
+	case strings.Contains(msg, "first record does not look like a TLS handshake"):
+		return &MySQLStreamingTransientError{error: err}
+	case strings.Contains(msg, "context deadline exceeded"):
```
serprex: can't use errors.Is for this?

flow/shared/exceptions/mysql.go
```diff
@@ -42,3 +43,30 @@ func (e *MySQLUnsupportedBinlogRowMetadataError) Error() string {
...
+	return e.error
+}
+
+func AsMySQLStreamingTransientError(err error) *MySQLStreamingTransientError {
+	if err == nil {
+		return nil
+	}
+	msg := err.Error()
+	switch {
+	case strings.Contains(msg, "first record does not look like a TLS handshake"):
```
serprex: https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/crypto/tls/conn.go;l=670

should be able to error.Is with https://pkg.go.dev/crypto/tls#RecordHeaderError

jgao54: The error.Is interface doesn't work super well with RecordHeaderError:

```
func Is(err, target error) bool {
	if err == nil || target == nil {
		return err == target
	}

	isComparable := reflectlite.TypeOf(target).Comparable()
	return is(err, target, isComparable)
}

func is(err, target error, targetComparable bool) bool {
	for {
		if targetComparable && err == target {
			return true
		}
		if x, ok := err.(interface{ Is(error) bool }); ok && x.Is(target) {
			return true
		}
		switch x := err.(type) {
		case interface{ Unwrap() error }:
			err = x.Unwrap()
			if err == nil {
				return false
			}
		case interface{ Unwrap() []error }:
			for _, err := range x.Unwrap() {
				if is(err, target, targetComparable) {
					return true
				}
			}
			return false
		default:
			return false
		}
	}
}

```
Because [RecordHeaderError](https://pkg.go.dev/crypto/tls#RecordHeaderError)  has to be exactly equal to the target error to return true. The struct consists of Msg/RecordHeader/Conn fields and those all have to match the error, which is not easily determined. So seems a bit fragile for this use case.  (it also does not implement unwrap, but i don't think this error is wrapped so that's less of an issue)

flow/shared/exceptions/mysql.go
jgao54: this can, updated

flow/shared/exceptions/mysql.go
serprex: My bad, you can use `errors.As` then

jgao54: ah yeah, that'd work.  edited class to define the error as retry-able vs not.

PR #3799, opened 2025-12-19
classification: pg replica identity index dropped error

jgao54:
This came up a few times. Can happen when user drops the index used for replica identity.  Should notify.


jgao54:
given these are the only two case now, will keep them as is, as they have different error message and it would be good to differentiate them. If more replica identity issues get introduced, we can use a generic message with an additional replica identity type attribute.

PR #3820, opened 2026-01-05
Reduce CI MySQL warnings

ilidemi:
Quote cleanup query properly - PG-style quoting produces 8812 warnings per run
```
"msg":"failed to parse QueryEvent",
"query":"DROP DATABASE IF EXISTS \"e2e_test_mychclg_nq7h7zy5\"",
"error":"line 1 column 51 near \"\"e2e_test_mychclg_nq7h7zy5\"\"
```


ilidemi:
Still about 500 our of 50000 lines per log left but better than 8800

PR #3833, opened 2026-01-10
add mongodb replication lag metric

jgao54:
Add a time-based replication lag metric for MongoDB. The metric is calculated by (latest oplog commit time  - latest offset cluster time). Offset cluster time is the cluster time extract from resume token which we store as offset.

Similar to replication slot size, this not a representation of end-to-end latency, since the cluster time represents when the data is processed, not when it's fully propagated to ClickHouse. The latter is is out-of-scope for this PR. The goal of this PR is to make the metric available in otel to set up observability/alerting, but does not touch persistence/UI, which will be worked on as follow-ups.

Note: evaluated tracking this metric in `recordSlotInformation`, however that piece is very specific to Postgres replication slot so it seems a bit convoluted to introduce it there.

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -1255,7 +1255,7 @@ func (a *FlowableActivity) RecordSlotSizes(ctx context.Context) error {
 			logger.Error("Failed to emit log retention hours", slog.Any("error", err))
 		}
 		if err := a.recordServerSideCommitLag(timeoutCtx, info, a.OtelManager.Metrics.ServerSideCommitLagGauge); err != nil {
-			logger.Error("Failed to record source-side commit lag", slog.Any("error", err))
+			logger.Error("Failed to record server-side commit lag", slog.Any("error", err))
 		}
 		cancel()
 	}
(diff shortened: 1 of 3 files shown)


flow/activities/flowable.go
```diff
@@ -1351,6 +1354,47 @@ func (a *FlowableActivity) emitLogRetentionHours(
...
+
+	flowName := info.config.FlowJobName
+	lagMicroseconds, err := srcConn.GetCommitLagMicroseconds(ctx, flowName)
+	if err != nil {
+		logger.Error("Failed to get replication time lag", slog.Any("error", err))
+		return err
+	}
+
+	if lagMicroseconds >= 0 {
+		commitLagGauge.Record(ctx, lagMicroseconds)
```
ilidemi: Negative is ok, gives us data

ilidemi (commented):
Why not both ways? Can introduce `server_side_commit_lag` gauge and do GetServerSideCommitLagMicroseconds() in the connector to be clear. The more data, the merrier

PR #3835, opened 2026-01-12
Introduce endpoint to reset LSN

Amogh-Bharadwaj:
To make step 7 in [PeerDB Postgres Upgrade process](https://docs.peerdb.io/bestpractices/postgres-upgrade#upgrading-the-postgresql-peer-of-an-ongoing-cdc-mirror) self-served

### DIFF
--- a/flow/cmd/handler.go
+++ b/flow/cmd/handler.go
@@ -870,9 +870,23 @@ func (h *FlowRequestHandler) ChangeCatalogLastOffsetForMirror(
 		lastOffset = *req.LastOffset
 	}
 
+	// Select existing job_name values to ensure flow exists
+	var existingJobName string
+	err := h.pool.QueryRow(
+		ctx,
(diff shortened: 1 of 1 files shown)


protos/route.proto
```diff
@@ -522,6 +522,18 @@ message CancelTableAdditionOutput {
   string run_id = 3;
 }
 
+message ChangeCatalogLastOffsetForMirrorRequest {
+  string flow_name = 1;
+  optional int64 last_offset = 2;
```
ilidemi, line 527: Would need to either introduce text for MySQL (mainly FilePos) or stick to just Postgres (might as well have the endpoint be reset to 0 in that case, dead simple)

ilidemi: Update: there are MySQL customers who need GTID modifications (on replica rebuild) so making it a generic change would be good

PR #3836, opened 2026-01-12
feat(metrics): propagate source hostname in downstream metrics

iamKunalGupta:


flow/activities/flowable.go
```diff
@@ -1763,6 +1776,18 @@ func (a *FlowableActivity) RemoveFlowDetailsFromCatalog(
 	return nil
 }
 
+func getPeerHostName(peerType protos.DBType, peer *protos.Peer) string {
+	switch peerType {
+	case protos.DBType_POSTGRES:
+		return shared.Val(peer.GetPostgresConfig()).Host
+	case protos.DBType_MYSQL:
+		return shared.Val(peer.GetMongoConfig()).TlsHost
```
ilidemi, line 1786:
```suggestion
	case protos.DBType_MYSQL:
		return shared.Val(peer.GetMysqlConfig()).Host
	case protos.DBType_MONGO:
		return shared.Val(peer.GetMongoConfig()).TlsHost
```

PR #3842, opened 2026-01-13
o11y: log user actions

jgao54:
Logging that provides an overview into user actions so we can quickly correlate user actions with pipe behavior.

An alternative I evaluated was embedding this logic into the requestLoggingMiddleware. Decided against it as it is a bit more flexible with the current implementation where we don't have a robust state machine. 

Testing: ran operations locally and checked for logs

### DIFF
--- a/flow/e2e/clickhouse_mysql_test.go
+++ b/flow/e2e/clickhouse_mysql_test.go
@@ -557,7 +557,7 @@ func (s ClickHouseSuite) Test_MySQL_Specific_Geometric_Types() {
 		require.Len(s.t, row, 8, "expected 8 columns")
 		for j := 1; j < 8; j++ {
 			geometryVal := row[j].Value()
-
+			//nolint:gosec
 			require.Equal(s.t, expectedValues[i][j-1], geometryVal, "geometry value mismatch at row %d column %d", i+1, j)
 		}
 	}


flow/cmd/user_actions.go
```diff
@@ -0,0 +1,86 @@
...
+func LogActionPauseFlow(ctx context.Context, flowName string) {
+	logUserAction(ctx, UserActionPauseFlow, flowName, "")
+}
+
+func LogActionResumeFlow(ctx context.Context, flowName string, tablesAdded, tablesRemoved []string) {
+	var extraAttrs []any
+	var logDetails []string
+	if len(tablesAdded) > 0 {
+		extraAttrs = append(extraAttrs, slog.Any("tablesAdded", tablesAdded))
+		logDetails = append(logDetails, "tables added")
```
ilidemi: Add counts to logDetails?

flow/cmd/user_actions.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+	"log/slog"
+	"strings"
+
+	"github.com/PeerDB-io/peerdb/flow/shared"
+)
+
+type UserActionType int
+
+// attribute to make efficient filtering for user action logs
```
ilidemi: Sorry to disappoint but LogAttributes are not indexed by key. Matching the message by prefix would be faster

flow/cmd/handler.go
```diff
@@ -438,14 +441,34 @@ func (h *FlowRequestHandler) FlowStateChange(
...
+					LogActionPauseFlow(ctx, req.FlowJobName)
+				}
 			}
 		case protos.FlowStatus_STATUS_RUNNING:
 			if currState == protos.FlowStatus_STATUS_PAUSED {
 				changeErr = model.FlowSignal.SignalClientWorkflow(ctx, h.temporalClient, workflowID, "", model.NoopSignal)
+				if changeErr == nil {
+					var tablesAdded, tablesRemoved []string
+					if req.FlowConfigUpdate != nil {
+						if cdcUpdate := req.FlowConfigUpdate.GetCdcFlowConfigUpdate(); cdcUpdate != nil {
```
ilidemi: It would make for a better O11y tool to log other fields too, if non-zero. Maybe even one entry here, one in processCDCFlowConfigUpdate where we know the before->after

flow/cmd/user_actions.go
```diff
@@ -0,0 +1,86 @@
...
+const userActionTag = "userAction"
+
+const (
+	UserActionCreateFlow UserActionType = iota
+	UserActionResyncFlow
+	UserActionPauseFlow
+	UserActionResumeFlow
+	UserActionTerminateFlow
+)
```
ilidemi: These are kind of redundant with function names. Can we just inline string constants there?

ilidemi (commented):
Let's add CreatePeer, DropPeer (for OSS sake) and StartMaintenance, EndMaintenance, SkipSnapshotWaitFlows so that they show up in context

flow/cmd/handler.go
jgao54: updating in `processCDCFlowConfigUpdate` is a good idea

PR #3849, opened 2026-01-17
Mitigate blank oids in catalog

ilidemi:
When a table has excluded columns, its oid doesn't get propagated and becomes zero. This makes table addition cancellation unsafe to run.
Fix the miss and add a check in the workflow to error out in such cases. After this change there won't be any new oids missed and the cancellation workflow will be safe to use, at most erroring out preemptively.

Repairing the existing mappings will be a separate change.

### DIFF
--- a/flow/workflows/cancel_table_addition_flow.go
+++ b/flow/workflows/cancel_table_addition_flow.go
@@ -28,15 +28,16 @@ func CancelTableAdditionFlow(ctx workflow.Context, input *protos.CancelTableAddi
 	}
 	getFlowConfigCtx := workflow.WithActivityOptions(ctx, getFlowConfigOptions)
 
-	var flowConfigFetchOutput *protos.GetFlowConfigAndWorkflowIdFromCatalogOutput
+	var flowConfigFetchOutput *protos.GetFlowInfoToCancelFromCatalogOutput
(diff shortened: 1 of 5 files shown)


flow/workflows/cancel_table_addition_flow.go
```diff
@@ -132,6 +133,11 @@ func CancelTableAdditionFlow(ctx workflow.Context, input *protos.CancelTableAddi
 
 	logger.Info("Retrieved PostgreSQL table OIDs", "flowName", flowJobName, "oidCount", len(tableOIDs))
 
+	if tableName, ok := tableOIDs[0]; ok && sourcePeerType == protos.DBType_POSTGRES {
```
jgao54, line 136: not sure if understand this logic here: `ok` indicates tableName found, but error is logging "PostgreSQL schema has zero OID".

jgao54: oh this is table with excluded columns

ilidemi: The key is 0 which is not supposed to happen for PG

jgao54: ahh right, read it like 0 index, but it's zero value

jgao54 (approved):

PR #3859, opened 2026-01-21
Flow config logging for better observability

jgao54:
Provide high-level flow config logging. Does not handle logging config changes, which will be addressed as a follow-up.

### DIFF
--- a/flow/shared/telemetry/activity_logging.go
+++ b/flow/shared/telemetry/activity_logging.go
@@ -7,12 +7,13 @@ import (
 	"log/slog"
 	"strings"
 
-	"github.com/PeerDB-io/peerdb/flow/generated/protos"
-	"github.com/PeerDB-io/peerdb/flow/internal"
-	"github.com/PeerDB-io/peerdb/flow/shared"
 	"github.com/jackc/pgx/v5"
 	"go.temporal.io/sdk/log"
 	"google.golang.org/protobuf/proto"
+
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/internal"
+	"github.com/PeerDB-io/peerdb/flow/shared"
 )
 
 const (
@@ -167,8 +168,8 @@ type TableMappingForLogging struct {
 	TableName     string   `json:"table_name"`
 	DestTableName string   `json:"destination_table_name"`
 	PartitionKey  string   `json:"partition_key"`
-	Exclude       []string `json:"excluded_columns"`
 	Engine        string   `json:"engine"`
+	Exclude       []string `json:"excluded_columns"`
 }
 
 func LogFlowConfigs(ctx context.Context, catalogPool shared.CatalogPool) error {


flow/shared/telemetry/activity_logging.go
```diff
@@ -142,3 +147,86 @@ func logActivity(ctx context.Context, action string, additionalAttrs ...any) {
...
+			TableMappings:               tableMappings,
+		}
+
+		configJSON, err := json.Marshal(flowConfig)
+		if err != nil {
+			logger.Error("failed to marshal flow configs", slog.Any("error", err))
+			continue
+		}
+		logger.Info("[flow config] "+string(configJSON),
+			slog.String("scheduledTask", "RecordFlowConfigs"))
```
ilidemi: Maybe log them the other way around, same body and config in a string property? Would be cheap to match and not noisy, and putting complex data into properties is how we do it in other places

flow/shared/telemetry/activity_logging.go
```diff
@@ -142,3 +147,86 @@ func logActivity(ctx context.Context, action string, additionalAttrs ...any) {
...
+type TableMappingForLogging struct {
+	SourceTable      string   `json:"source_table"`
+	DestinationTable string   `json:"destination_table"`
+	PartitionKey     string   `json:"partition_key"`
+	Exclude          []string `json:"excluded_columns"`
+}
+
+func LogFlowConfigs(ctx context.Context, catalogPool shared.CatalogPool, logger log.Logger) error {
+	rows, err := catalogPool.Query(ctx,
+		`SELECT DISTINCT ON (name) name, config_proto FROM flows WHERE config_proto IS NOT NULL`)
```
ilidemi, line 179: Let's add peer info too? Most of the fields from the source (except credentials) and destination database
Would help to see, is it ssh, is it neon, private link, custom TLS, are different pipes going from/to same db or different

flow/shared/telemetry/activity_logging.go
```diff
@@ -142,3 +147,86 @@ func logActivity(ctx context.Context, action string, additionalAttrs ...any) {
...
+		return cfg, nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to collect flow configs: %w", err)
+	}
+
+	for _, cfg := range configs {
+		tableMappings := make([]TableMappingForLogging, 0, len(cfg.TableMappings))
+		for _, tm := range cfg.TableMappings {
+			tableMappings = append(tableMappings, TableMappingForLogging{
```
ilidemi: I am not entirely sure if table mappings can stop fitting in one log line. Maybe separate them out, at least for resilience, if not test the limits?
Logging pipeline definitely splits the line into multiple past some limit so matching would break

flow/shared/telemetry/activity_logging.go
jgao54: the `scheduledTask`  attribute here is for consistency with other scheduled tasks. Grafana supports a log json parsing, and transformation and extracting fields into a table. so was going with that approach. 

i hesitated on putting it in attribute due to high cardinality of data, in terms of number of services, and every time a modification happens.

flow/shared/telemetry/activity_logging.go
jgao54: yes, just want to start something basic (i.e. FlowConfig) and then add more fields later :)

flow/shared/telemetry/activity_logging.go
ilidemi: Yes either way it would be JSON parsing, just with body you'd also need to strip the prefix, and it would be the least nice looking row when looking through logs for something else. The direction everywhere else has been to remove custom formatting from the body and putting data in properties

jgao54: discussed offline and realized cardinality bloat is not a concern here. this makes sense.

flow/shared/telemetry/activity_logging.go
ilidemi: DBI-458 so we don't forget

PR #3862, opened 2026-01-24
Change normalize buffer to 24h

ilidemi:
Our mantra is "always read the slot" and the current buffer of 128 is not enough for some issues - for 60 second sync interval it's only 2 hours before slot lag starts going up.

Make the size of the buffer approximately 24h (configurable), based on sync interval.
The real value will be longer when the batches are effectively longer:
* pull idling (so no load anyway)
* pull processing big transactions (rare)
* sync interval is so short that the time between batches becomes visible (x1.66 on 1-second batches in ClickPipes 1SU environment)
And shorter when the batches are effectively shorter:
* batch is cut by size rather than time

Remove the `PEERDB_NORMALIZE_CHANNEL_BUFFER_SIZE` setting (to be mentioned in release notes)

QA: tested with the setting at 1h and sync intervals of 1s, 60s, 60m, 70m - worked as expected

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -352,8 +352,9 @@ func (a *FlowableActivity) SyncFlow(
 	idleTimeout := cdcIdleTimeout(int(options.IdleTimeoutSeconds))
 	// normBufferSize allows _approximately_ normBufferHours delay between pull/sync and normalize
 	// under normal steady operation where the batch hits idle timeout every time it will match the hours very closely
-	// effective hours will be longer if pull is idling, or there are waits on big transactions, or the sync interval is so small that start/stop overhead starts being visible
-	// shorter if the batches hit the size limit rather rather than idle timeout
+	// effective hours will be longer if pull is idling, or there are waits on big transactions,
+	// or the sync interval is so small that start/stop overhead starts being visible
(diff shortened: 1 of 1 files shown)


flow/activities/flowable.go
```diff
@@ -337,11 +344,20 @@ func (a *FlowableActivity) SyncFlow(
...
 	}
+	idleTimeout := cdcIdleTimeout(int(options.IdleTimeoutSeconds))
+	// normBufferSize allows _approximately_ normBufferHours delay between pull/sync and normalize
+	// under normal steady operation where the batch hits idle timeout every time it will match the hours very closely
+	// effective hours will be longer if pull is idling, or there are waits on big transactions,
+	// or the sync interval is so small that start/stop overhead starts being visible
+	// will be shorter if the batches hit the size limit rather rather than idle timeout
+	normBufferSize := normBufferHours * 3600 / int64(idleTimeout.Seconds())
+	// Normalize is always 1 batch behind, allow 2 to still run in parallel with pull-sync
+	normBufferSize = max(normBufferSize, 2)
```
jgao54, line 360: this should also account for the scenario where idleTimeout is set to a very high value, say, 48 hours or more. normBufferSize would return 2 here instead of 0.

PR #3867, opened 2026-01-27
Add handling of error code 1226

jgao54:
Add handling of error code 1226

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -605,7 +605,7 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 
 	var myErr *mysql.MyError
 	if errors.As(err, &myErr) {
-		// https://mariadb.com/kb/en/mariadb-error-code-reference
+		// https://mariadb.com/docs/server/reference/error-codes/mariadb-error-code-reference
 		myErrorInfo := ErrorInfo{
 			Source: ErrorSourceMySQL,
 			Code:   strconv.Itoa(int(myErr.Code)),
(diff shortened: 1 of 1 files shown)


flow/alerting/classifier.go
```diff
@@ -946,12 +894,31 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
...
+		}
+
+		var underlyingMyErr *mysql.MyError
+		if errors.As(mysqlStreamingError, &underlyingMyErr) {
+			return classifyMySQLError(underlyingMyErr)
+		}
+
+		// Fallback: try to parse MySQL error code from the error message string
+		// This handles cases where the error chain is broken
+		// (e.g., intermediate wrappers that don't implement Unwrap properly)
```
ilidemi: Do we have any specific cases where that fires? Should there be an issue on go-mysql-org to wrap those?

jgao54: the alert seem to suggest that this case happened, since 
```
	var myErr *mysql.MyError
	if errors.As(err, &myErr) {
             ...
        }
```
should have caught it if the error chain is not broken. however, from looking at the code it was not clear where the chain was broken. doesn't hurt to raise an upstream issue yes.

ilidemi: It was caught? The alert linked from the task says code 1226, so it was parsed just not classified. We don't need the string parsing and the special streaming error handling then, just add the code to the NOTIFY_CONNECTIVITY branch.

jgao54: ah, misread. thanks!

PR #3868, opened 2026-01-27
better mysql column drift error handling

jgao54:
Better handling of mysql column position drift by detecting and erroring early when the following scenario is hit when mysql does not support `row_metadata_data = full`:
- processed a drop column DDL
- processed an add column DDL when the column position is specified (`AFTER`/`FIRST`)

Testing:
- added an e2e test
- Locally tested mysql 5.7 and confirmed that `ATLER TABLE t DROP COLUMN` and `ALTER TABLE t ADD COLUMN xxx AFTER yyy` result in surfacing an expected error message in the logs; but `ALTER TABLE t ADD COLUMN xxx` works as expected.

### DIFF
--- a/flow/e2e/clickhouse_mysql_test.go
+++ b/flow/e2e/clickhouse_mysql_test.go
@@ -968,7 +968,7 @@ func (s ClickHouseSuite) Test_MySQL_Column_Position_Shifting_DDL_Error() {
 
 	for _, tc := range testCases {
 		s.t.Run(tc.name, func(t *testing.T) {
-			srcTableName := fmt.Sprintf("test_position_shift_%s", tc.name)
+			srcTableName := "test_position_shift_" + tc.name
 			srcFullName := s.attachSchemaSuffix(srcTableName)
 			dstTableName := fmt.Sprintf("test_position_shift_%s_dst", tc.name)
 


flow/e2e/clickhouse_mysql_test.go
```diff
@@ -940,3 +942,73 @@ func (s ClickHouseSuite) Test_MySQL_Coercion() {
...
+	require.NoError(s.t, err)
+	if cmp >= 0 {
+		s.t.Skip("only applies to mysql versions WITHOUT binlog_row_metadata support")
+	}
+
+	srcTableName := "test_position_shift_err"
+	srcFullName := s.attachSchemaSuffix(srcTableName)
+	dstTableName := "test_position_shift_err_dst"
+
+	// Create initial table with columns: id, c1, c2
```
ilidemi:
```suggestion
	// Create initial table with columns: id, c1
```

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -940,3 +942,73 @@ func (s ClickHouseSuite) Test_MySQL_Coercion() {
 	env.Cancel(s.t.Context())
 	RequireEnvCanceled(s.t, env)
 }
+
+func (s ClickHouseSuite) Test_MySQL_Column_Position_Shifting_DDL_Error() {
```
ilidemi, line 947: ~~ALTER~~ ADD/CHANGE/MODIFY COLUMN ... FIRST/AFTER also need tests - it's an explicit behavior now. Ok to do as a follow-up

jgao54: good catch, think we could also fail fast for MODIFY/CHANGE column without position specified (as in, block it altogether instead of letting it fail in a confusing way).. but require a bit more thinking so will leave that as a separate exercise.

PR #3883, opened 2026-02-02
Revert "Bring back "Adding support of TIME64 datatype"  (#3876)"

jgao54:
This reverts commit a7ac43e1392d6052ae95e21bb3c85932a1c8e6de.

Will re-apply once cloud is on 25.12


jgao54:
Discussed offline and decide to keep this change. If another internal version is introduced, we can handle it on a case-by-case basis.

PR #3889, opened 2026-02-04
Use more partitions for snapshotting

ilidemi:
The current logic multiplies the rows per partition by an integer until the partition count starts to fit in <= 1000 to not exceed Temporal size limits. This results in the overall 101..1000 expected count, the lower end of this is not enough when there are many workers and especially if partitions are unbalanced.

Instead try to land in the 901..1000 range by discarding the desired row count and picking a round number with 1-2 significant digits, still only when the requested values would've resulted in >1000 partitions.


flow/shared/math.go
```diff
@@ -18,47 +16,34 @@ type AdjustedPartitions struct {
 
 // AdjustNumPartitions takes the total number of rows and the desired number of rows per partition,
 // and returns the adjusted number of partitions and rows per partition so that the partition count does not exceed 1000.
-// It does so by increasing the rows-per-partition by a power-of-10 multiplier when necessary.
+// When adjustment is needed, resulting partition count will be in [901, 1000] range for 9K+ rows and [501, 1000] range for 1-9K rows.
```
jgao54, line 19: General observation, not specific to changes in this PR as it was the case prior as well: since we currently expose `numRowsPerPartition` to user, it does open up room for very inefficient adjusted partitions for small tables if user decides to use a small `numRowsPerPartition`. 

Example: user has a table with a little more than 1M rows,  they set `numRowsPerPartition` to 1k, where they'd end up with ~1k partitions. With our default parallelism of 4 workers, each worker is only processing 10k rows each time, but would have to do this for ~250 times. 

The overhead of scheduling 250 child workflows seem to outweigh the benefit of more granular partitions here when the table size is small. So may be worth handling it like a special case. 

e.g.  if totalRows < 1M, return AdjustedNumPartitions as numWorkers * 10, or some sensible constant

Alternatively, set numRowsPerPartition = max(numRowsPerPartition, 100k), or some sensible min value

jgao54: long-term, echo @serprex , we should just remove this setting -- it's a bit misleading anyways given the value is adjusted anyways. It may be worth to instead simplify it by using a default `numRowsPerPartition` and expose `numPartitionsOverride` as an advanced setting in the UI (which is what `numRowsPerPartition` is used to derive anyways)

ilidemi: The number of workflows will still be 4 in this case, and looping through partitions will stay within the same activity and reuse the same connection. We are losing some efficiency with cutting the data streaming pipeline, especially if the database is in another region, but I wouldn't say it's a giant overhead.

Agreed on the setting becoming meaningless - when we switch to block-based CTID partitioning, there's even less sense in it. The only corner case it's legitimately serving (that I remember seeing) is having very big (hundred MB to GB) rows and wanting extra small partitions not to OOM, although that would practically only work with few rows.

jgao54: overhead of creating a childflow is probably in the hundreds of miliseconds (considering scheduling and dispatching requires a roundtrip to the server). so at 250 sequential child workflow, let's assume 0.5s per schedule (not sure though, would require some measurement), would be an extra 125 seconds spent on overhead. Agree it's not  a big amount, but still nice to shave if this is the case. (not saying we need to now, but just something to keep in mind)

ilidemi: There are no sequential child workflows! It's one workflow and one activity per worker, within a table.

jgao54: my mental model was wrong here, just checked the code and realized we create one child workflow per batch, not per partition. 

ooc, what limit does "partition count starts to fit in <= 1000 to not exceed Temporal size limits." refer to?

ilidemi: Possible there were other limits originally that had more need for caution, the current one would be that the output of an activity can't exceed 2MB. Looking at how big one partition descriptor is, we could probably get away with low thousands but not 50K

PR #3892, opened 2026-02-04
PG to PG Normalize: Account for types in different schemas, quoting

Amogh-Bharadwaj:
## Why
Consider a source Postgres table such as:
```sql
CREATE TABLE funny_types(
    id SERIAL PRIMARY KEY,
    mixey "mixedCaseType",
    coolschema "coolSchema"."coolType"
);
```

When we generate our normalize statements for Postgres to Postgres CDC, we perform type-casting in the commands, where we do `::pgType`. Here, in the case of a mirror where PG type system is enabled, `pgType` could be, for example, an enum in a particular schema, and with particular casing. Our commands fail in these cases. For example:
```
error executing normalize statement for table test.funny_table: ERROR: type "my_type" does not exist 
```
(because it was created in a place not covered by the search path)

## What
The goal is to obtain the schema name of each type of the columns we deal with in normalize.
- Added a new field to `FieldDescription` proto - `typeSchemaName`.
- Introduced utility function in the Postgres connection to hit Postgres catalog tables and give a map of type OIDs to their schema names.
- Populates `typeSchemaName` in two relevant places - `GetTableSchema` and `processRelationMessage` - using above utility.
- In Postgres normalize, checks the above field for each column and if it is not empty and not `pg_catalog`, schema qualify the type name to use in commands.
- Adds E2E test

> [!NOTE]
> Changes this PR makes to creation of target tables and inserts to target in normalize are only in PG type system code paths. QValue system mirrors are not in the scope of this change, apart from the fact their behaviour is untouched.

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -1025,12 +1025,11 @@ func (c *PostgresConnector) GetSchemaNameOfColumnTypeByOID(ctx context.Context, 
 	result := make(map[uint32]string, len(typeOIDs))
 	var oid uint32
 	var schemaName string
-	_, err = pgx.ForEachRow(rows, []any{&oid, &schemaName}, func() error {
+	if _, err := pgx.ForEachRow(rows, []any{&oid, &schemaName}, func() error {
 		result[oid] = schemaName
 		return nil
-	})
-	if err != nil {
-		return nil, fmt.Errorf("error iterating type schemas: %w", err)
+	}); err != nil {
+		return nil, fmt.Errorf("error scanning rows for schema of column types: %w", err)
 	}
 
 	return result, nil


flow/connectors/postgres/client.go
```diff
@@ -601,13 +601,23 @@ func generateCreateTableSQLForNormalizedTable(
 	createTableSQLArray := make([]string, 0, len(tableSchema.Columns)+2)
 	for _, column := range tableSchema.Columns {
 		pgColumnType := column.Type
-		if tableSchema.System == protos.TypeSystem_Q {
+
+		// handle schema-qualified custom types first (for TypeSystem_PG)
+		if tableSchema.System == protos.TypeSystem_PG && column.TypeSchemaName != "" && column.TypeSchemaName != "pg_catalog" {
```
serprex: why omit pg_catalog? ideally search_path would be empty

flow/connectors/postgres/postgres.go
```diff
@@ -886,9 +887,9 @@ func (c *PostgresConnector) GetTableSchema(
 	tableMapping []*protos.TableMapping,
 ) (map[string]*protos.TableSchema, error) {
 	res := make(map[string]*protos.TableSchema, len(tableMapping))
-
+	typeSchemaNameMapping := make(map[uint32]string)
```
serprex:
```suggestion
	typeSchemaNameMapping := make(map[uint32]string, len(tableMapping))
```

flow/connectors/postgres/postgres.go
```diff
@@ -1001,12 +1002,50 @@ func (c *PostgresConnector) GetTablesFromPublication(
...
+		JOIN pg_namespace n ON t.typnamespace = n.oid
+		WHERE t.oid = ANY($1)
+	`, typeOIDs)
+	if err != nil {
+		return nil, fmt.Errorf("error getting schema of column types: %w", err)
+	}
+	defer rows.Close()
+
+	result := make(map[uint32]string, len(typeOIDs))
+	for rows.Next() {
```
serprex: error handling gets simplified here using pgx.ForEach

flow/connectors/postgres/postgres.go
```diff
@@ -1067,11 +1106,43 @@ func (c *PostgresConnector) getTableSchemaForTable(
 	if err != nil {
 		return nil, fmt.Errorf("error getting table schema for table %s: %w", schemaTable, err)
 	}
-	defer rows.Close()
 
-	fields := rows.FieldDescriptions()
+	// Make a copy of field descriptions since pgx may reuse the underlying array
+	fieldDescriptionsSlice := rows.FieldDescriptions()
+	fields := make([]pgconn.FieldDescription, len(fieldDescriptionsSlice))
+	copy(fields, fieldDescriptionsSlice)
```
serprex:
```suggestion
	fields := slices.Clone(fieldDescriptionsSlice)
```

flow/connectors/postgres/postgres.go
```diff
@@ -1067,11 +1106,43 @@ func (c *PostgresConnector) getTableSchemaForTable(
...
+		if _, exists := typeSchemaNameMapping[fieldDescription.DataTypeOID]; !exists {
+			unfetchedOIDsMap[fieldDescription.DataTypeOID] = struct{}{}
+		}
+	}
+
+	// Convert map to slice
+	unfetchedOIDs := make([]uint32, 0, len(unfetchedOIDsMap))
+	for oid := range unfetchedOIDsMap {
+		unfetchedOIDs = append(unfetchedOIDs, oid)
+	}
```
serprex:
```suggestion
	unfetchedOIDs := slices.Collect(maps.Keys(unfetchedOIDsMap))
```

flow/connectors/postgres/client.go
Amogh-Bharadwaj: Got it, let me test with pg_catalog not omitted

flow/connectors/postgres/postgres.go
```diff
@@ -1001,12 +1003,46 @@ func (c *PostgresConnector) GetTablesFromPublication(
...
+
+	result := make(map[uint32]string, len(typeOIDs))
+	var oid uint32
+	var schemaName string
+	_, err = pgx.ForEachRow(rows, []any{&oid, &schemaName}, func() error {
+		result[oid] = schemaName
+		return nil
+	})
+	if err != nil {
```
serprex:
```suggestion
	if _, err := pgx.ForEachRow(rows, []any{&oid, &schemaName}, func() error {
		result[oid] = schemaName
		return nil
	}); err != nil {
```

PR #3897, opened 2026-02-05
Postgres to Postgres: Normalize one batch at a time

Amogh-Bharadwaj:
Normalizing change data across batches for Postgres to Postgres mirrors is tricky due to the possibility of there being updates with unchanged toast columns across multiple batches, where we can only backfill data when pulling in the scope of a single batch and not across batches. This can result in errors where normalize tries to perform an update on a row which does not exist and error out.

This is a stop-gap PR where we now normalize one batch at a time

TODO:
E2E test for the error this fixes

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -747,7 +747,8 @@ func (c *PostgresConnector) NormalizeRecords(
 						slog.String("destinationTableName", destinationTableName),
 						slog.Any("error", err),
 					)
-					return model.NormalizeResponse{}, fmt.Errorf("error executing normalize statement for table %s: %w", destinationTableName, err)
+					return model.NormalizeResponse{},
+						fmt.Errorf("error executing normalize statement for table %s: %w", destinationTableName, err)
 				}
 				totalRowsAffected += int(ct.RowsAffected())
 			}


flow/connectors/postgres/postgres.go
```diff
@@ -733,21 +733,24 @@ func (c *PostgresConnector) NormalizeRecords(
...
+	for batchId := normBatchID + 1; batchId <= req.SyncBatchID; batchId++ {
+		for _, destinationTableName := range destinationTableNames {
+			normalizeStatements := normalizeStmtGen.generateNormalizeStatements(destinationTableName)
+			for _, normalizeStatement := range normalizeStatements {
+				ct, err := normalizeRecordsTx.Exec(ctx, normalizeStatement, batchId, destinationTableName)
+				if err != nil {
+					c.logger.Error("error executing normalize statement",
+						slog.String("statement", normalizeStatement),
+						slog.Int64("currentBatchId", batchId),
+						slog.Int64("normBatchID", normBatchID),
```
heavycrystal, line 745: logs need to be adjusted

Amogh-Bharadwaj: What is the adjustment?

Amogh-Bharadwaj: Changed the casing

PR #3899, opened 2026-02-06
Remove flow-api gRPC dependency from e2e tests

iskakaushik:
## Summary
Replaces all gRPC `FlowServiceClient` calls (connecting to flow-api on port 8112) in e2e tests with direct in-process calls to `*cmd.FlowRequestHandler`. This removes the need to have a running flow-api server when running e2e tests, since `FlowRequestHandler` has the same method signatures and its `APIError` return type satisfies the `error` interface.


ilidemi:
Great direction @iskakaushik! As this is an infra change, please make sure to add relevant reviewers when it's ready to be taken a look at

PR #3901, opened 2026-02-06
Add TILT setup

ilidemi:
We are pretty much able to piggyback on docker-compose-dev that TILT wonderfully integrates with.

Local UI port changed to 3030 to avoid conflict with a ClickHouse Inc's internal service in e2e setup.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,0 +1,26 @@
+docker_compose('./docker-compose-dev.yml')
+
+local_resource(
+    'proto-gen',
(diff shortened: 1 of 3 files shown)


Tiltfile
```diff
@@ -0,0 +1,26 @@
+docker_compose('./docker-compose-dev.yml')
```
jgao54, line 1: nice

jgao54 (approved):

PR #3904, opened 2026-02-09
Heartbeat in Postgres: Make pg_logical_emit_message transactional

Amogh-Bharadwaj:
We have noticed cases where in freshly spun up Postgres instances which have a replication slot and without having a single transaction performed on them by the user, our logical decoding messages emitted via the heartbeat workflow can cause WAL files to accumulate over time, taking up disk space, even though the slot size is low.

This is due to a Postgres issue where non-transactional changes in the WAL do not trigger recomputation of global required LSN across all slots. Thus this PR makes a change where it changes the first parameter in pg_emit_logical_message to true. The first parameter denotes whether the messages are transactional (i.e increment xid):
https://pgpedia.info/p/pg_logical_emit_message.html

### DIFF
--- a/flow/internal/dynamicconf.go
+++ b/flow/internal/dynamicconf.go
@@ -98,7 +98,7 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 	},
 	{
 		Name:             "PEERDB_WAL_HEARTBEAT_QUERY",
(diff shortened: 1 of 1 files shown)


ilidemi:
@serprex could you revert and leave a reapply PR open so we can merge it after the next release?

PR #3910, opened 2026-02-09
Add a Claude skill for updating dependencies

ilidemi:
We're doing this every few weeks, it requires fuzzy manual steps that can be automated and workflow'd.

Tested in https://github.com/PeerDB-io/peerdb/pull/3907


jgao54 (approved):
PR looks good.

Nit: maybe add an `allowed-tools` to scope the set of tools (i haven't tested it out though, so not sure if it's more effort than it's worth)

PR #3911, opened 2026-02-09
Fix precision bug in FormatExtendedTimeDuration

jgao54:
@ilidemi raised a [precision error ](https://github.com/PeerDB-io/peerdb/actions/runs/21835661492/job/63005868867?pr=3905)in our e2e test and after some investigation we did it turns out  `toDateTime64` multiplies the float by 1000000 then static casts to Int64 which throws out the 99999 without rounding, this can cause precision issue by a microsecond.  The fix is to stay away from floats :)

### DIFF
--- a/flow/connectors/clickhouse/normalize_query_test.go
+++ b/flow/connectors/clickhouse/normalize_query_test.go
@@ -33,6 +33,7 @@ func TestExtendedTimeToDateTime(t *testing.T) {
 		{"negative", -1 * time.Minute},
 		{"negative_extended", -123*time.Hour - 45*time.Minute - 7*time.Second - 899999*time.Microsecond},
 		{"precision_test", 18*time.Hour + 18*time.Minute + 8*time.Second + 511787*time.Microsecond},
+		{"trailing_zeros_test", 18*time.Hour + 18*time.Minute + 8*time.Second + 500000*time.Microsecond},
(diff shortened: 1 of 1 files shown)


flow/connectors/clickhouse/normalize_query_test.go
```diff
@@ -32,6 +32,7 @@ func TestExtendedTimeToDateTime(t *testing.T) {
 		{"extended_hours", 838*time.Hour + 59*time.Minute + 59*time.Second + 999999*time.Microsecond},
 		{"negative", -1 * time.Minute},
 		{"negative_extended", -123*time.Hour - 45*time.Minute - 7*time.Second - 899999*time.Microsecond},
+		{"precision_test", 18*time.Hour + 18*time.Minute + 8*time.Second + 511787*time.Microsecond},
```
ilidemi, line 35:
```suggestion
		{"precision_test", 18*time.Hour + 18*time.Minute + 8*time.Second + 511787*time.Microsecond},
		{"trailing_zeros_test", 18*time.Hour + 18*time.Minute + 8*time.Second + 500000*time.Microsecond},
```

ilidemi (approved):
Go formatting can drop trailing zeros, but we're not using it here, so just making it explicit that this is handled.

PR #3912, opened 2026-02-10
Limit max byte length for PG handshake

ilidemi:
When the endpoint is misbehaved (e.g. a TCP tunnel started pointing at something new), the random initial sequence of bytes may be misinterpreted as a multi-GB message and crash the process. Avoid that by limiting the max body length just for the initial handshake. The param used here is meant exactly for that.

### DIFF
--- a/flow/connectors/postgres/ssh_wrapped_conn.go
+++ b/flow/connectors/postgres/ssh_wrapped_conn.go
@@ -58,7 +58,7 @@ func NewPostgresConnFromConfig(
 	// crash the process. Avoid that by limiting the max body length just for the initial handshake.
 	connConfig.BuildFrontend = func(r io.Reader, w io.Writer) *pgproto3.Frontend {
 		frontend := pgproto3.NewFrontend(r, w)
-		frontend.SetMaxBodyLen(1024 * 1024)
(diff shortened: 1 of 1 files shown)


flow/connectors/postgres/ssh_wrapped_conn.go
```diff
@@ -50,6 +52,15 @@ func NewPostgresConnFromConfig(
 		connConfig = connConfig.Copy()
 		connConfig.Password = token
 	}
+
+	// If the endpoint is misbehaved (e.g. a TCP tunnel started pointing at something new),
+	// the random initial sequence of bytes may be misinterpreted as a multi-GB message and
+	// crash the process. Avoid that by limiting the max body length just for the initial handshake.
+	connConfig.BuildFrontend = func(r io.Reader, w io.Writer) *pgproto3.Frontend {
+		frontend := pgproto3.NewFrontend(r, w)
+		frontend.SetMaxBodyLen(1024 * 1024)
```
jgao54: ooc: is 1MB used here as a safe upper bound because correct handshake would be guaranteed to be less than this?

ilidemi: LLMs are saying many times over. Bumping to 16MB just in case.

PR #3913, opened 2026-02-11
better mongodb tls handling

jgao54:
- Update Mongo's TLS handling such that we assess it in the following order:
(1) if user disabled TLS via toggle —  honor it
(2) if user set tls=false in URI param — honor it
(3) if user pass in tlsInsecure=true -- honor it when creating tls config
(4) default to creating the tls config
- Create `BuildClientOption` in pkg so it can be shared
- Move `CreateTlsConfig` to pkg so it can be shared

### DIFF
--- a/flow/pkg/mongo/client_options_test.go
+++ b/flow/pkg/mongo/client_options_test.go
@@ -51,8 +51,8 @@ func TestBuildClientOptions_TLS(t *testing.T) {
 			name:            "default TLS config with ServerName set",
 			uri:             "mongodb://localhost:27017/",
 			disableTls:      false,
-			expectTLSConfig: true,
 			tlsHost:         "custom.host.com",
+			expectTLSConfig: true,
 		},
 	}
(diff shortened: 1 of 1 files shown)


flow/pkg/mongo/client_options_test.go
```diff
@@ -0,0 +1,87 @@
...
+			disableTls:               false,
+			expectTLSConfig:          true,
+			expectInsecureSkipVerify: false,
+		},
+		{
+			name:            "default TLS config with ServerName set",
+			uri:             "mongodb://localhost:27017/",
+			disableTls:      false,
+			expectTLSConfig: true,
+			tlsHost:         "custom.host.com",
```
ilidemi:
```suggestion
			tlsHost:         "custom.host.com",
			expectTLSConfig: true,
```

ilidemi (approved):
Aligning pkg is great

PR #3914, opened 2026-02-11
o11y: add a normalize lag for observability

jgao54:
Add a NormalizeLag to detect delay in the normalize side by comparing the current timestamp against the timestamp of when a batch was synced. Use sync.Map for thread-safety since normalize is running asynchronously in a separate go-routine. Note this is not e2e latency. So if this metric goes up it strict indicates the delay in the NormalizeLoop (which compliments commitLag for source lag). 

The metric is recorded at the start of every loop iteration (and batchId is deleted from the map when batch completes). This means a huge lag will require memory to buffer these timestamps, but in practice this map size should be reasonable given our 24 hour limit on back-pressure (86400 entries at max).

Alternative evaluated was measuring e2e latency, however it means every pipe will end up making a database call on every batch, which seems quite expensive, so leaning into this cheaper in-memory approach.

### DIFF
--- a/flow/connectors/utils/monitoring/monitoring.go
+++ b/flow/connectors/utils/monitoring/monitoring.go
@@ -120,7 +120,7 @@ func GetPendingNormalizeLagByFlow(
 	ctx context.Context,
 	pool shared.CatalogPool,
 ) (map[string]int64, error) {
-	// using server-side time delta to avoid clock skew
+	// using catalog time to avoid clock skew with ScheduledTasks
 	rows, err := pool.Query(ctx,
 		`SELECT flow_name, (EXTRACT(EPOCH FROM (NOW() - MIN(sync_time))) * 1000000)::bigint
 		FROM peerdb_stats.cdc_batches


masterashu (approved):

flow/activities/flowable.go
```diff
@@ -1297,8 +1304,21 @@ func (a *FlowableActivity) RecordSlotSizes(ctx context.Context) error {
...
 		}
 		cancel()
+
+		if qryErr == nil {
+			flowName := info.config.FlowJobName
+			// record 0 as the lag if the flow is not in the normalizeLagByFlow map,
+			var lagMicroseconds int64
+			if lag, ok := normalizeLagByFlow[flowName]; ok {
+				lagMicroseconds = lag
+			}
```
ilidemi:
```suggestion
			lagMicroseconds := normalizeLagByFlow[flowName] // record 0 as the lag if the flow is not in the map
```

nexus/catalog/migrations/V50__cdc_batches_sync_time.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE peerdb_stats.cdc_batches ADD COLUMN IF NOT EXISTS sync_time TIMESTAMP;
```
ilidemi, line 1:
```suggestion
ALTER TABLE peerdb_stats.cdc_batches ADD COLUMN IF NOT EXISTS sync_time TIMESTAMP;
COMMENT ON COLUMN peerdb_stats.cdc_batches.sync_time IS 'Set via PG NOW() for clock-skew-safe lag calculation';
```

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -114,6 +116,33 @@ func UpdateEndTimeForCDCBatch(
 	return nil
 }
 
+func GetPendingNormalizeLagByFlow(
+	ctx context.Context,
+	pool shared.CatalogPool,
+) (map[string]int64, error) {
+	// using server-side time delta to avoid clock skew
```
ilidemi:
```suggestion
	// using catalog time to avoid clock skew with ScheduledTasks
```

PR #3919, opened 2026-02-12
Handle "could not import the requested snapshot" as irrecoverable

ilidemi:
failed to get partitions from source: failed to set transaction snapshot: failed to set transaction snapshot: ERROR: could not import the requested snapshot (SQLSTATE 55000)

Seen on Aurora after looping on NTILE error for a week

### DIFF
--- a/flow/connectors/postgres/qrep.go
+++ b/flow/connectors/postgres/qrep.go
@@ -153,6 +153,10 @@ func (c *PostgresConnector) setTransactionSnapshot(ctx context.Context, tx pgx.T
 			if shared.IsSQLStateError(err, pgerrcode.UndefinedObject, pgerrcode.InvalidParameterValue) {
 				return temporal.NewNonRetryableApplicationError("failed to set transaction snapshot",
 					exceptions.ApplicationErrorTypeIrrecoverableInvalidSnapshot.String(), err)
+			} else if shared.IsSQLStateErrorSubstring(err,
+				pgerrcode.ObjectNotInPrerequisiteState, "could not import the requested snapshot") {
+				return temporal.NewNonRetryableApplicationError("failed to set transaction snapshot",
+					exceptions.ApplicationErrorTypeIrrecoverableCouldNotImportSnapshot.String(), err)
 			}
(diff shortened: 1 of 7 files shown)


flow/connectors/postgres/qrep.go
```diff
@@ -153,6 +153,10 @@ func (c *PostgresConnector) setTransactionSnapshot(ctx context.Context, tx pgx.T
 			if shared.IsSQLStateError(err, pgerrcode.UndefinedObject, pgerrcode.InvalidParameterValue) {
 				return temporal.NewNonRetryableApplicationError("failed to set transaction snapshot",
 					exceptions.ApplicationErrorTypeIrrecoverableInvalidSnapshot.String(), err)
+			} else if shared.IsSQLStateErrorSubstring(err,
```
jgao54, line 156: more of a general pattern in the codebase to discuss: currently there's two ways to surface postgres errors: one is by wrapping it around an temporal error as did here; another is by letting the error bubble up, and then let it get caught in the PG error path via error code. 

the classifier currently handles application error before pg error, so this would be handled by line 427 in classifier.go. 

when implementing https://github.com/PeerDB-io/peerdb/pull/3967, i started with this approach as well but then switched to let it fall thru and have pg error handle it -- so we can have the PG handle all error code in one place. 

so wondering if you have any thoughts here.

jgao54: iiuc, the temporal NonRetryableAppliacationError implies it won't retry on the _activity_ level, but will still retry the workflow (technically not retried, but in practice, with ContinueAsNew), so we are not necessarily preventing retry with this approach; so iiuc the only purpose here is for classification.

ilidemi: Snapshot id errors can only happen in QRep flows that don't have ContinueAsNew, the named NonRetryableApplicationErrors is our design for having the activities and workflows fail all the way up as PG won't allow to retry the snapshot queries. If preventing retries wasn't a concern, classification should be in classifier.go as much as possible

PR #3922, opened 2026-02-12
UI: Fix table deselection upon schema expand and showing of hard delete

Amogh-Bharadwaj:
- Makes sure expanding schema, selecting all tables, collapsing and expanding schema again still results in all table checkboxes being checked
- Makes sure Hard delete is only shown for ClickHouse destination mirrors

- [x] Functionally tested


ui/app/mirrors/create/cdc/cdc.tsx
```diff
@@ -163,7 +163,7 @@ export default function CDCConfigForm({
   };
 
   const shoudlHideHardDelete = (label: string) => {
```
serprex, line 165:
```suggestion
  const shouldHideHardDelete = (label: string) => {
```

PR #3936, opened 2026-02-16
Handle inherited tables with different schemas

### LINKED ISSUE #3935: Postgres CDC: error when syncing inherited tables where child has different schema than parent
### Summary
When running Postgres CDC with `PEERDB_POSTGRES_CDC_HANDLE_INHERITANCE_FOR_NON_PARTITIONED_TABLES` enabled and the source uses old-style `INHERITS` tables (not declarative partitioning), child tables with **extra columns** beyond the parent's schema cause tuple decoding failures or cross-child data corruption.

Unlike partitioned tables where all partitions share an identical column layout, `INHERITS` children can add their own columns (e.g. `stripe_payment` inherits 18 columns from `payment` and adds `fk_stripe_payment` and `stripe_account_id`). The WAL emits tuples using the child's column layout, but PeerDB was decoding them using the wrong RelationMessage.

### Root cause
When a child table's `RelationMessage` arrived in the WAL stream, `processMessage` mutated `msg.RelationID` from the child's ID to the parent's:
```go
msg.RelationID, err = p.checkIfUnknownTableInherits(ctx, msg.RelationID)
```
This caused `processRelationMessage` to store the child's RelationMessage under the **parent's** relation ID:
```go
p.relationMessageMapping[currRel.RelationID] = currRel  // stored under parent's ID
```
Then in `processInsertMessage` / `processUpdateMessage` / `processDeleteMessage`, the RelationMessage was looked up using the parent's ID:
```go
rel, ok := p.relationMessageMapping[relID]  // relID = parent's ID
```
This had two consequences:
1. **Cross-child contamination**: multiple children's RelationMessages overwrote each other under the same parent key — whichever child sent last determined the column layout used for all children.
2. **Positional mismatch**: a child with extra columns (e.g. 20 columns) had its tuple decoded using a RelationMessage from a different child or the parent (e.g. 18 columns), causing columns to be decoded with the wrong types.

### Fix
- Stop mutating `msg.RelationID` in the `RelationMessage` handler. Store each child's RelationMessage under its **own** relation ID.
- In `processInsertMessage` / `processUpdateMessage` / `processDeleteMessage`, use the child's actual relation ID (`actualRelID`) for RelationMessage lookup (correct tuple decoding), and the parent's relation ID (`relID`) only for table name and destination routing.
- In `processRelationMessage`, resolve to the parent via `getParentRelIDIfPartitioned` for name/schema lookups, but leave `currRel.RelationID` untouched for storage.

lukejudd-lux:
## Summary
Fixes Postgres CDC tuple decoding failures and cross-child data corruption when syncing old-style `INHERITS` tables where children have additional columns beyond the parent's schema.

## Problem
When a child table's `RelationMessage` arrived, `processMessage` mutated `msg.RelationID` to the parent's ID, causing the child's RelationMessage to be stored under the parent's key in `relationMessageMapping`. This had two consequences:

1. **Cross-child contamination**: multiple children overwrote each other's RelationMessage under the same parent key — whichever child sent last determined the column layout used for decoding all children's tuples.
2. **Positional column mismatch**: a child with extra columns (e.g. `stripe_payment` with 20 columns including `fk_stripe_payment`) had its tuple decoded using the parent's or another child's RelationMessage (e.g. 18 columns), causing columns to be decoded with wrong types.

## Solution
- **RelationMessage handling**: stop mutating `msg.RelationID` — store each child's RelationMessage under its own relation ID. Use the parent's ID only for the "do we care?" check and table/schema lookups.
- **Tuple decoding**: in `processInsertMessage` / `processUpdateMessage` / `processDeleteMessage`, use the child's actual relation ID (`actualRelID`) for RelationMessage lookup so column types match the WAL tuple. The parent's ID is used only for table name and destination routing.

## Testing
- `TestInheritedTableWithExtraColumns`: parent `payment` (4 columns), child `stripe_payment` (6 columns including `fk_stripe_payment` TEXT and `stripe_account_id` TEXT). Inserts from child with Stripe charge ID `ch_3T1KxCFUtwYrZPVC0M6OeTpy` — verifies extra columns are decoded as text, not misinterpreted due to positional mismatch.
- `TestMultipleInheritedChildrenNoContamination`: two children (`stripe_payment`, `paypal_payment`) with different extra columns. Verifies inserts from both are decoded using their own RelationMessages without cross-child contamination.

Fixes #3935

### DIFF
--- a/flow/connectors/postgres/cdc_inheritance_test.go
+++ b/flow/connectors/postgres/cdc_inheritance_test.go
@@ -182,12 +182,12 @@ func TestMultipleInheritedChildrenNoContamination(t *testing.T) {
 
 	var parentOID, stripeOID, paypalOID uint32
 	for _, tbl := range []struct {
-		name string
 		oid  *uint32
+		name string
 	}{
-		{"payment", &parentOID},
-		{"stripe_payment", &stripeOID},
-		{"paypal_payment", &paypalOID},
+		{&parentOID, "payment"},
+		{&stripeOID, "stripe_payment"},
+		{&paypalOID, "paypal_payment"},
 	} {
 		err = connector.conn.QueryRow(ctx,
 			`SELECT c.oid FROM pg_class c JOIN pg_namespace n ON c.relnamespace = n.oid
@@ -326,12 +326,12 @@ func TestChildTableSchemaDeltaNotFalsePositive(t *testing.T) {
 
 	var parentOID, stripeOID, promoOID uint32
 	for _, tbl := range []struct {
-		name string
 		oid  *uint32
+		name string
 	}{
-		{"payment", &parentOID},
-		{"stripe_payment", &stripeOID},
-		{"promo_payment", &promoOID},
+		{&parentOID, "payment"},
+		{&stripeOID, "stripe_payment"},
+		{&promoOID, "promo_payment"},
 	} {
 		err = connector.conn.QueryRow(ctx,
 			`SELECT c.oid FROM pg_class c JOIN pg_namespace n ON c.relnamespace = n.oid


lukejudd-lux:
@heavycrystal after looking a little closer its not that they are _different_ types at source

Its that the inherited table has more columns than the parent (as can happen under relkind 'r'). So the columns are out of alignment at parsing time.

I've updated the test to address the real issue, the code changes still address the actual problem.

heavycrystal:
@lukejudd-lux child tables having mismatched schemas from the parent table from inheritance is something we've hit before. It is a niche feature that is tricky to support reliably during CDC.

As an example, it's possible for a column to be added and then dropped almost immediately during CDC. This should still create the column on the destination table and populate a few rows with the data. Your change makes it such that if `pg_attribute` doesn't contain the column when querying, the schema change won't be propagated.

PR #3938, opened 2026-02-16
feat(clickhouse): add directory-based TLS certificate loading

jkaflik:
Add `tls_certificate_directory` option to ClickHouse peers for loading TLS certificates from a directory (e.g. a Kubernetes Secret mounted by cert-manager).

- Client certs are re-read on every TLS handshake via `GetClientCertificate`, picking up rotations without restart
- CA cert is loaded once at connection time
- Falls back to existing inline cert fields when directory is not set
- Fix POSIX compatibility bug in `generate-protos.sh` (`&>` bashism broke proto generation on cache miss)

### DIFF
--- a/flow/connectors/clickhouse/clickhouse.go
+++ b/flow/connectors/clickhouse/clickhouse.go
@@ -256,8 +256,13 @@ func configureDirectoryTLS(tlsConfig *tls.Config, dir string) error {
 
 	// Load CA certificate upfront — RootCAs must be set before the TLS
 	// handshake so the server certificate can be verified.
+	// ca.crt is optional: when absent, the system CA pool is used instead.
 	caCertPEM, err := os.ReadFile(caPath)
-	if err == nil && len(caCertPEM) > 0 {
+	if errors.Is(err, os.ErrNotExist) {
+		// ca.crt is optional — use system CA pool
+	} else if err != nil {
+		return fmt.Errorf("failed to read CA certificate from %q: %w", caPath, err)
+	} else if len(caCertPEM) > 0 {
 		caPool := x509.NewCertPool()
 		if !caPool.AppendCertsFromPEM(caCertPEM) {
 			return fmt.Errorf("failed to parse CA certificate from %q", caPath)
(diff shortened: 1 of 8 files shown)


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -232,32 +234,114 @@ func (c *ClickHouseConnector) ValidateCheck(ctx context.Context) error {
...
+	if _, err := os.ReadFile(certPath); err != nil {
+		return fmt.Errorf("failed to read TLS certificate from %q: %w", certPath, err)
+	}
+	if _, err := os.ReadFile(keyPath); err != nil {
+		return fmt.Errorf("failed to read TLS private key from %q: %w", keyPath, err)
+	}
+
+	// Load CA certificate upfront — RootCAs must be set before the TLS
+	// handshake so the server certificate can be verified.
+	caCertPEM, err := os.ReadFile(caPath)
```
ilidemi, line 260: Do we need to error on err != nil here?

ilidemi (approved):

flow/connectors/clickhouse/clickhouse.go
whites11: Good catch — ca.crt is intentionally optional (system CA pool is used when it's absent), so `os.ErrNotExist` should be silently skipped. But you're right that other errors (e.g. permission denied) would be silently swallowed here, which could cause confusing TLS failures downstream.

I'll push a fix that distinguishes "not found" (skip) from unexpected errors (surface):

```go
caCertPEM, err := os.ReadFile(caPath)
if errors.Is(err, os.ErrNotExist) {
    // ca.crt is optional — use system CA pool
} else if err != nil {
    return fmt.Errorf("failed to read CA certificate from %q: %w", caPath, err)
} else if len(caCertPEM) > 0 {
    caPool := x509.NewCertPool()
    if !caPool.AppendCertsFromPEM(caCertPEM) {
        return fmt.Errorf("failed to parse CA certificate from %q", caPath)
    }
    tlsConfig.RootCAs = caPool
}
```

PR #3953, opened 2026-02-18
Parse Time64 value without setting it during table creation

jgao54:
- Adds ClickHouse Time64(6) type support for TIME/TIMETZ columns, gated by a read-only capability flag (clickhouse_time64_enabled) detected at flow creation time and persisted to the flow config protobuf.
- Replaces the previous InternalVersion-based approach with a map<string, bool> flags field on FlowConnectionConfigs, FlowConnectionConfigsCore, QRepConfig, and SetupNormalizedTableBatchInput. This decouples connector-specific capabilities from the global linear version ordering.
- Introduces a GetFlagsConnector interface. ClickHouse's implementation queries system.settings to check if enable_time_time64_type is already enabled on the server -- PeerDB never attempts to set this value.
- Flags are immutable after flow creation: if the flag was false and the admin enables it later, only new flows will use Time64, existing flows retain their original behavior, including resync. If the flag was true and the admin disables the setting, queries will error out if it depends on Time64.

Testing: e2e tests + test migration path in in CHC + test non-supported CHC environments + validated resync blocks when setting is updated from enabled -> disabled.

### DIFF
--- a/ui/app/mirrors/create/helpers/common.ts
+++ b/ui/app/mirrors/create/helpers/common.ts
@@ -49,7 +49,7 @@ export const blankCDCSetting: CDCConfig = {
   env: {},
   envString: '',
   version: 0,
-  flags: {},
+  flags: [],
 };
 
 export const cdcSourceDefaults: { [index: string]: Partial<CDCConfig> } = {
@@ -87,5 +87,5 @@ export const blankQRepSetting: QRepConfig = {
   exclude: [],
   columns: [],
   sourceType: DBType.DBTYPE_UNKNOWN,
-  flags: {},
+  flags: [],
 };


flow/connectors/clickhouse/clickhouse.go
```diff
@@ -406,6 +406,21 @@ func (c *ClickHouseConnector) GetVersion(ctx context.Context) (string, error) {
...
 }
 
+func (c *ClickHouseConnector) GetFlags(ctx context.Context) map[string]bool {
+	flags := make(map[string]bool)
+
+	var time64Setting string
+	err := c.queryRow(ctx,
+		"SELECT value FROM system.settings WHERE name = 'enable_time_time64_type'",
+	).Scan(&time64Setting)
+	if err == nil && time64Setting == "1" {
```
ilidemi: The error should be propagated, random failure shouldn't silently change the schema

flow/cmd/handler.go
```diff
@@ -61,32 +61,21 @@ func (h *FlowRequestHandler) getPeerID(ctx context.Context, peerName string) (in
 	return id.Int32, nil
 }
 
-func (h *FlowRequestHandler) determineInternalVersion(
+func (h *FlowRequestHandler) determineFlags(
```
ilidemi, line 64: Resync should call determineFlags and error out with NewFailedPreconditionApiError and a user-friendly message if they don't match. If the pipe error-loops in setup instead, the user can't change it to remove the table if it's not needed anymore, or take time to consider what to do

jgao54: good idea

flow/cmd/validate_mirror.go
```diff
@@ -125,3 +137,55 @@ func (h *FlowRequestHandler) checkIfMirrorNameExists(ctx context.Context, mirror
 
 	return nameExists, nil
 }
+
+// checkFlagsCompatibility blocks resync when a destination feature flag that was
+// enabled at mirror creation is now disabled and the stored schema contains
+// column types whose type mapping depends on that flag.
```
ilidemi: User might've deleted the time column on the source or changed it to string, we need to validate new source schema rather than the stored one

flow/cmd/validate_mirror.go
```diff
@@ -125,3 +146,53 @@ func (h *FlowRequestHandler) checkIfMirrorNameExists(ctx context.Context, mirror
...
+			for _, col := range schema.Columns {
+				if slices.Contains(colTypes, types.QValueKind(col.Type)) {
+					return true, nil
+				}
+			}
+		}
+		return false, nil
+	}
+
+	for flag, wasEnabled := range cfg.Flags {
```
ilidemi: Resyncing an existing mirror/pipe would skip this loop and potentially change the schema, so it needs to be more like set union or iterating over the list of possible flags. Also `map[string]struct{}` would make for fewer cases to think through

flow/cmd/validate_mirror.go
```diff
@@ -125,3 +146,53 @@ func (h *FlowRequestHandler) checkIfMirrorNameExists(ctx context.Context, mirror
 
 	return nameExists, nil
 }
+
+// checkFlagsCompatibility blocks resync when a destination feature flag that was
+// enabled at mirror creation is now disabled and the stored schema contains
```
ilidemi, line 151: Flag going from disabled to enabled is also a problem, if that type is present. In that case we would want to remove the flag and continue (the current logic is fetching them twice)

flow/cmd/validate_mirror.go
```diff
@@ -125,3 +146,53 @@ func (h *FlowRequestHandler) checkIfMirrorNameExists(ctx context.Context, mirror
...
+	if err != nil {
+		return NewInternalApiError(fmt.Errorf("failed to determine destination flags: %w", err))
+	}
+
+	schemaHasColumnTypes := func(colTypes []types.QValueKind) (bool, error) {
+		tableNames := make([]string, 0, len(cfg.TableMappings))
+		for _, tm := range cfg.TableMappings {
+			tableNames = append(tableNames, tm.DestinationTableIdentifier)
+		}
+		schemas, err := internal.LoadTableSchemasFromCatalog(ctx, h.pool, cfg.FlowJobName, tableNames)
```
ilidemi, line 167: There is a nit case where the type was present in the mappings but is not on the source anymore (column dropped/type changed), but going into that rabbithole requires adding a single-query "get a set of types from these tables" method on all sources, and even then we wouldn't catch cases like, remove the table with time - disable the setting - resync - add the table with time, or move to a different cdc-only mirror into the same destination. So I'd say fine to keep as is until it proves to be problematic, as the current approach solves for 99%.

flow/cmd/validate_mirror.go
```diff
@@ -125,3 +146,53 @@ func (h *FlowRequestHandler) checkIfMirrorNameExists(ctx context.Context, mirror
...
+				if slices.Contains(colTypes, types.QValueKind(col.Type)) {
+					return true, nil
+				}
+			}
+		}
+		return false, nil
+	}
+
+	for flag, wasEnabled := range cfg.Flags {
+		if !wasEnabled || newFlags[flag] {
```
ilidemi: `map[string]struct{}` would make this easier to think through, and we're sticking to that pattern throughout the codebase (which I missed on the last round of the review). Mind making the change?

flow/cmd/validate_mirror.go
jgao54: Not sure if i understand the concern here.

The flag is immutable, so once it's stored into catalog the first time, we don't want to change the flag on resync (and if the flag was not there, it's treated the same as disabled). If the settings for time64 is enabled on resync, we ignore the setting change and continue to replicated based on the original flag setting. So it is only problematic when the setting goes from enabled -> disabled, because subsequent `CREATE TABLE` with Time64 column will fail since the setting disabled it.

flow/cmd/validate_mirror.go
jgao54: sounds good, this way flag not existing on older version of clickhouse and flag disabled can be handled the same way which is nice.

jgao54: proto does not support `map[string]struct{}`, simplifying it with `repeated string` instead (the added ceremony of converting to `map[string]struct{}`  everywhere is hard to justify.

flow/cmd/validate_mirror.go
ilidemi: Right, makes sense. I missed the part where us not modifying the original value of the flag is actually a feature, and that CH with TIME64 enabled still allows us to take the non-TIME64 path.

ilidemi (approved):

PR #3956, opened 2026-02-18
Enable SSH keepalive for Postgres

ilidemi:
Reusing the MySQL implementation with less connection juggling

Tested on a customer branch

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -54,7 +54,6 @@ type PostgresConnector struct {
 	relationMessageMapping model.RelationMessageMapping
 	typeMap                *pgtype.Map
 	rdsAuth                *utils.RDSAuth
-	done                   chan struct{}
 	connStr                string
 	metadataSchema         string
 	replLock               sync.Mutex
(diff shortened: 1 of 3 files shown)


flow/connectors/postgres/postgres.go
```diff
@@ -122,10 +124,42 @@ func NewPostgresConnector(ctx context.Context, env map[string]string, pgConfig *
...
 		replLock:               sync.Mutex{},
+		done:                   done,
 		pgVersion:              0,
 		typeMap:                pgtype.NewMap(),
 		rdsAuth:                rdsAuth,
-	}, nil
+	}
+
+	if tunnel != nil {
+		go func() {
```
jgao54: With this change we have 2 goroutines:
- GetKeepaliveChan introduces one for the keepalive itself
- this goroutine here, which receives a signal on error and then close the connection

An alternative approach that wouldn't require introducing a second goroutine:
```
func (tunnel *SSHTunnel) StartKeepalive(ctx context.Context, logger *slog.Logger, onFailure func()) { ... }
```

then can `onFailure()` callback inside `GetKeepaliveChan`'s loop rather than needing this `keepaliveChan` channel at all.

Doesn't block this PR, can be implemented separately if it makes sense.

ilidemi: You're right, no need to penalize Postgres if MySQL is complicated. Added a callback API

PR #3964, opened 2026-02-19
Notify on ECONNRESET during peer creation

ilidemi:
ECONNRESET is currently ignored because it often happens on idle connections. However, if we get it during peer creation, we know it's because the peer is completely unavailable and that's worthy of notifying.

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -351,7 +351,23 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
-	// Connection reset errors can mostly be ignored
+	var peerCreateError *exceptions.PeerCreateError
+	if errors.As(err, &peerCreateError) {
+		if errors.Is(peerCreateError, context.DeadlineExceeded) {
+			return ErrorNotifyConnectivity, ErrorInfo{
+				Source: ErrorSourceOther,
+				Code:   "CONTEXT_DEADLINE_EXCEEDED",
+			}
+		}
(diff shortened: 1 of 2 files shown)


flow/alerting/classifier.go
```diff
@@ -351,7 +351,23 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
...
-	// Connection reset errors can mostly be ignored
+	var peerCreateError *exceptions.PeerCreateError
+	if errors.As(err, &peerCreateError) {
+		if errors.Is(peerCreateError, context.DeadlineExceeded) {
+			return ErrorNotifyConnectivity, ErrorInfo{
+				Source: ErrorSourceOther,
+				Code:   "CONTEXT_DEADLINE_EXCEEDED",
+			}
+		}
+		if errors.Is(peerCreateError, syscall.ECONNRESET) {
```
jgao54, line 362: iiuc econnreset is possible only when an open connection is abruptly closed. for an already unavailable peer, i would not expect this to happen.

jgao54: wondering if this would make more sense to be a retryable error

ilidemi: In the case we saw it was during TCP handshake when one party was expecting something that the other wasn't doing, then gave up. NOTIFY_ errors do wait for a few occurrences before notifying, unless you meant it should go to alerting instead?

jgao54: if the TCP handshake is failing with ECONNRESET repeatedly under some scenarios, then this makes sense 👍

PR #3970, opened 2026-02-20
Switch to rows_synced for InitialLoadSummary and TotalRowsSyncedByMirror

ilidemi:
BigQuery doesn't populate rows_in_partition and the rows synced should be actually rows synced.

Ensured tests for both paths for four data sources.


jkaflik (approved):
Nice! Thank you.

PR #3972, opened 2026-02-20
reject pipe with empty tuple during validation

jgao54:
In ClickHouse 25.12, `ORDER BY TUPLE()` is no longer supported. This has never worked properly in the past because snapshot would end up collapsing into a single row when PK is not specified in source database. So this PR will throw a validation error when order by empty tuple is detected. 

The scenarios where empty tuple is allowed:
- NullEngine: since it just forwards the data to MV without persisting them (the empty tuple error only applies to MergeTree)
- When `allow_suspicious_primary_key` is enabled

### DIFF
--- a/flow/e2e/api_test.go
+++ b/flow/e2e/api_test.go
@@ -335,6 +335,11 @@ func (s APITestSuite) TestClickHouseMirrorValidation_NoPrimaryKey() {
 			expectErr: isAtLeast25_12,
 		},
 		{
+			name:     "MergeTree engine allows empty ordering",
+			dstTable: "no_pkey_mt",
+			engine:   protos.TableEngine_CH_ENGINE_MERGE_TREE,
+		},
+		{
 			name:     "Null engine allows empty ordering",
 			dstTable: "no_pkey_null",
 			engine:   protos.TableEngine_CH_ENGINE_NULL,


flow/connectors/clickhouse/validate.go
```diff
@@ -74,10 +74,24 @@ func (c *ClickHouseConnector) ValidateMirrorDestination(
...
+		if !ok {
 			// if destination table is not a key, that means source table was not a key in the original schema mapping(?)
 			return fmt.Errorf("source table %s not found in schema mapping", tableMapping.SourceTableIdentifier)
 		}
+
+		hasOrderingKeys := len(processedSchema.PrimaryKeyColumns) > 0 ||
+			slices.ContainsFunc(tableMapping.Columns, func(col *protos.ColumnSetting) bool {
+				return col.Ordering > 0
+			})
+		if !hasOrderingKeys && tableMapping.Engine != protos.TableEngine_CH_ENGINE_NULL {
```
ilidemi: `&& tableMapping.Engine != protos.TableEngine_CH_ENGINE_MERGE_TREE`
(if we're following the [breaking change](https://github.com/ClickHouse/ClickHouse/pull/91569))

flow/pkg/clickhouse/validation.go
```diff
@@ -140,3 +141,23 @@ func storeColumnInfoForTable(ctx context.Context, logger log.Logger, conn clickh
...
+func CheckEmptyOrderingKeySupported(ctx context.Context, logger log.Logger, conn clickhouse.Conn,
+	chVersion *chproto.Version, sourceTable string,
+) error {
+	if chVersion == nil || !chproto.CheckMinVersion(chproto.Version{Major: 25, Minor: 12, Patch: 0}, *chVersion) {
+		return nil
+	}
+	var settingVal string
+	if err := QueryRow(ctx, logger, conn,
+		"SELECT value FROM system.settings WHERE name = 'allow_suspicious_primary_key'",
+	).Scan(&settingVal); err == nil && settingVal == "1" {
```
ilidemi: A network error here would result in a confusing message

flow/pkg/clickhouse/validation.go
```diff
@@ -140,3 +141,23 @@ func storeColumnInfoForTable(ctx context.Context, logger log.Logger, conn clickh
...
+		return nil
+	}
+
+	return fmt.Errorf(
+		"source table %s has no primary key columns; add a primary key to the source table, "+
```
ilidemi: This function doesn't really know about primary keys on the source, it only checks a setting? It also implicitly depends on the caller validating the dest sorting keys and table engine

flow/connectors/clickhouse/validate.go
```diff
@@ -74,10 +74,24 @@ func (c *ClickHouseConnector) ValidateMirrorDestination(
...
+		processedSchema, ok := processedMapping[dstTableName]
+		if !ok {
 			// if destination table is not a key, that means source table was not a key in the original schema mapping(?)
 			return fmt.Errorf("source table %s not found in schema mapping", tableMapping.SourceTableIdentifier)
 		}
+
+		hasOrderingKeys := len(processedSchema.PrimaryKeyColumns) > 0 ||
+			slices.ContainsFunc(tableMapping.Columns, func(col *protos.ColumnSetting) bool {
+				return col.Ordering > 0
+			})
```
ilidemi, line 86: What would the behavior be for missing primary key, missing sorting keys but CDC on and replica identity full? As I understand it, setup flow would just take all columns as src primary key + dest sorting key in this case, and this change would validate away that scenario (but I might be wrong). Not sure if there's a way to unify all that logic somehow, possible there's not.

flow/connectors/clickhouse/validate.go
jgao54: interesting, good find

flow/pkg/clickhouse/validation.go
jgao54: wanted to have a user-facing error that is useful, but yeah will simplify and make less assumptions / recommendations

flow/connectors/clickhouse/validate.go
jgao54: currently if table is replica identity full, we populate pk with all the columns:
```
	// if we have no pkey, we will use all columns as the pkey for the MERGE statement
	if replicaIdentityType == ReplicaIdentityFull && len(pKeyCols) == 0 {
		pKeyCols = columnNames
	}
```

so `hasOrderingKeys` would always return true.

jgao54: added a test case for replica identity full

jgao54: ah, i realized you mean that this may not _yet_ be populated during validation -- checking

jgao54: okay, yeah it does get populated during validation -- because `ValidateCDCMirror` calls `  srcConn.GetTableSchema()` before it calls `dstConn.ValidateMirrorDestination` which runs the above logic.

flow/e2e/api_test.go
```diff
@@ -303,6 +304,104 @@ func (s APITestSuite) TestClickHouseMirrorValidation_Pass() {
...
+			name:      "ReplacingMergeTree rejects empty ordering on 25.12+",
+			dstTable:  "no_pkey_rmt",
+			engine:    protos.TableEngine_CH_ENGINE_REPLACING_MERGE_TREE,
+			expectErr: isAtLeast25_12,
+		},
+		{
+			name:     "Null engine allows empty ordering",
+			dstTable: "no_pkey_null",
+			engine:   protos.TableEngine_CH_ENGINE_NULL,
+		},
```
ilidemi, line 346:
```suggestion
		{
			name:     "MergeTree engine allows empty ordering",
			dstTable: "no_pkey_mt",
			engine:   protos.TableEngine_CH_ENGINE_MERGE_TREE,
		},
		{
			name:     "Null engine allows empty ordering",
			dstTable: "no_pkey_null",
			engine:   protos.TableEngine_CH_ENGINE_NULL,
		},
```

ilidemi (approved):

PR #3977, opened 2026-02-21
Export OTel traces from Temporal

ilidemi:
First step towards having traces, exporting them from Temporal is basically free and we have a lot of interesting activity there that can be O11d.

- Wire up TracerProvider with OTLP gRPC exporter in flow-api, flow-worker, and snapshot-worker
- Instrument Temporal worker client with an interceptor
- Gated behind `ENABLE_OTEL_TRACES` env var, off by default

### DIFF
--- a/flow/cmd/snapshot_worker.go
+++ b/flow/cmd/snapshot_worker.go
@@ -56,9 +56,11 @@ func SnapshotWorkerMain(ctx context.Context, opts *SnapshotWorkerOptions) (*Work
 		return nil, fmt.Errorf("unable to setup tracer provider: %w", err)
 	}
 	if opts.EnableOtelTraces {
-		tracingInterceptor, tracingErr := temporalotel.NewTracingInterceptor(temporalotel.TracerOptions{})
-		if tracingErr != nil {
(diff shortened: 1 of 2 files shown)


flow/cmd/api.go
```diff
@@ -38,6 +38,7 @@ type APIServerParams struct {
 	Port              uint16
 	GatewayPort       uint16
 	EnableOtelMetrics bool
+	EnableOtelTraces  bool
```
iamKunalGupta, line 41: we have a request id interceptor in API which pushes the id in the context, maybe the interceptor for temporal client can re-use that request id as trace/span id for better stitching

ilidemi: I want to do the APIs as an e2e project, just so that it all works together when we go for it and doesn't produce noise in the meantime (wouldn't want to parent all workflows under the originating request id for example, some places should be parents and some links)

Rough overall plan (if I'll have time for everything):
1. Infra, temporal as-is
2. Span events for snapshot partitions, spans for CDC batches
3. Clickpipes temporal
4. E2E request tracing in clickpipes

PR #3978, opened 2026-02-21
Switchboard - a Go tool to query peers (Postgres)

ilidemi:
Same kind of admin tool as nexus wrt diagnosing peers, exposed via flow-api. Differences:
* Go - no Rust skill required, allows us to cut down on compile times in the future
* No need for separate SSH or weird auth reimplementation, if it's supported in QRep/CDC it's supported here
* Sticks to string in strings out, no SQL parsing - easy to add more connectors with other dialects or NoSQL
* Access security is exactly the same, plus disabled by default
* Potentially damaging queries are restricted via RO transaction and token-level parsing - a mistyped or wrongly pasted query would be rejected but not expecting to stop an attacker. Main obstacle to tighter detection is lack of a native postgres dialect parser library without CGO
* Integration tests are pretty thorough and use real psql

Readme is designed to be self-documenting so that LLMs can maintain it. 36% of lines are tests.
Will be total of three stacked PRs - Postgres (this one), MySQL, MongoDB.

### DIFF
--- a/flow/e2e/switchboard_pg_test.go
+++ b/flow/e2e/switchboard_pg_test.go
@@ -971,6 +971,6 @@ func (s SwitchboardPostgresSuite) Test_Catalog_Writable() {
 	require.NoError(s.t, err)
 	require.Equal(s.t, "hello", output)
 
-	err = s.psqlExecRW(fmt.Sprintf("DROP TABLE %s", tableName))
+	err = s.psqlExecRW("DROP TABLE " + tableName)
 	require.NoError(s.t, err, "DROP TABLE should succeed on catalog")
 }
--- a/flow/switchboard/upstream_postgres.go
+++ b/flow/switchboard/upstream_postgres.go
@@ -60,7 +60,9 @@ type PostgresUpstream struct {
 }
 
 // NewPostgresUpstream creates a new PostgreSQL upstream connection
-func NewPostgresUpstream(ctx context.Context, config *protos.PostgresConfig, queryTimeout time.Duration, readOnly bool) (*PostgresUpstream, error) {
+func NewPostgresUpstream(
+	ctx context.Context, config *protos.PostgresConfig, queryTimeout time.Duration, readOnly bool,
+) (*PostgresUpstream, error) {
 	conn, err := connpostgres.NewPostgresConnector(ctx, nil, config)
 	if err != nil {
 		return nil, err


flow/switchboard/upstream_postgres.go
```diff
@@ -0,0 +1,268 @@
...
+			ConstraintName:   pgErr.ConstraintName,
+			File:             pgErr.File,
+			Line:             pgErr.Line,
+			Routine:          pgErr.Routine,
+		}}
+	}
+	// Non-PgError: wrap as internal error
+	return &UpstreamError{Resp: &pgproto3.ErrorResponse{
+		Severity: "ERROR",
+		Code:     "XX000",
```
jgao54, line 49: nit: to differentiate postgres's known internal error with "XX000" from other errors, maybe better to have code `UNKNOWN` here

flow/switchboard/upstream_postgres.go
```diff
@@ -0,0 +1,268 @@
...
+	lower := strings.ToLower(query)
+	if strings.Contains(lower, "default_transaction_read_only") {
+		return errors.New("cannot modify read-only mode")
+	}
+	if strings.Contains(lower, "set_config") {
+		return errors.New("set_config is not allowed")
+	}
+
+	// Check each statement against blocked commands using pgsplit
+	statements, err := pgsplit.SplitStatements(query)
```
jgao54, line 197: pgsplit doesn't look maintained so I'm a bit wary here: https://github.com/nickbruun/pgsplit

Alternative (and more robust) library we could use here is https://github.com/pganalyze/pg_query_go's `pg_query.Parse`, which returns typed statement nodes, and which we could support an allowlist rather than a blocklist. e.g.

```
	result, err := pg_query.Parse(query)
	for _, stmt := range result.Stmts {
		switch stmt.Stmt.Node.(type) {
		case *pg_query.Node_SelectStmt:
            // SELECT by itself isn't necessarily read-only, but safe in combination with `default_transaction_read_only = on`
		case *pg_query.Node_ExplainStmt: 
            // read-only
		case *pg_query.Node_VariableShowStmt:
            // read-only
		default:
			return fmt.Errorf("statement not allowed: %T", stmt.Stmt.Node)
		}
	}
```

List above is conservative. If we discover more useful commands, they can be explicitly add it here and it makes it clear what commands are supported.

flow/switchboard/upstream_postgres.go
```diff
@@ -0,0 +1,268 @@
...
+
+// PostgresUpstream implements Upstream for PostgreSQL databases
+type PostgresUpstream struct {
+	conn   *connpostgres.PostgresConnector
+	pid    uint32
+	secret uint32
+}
+
+// postgresBlockedCommands are statements that are always denied for PostgreSQL
+var postgresBlockedCommands = map[string]string{
```
jgao54: I have a separate comment below advocating for allowlist over blocklist. so this comment is less relevant, but as a case in point I think it is worth to be defensive on top of `default_transaction_read_only = on` and also block commands like: CREATE, UPDATE, DELETE, DROP, TRUNCATE, GRANT, REVOKE, ALTER, etc. The more defensive here the better given the risk/consequences is quite high

flow/switchboard/session.go
```diff
@@ -0,0 +1,477 @@
...
+// parseStartupOptions parses PostgreSQL options string (-c key=value format)
+// and extracts peerdb.* parameters
+func parseStartupOptions(options string) map[string]string {
+	result := make(map[string]string)
+	if options == "" {
+		return result
+	}
+
+	// Split on -c, handling the format: -c key=value -c key2=value2
+	parts := strings.Split(options, "-c")
```
jgao54: " -c " just in case  `-c` is part of key/value

flow/switchboard/session.go
```diff
@@ -0,0 +1,477 @@
...
+
+		// Split on first = sign
+		if idx := strings.Index(part, "="); idx > 0 {
+			key := strings.TrimSpace(part[:idx])
+			value := strings.TrimSpace(part[idx+1:])
+			result[key] = value
+		}
+	}
+
+	return result
```
jgao54, line 62:
```suggestion
func parseStartupOptions(options string) map[string]string {
	result := make(map[string]string)
	fields := strings.Fields(options)
	for i := 0; i < len(fields); i++ {
		if fields[i] == "-c" && i+1 < len(fields) {
			i++
			if key, value, ok := strings.Cut(fields[i], "="); ok {
				result[key] = value
			}
		}
	}
	return result
}
```

flow/switchboard/session.go
```diff
@@ -0,0 +1,477 @@
...
+	startTime := time.Now()
+
+	// Handle empty queries as no-op (return EmptyQueryResponse)
+	trimmed := strings.TrimSpace(query)
+	if trimmed == "" || strings.Trim(trimmed, ";") == "" {
+		if err := writeBackendMessage(s.clientConn, &pgproto3.EmptyQueryResponse{}, s.writeTimeout); err != nil {
+			return err
+		}
+		return writeReadyForQuery(s.clientConn, s.txStatus(), s.writeTimeout)
+	}
```
jgao54, line 296: can be implemented inside CheckQuery:

```
func CheckQuery(query string) {
	if len(query) == 0 {
		return nil
	}

	result, err := pg_query.Parse(query) // see comment below on why `pg_query.Parse`
	if err != nil {
		return fmt.Errorf("failed to parse SQL: %w", err)
	}

	if len(result.Stmts) == 0 {
		return nil
	}
    ...
}
```

flow/e2e/switchboard_pg_test.go
```diff
@@ -0,0 +1,922 @@
...
+
+	"github.com/jackc/pgx/v5"
+	"github.com/stretchr/testify/require"
+
+	"github.com/PeerDB-io/peerdb/flow/e2eshared"
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/shared"
+)
+
+type SwitchboardPostgresSuite struct {
```
jgao54, line 20: I wasn't able to find tests for `CheckQuery`, would be useful to add some unit tests here to make sure they are blocked (or just `s.psql` to test general queries that triggers error in `CheckQuery`)

flow/e2e/switchboard_pg_test.go
```diff
@@ -0,0 +1,922 @@
...
+	t      *testing.T
+	peer   *protos.Peer
+	suffix string
```
jgao54, line 23: nit: suffix can be removed (not used)

flow/switchboard/upstream_postgres.go
ilidemi: We would want to use a robust PG SQL parser if one is available but pg_query_go brings in CGO which we don't want. To my knowledge, pgsplit is the best option available and all it does is splitting on semicolons outside strings/comments. I wouldn't expect that part to change much.

flow/e2e/switchboard_pg_test.go
ilidemi: Are these the ones you're thinking of? https://github.com/PeerDB-io/peerdb/blob/41969732439e9f855fc8c73fdf8868063163bb6a/flow/e2e/switchboard_pg_test.go#L474-L478

flow/switchboard/upstream_postgres.go
jgao54: For our use case I agree that splitting + string matching is probably sufficient, as it's not intended to be a full-feature parser. And not wanting CGO is a good call (we have a couple other deps that uses cgo today, `CGO_ENABLED=0  go build .` does not work at the moment. maybe there is a path where we could get them replaced as well so we can enforce no cgo in CI to avoid accidentally introducing more cgo libraries, but yeah that's a separate topic)

jgao54: and would still advocate for using allowlist over blocklist. wdyt?

flow/e2e/switchboard_pg_test.go
jgao54: ah yep! missed this during the review

flow/switchboard/upstream_postgres.go
ilidemi: Did not realize we have a bunch of CGo dependencies already. Tested the build time for pg_query_go, it would add 20 seconds and the dep itself is 400KLoc, so string matching it is. Allowlist sounds good!

flow/switchboard/session.go
ilidemi: CheckQuery returning nil means the query is good to execute?

flow/switchboard/upstream_postgres.go
ilidemi: Would prefer to match postgres behavior as much as possible, `XX000` is a catch-all internal error there

PR #3979, opened 2026-02-21
Switchboard - a Go tool to query peers (MySQL)

ilidemi:
Stacked on top of #3978

No surprises here, a straightforward implementation of the interface plus tests. Query guarding is done via a RO transaction + running the query through a parser and filtering statements. 60% of lines are tests.


.github/workflows/flow.yml
```diff
@@ -2,11 +2,11 @@ name: Flow build and test
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, switchboard-pg]
   push:
     branches: [main]
   pull_request_target:
-    branches: [main]
+    branches: [main, switchboard-pg]
```
ilidemi: Undo before merging

.github/workflows/dev-docker.yml
```diff
@@ -2,11 +2,11 @@ name: Dev Docker Images
...
 on:
   push:
-    branches: [main]
+    branches: [main, switchboard-pg]
   pull_request:
-    branches: [main]
+    branches: [main, switchboard-pg]
   pull_request_target:
-    branches: [main]
+    branches: [main, switchboard-pg]
```
ilidemi: Undo before merging

.github/workflows/codeql-analysis.yml
```diff
@@ -2,9 +2,9 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "main", "customer-*", "stable" ]
+    branches: [ "main", "customer-*", "stable", "switchboard-pg" ]
   pull_request:
-    branches: [ "main", "customer-*", "stable" ]
+    branches: [ "main", "customer-*", "stable", "switchboard-pg" ]
```
ilidemi: Undo before merging

.github/workflows/golang-lint.yml
```diff
@@ -2,7 +2,7 @@ name: GolangCI-Lint
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, switchboard-pg]
```
ilidemi: Undo before merging

.github/workflows/flow.yml
ilidemi: We could have a convention of `stacked-*` that also trigger all the checks. Not in this PR though

flow/switchboard/upstream_mysql.go
```diff
@@ -0,0 +1,543 @@
...
+)
+
+// wrapMySQLError converts a MySQL error to an UpstreamError
+func wrapMySQLError(err error) error {
+	if err == nil {
+		return nil
+	}
+	var mysqlErr *mysql.MyError
+	if errors.As(err, &mysqlErr) {
+		return &UpstreamError{Resp: &pgproto3.ErrorResponse{
```
jgao54, line 28: nit: is it intended that we are using pgproto3 error response here for mysql?

flow/switchboard/upstream_mysql.go
```diff
@@ -0,0 +1,543 @@
...
+	for _, stmt := range stmts {
+		if err := checkMySQLStatement(stmt); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// checkMySQLStatement validates a single statement against the allowlist
+func checkMySQLStatement(stmt ast.StmtNode) error {
```
jgao54, line 194: did not see this when i was reviewing PG, but yeah this one is using allowlist (with blocklist within SELECT which matches what i had in mind) :)

flow/switchboard/upstream_mysql.go
```diff
@@ -0,0 +1,543 @@
...
+			return err
+		}
+	}
+
+	// Check WHERE clause for subqueries
+	if s.Where != nil {
+		if err := checkExprForSubqueries(s.Where); err != nil {
+			return err
+		}
+	}
```
jgao54, line 287: nit:  just in case you want to make this more thorough, can also check `s.OrderBy`, `s.GroupBy`, `s.Having`, etc. But could be overkill here so leaving as is is fine as well

flow/switchboard/upstream_mysql.go
```diff
@@ -0,0 +1,543 @@
...
+		config:       config,
+		connectionID: uint32(connectionID),
+		pid:          pid,
+		secret:       secret,
+	}, nil
+}
+
+// Exec executes a query and returns results for streaming
+func (u *MySQLUpstream) Exec(ctx context.Context, query string) (ResultIterator, error) {
+	result, err := u.conn.Execute(ctx, query)
```
jgao54: Our mysql connector's `Execute` method has built-in retry. on connection retry this will erase existing settings like `SET SESSION TRANSACTION READ ONLY` and session timeout settings (because of `c.conn.CompareAndSwap(conn, nil)`), and the connectionId recorded in `MySQLUpstream` would become stale. So to be safe here we'd want to disable retry for switchboard.

ilidemi: Awesome catch!

flow/switchboard/upstream_mysql.go
ilidemi: Yes, we're pretending to be a postgres to the psql client

PR #3980, opened 2026-02-21
Switchboard - a Go tool to query peers (MongoDB)

ilidemi:
Stacked on top of #3979 and #3978

Queries are Extended JSON, psql still treats those as strings and they can be turned into BSON by mongo driver. 
Has a help command that lists everything supported and links out to the docs.


.github/workflows/codeql-analysis.yml
```diff
@@ -2,9 +2,9 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "main", "customer-*", "stable", "switchboard-pg" ]
+    branches: [ "main", "customer-*", "stable", "switchboard-pg", "switchboard-mysql" ]
   pull_request:
-    branches: [ "main", "customer-*", "stable", "switchboard-pg" ]
+    branches: [ "main", "customer-*", "stable", "switchboard-pg", "switchboard-mysql" ]
   schedule:
```
ilidemi, line 8: Undo before merging

.github/workflows/dev-docker.yml
```diff
@@ -2,11 +2,11 @@ name: Dev Docker Images
...
 on:
   push:
-    branches: [main, switchboard-pg]
+    branches: [main, switchboard-pg, switchboard-mysql]
   pull_request:
-    branches: [main, switchboard-pg]
+    branches: [main, switchboard-pg, switchboard-mysql]
   pull_request_target:
-    branches: [main, switchboard-pg]
+    branches: [main, switchboard-pg, switchboard-mysql]
```
ilidemi: Undo before merging

.github/workflows/flow.yml
```diff
@@ -2,11 +2,11 @@ name: Flow build and test
 
 on:
   pull_request:
-    branches: [main, switchboard-pg]
+    branches: [main, switchboard-pg, switchboard-mysql]
   push:
     branches: [main]
   pull_request_target:
-    branches: [main, switchboard-pg]
+    branches: [main, switchboard-pg, switchboard-mysql]
```
ilidemi: Undo before merging

.github/workflows/golang-lint.yml
```diff
@@ -2,7 +2,7 @@ name: GolangCI-Lint
 
 on:
   pull_request:
-    branches: [main, switchboard-pg]
+    branches: [main, switchboard-pg, switchboard-mysql]
```
ilidemi: Undo before merging

flow/switchboard/mongosh/command/allowlist.go
```diff
@@ -0,0 +1,163 @@
...
+// AllowedCommand defines an allowed MongoDB wire protocol command.
+type AllowedCommand struct {
+	Name            string
+	Required        []string
+	Optional        []string
+	SupportsComment bool // whether the command supports the comment field for cancel tracking
+}
+
+// AllowedCommands defines allowed MongoDB commands by category.
+var AllowedCommands = map[string][]AllowedCommand{
```
jgao54: hopefully user is set up with read-only role per our guidelines (e.g. readAnyDatabase) which would be even stricter than pg's `default_transaction_read_only`. that said, great to have this allowlist in case more permissions were granted to the user 👍

flow/switchboard/mongosh/README.md
```diff
@@ -0,0 +1,190 @@
+# mongosh — MongoDB Shell Compiler
+
+Compiles mongosh-style JavaScript statements into BSON wire commands for execution via the MongoDB Go driver. This is not a JavaScript runtime — it parses the AST statically, evaluates only literal expressions, and produces a `bson.D` command document that `upstream_mongodb.go` sends with `RunCommand` or `RunCommandCursor`.
```
jgao54: General question: wondering if you considered using raw json as input vs. mongosh-style javascript statement. 

For example, instead of supporting:
```
db.runCommand({ find: "users", filter: { active: true } })
```
we just support json:
```
{"find": "users", "filter": {"active": true}}
```
and then parse the json into bson.D with `bson.UnmarshalExtJSON`, validate the bson object (or validate the json before unmarshalling), and call `RunCommand` to execute the command.

Benefit here is simplicity, and should be able to achieve feature parity for our use cases today.

ilidemi: I gave it some consideration but not a ton. The current thinking is, documentation online would mostly focus on mongosh style queries, this tool would only be used once in a while so investment in learning quirks is not too justified for every on-call operator, and we have an example of the current tool going through lack of adoption because it's cumbersome to deal with. I'd rather make sure the UX is good, it's there when we need it, and LLMs can maintain it without us having to get into much context. IMO the current implementation is one-and-done kind of thing with minimal changes in the future, and if this proves not to be the case, we can delete that code in a self-contained way.

jgao54: my 2c here is that the UX isn't worse in JSON land, in fact `runCommand` is pretty well-documented in MongoDB Atlas [[example](https://www.mongodb.com/docs/manual/reference/command/find/#syntax)]), and json would be simpler to maintain. LLM does make maintenance easier, but less code means les surface area for bugs as well.

jgao54: if you feel there is a notable UX degradation, or if you just want to experiment with this approach first and delete if it becomes unmaintainable, that's fine as well.

ilidemi: I see how `find` command removes a lot of complexity with parsing chained statements. What do you think of the other pieces? E.g. without object constructors we'd have to type out `{"$timestamp": {"t": <t>, "i": <i>}}` for timestamps, would you see a way around that with EJSON?

jgao54: in practice this may not be an issue because [timestamp](https://www.mongodb.com/docs/manual/reference/bson-types/#timestamps) is mongodb internal data structure and i have yet had the need to query it directly in the past.

As for chained statements, i find json to be relatively straight-foward as well, for example, instead of
`db.find(..).filter(..).sort(..).limit(..)`
it would just be
`{"find": .., "filter":..,  "sort": ..., "limit":, ..}`
but definitely let me know if there are things I am not considering.

Edit: oh ignore the chained command reply, i think i mis-interpreted your statement. you are saying json for find removes the complexity of the parser, which i agree

ilidemi: Claude is flagging these, I am not sure how important they are:
```
  - findOne sugar gone — becomes {find: "...", filter: {...}, limit: 1, singleBatch: true}
  - explain() wrapping gone — becomes {explain: {find: "...", filter: {...}}} (which is actually how MongoDB does it)
  - countDocuments gone — becomes an aggregate pipeline
  - Less "friendly" for users used to mongosh, more natural for users who know the wire protocol
 ```
If you think a simpler proposal covers our needs, could you sketch out what the UX looks like for all the operations the operator might need? I want to make sure we learn from the issues with the current tool.

jgao54: Personally i haven't felt the need to run complex upstream queries in the past when working with the MongoDB connector. hence i was trying to lead us towards a simpler implementation to start, and then fallback on a more complex parser if it becomes necessary. 

if you feel strongly about going with the opposite direction, start with a parser, and reduce if it becomes too complex, that works as well.

ilidemi: > timestamp is mongodb internal data structure and i have yet had the need to query it directly in the past.

Regular expressions, binary data, dates before 1970 also fall in this bucket.

jgao54: these are still edge cases imo, and my gut feeling is that they are not worth the trade-off of the code complexity upfront without clear use cases. so I am leaning towards simplicity to start. But as mentioned, will let you decide.

ilidemi: Extended JSON is a lot simpler. Would appreciate it if you can give it another look, as the code is all different now.
UX concerns can be left in the past 👍

Issue #3951, opened 2026-02-17
MySQL: support NULL values in a custom partitioning column


ilidemi:
The only way to do partitioning in MySQL is via a custom column, where we do MIN/MAX and range queries. NULL values circumvent all of those and get skipped. If the column type is nullable, add a special partition that queries just for that. We limit the count of partitions to 1000 but here it's ok if the total becomes 1001.

Bonus points for handling it in Postgres when the count of partitions is passed along with the custom column, so there it also defaults to min-max.

PR #3983, opened 2026-02-23
Fix/handle null partition for pg/mysql minmax and pg NTILE

dtunikov:
## Summary
The PR implements this issue - https://github.com/PeerDB-io/peerdb/issues/3951

- Add a special "null partition" to QRep partitioning so rows with NULL watermark column values are no longer silently skipped during initial snapshot
- Apply the fix to both MySQL (MinMax) and Postgres (NTILE + MinMax) partitioning strategies
- Add `NullPartitionRange` proto type and `base_query` field to `QRepConfig` for building the `WHERE col IS NULL` query cleanly

## Context

When QRep partitioning uses MIN/MAX and range queries (`col BETWEEN start AND end`) to split tables into chunks, rows with NULL in the watermark column are silently lost:
- `MIN()`/`MAX()` ignore NULLs
- Range comparisons (`col >= X AND col < Y`) exclude NULLs

This only matters when a custom partitioning key is set (not the default `ctid` on Postgres, which scans physical blocks and includes all rows).

## Changes

- **protos/flow.proto**: Added `NullPartitionRange` message, `null_range` in `PartitionRange` oneof, `base_query` in `QRepConfig`
- **flow/connectors/utils/partition.go**: Added `AddNullPartition()` to shared `PartitionHelper`
- **flow/connectors/mysql/qrep.go**: Append null partition in `GetQRepPartitions` when `InitialCopyOnly`, handle `NullRange` in `PullQRepRecords`, refactored MIN/MAX query building
- **flow/connectors/postgres/qrep_partition.go**: Added `addNullPartition` to `PartitionParams`, append null partition in both NTILE and MinMax strategies
- **flow/connectors/postgres/qrep.go**: Set `addNullPartition` from `InitialCopyOnly`, handle `NullRange` in `PullQRepRecords`
- **flow/workflows/snapshot_flow.go**: Extract `baseQuery` and set `BaseQuery` in QRepConfig
- **flow/connectors/postgres/qrep_partition_test.go**: Updated tests to cover nullable watermark column with NULL rows

## Design decisions

I decided that it would be easier to always have null partition (regardless of column nullability), it will be no-op anyways later during ingestion
But we can also add a check to partition creation code and make it optional depending on watermarkColumn type

- **Only for `InitialCopyOnly=true`**: Continuous QRep would re-replicate NULL rows every iteration since there's no way to track progress on them
- **`BaseQuery` proto field**: Avoids fragile string manipulation on the query template to build the `WHERE col IS NULL` query
- **Not centralized in FlowableActivity**: BigQuery uses `ObjectIdPartitionRange` and MongoDB uses `_id` (never null) — adding null partitions generically would break other connectors
- **CTID not affected**: Scans physical blocks, includes all rows regardless of NULL values

### DIFF
--- a/flow/connectors/postgres/qrep_partition_test.go
+++ b/flow/connectors/postgres/qrep_partition_test.go
@@ -197,7 +197,6 @@ func TestGetQRepPartitions(t *testing.T) {
 			assert.Len(t, got, expected)
 		})
 	}
-
 }
 
 func TestGetQRepPartitionsWithNulls(t *testing.T) {
@@ -254,7 +253,6 @@ func TestGetQRepPartitionsWithNulls(t *testing.T) {
 			assert.Len(t, got, tc.expectedNumPartitions)
 		})
 	}
-
 }
 
 // returns the number of rows inserted
--- a/ui/app/mirrors/create/helpers/common.ts
+++ b/ui/app/mirrors/create/helpers/common.ts
@@ -68,7 +68,7 @@ export const blankQRepSetting: QRepConfig = {
   watermarkTable: '',
   watermarkColumn: '',
   baseQuery: '',
-  watermarkColumnNullable: false,
+  addNullPartition: false,
   initialCopyOnly: false,
   maxParallelWorkers: 4,
   waitBetweenBatchesSeconds: 30,


flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -21,7 +21,6 @@ type testCase struct {
 	name                  string
 	config                *protos.QRepConfig
 	last                  *protos.QRepPartition
-	want                  []*protos.QRepPartition
```
dtunikov, line 24: this hasn't been used by any of test cases, so I removed it

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -354,6 +354,9 @@ func addPartitionToQRepRun(ctx context.Context, tx pgx.Tx, flowJobName string,
 		case *protos.PartitionRange_ObjectIdRange:
 			rangeStart = x.ObjectIdRange.Start
 			rangeEnd = x.ObjectIdRange.End
+		case *protos.PartitionRange_NullRange:
```
dtunikov, line 381: otherwise this monitoring insertion would fail
rangeStart and rangeEnd are non nullable fields
the other option would be to make them nullable with a migration, but I thought it would be too much for this issue

flow/connectors/mysql/qrep.go
```diff
@@ -60,22 +60,16 @@ func (c *MySqlConnector) GetQRepPartitions(
 		return nil, fmt.Errorf("failed to parse watermark table %s: %w", config.WatermarkTable, err)
 	}
 
-	var minmaxQuery string
+	minmaxQuery := fmt.Sprintf("SELECT MIN(`%[2]s`),MAX(`%[2]s`) FROM %[1]s",
```
dtunikov, line 60: simplified if/else statements here a bit as this condition "WHERE `%[2]s` > " has been moved below

flow/connectors/mysql/qrep.go
```diff
@@ -108,14 +102,17 @@ func (c *MySqlConnector) GetQRepPartitions(
 		case *protos.PartitionRange_TimestampRange:
 			time := lastRange.TimestampRange.End.AsTime()
 			minVal = "'" + time.Format("2006-01-02 15:04:05.999999") + "'"
+		case *protos.PartitionRange_NullRange:
+			// handling it just in case to keep the query correct, but this case should never happen because we only add null partition for InitialCopyOnly replication
+			minVal = ""
```
ilidemi: It's better to error out here to be explicit that this situation is not supposed to happen. As-is this branch is a no-op, as minVal is already an empty string and the switch is non-exhaustive.

flow/connectors/utils/monitoring/monitoring.go
ilidemi: DB migration is preferable here, they're reliable and easy to do. Want to do it?

flow/workflows/snapshot_flow.go
```diff
@@ -224,6 +225,7 @@ func (s *SnapshotFlowExecution) cloneTable(
 		SourceType:                 sourcePeerType,
 		DestinationName:            s.config.DestinationName,
 		Query:                      query,
+		BaseQuery:                  baseQuery,
 		WatermarkColumn:            mapping.PartitionKey,
```
ilidemi, line 242: We have the information on whether the watermark column is nullable here. Let's pass it too and be precise

flow/connectors/utils/partition.go
```diff
@@ -381,6 +388,8 @@ func (p *PartitionHelper) updatePartitionHelper(partition *protos.QRepPartition)
 			BlockNumber:  r.TidRange.End.BlockNumber,
 			OffsetNumber: uint16(r.TidRange.End.OffsetNumber),
 		}
+	case *protos.PartitionRange_NullRange:
+	// null partition is always last, no need to track prev start/end
```
ilidemi: If this code is never called with null, can return an explicit error here

docs/architecture.md
ilidemi, line 1: A welcome addition. Let's submit this as a separate PR.

flow/workflows/snapshot_flow.go
dtunikov: added it
though, it is going to be 'false' if QRep is triggered manually via CreateQRepFlow method..
should we care about that one too?

i think i have more generic question here - whether we want to keep feature parity between manually triggered QRep and the one triggered during initial cdc snapshot

but yeah
it's a bit messy that we have to expose things like BaseQuery and WatermarkColumnNullable on the API level (this seems more like an internal service logic that we shouldn't be exposed to the outside world)
but I think we can discuss it later once I have a bit more exposure to the source code and currently established patterns

flow/connectors/utils/monitoring/monitoring.go
dtunikov: yep, done the migration

volumes/docker-entrypoint-initdb.d/create-extensions.sql
```diff
@@ -0,0 +1,2 @@
+CREATE EXTENSION IF NOT EXISTS hstore;
```
dtunikov: otherwise e2e pg tests fail
i think it makes sense to keep these extensions in the repo instead of creating them manually every time you want to run e2e tests

flow/e2e/clickhouse.go
```diff
@@ -70,7 +78,7 @@ func (s ClickHouseSuite) Peer() *protos.Peer {
 			Type: protos.DBType_CLICKHOUSE,
 			Config: &protos.Peer_ClickhouseConfig{
 				ClickhouseConfig: &protos.ClickhouseConfig{
-					Host:       "localhost",
+					Host:       clickhouseHost(),
```
dtunikov, line 81: default behaviour stays the same, so it should still work on CI
but locally we want to configure CI_CLICKHOUSE_HOST=host.docker.internal 
check docker-compose-e2e.yml below to see how I ran clickhouse/mysql/mongo

docker-compose-dev.yml
```diff
@@ -41,7 +41,7 @@ x-flow-worker-env: &flow-worker-env
 services:
   catalog:
     container_name: catalog
-    image: postgres:18-alpine@sha256:035b9ab53cfa147d7202b61f5f7782b939ae815b7d6bc81c96b7b42ff1fca950
+    image: imresamu/postgis:18-3.5-alpine
```
dtunikov: this aligns dev pg image with the one used on CI:
```
    services:
      catalog:
        image: imresamu/postgis:${{ matrix.db-version.pg }}-3.5-alpine
```

again, otherwise e2e tests fail locally with postgis related errors

volumes/docker-entrypoint-initdb.d/create-extensions.sql
dtunikov: the only possible issue is that the same docker-entrypoint-initdb.d is used inside docker-compose.yml as well
so, these extensions will also get enabled for "prod" peerdb deployments (those that use docker compose approach)
i can move this script to a separate entrypoint if that is a problem..

flow/connectors/postgres/qrep_partition.go
```diff
@@ -68,6 +69,9 @@ func NTileBucketPartitioningFunc(ctx context.Context, pp PartitionParams) ([]*pr
 	if err := rows.Err(); err != nil {
 		return nil, fmt.Errorf("failed to read rows: %w", err)
 	}
+	if pp.addNullPartition {
+		partitionHelper.AddNullPartition()
+	}
```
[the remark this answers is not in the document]
dtunikov: extra NULL-ranged buckets are discarded inside AddPartition anyways
so, it shouldn't cause any issues

volumes/docker-entrypoint-initdb.d/create-extensions.sql
ilidemi: We only have a handful of tests that need the extensions (I think type tests), maybe the setup code for those would be the best place?

flow/connectors/mysql/qrep.go
```diff
@@ -107,14 +99,15 @@ func (c *MySqlConnector) GetQRepPartitions(
 		case *protos.PartitionRange_TimestampRange:
 			time := lastRange.TimestampRange.End.AsTime()
 			minVal = "'" + time.Format("2006-01-02 15:04:05.999999") + "'"
+		case *protos.PartitionRange_NullRange:
+			//nolint:lll // this case should never happen because we only add null partition for InitialCopyOnly replication (so there shouldn't be a resume scenario with null range),
```
ilidemi:
```suggestion
			// this case should never happen because we only add null partition for InitialCopyOnly replication
			// (so there shouldn't be a resume scenario with null range)
```

flow/connectors/mysql/qrep.go
```diff
@@ -163,6 +156,11 @@ func (c *MySqlConnector) GetQRepPartitions(
 		return nil, fmt.Errorf("failed to add partitions: %w", err)
 	}
 
+	//nolint:lll // add null values partition to the end, if nulls aren't present it will be an empty partition that gets skipped during replication
```
ilidemi:
```suggestion
	// add null values partition to the end, if nulls aren't present it will be an empty partition
	// that gets skipped during replication
```

flow/connectors/postgres/qrep_partition.go
```diff
@@ -98,6 +102,12 @@ func MinMaxRangePartitioningFunc(ctx context.Context, pp PartitionParams) ([]*pr
 	if err := partitionHelper.AddPartitionsWithRange(start, end, pp.numPartitions); err != nil {
 		return nil, fmt.Errorf("failed to add partitions: %w", err)
 	}
+
+	//nolint:lll // add null values partition to the end, if nulls aren't present it will be an empty partition that gets skipped during replication
```
ilidemi:
```suggestion
	// add null values partition to the end, if nulls aren't present it will be an empty partition
	// that gets skipped during replication
```

flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -17,11 +17,11 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared"
 )
 
+//nolint:govet
```
ilidemi:
```suggestion
//nolint:govet // field alignment not important
```

flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -217,16 +202,27 @@ func prepareTestData(t *testing.T, pool *pgx.Conn, schema string) int {
...
-		`, schema), times, tm)
+		`, schema), rowsCount, tm)
 		if err != nil {
 			t.Fatalf("Failed to insert test data: %v", err)
 		}
 	}
 
-	return times
+	// add some rows with null "from" value to ensure they get partitioned correctly as well
+	for i := range 12 {
```
ilidemi: We need to exercise non-null flows too. Let's make null generation optional and have the pre-existing tests skip it, then add a separate one or two for null behavior.

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -347,18 +347,18 @@ func addPartitionToQRepRun(ctx context.Context, tx pgx.Tx, flowJobName string,
...
 
-	var rangeStart, rangeEnd string
+	var rangeStart, rangeEnd *string
 	if partition.Range != nil {
 		switch x := partition.Range.Range.(type) {
 		case *protos.PartitionRange_IntRange:
-			rangeStart = strconv.FormatInt(x.IntRange.Start, 10)
-			rangeEnd = strconv.FormatInt(x.IntRange.End, 10)
+			s, e := strconv.FormatInt(x.IntRange.Start, 10), strconv.FormatInt(x.IntRange.End, 10)
+			rangeStart, rangeEnd = &s, &e
```
ilidemi: After #3946 can do 
```go
			rangeStart = new(strconv.FormatInt(x.IntRange.Start, 10))
			rangeEnd = new(strconv.FormatInt(x.IntRange.End, 10))
```

flow/workflows/snapshot_flow.go
ilidemi: I'd say it's fine to keep the same contract for internal and external (parent mirror name, version, flags are also internal). Can add a comment on all five that they're internal. We did a two-contracts + codegen setup for FlowConnectionConfigs in #3605 but there the purpose was removing fields in internal to stay type-safe, here it'd be overkill.

Re feature parity - we may actually want it down the line (for source connectors that don't have CDC at all) but that would need e2e design, including clickpipes, so it's fine to defer that one.

Can also rename WatermarkColumnNullable to AddNullPartition, this way it's non-confusingly false in CreateQRepFlow

flow/workflows/snapshot_flow.go
```diff
@@ -218,13 +219,28 @@ func (s *SnapshotFlowExecution) cloneTable(
 		}
 	}
 
+	watermarkColumnNullable := false
+	if mapping.PartitionKey != "" {
+		if err := initTableSchema(); err != nil {
```
ilidemi, line 224: cool to see there's already a pattern for not double-querying

flow/connectors/postgres/qrep_partition_test.go
dtunikov: done

flow/connectors/utils/monitoring/monitoring.go
dtunikov: done

flow/workflows/snapshot_flow.go
dtunikov: kk
renamed to AddNullPartition and put // internal comment for those fields

thanks for clarification!

ilidemi (approved):
Nice work!

PR #3994, opened 2026-02-27
o11y: small refactor on user-facing logs

jgao54:
Initial goal was just to add an attribute for user-facing logs  (i.e. slog.String("logAudience", "user")) for easy dashboard filtering. However the code in `flow/alerting/alerting.go` is a bit more confusing than I would like to see. Introduced some refactoring/renaming to clarify user-facing logging (insert into flow_errors), telemetry emission, and internal logging.

### DIFF
--- a/flow/alerting/alerting.go
+++ b/flow/alerting/alerting.go
@@ -460,8 +460,8 @@ func (a *Alerter) recordFlowErrorInternal(
 	err error,
 ) {
 	logger := internal.LoggerFromCtx(ctx)
-	errorClass, errInfo := GetErrorClass(ctx, err)
-	errWithStack := fmt.Sprintf("%+v", err)
+	errClass, errInfo := GetErrorClass(ctx, err)
+	errMessage := fmt.Sprintf("%+v", err)
 
 	// 1. Log the error internally
 	logFn := logger.Error
@@ -469,15 +469,14 @@ func (a *Alerter) recordFlowErrorInternal(
 		logFn = logger.Warn
 	}
 	logFn(err.Error(),
-		slog.Any("errorClass", errorClass),
+		slog.Any("errorClass", errClass),
 		slog.Any("errorInfo", errInfo),
-		slog.Any("stack", errWithStack),
+		slog.Any("errorMessage", errMessage),
 		slog.String("flowErrorType", errorType.String()),
-		slog.String("logAudience", "user"),
 	)
 
 	// 2. Insert log to flow_errors table
-	if err := InsertFlowLog(ctx, a.CatalogPool, flowName, errWithStack, errorType); err != nil {
+	if err := InsertFlowLog(ctx, a.CatalogPool, flowName, errMessage, errorType); err != nil {
 		logger.Error("failed to insert flow error", slog.Any("error", err))
 	}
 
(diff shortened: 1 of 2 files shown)


flow/alerting/alerting.go
```diff
@@ -526,48 +534,46 @@ func (a *Alerter) logFlowErrorInternal(
...
-	a.logFlowErrorInternal(ctx, flowName, flowErrorTypeError, inErr, logger.Error)
-	return inErr
-}
-
-func (a *Alerter) LogFlowWarning(ctx context.Context, flowName string, inErr error) {
-	logger := internal.LoggerFromCtx(ctx)
-	a.logFlowErrorInternal(ctx, flowName, flowErrorTypeWarn, inErr, logger.Warn)
-}
-
-func (a *Alerter) LogFlowEvent(ctx context.Context, flowName string, info string) {
```
jgao54, line 559: renamed to `EmitFlowInfoTelemetryEvent` above

flow/alerting/alerting.go
```diff
@@ -412,109 +411,118 @@ func (a *Alerter) sendTelemetryMessage(
...
-	if internal.PeerDBTelemetrySenderSendErrorAlertsEnabled(ctx) {
-		a.sendTelemetryMessage(ctx, logger, flowName, inErrWithStack, telemetry.ERROR, tags...)
+	// 1. Log the error internally
+	logFn := logger.Error
+	if errorType == FlowErrorTypeWarn {
+		logFn = logger.Warn
 	}
-	loggerFunc(fmt.Sprintf("Emitting error/warning metric: '%s'", errError),
-		slog.Any("error", inErr),
+	logFn(err.Error(),
```
jgao54, line 471: note that in the past this log shows up twice, once on line 459 (`loggerFunc(errError, slog.String("stack", inErrWithStack))` before this PR), and again here. 

combine them into a single log line to remove redundancy, and remove the `Emitting error/warning metric:` prefix which does not bring much value.

flow/alerting/alerting.go
```diff
@@ -412,109 +411,118 @@ func (a *Alerter) sendTelemetryMessage(
...
 	logger := internal.LoggerFromCtx(ctx)
-	inErrWithStack := fmt.Sprintf("%+v", inErr)
-	errError := inErr.Error()
-	loggerFunc(errError, slog.String("stack", inErrWithStack))
-	if _, err := a.CatalogPool.Exec(
-		ctx, "INSERT INTO peerdb_stats.flow_errors(flow_name,error_message,error_type) VALUES($1,$2,$3)",
-		flowName, inErrWithStack, errorType.String(),
-	); err != nil {
-		logger.Error("failed to insert flow error", slog.Any("error", err))
-		return
```
jgao54, line 465: i removed this silent early return here -- if we failed to insert to the database, it would still be good to include telemetry and metrics below.

flow/alerting/alerting.go
ilidemi: I would keep the prefix to make it obvious which errors are user-visible

jgao54: Added the "logAudience" attribute as a filter for this purpose (and plan to have a separate panel in dashboard for user-facing logs), or do you think there's other benefits in having this prefix in the message itself.

ilidemi: Pretty much just scanning through the logs. I'm leaning 80% towards keep but ok both ways, we can see if anyone gets confused about what's visible from the outside.

jgao54: updated PR to have a separate user-facing log line that is encapsulated in `InsertFlowLog`, making it easy to spot these logs consistently (and accepting the bit of log duplication here for internal errors)

PR #3995, opened 2026-02-27
o11y: add more table-level stats

jgao54:
Add table-level stats for
(1) whether it is using custom sort key 
(2) whether is replica identity full 
(3) inserts/updates/deletes stats

Using batched query to reduce network roundtrips

### DIFF
--- a/flow/shared/telemetry/activity_logging.go
+++ b/flow/shared/telemetry/activity_logging.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"slices"
 	"strings"
 
 	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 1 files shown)


flow/shared/telemetry/activity_logging.go
```diff
@@ -191,23 +192,35 @@ type FlowConfigForLogging struct {
...
+	TableName           string   `json:"table_name"`
+	DestTableName       string   `json:"destination_table_name"`
+	PartitionKey        string   `json:"partition_key"`
+	Engine              string   `json:"engine"`
+	Exclude             []string `json:"excluded_columns"`
+	ReplicaIdentityFull bool     `json:"replica_identity_full"`
+	UseCustomSortKey    bool     `json:"use_custom_sort_key"`
+	TotalInserts        int64    `json:"total_inserts"`
+	TotalUpdates        int64    `json:"total_updates"`
+	TotalDeletes        int64    `json:"total_deletes"`
```
ilidemi, line 204: What are we looking for with the aggregate counts? One would need to refresh the page and remember the previous numbers, or fiddle with the time window to get a sense of what's going on. Maybe better as a metric over time, if we need them?

jgao54: A few things that come to mind:
- it provides a better heuristics than individual batch's inserts/updates/deletes on table size and change event volume (since individual batches can be spikier depending on time-of-day)
- it can tell us whether the table is append-only vs contains updates. 
- similar to user-logging, it may be useful to see what the user see, since these aggregates show up in their dashboard (and i have actually recently had customer mentioning this as well)

Because these values are extracted on the hourly basis, if we just naively fetch the current/latest inserts,updates,deletes it does not capture everything over time, so may be less useful.  

You brought up a good point (iiuc) that it is worth adding o11y for per-batch metrics without logging into the peerdb UI, which would be a useful metric that get logged as it happen rather than hourly metrics here.

ilidemi: All good points, especially enabling the support to see what customer sees.

ilidemi (approved):

PR #3996, opened 2026-02-27
Classify aurora failover, remove temporal app errors from sync flow

ilidemi:
Aurora failover was wrapped in a temporal app error which made it hard to classify
We also don't need temporal app errors in sync flow anymore as their purpose was not to retry the activity and we're already never retrying the sync activity

* Aurora failover classified as recoverable
* Publication missing, slot missing, repl state desync moved to their own error types with same classification
* Connection to source down wasn't going through classification, fixed that and made sure that all notifications it'd send would be deduped with other network ones - no extra services would start to get notified, based on 30 days of logs

### DIFF
--- a/flow/alerting/classifier_test.go
+++ b/flow/alerting/classifier_test.go
@@ -842,7 +842,7 @@ func TestMySQLUnsupportedDDLShouldNotifyUser(t *testing.T) {
 	}, errInfo)
 }
 
-func TestPublicationMissingErrorShouldBePublicationMissing(t *testing.T) {
+func TestPublicationMissingError(t *testing.T) {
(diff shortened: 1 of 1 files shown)


flow/alerting/classifier_test.go
```diff
@@ -855,3 +841,38 @@ func TestMySQLUnsupportedDDLShouldNotifyUser(t *testing.T) {
 		},
 	}, errInfo)
 }
+
+func TestPublicationMissingErrorShouldBePublicationMissing(t *testing.T) {
```
jgao54: lol function name is a tautology (same as the one below), it's not wrong though

PR #3997, opened 2026-02-27
o11y: add logging to improve source peer observability

jgao54:
Add the following peer info for source peers that are being used by existing pipes:
- Database type 
- Database version
- Database variant
- Whether it uses custom tls
- whether it uses ssh tunnel

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -959,7 +959,7 @@ func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
 		}
 		for _, peer := range peers {
 			version, variant := a.getRuntimePeerInfo(ctx, peer)
-			telemetry.LogPeerConfig(ctx, peer, version, variant)
+			telemetry.LogPeerInfo(ctx, peer, version, variant)
 		}
 		return nil
 	}))()
--- a/flow/connectors/mongo/mongo.go
+++ b/flow/connectors/mongo/mongo.go
@@ -150,9 +150,9 @@ func (c *MongoConnector) GetVersion(ctx context.Context) (string, error) {
 }
 
 func (c *MongoConnector) GetDatabaseVariant(ctx context.Context) (protos.DatabaseVariant, error) {
-	// matches private link connections as well
+	// AtlasDomain matches private link connections as well
 	// https://www.mongodb.com/docs/atlas/security-private-endpoint/?cloud-provider=aws#private-endpoint-aware-connection-strings
-	if strings.Contains(c.config.Uri, ".mongodb.net") {
+	if strings.Contains(c.config.Uri, peerdb_mongo.AtlasDomain) {
 		return protos.DatabaseVariant_MONGODB_ATLAS, nil
 	}
 	ss, err := peerdb_mongo.GetServerStatus(ctx, c.client)
--- a/flow/shared/telemetry/activity_logging.go
+++ b/flow/shared/telemetry/activity_logging.go
@@ -351,7 +351,7 @@ func LogFlowConfigs(
 	return flowConfigs, nil
 }
 
-func LogPeerConfig(ctx context.Context, peer *protos.Peer, version string, variant string) {
+func LogPeerInfo(ctx context.Context, peer *protos.Peer, version string, variant string) {
 	logger := log.With(internal.LoggerFromCtx(ctx), slog.String("scheduledTask", "LogFlowConfigs"))
 	info := &sourcePeerInfoForLogging{
 		PeerName:        peer.Name,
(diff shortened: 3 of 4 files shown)


flow/connectors/mongo/mongo.go
```diff
@@ -148,6 +149,20 @@ func (c *MongoConnector) GetVersion(ctx context.Context) (string, error) {
 	return buildInfo.Version, nil
 }
 
+func (c *MongoConnector) GetDatabaseVariant(ctx context.Context) (protos.DatabaseVariant, error) {
+	if strings.Contains(c.config.Uri, ".mongodb.net") {
```
ilidemi: Will this support private link as well?

flow/activities/flowable.go
```diff
@@ -952,6 +952,37 @@ func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
 	return nil
 }
 
+func (a *FlowableActivity) populateRuntimePeerInfo(
+	ctx context.Context, peer *protos.Peer, info *telemetry.SourcePeerInfoForLogging,
+) {
+	logger := internal.LoggerFromCtx(ctx)
+
+	conn, err := connectors.GetConnector(ctx, nil, peer)
+	if err != nil {
```
ilidemi, line 980: Idea: populate fields with `N/A` or some sort of error code on errors, so that it's clear what's going on

flow/shared/telemetry/activity_logging.go
```diff
@@ -198,7 +199,26 @@ type TableMappingForLogging struct {
...
 }
 
-func LogFlowConfigs(ctx context.Context, catalogPool shared.CatalogPool) error {
+type SourcePeerInfoForLogging struct {
+	PeerName        string `json:"peer_name"`
+	PeerType        string `json:"peer_type"`
+	DatabaseVersion string `json:"database_version,omitempty"`
+	DatabaseVariant string `json:"database_variant,omitempty"`
+	CustomTLS       bool   `json:"custom_tls"`
+	SSHTunnel       bool   `json:"ssh_tunnel"`
```
ilidemi: Let's add host, break out (or replace SSH with) SSH host, and add db name where applicable

ilidemi: Postgres: auth type, possibly assume_role_arn and chained_role_arn if we're shooting for full user-provided payload
MySQL: compression, flavor, replication mechanism (we have AUTO here, could keep it or resolve from metadata_last_sync_state.last_text format), auth type, same as above for aws auth
Mongo: read preference?
BigQuery: @jkaflik what's helpful to display on a dashboard and not private?

Less certain about expanding custom TLS but probably wouldn't hurt

flow/activities/flowable.go
```diff
@@ -942,7 +942,7 @@ func (a *FlowableActivity) ScheduledTasks(ctx context.Context) error {
 		return nil
 	}))()
 	defer common.Interval(ctx, 1*time.Hour, wrapWithLog("LogFlowConfigs", func() error {
-		return telemetry.LogFlowConfigs(ctx, a.CatalogPool)
+		return telemetry.LogFlowConfigs(ctx, a.CatalogPool, connectors.LoadPeers, a.populateRuntimePeerInfo)
```
ilidemi: Have you considered a design without dependency injection? E.g. have telemetry.LogFlowConfigs return a list of peers, load peers and populate runtime info happen here in flowable.go, and call out to helpers from shared/telemetry as needed (for individual steps or assemble a big struct struct for telemetry.LogPeerConfigs)

flow/connectors/mongo/mongo.go
jgao54: Not quite sure. I would expect private link (from Atlas) would still use `.mongodb.net` host names.

jgao54: https://www.mongodb.com/docs/atlas/security-private-endpoint/?cloud-provider=aws#private-endpoint-aware-connection-strings seem to suggest this.

flow/activities/flowable.go
jgao54: great idea

flow/shared/telemetry/activity_logging.go
jgao54: i don't recall an issue where the role was needed for debugging, so leaning towards keeping it out for now. it contains account id which can be sensitive, and redacting adds more logic and can be added if the fields  become useful.

rest make sense, useful to have these additional information 👍

flow/connectors/mongo/mongo.go
ilidemi:
```suggestion
	// matches private link connections as well
	// https://www.mongodb.com/docs/atlas/security-private-endpoint/?cloud-provider=aws#private-endpoint-aware-connection-strings
	if strings.Contains(c.config.Uri, ".mongodb.net") {
```

PR #3998, opened 2026-02-27
Support partitioned table with ctid partitioning strategy

jgao54:
Adds CTID block partitioning support for PostgreSQL _partitioned_ tables during snapshot.

Previously, CTID partitioning only worked on regular tables because pg_relation_size() on a partitioned parent returns 0.

This PR:
- detects partitioned tables by checking `pg_class.relkind = 'p'` on the parent
- handles multi-level partitioning by recursively find out all the "leaf" partitioned tables
- generates snapshot partitions using a greedy block-packing algorithm that distributes child table blocks across computedNumPartitions, allowing even distribution.
- introduce ChildTableRange to QRepPartition to support multiple partition ranges in a single qrep partition
- at pull time, iterates over ChildTableRange entries, substituting the child table name into the
  query template for each range

Testing: add e2e tests for:
- single-level partitioned table 
- multi-level partitioned table
- empty partitioned tables
- tested with 10k partitioned tables across 1k partitions.

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -833,6 +833,15 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
+	// ErrClientDisconnected is only produced when the topology is explicitly closed via
+	// Client.Disconnect(); classify as internal to flag client lifecycle bugs in our code.
+	if errors.Is(err, mongo.ErrClientDisconnected) {
+		return ErrorInternal, ErrorInfo{
+			Source: ErrorSourceMongoDB,
+			Code:   "CLIENT_DISCONNECTED",
+		}
+	}
+
 	// MongoDB can leak error without properly encapsulate it into a pre-defined error type.
 	// Use string matching as a catch-all for poolClearedErrors. These errors occur when a
 	// connection pool is cleared due to another operation failure; they are always retryable
--- a/flow/alerting/classifier_test.go
+++ b/flow/alerting/classifier_test.go
@@ -921,6 +921,16 @@ func TestSlotMissingError(t *testing.T) {
 	}, errInfo)
 }
 
+func TestMongoClientDisconnectedShouldBeInternal(t *testing.T) {
+	err := fmt.Errorf("operation failed: %w", mongo.ErrClientDisconnected)
+	errorClass, errInfo := GetErrorClass(t.Context(), err)
+	assert.Equal(t, ErrorInternal, errorClass, "Unexpected error class")
+	assert.Equal(t, ErrorInfo{
+		Source: ErrorSourceMongoDB,
+		Code:   "CLIENT_DISCONNECTED",
+	}, errInfo, "Unexpected error info")
+}
+
 func TestAuroraFailoverRONodeShouldBeRecoverable(t *testing.T) {
 	pgErr := &pgconn.PgError{
 		Severity: "ERROR",


flow/connectors/postgres/qrep.go
```diff
@@ -421,6 +426,65 @@ func corePullQRepRecords(
...
+	}
+
+	baseQuery, err := BuildQuery(c.logger, config.Query)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	var totalRecords, totalBytes int64
+	for _, child := range partition.ChildTableRanges {
+		query := strings.ReplaceAll(baseQuery, config.WatermarkTable, child.Table)
```
ilidemi: String manipulation doesn't sound great here. Zooming all the way out, I think the intention behind passing the query as a string and not a template is to allow joins and such, which would also break with this approach if it's enabled by default. I'd go towards adding a structured query template to QRepConfig, passing it from the SnapshotFlow and ignoring string query in QRepFlow if it's passed.

Conversation in the [MySQL enums PR](https://github.com/PeerDB-io/peerdb/pull/4025#discussion_r2912071930) also went in that direction, would you be able to work out with @dtunikov what's a good struct for both cases?

flow/connectors/postgres/qrep_partition.go
```diff
@@ -178,6 +211,146 @@ func CTIDBlockPartitioningFunc(ctx context.Context, pp PartitionParams) ([]*prot
...
+		return false, fmt.Errorf("failed to check relkind for %s: %w", table, err)
+	}
+	return relkind == "p", nil
+}
+
+// getPartitionedTables returns a map of partitioned child table names to their block counts,
+// recursively handle multi-level partitioned tables.
+func getPartitionedTables(ctx context.Context, tx pgx.Tx, table string) (map[string]int64, error) {
+	rows, err := tx.Query(ctx, `
+		SELECT format('%I.%I', n.nspname, c.relname), c.relkind::text,
```
ilidemi: This would be pretty hard to find when we'll get on dots in table/schema names and attempt a sweep. Best I could come up with is doing `fmt.Sprintf("%s.%s")` on Go side so at least it's `%s.%s` everywhere

flow/connectors/postgres/qrep_partition.go
```diff
@@ -178,6 +211,146 @@ func CTIDBlockPartitioningFunc(ctx context.Context, pp PartitionParams) ([]*prot
...
+//   - partition 1: [T1:0-24]
+//   - partition 2: [T1:25-29, T2:0-19]
+//   - partition 3: [T3:0-9]
+func ctidPartitionsForPartitionedTable(
+	pp PartitionParams,
+	blocksPerTable map[string]int64,
+	totalChildBlocks int64,
+) ([]*protos.QRepPartition, error) {
+	sortedTables := slices.Sorted(maps.Keys(blocksPerTable))
+	blocksPerPartition := max(int64(1), (totalChildBlocks+pp.numPartitions-1)/pp.numPartitions) // ceiling division
```
ilidemi: shared.DivCeil

flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -209,6 +211,482 @@ func TestGetQRepPartitions(t *testing.T) {
...
+	defer tunnel.Close()
+
+	conn, err := NewPostgresConnFromConfig(t.Context(), config, "", nil, tunnel)
+	if err != nil {
+		t.Fatalf("Failed to create connection: %v", err)
+	}
+	defer conn.Close(t.Context())
+
+	//nolint:gosec
+	schemaName := fmt.Sprintf("test_%d", rand.Uint64())
```
ilidemi: Let's follow the other tests' convention `strings.ToLower(shared.RandomString(8))`

flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -209,6 +211,482 @@ func TestGetQRepPartitions(t *testing.T) {
...
+	for _, p := range partitions {
+		require.NotEmpty(t, p.ChildTableRanges, "partitions from partitioned table must have ChildTableRanges")
+		for _, ctr := range p.ChildTableRanges {
+			require.NotNil(t, ctr.Range, "each ChildTableRange must have a TID range")
+			require.NotEmpty(t, ctr.Table, "each ChildTableRange must name a child table")
+			childTableCounts[ctr.Table]++
+		}
+	}
+
+	// Every non-empty child should appear in at least one ChildTableRange
```
ilidemi: Can add an explicit assert on nothing in the root

flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -209,6 +211,482 @@ func TestGetQRepPartitions(t *testing.T) {
...
+	for _, p := range partitions {
+		require.NotEmpty(t, p.ChildTableRanges)
+		for _, ctr := range p.ChildTableRanges {
+			require.NotNil(t, ctr.Range)
+			require.NotEmpty(t, ctr.Table)
+			childTableCounts[ctr.Table]++
+		}
+	}
+
+	// All 6 leaf tables should appear, no mid-level tables
```
ilidemi: Can add an explicit assert on nothing in the root

flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -209,6 +211,482 @@ func TestGetQRepPartitions(t *testing.T) {
...
+func TestCtidPartitionsForPartitionedTableOffsetNumberBounds(t *testing.T) {
+	t.Parallel()
+	pp := PartitionParams{
+		numPartitions: 3,
+		logger:        log.NewStructuredLogger(slog.With(slog.String(string(shared.FlowNameKey), "testOffsetBounds"))),
+	}
+	leafBlocks := map[string]int64{
+		"public.t1": 100,
+		"public.t2": 50,
+		"public.t3": 10,
```
ilidemi, line 561: As we're passing a specific input and there's two-level iteration going on, let's assert a specific output

protos/flow.proto
```diff
@@ -418,10 +419,19 @@ message QRepConfig {
...
+}
+
 message QRepPartition {
   string partition_id = 2;
   PartitionRange range = 3;
   bool full_table_partition = 4;
+  // For partitioned tables with CTID partitioning: each entry specifies a
+  // child table and TID range to query instead of the parent watermark table.
+  // When populated, the top-level `range` field is unused.
+  repeated ChildTableRange child_table_ranges = 5;
```
ilidemi, line 441: When there are a lot of partitions (like thousands in that TimescaleDB case we had), this could overflow 2MB temporal input limit and fail at runtime. Ideas:
* Try to work with the limits - shorten the payload as much as possible (e.g. just table name for full child range), reasonably calculate that 99% of sane configurations wouldn't trip them, add a size estimator to clickpipes to fail early
* Just move this payload to catalog, like we'll do for table mappings when it's time to support many of them. Except here they don't even need to be versioned, just dump a proto, pass around the row id

flow/connectors/postgres/qrep_partition.go
```diff
@@ -178,6 +211,146 @@ func CTIDBlockPartitioningFunc(ctx context.Context, pp PartitionParams) ([]*prot
...
+	err := tx.QueryRow(ctx,
+		`SELECT c.relkind::text FROM pg_class c WHERE c.oid = to_regclass($1)`,
+		table).Scan(&relkind)
+	if err != nil {
+		if err == pgx.ErrNoRows {
+			return false, fmt.Errorf("table %s not found in pg_class", table)
+		}
+		return false, fmt.Errorf("failed to check relkind for %s: %w", table, err)
+	}
+	return relkind == "p", nil
```
ilidemi, line 302: We also need to support inherited tables (relkind = 'r', relhassubclass = true). They could have rows at the root/middle levels too. CDC for them is weird as well, we query pg_inherits there too in getChildToParentRelIDMap

flow/connectors/postgres/qrep_partition.go
jgao54: we can use "%s.%s" for now for consistency with the rest of codebase

flow/connectors/postgres/qrep_partition.go
jgao54: great catch!

I'll defer inherited tables as a follow-up and add appropriate e2e tests to not bloat this PR too much, and before we enabling ctid as default. Also, the current proto extension for QRepPartition should work nicely with inherit table as well since it has both PartitionRange and ChildPartitionRanges.

flow/connectors/postgres/qrep.go
jgao54: +1 on using structured query. 

for this PR, I can make this logic a bit safer by using regex to look for `... FROM <TABLE> WHERE...`, since we know the query shape in advance.  once @dtunikov  introduce the query object we can move this over as well.

jgao54: depending on when this is merged, will either update with StructuredQuery if it's committed, or merge as is for now. since we are not shipping this yet until inherited table is sorted.

protos/flow.proto
jgao54: A bit of napkin math, for 10k child tables: since we limit max of 1k qrep partitions, the theoretical max for for total number of ChildTableRange is ~11k (each child table produce at least one ChildTableRange totalling 10k ChildTableRange, and then 1k tables at partition boundaries totalling 1k ChildTableRange).

Assume each table name estimated at 50 bytes, TIDPartitionRange estimated at 16 bytes, that's ~66 bytes per ChildTableRange, which totals to 66*11000 is less than 800kb, plus the other fields in the input should still give us a good buffer under 2MB.

So seems okay to leave as is for now, but lmk if i mis-considered anything.

ilidemi: We're using JSON though. So the output from the unit test would be serialized as
```
[{"table": "public.t1", "range": {"start": {"blockNumber": 0}, "end": {"blockNumber": 19, "offsetNumber": 65535}}}]
```
which is a bit longer than 66 while not hitting the upper bounds yet. It would give me more confidence if we did a real test with cloud dev temporal with the upper bound numbers and see for sure if it goes through. There is also an event history limit at 4MB per workflow, and I wouldn't be able to tell off the top of my head how many times one payload will get duplicated.

Alternatively, have a plan on what to do when someone gets paged on the workflow stuck and if we're comfortable saying we don't support more partitions than that, vs just storing data outside of Temporal

jgao54: Good call on the JSON and the concern for event history.

Because the partition ranges are split across the workers, with default of 4 workers, each template in practice only stored 1/4 of the ranges.

I ran a local test with 10k tables, entry looks as follows:
```
{"table":"ts_test_schema_padding_aaaaaa.part_tbl_p00001", "range":{"start":{"blockNumber":338}, "end":{"blockNumber":387, "offsetNumber":65535}}},
```

Given the offset always start at 0 and end at 65535, shortened payload to just start an end block number (and start field is omitted if 0). 

After the change, a entry looks as follows
```
{"table":"ts_test_schema_padding_aaaaaa.part_tbl_p00000", "start":303578, "end":306116}
```

Ran this test with snapshot parallelism = 1 to force partition ranges in the same workflow. The payload is ~840KB. With the default snapshot parallelism is 4, it came down to ~250KB. 

In terms of handling, user will hit this error during initial snapshot, which would be at the very beginning of setting up a pipe so pipe is cheap to recreate, and would require disabling ctid on our end (or instead of delete and recreate, we can manually recreate their workflow, but less ideal). If during table addition, we already have able addition cancellation mechanism.

Open to evaluate db as an option later, agree it feels like the right direction.

jgao54: confirmed partition start/end are no longer NOT NULL so setting to empty string is not necessary

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -387,12 +388,23 @@ func addPartitionToQRepRun(ctx context.Context, tx pgx.Tx, flowJobName string,
 		internal.LoggerFromCtx(ctx).Warn("[monitoring]: partition "+partition.PartitionId+" has nil range",
 			slog.String(string(shared.FlowNameKey), parentMirrorName))
 	}
+
+	var childTableBlockRangesJSON []byte
+	if len(partition.ChildTableRanges) > 0 {
+		internal.LoggerFromCtx(ctx).Info("Child table ranges detected")
+		var err error
+		childTableBlockRangesJSON, err = json.Marshal(partition.ChildTableRanges)
```
jgao54, line 396: using JSON instead of proto here for easier readability when manually inspecting rows

flow/connectors/postgres/qrep.go
```diff
@@ -214,6 +215,10 @@ func (c *PostgresConnector) getPartitions(
 		if err != nil {
 			return nil, err
 		}
+		if computedNumPartitions == 0 {
```
jgao54, line 218: `ComputeNumPartitions` can return 0.

Previously NTILE-partitioning handled this case. Instead of letting each partitioning strategy handle it, just do this consistently before picking strategies.

nexus/catalog/migrations/V52__qrep_partitions_child_table_ranges.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE peerdb_stats.qrep_partitions ADD COLUMN child_table_ranges JSONB;
```
jgao54, line 1: Technically, we could have used this table to avoid introducing the proto field and hitting limits; but using `peerdb_stats.*` table as source-of-truth for non-monitoring code path feels wrong.  Using `public.metadata_qrep_partitions` table would be more appropriate, but today the `metadata_qrep_partitions` table is inserted at `FinishQRepPartition`. So requires more careful handling for a db-based approach (e.g. handling row update correctly when encountering failed snapshot.)

Also would be confusing to have non-partitioned table using temporal as source-of-truth, partitioned table using db as source of truth, and inherited table using a mix of the two.  So for db-approach would be better if we deprecate the Temporal-based approach completely when we make that change. 

So decided with keep things consistent for now, especially since we have confidence that 10k partitions can be handled by the temporal-based approach comfortably.

PR #4000, opened 2026-02-28
fix qrep partition replication pipe deadlock

iskakaushik:
When the sync side errors during qrep partition replication, the pull goroutine blocks forever in io.Pipe.Write() because nobody closes the read end. This extracts orchestratePullAndSync with an onSyncFailure callback to close the pipe on error, and makes QRecordStream/QObjectStream.Close thread-safe with sync.Once.

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -55,21 +55,11 @@ type StreamCloser interface {
 	Close(error)
 }
 
-func drainQRecordStreamOnSyncFailure(logger log.Logger, stream *model.QRecordStream) func(error) {
+func drainChannelOnSyncFailure[T any](logger log.Logger, msg string, ch <-chan T) func(error) {
 	return func(syncErr error) {
-		logger.Warn("draining record stream to unblock pull after sync failure", "error", syncErr)
+		logger.Warn(msg, "error", syncErr)
 		go func() {
-			for range stream.Records {
-			}
-		}()
-	}
-}
-
-func drainQObjectStreamOnSyncFailure(logger log.Logger, stream *model.QObjectStream) func(error) {
-	return func(syncErr error) {
-		logger.Warn("draining object stream to unblock pull after sync failure", "error", syncErr)
-		go func() {
-			for range stream.Objects {
+			for range ch {
 			}
 		}()
 	}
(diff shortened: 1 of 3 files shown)


flow/activities/flowable.go
```diff
@@ -55,6 +55,26 @@ type StreamCloser interface {
...
+func drainQObjectStreamOnSyncFailure(logger log.Logger, stream *model.QObjectStream) func(error) {
+	return func(syncErr error) {
+		logger.Warn("draining object stream to unblock pull after sync failure", "error", syncErr)
+		go func() {
+			for range stream.Objects {
+			}
+		}()
+	}
+}
+
```
jgao54: nit:
```suggestion
func drainChannelOnSyncFailure[T any](logger log.Logger, msg string, ch <-chan T) func(error) {
    return func(syncErr error) {
        logger.Warn(msg, "draining stream to unblock pull after sync failure", syncErr)
        go func() { for range ch { } }()
    }
}

```

jgao54 (approved):

flow/activities/flowable_core.go
```diff
@@ -461,42 +511,35 @@ func replicateQRepPartition[TRead any, TWrite StreamCloser, TSync connectors.QRe
...
-	errGroup.Go(func() error {
-		numRecords, numBytes, err := pullRecords(srcConn, errCtx, a.OtelManager, config, dstType, partition, stream)
-		stream.Close(err)
-		if err != nil {
-			return a.Alerter.LogFlowWrappedError(ctx, config.FlowJobName, "[qrep] failed to pull records", err)
-		}
-
-		// for Postgres source, reports all bytes fetched from source
-		// for MySQL and MongoDB source, connector reports bytes fetched but some bytes are counted here
-		// since the reporting is asynchronous (goroutine)
```
jgao54, line 475: nit: bring back comment here for context

flow/go.mod
```diff
@@ -65,41 +65,44 @@ require (
...
+	go.temporal.io/sdk/contrib/opentelemetry v0.7.0
 	go.uber.org/automaxprocs v1.6.0
 	golang.org/x/crypto v0.48.0
-	golang.org/x/exp v0.0.0-20260112195511-716be5621a96
+	golang.org/x/exp v0.0.0-20260218203240-3dfff04db8fa
 	golang.org/x/sync v0.19.0
-	google.golang.org/api v0.257.0 // PINNED(DBI-443): v0.258.0 deprecates WithCredentialsJSON, migrate to WithAuthCredentialsJSON
-	google.golang.org/genproto/googleapis/api v0.0.0-20260203192932-546029d2fa20
-	google.golang.org/grpc v1.78.0
+	google.golang.org/api v0.269.0 // PINNED(DBI-443): v0.258.0 deprecates WithCredentialsJSON, migrate to WithAuthCredentialsJSON
```
jgao54, line 75: re: 
> // PINNED(DBI-443): v0.258.0 deprecates WithCredentialsJSON, migrate to WithAuthCredentialsJSON

I missed this during the review and @ilidemi pointed this out.

For pinning versions, would be good to add a CI test to check to pin version and make sure it doesn't get accidentally bumped when upgrading deps.

Same as the ones above. 

Edit: given the alternative PR, this is just a general comment -- i will follow-up with action here

PR #4005, opened 2026-03-02
QRep: Make Pull aware of Sync errors via context cancellation

Amogh-Bharadwaj:
This PR tries to solve the same issue as #4000 via a different approach:

- Add `Send` methods to QRecord and QObject streams which receives from the channel modulo context cancellation. 
- Add a `ReportSyncError` method on QRecord, QObject and PgCopyWriter streams where it is no-op for the Q streams. In the case of PgCopyWriter (used for PG type system PG to PG mirrors), it closes the I/O pipe used by pgx CopyTo, making it error pull-side.
- Adds a unit test for Send functionality
- Ports over improved Close logic for streams from #4000 with closeOnce -- though technically not used here, can still be useful in the future and no harm by having it.

Motivation:
https://github.com/PeerDB-io/peerdb/pull/4000#pullrequestreview-3878211117

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -52,7 +52,7 @@ type FlowableActivity struct {
 }
 
 type QRepStreamCloser interface {
-	ReportQRepSyncError(error)
+	HandleQRepSyncError(error)
 	Close(error)
 }
 
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -488,7 +488,7 @@ func replicateQRepPartition[TRead any, TWrite QRepStreamCloser, TSync connectors
 		var err error
 		rowsSynced, warnings, err = syncRecords(dstConn, errCtx, config, partition, outstream)
 		if err != nil {
-			stream.ReportQRepSyncError(err)
+			stream.HandleQRepSyncError(err)
 			return a.Alerter.LogFlowWrappedError(ctx, config.FlowJobName, "failed to sync records", err)
 		}
 		for _, warning := range warnings {
@@ -605,7 +605,7 @@ func replicateXminPartition[TRead any, TWrite QRepStreamCloser, TSync connectors
 		var err error
 		rowsSynced, warnings, err = syncRecords(dstConn, ctx, config, partition, outstream)
 		if err != nil {
-			stream.ReportQRepSyncError(err)
+			stream.HandleQRepSyncError(err)
 			return a.Alerter.LogFlowWrappedError(ctx, config.FlowJobName, "failed to sync records", err)
 		}
 		for _, warning := range warnings {
--- a/flow/model/qobject_stream.go
+++ b/flow/model/qobject_stream.go
@@ -90,7 +90,7 @@ func (s *QObjectStream) Send(ctx context.Context, object *Object) error {
 	}
 }
 
-func (s *QObjectStream) ReportQRepSyncError(err error) {
+func (s *QObjectStream) HandleQRepSyncError(err error) {
 	// no-op for QObjectStream
 }
 
--- a/flow/model/qrecord_stream.go
+++ b/flow/model/qrecord_stream.go
@@ -48,7 +48,7 @@ func (s *QRecordStream) SchemaChan() <-chan struct{} {
 	return s.schemaLatch.Chan()
 }
 
-func (s *QRecordStream) ReportQRepSyncError(err error) {
+func (s *QRecordStream) HandleQRepSyncError(err error) {
 	// no-op for QRecordStream
 }
 
--- a/flow/connectors/postgres/sink_pg.go
+++ b/flow/connectors/postgres/sink_pg.go
@@ -45,7 +45,7 @@ func (p PgCopyWriter) SetSchema(schema []string) {
 	p.schema.schemaLatch.Set(schema)
 }
 
-func (p PgCopyWriter) ReportQRepSyncError(err error) {
+func (p PgCopyWriter) HandleQRepSyncError(err error) {
 	p.Close(err)
 }
 


flow/model/qrecord_batch.go
```diff
@@ -12,11 +12,11 @@ type QRecordBatch struct {
 
 func (q *QRecordBatch) ToQRecordStream(buffer int) *QRecordStream {
 	stream := NewQRecordStream(min(buffer, len(q.Records)))
-	go q.FeedToQRecordStream(stream)
+	go q.feedToQRecordStream(stream)
 	return stream
 }
 
-func (q *QRecordBatch) FeedToQRecordStream(stream *QRecordStream) {
+func (q *QRecordBatch) feedToQRecordStream(stream *QRecordStream) {
```
Amogh-Bharadwaj, line 19: Not relevant to this PR, just noticed this was being unnecessarily exported

flow/activities/flowable_core.go
```diff
@@ -488,6 +488,7 @@ func replicateQRepPartition[TRead any, TWrite StreamCloser, TSync connectors.QRe
 		var err error
 		rowsSynced, warnings, err = syncRecords(dstConn, errCtx, config, partition, outstream)
 		if err != nil {
+			stream.ReportSyncError(err)
```
ilidemi: nit: XMin too for uniformity

flow/connectors/bigquery/qrep_object_pull.go
```diff
@@ -65,13 +65,16 @@ func (c *BigQueryConnector) PullQRepObjects(
...
 	var totalBytes int64
 
-	processObject := func(attrs *storage.ObjectAttrs) {
-		stream.Objects <- &model.Object{
+	processObject := func(attrs *storage.ObjectAttrs) error {
+		if err := stream.Send(ctx, &model.Object{
 			URL:  fmt.Sprintf("https://storage.googleapis.com/%s/%s", bucketName, url.PathEscape(attrs.Name)),
 			Size: attrs.Size,
+		}); err != nil {
+			return err
```
ilidemi:
```suggestion
			return fmt.Errorf("failed to send object to stream: %w", err)
```

flow/connectors/postgres/qrep_query_executor.go
```diff
@@ -314,7 +314,12 @@ func (qe *QRepQueryExecutor) processRowsStream(
 			qe.logger.Error("[pg_query_executor] failed to map row to QRecord", slog.Any("error", err))
 			return numRows, numBytes, fmt.Errorf("failed to map row to QRecord: %w", err)
 		}
-		stream.Records <- record
+
+		if err := stream.Send(ctx, record); err != nil {
+			qe.logger.Info("Context canceled while sending record to stream, exiting processRowsStream early")
+			return numRows, numBytes, err
```
ilidemi:
```suggestion
			return numRows, numBytes, fmt.Errorf("failed to send record to stream: %w", err)
```

flow/model/qrecord_stream.go
```diff
@@ -44,19 +48,30 @@ func (s *QRecordStream) SchemaChan() <-chan struct{} {
 	return s.schemaLatch.Chan()
 }
 
+func (s *QRecordStream) ReportSyncError(err error) {
+	// no-op for QRecordStream
+}
+
+func (s *QRecordStream) Send(ctx context.Context, record []types.QValue) error {
+	select {
```
jgao54: in `qobject_stream.go` check for error first:
```
	if s.err != nil {
		return s.err
	}
```

should we also be doing the same check here

flow/model/qrecord_stream.go
```diff
@@ -44,19 +48,30 @@ func (s *QRecordStream) SchemaChan() <-chan struct{} {
 	return s.schemaLatch.Chan()
 }
 
+func (s *QRecordStream) ReportSyncError(err error) {
+	// no-op for QRecordStream
+}
+
+func (s *QRecordStream) Send(ctx context.Context, record []types.QValue) error {
```
ilidemi, line 56:
```suggestion
// Sends the record into the channel, erroring out on context cancellation instead of waiting for the reader indefinitely
func (s *QRecordStream) Send(ctx context.Context, record []types.QValue) error {
```

flow/model/qobject_stream.go
```diff
@@ -75,6 +75,22 @@ func (s *QObjectStream) Err() error {
 	return s.err
 }
 
+func (s *QObjectStream) Send(ctx context.Context, object *Object) error {
```
ilidemi, line 81:
```suggestion
// Sends the object into the channel, erroring out on context cancellation instead of waiting for the reader indefinitely
func (s *QObjectStream) Send(ctx context.Context, object *Object) error {
```

flow/model/qrecord_stream.go
ilidemi, line 60:
```suggestion
	if s.err != nil {
		return s.err
	}
	select {
```

flow/activities/flowable.go
```diff
@@ -52,6 +52,7 @@ type FlowableActivity struct {
 }
 
 type StreamCloser interface {
+	ReportSyncError(error)
```
ilidemi: Just to make it clear how generic the interface is (hopefully should work as an auto-rename)
```suggestion
type QRepStreamCloser interface {
	ReportQRepSyncError(error)
```

ilidemi (approved):
Appreciate working together with the abstractions we already have 🙏

flow/activities/flowable.go
```diff
@@ -51,7 +51,8 @@ type FlowableActivity struct {
 	TemporalClient client.Client
 }
 
-type StreamCloser interface {
+type QRepStreamCloser interface {
+	ReportQRepSyncError(error)
```
jgao54: nit: this name feels a bit misleading given the method closes the writer but doesn't actually report anything here.

Amogh-Bharadwaj: I had a similar interpretation to you initially as well @jgao54 , @ilidemi maybe we can change this to `HandleQRepSyncError` or something :)

flow/model/qobject_stream.go
```diff
@@ -75,6 +75,23 @@ func (s *QObjectStream) Err() error {
...
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (s *QObjectStream) ReportQRepSyncError(err error) {
+	// no-op for QObjectStream
+}
+
 func (s *QObjectStream) Close(err error) {
```
jgao54, line 97: nit: while we are at it, would be good to also use `closeOnce` here similar to `qrecord_stream` to avoid accidentally causing panic

flow/activities/flowable.go
ilidemi: SGTM!

PR #4007, opened 2026-03-03
add a CI test for pinned versions

jgao54:
To avoid accidentally overwriting pinned versions

See https://github.com/PeerDB-io/peerdb/actions/runs/22981746313/job/66722909076?pr=4007  for successful run.

Failure cases tested locally with (1) pinned version exist in go.mod but not in script (2) pinned version exist in script but not go.mod (3) exist in both but not matched

### DIFF
--- a/flow/cmd/check_pinned_versions/main.go
+++ b/flow/cmd/check_pinned_versions/main.go
@@ -24,7 +24,11 @@ var pinnedVersions = map[string]string{
 }
 
 func main() {
-	_, thisFile, _, _ := runtime.Caller(0)
+	_, thisFile, _, ok := runtime.Caller(0)
+	if !ok {
+		fmt.Fprintln(os.Stderr, "failed to determine source file path")
+		os.Exit(1)
+	}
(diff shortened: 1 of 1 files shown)


flow/go_mod_ci_test.go
```diff
@@ -0,0 +1,72 @@
...
+// To update a pinned dependency, change both go.mod AND this map.
+var pinnedVersions = map[string]string{
+	"cloud.google.com/go/bigquery":                    "v1.72.0",
+	"cloud.google.com/go/pubsub/v2":                   "v2.3.0",
+	"github.com/aws/aws-sdk-go-v2/feature/s3/manager": "v1.21.0",
+	"google.golang.org/api":                           "v0.257.0",
+	"github.com/tikv/pd/client":                       "v0.0.0-20251229071808-6173d50c004c",
+}
+
+func TestPinnedDependencies(t *testing.T) {
```
ilidemi: As mentioned offline, a workflow would be good.
Could you update the Claude skill too?

PR #4052, opened 2026-03-13
support dots in identifiers

serprex:
double dot is used to escape dots, like %% in format strings


ilidemi (changes requested):
As discussed before, the proper solution to this is a contract migration to separate schema and table. Here's an example of a recent one doing the db/protos side of it: https://github.com/PeerDB-io/peerdb/pull/3704

serprex:
protobuf doesn't support map[type], keys must be strings

serprex:
been pointed out this has a gap: `schema...table` ambiguous whether `..` is at end of schema name or start of table name

PR #4101, opened 2026-03-27
(feat) Local dev env: Tilt orchestrated data store initialization and connection coordinates propagation

pfcoperez:
## Description

This PR uses Tilt+Docker Compose to start ancillary services required for local e2e testing.

image

The idea is simple: Make tests read their input parameters from the same source of truth that generated the dev environment.

The local  `.env` file contains the ports, users and other parameters used to fill in the docker compose template defining the services.

e.g:

```
MONGO_HOST=localhost
MONGO_PORT=27017
MONGO_ADMIN_USERNAME=admin
MONGO_ADMIN_PASSWORD=admin
MONGO_USERNAME=csuser
MONGO_PASSWORD=cspass

CLICKHOUSE_HOST=localhost
CLICKHOUSE_HTTP_PORT=8123
CLICKHOUSE_NATIVE_PORT=9000

PG_HOST=localhost
PG_PORT=5436
PG_USER=postgres
PG_PASSWORD=postgres
PG_DATABASE=postgres
``` 

## Notes for reviewers

- This PR version extracts versions from GH Workflow test flow. I think we can re-use this infrastructure definition in the flow itself. If we do that, we can just use a different way of defining the test matrix images that works both locally and in CI without the need of brittle extraction scripts. Alternatively we can just hard-code versions in the local env too. 
- The dev env adds an additional PSQL service to use as data source but the existing e2e tests use the catalog PSQL instance. We can disable the extra resource. 
- Some PG tests fail because the catalog has not been provisioned completely. (e.g: `env -f ../.env go test -v -run TestGenericCH_PG/Test_Simple_Flow ./e2e/`)  I am following up with a PR in the tests to use the additional PSQL instance which is intended to be a data source instead of the catalog.

## ~WIP~

- [x] Update README.md with instructions.
- [x] Add MySQL/MariaDB f6ef96d5f0256618e11764df90dbb7633abf180b
- [x] Make provisioning scripts idempotent: 2d8a2ccc2a548cae48d8d1f53f6ad13455e71cab

## Related work

Follows: [PeerDB-io/peerdb#3901](https://github.com/PeerDB-io/peerdb/pull/3901)
Replaces: [PeerDB-io/peerdb#4095](https://github.com/PeerDB-io/peerdb/pull/4095) (same changes, moving source repository)


README.md
```diff
@@ -91,6 +91,53 @@ You can use Postgres’ ecosystem to manage your ETL —
...
+
+Or local debugging sessions.
+
+These tests require both PeerDB services, source and destination stores to be running. We provide a local environment with all the necessary services and dependencies to run these tests.
+
+This is done through [Tilt](https://tilt.dev/) orchestrated Docker compose.
+
+To get the environment up you first need to specify the shared environment variables for both the test and the test environment in your local `.env` file. You can use the provided `.env.example` as a template: `cp .env.example .env `.
+
+:memo: In the template, services URLs are set to `172.18.0.1`, which is the default gateway for Docker containers. This allows both test processes and services running inside Docker to access services running on the host machine. If your setup differs, update these URLs accordingly.
```
ilidemi: Let's default to `host.docker.internal` so it's at least some sort of standard. Then IPs can be substituted via hosts file or editing local .env, unless there's some way to make it linux-compatible by default

README.md
```diff
@@ -91,6 +91,53 @@ You can use Postgres’ ecosystem to manage your ETL —
...
+
+Then you can just run:
+
+```bash
+./tilt.sh
+```
+
+And follow the status of the services and access logs through the Tilt UI at http://localhost:10350/.
```
ilidemi: Can we make :10352 the default so it works with clickpipes e2e?

ancillary-docker-compose.yml
```diff
@@ -0,0 +1,79 @@
+name: peerdb-ancillary-services
+
+services:
+  mongodb:
+    container_name: mongodb
```
ilidemi: Let's make the container names specific to this setup, like `peerdb-mongodb`, `peerdb-clickhouse`, etc, to avoid conflicts with other potentially running services

generate-test-environment.sh
```diff
@@ -0,0 +1,51 @@
...
+
+#export POSTGRES_VERSION=18;
+#VERSIONS=$(cat .github/workflows/flow.yml | yq '.jobs.flow_test.strategy.matrix."db-version"[] | select((.pg | tostring) == strenv(POSTGRES_VERSION))')
+VERSIONS=$(cat .github/workflows/flow.yml | yq '.jobs.flow_test.strategy.matrix."db-version"[-1]') # Select last version row in the matrix.
+
+# Version extraction is conditional on the presence of overriding values
+if [ -z "$MYSQL_VERSION" ]; then
+    #MYSQL_VERSION=$(echo "$VERSIONS" | yq -r '.mysql')
+    # Hardcode MYSQL version to the most popular one since we might replace the flow version
+    MYSQL_VERSION="9.5"
```
ilidemi, line 17: MySQL has 5.7 for mysql-pos and a mariadb, it would be good to be able to test those locally too - the behavior can be annoyingly different
https://github.com/PeerDB-io/peerdb/pull/4082
https://github.com/PeerDB-io/peerdb/pull/4075/changes/449c9cb3799219fe6a7b44998d435d10160e2ced (clickhouse_mysql_test.go changes for maria)

ancillary-docker-compose.yml
pfcoperez: Great idea, done in 8cf2a5f85051c6796b4861f4700c491695dfaa82

pfcoperez: Great idea, done in 8cf2a5f85051c6796b4861f4700c491695dfaa82

README.md
pfcoperez: Sure, done in 4bdc805963c899312c1224ee025c22ee9653c68b

I tried to set it up at Tiltfile level but it is not possbile, done at launcher script level. If someone invokes tilt directly with `tilt up` the default port will not change.

README.md
pfcoperez: :+1: Done at 8132ad10680e17e95fb93e11247bd75691e3f375

generate-test-environment.sh
pfcoperez: Done in e5c55fa0812d4e9f68facb82c0cdb99df5ed3886, 

Each flavor has its own port, qtid is selected by default for CI compatibility. I want to follow this PR with at least two more:

1. Changes to make the e2e tests point to the ancillary PSQL instance rather than the catalog when running the local dev-env.
2. (less priority) Unify CI and local-dev env ancillary services provisioning, this will open the door for test code simplification and make the whole set-up more robust.

ancillary-docker-compose.yml
```diff
@@ -0,0 +1,123 @@
+name: peerdb-ancillary-services
```
jgao54, line 1: @pfcoperez  would be better if we disable ancillary containers by default when starting `./tilt.sh`. we almost never need all the source databases when running locally peerdb, so would be more resource-efficient to manually start what is needed -- whether for local development or e2e testing

README.md
```diff
@@ -91,6 +91,73 @@ You can use Postgres’ ecosystem to manage your ETL —
...
+:memo: In the template, services URLs are set to `host.docker.internal`, which is the name for the default Docker gateway in Docker Desktop set-ups such as macOS and Windows. Using the default gateway address allows both test processes and services running inside Docker to access services on the host machine. In native Docker (Linux) this name is not resolved by default, you might replace it with the default gateway IP (e.g., `172.18.0.1`) or add a custom entry to your `/etc/hosts` file to resolve `host.docker.internal` to the appropriate IP address. e.g:
+
+```bash
+echo "172.18.0.1 host.docker.internal" | sudo tee -a /etc/hosts
+```
+
+Then you can just run:
+
+```bash
+./tilt.sh
```
jgao54, line 123: I checked out your branch and ran the following per instruction:
```
cp .env.example .env 
./tilt.sh
```
but I am still getting the following warning and not sure if this is expected (i use Docker Desktop).
```
Tilt started on http://localhost:10352/
v0.36.1, built 2026-01-20

(space) to open the browser
(s) to stream logs (--stream=true)
(t) to open legacy terminal mode (--legacy=true)
(ctrl-c) to exit
WARN[0000] The "PG_PORT" variable is not set. Defaulting to a blank string.
WARN[0000] The "PG_USER" variable is not set. Defaulting to a blank string.
WARN[0000] The "PG_PASSWORD" variable is not set. Defaulting to a blank string.
WARN[0000] The "PG_DATABASE" variable is not set. Defaulting to a blank string.
WARN[0000] The "CI_MONGO_PORT" variable is not set. Defaulting to a blank string.
WARN[0000] The "CI_CLICKHOUSE_HTTP_PORT" variable is not set. Defaulting to a blank string.
WARN[0000] The "CI_CLICKHOUSE_NATIVE_PORT" variable is not set. Defaulting to a blank string.
WARN[0000] The "CI_MYSQL_GTID_PORT" variable is not set. Defaulting to a blank string.
WARN[0000] The "CI_MYSQL_ROOT_PASSWORD" variable is not set. Defaulting to a blank string.
WARN[0000] The "CI_MYSQL_POS_PORT" variable is not set. Defaulting to a blank string.
WARN[0000] The "CI_MYSQL_ROOT_PASSWORD" variable is not set. Defaulting to a blank string.
WARN[0000] The "CI_MYSQL_ROOT_PASSWORD" variable is not set. Defaulting to a blank string.
WARN[0000] The "CI_MARIADB_PORT" variable is not set. Defaulting to a blank string.
```

local_provision_scripts/mongodb.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
```
jgao54, line 1: when testing your branch locally, it looks like mongo is not successfully provisioning on my end and it's also resulting in other services (including peerdb services) to be waiting on `provision-mongodb`.

Image

peerdb services, as well as other provision-* services, should not need to wait for provision-mongodb.

jgao54: the ancilliary mongodb container is completed but not sucessfully authenticated: i see that it's `Failed to authenticate`:

>{"code":13,"codeName":"Unauthorized","errmsg":"not authorized on admin to execute command { getParameter: 1, featureCompatibilityVersion: 1, lsid: { id: UUID(\"635caa3c-704d-40d6-b22a-fc5e52cfa909\") }, $readPreference: { mode: \"primaryPreferred\" }, $db: \"admin\" }"}}}

README.md
pfcoperez: Thanks for testing this @jgao54 ! It's great to have a variety of configurations to find these differences.

This means that Docker Compose is unable to find the temporary generated environment file (`ancillary.env`)  .  With these messages, your environment won't boot correctly.

I've pushed https://github.com/PeerDB-io/peerdb/pull/4101/changes/8ff6b9e2631a92cf102cfb98f58be62c9500ec3a to pass the path explicitly based on the location of the Tiltfile itself (`ancillary.env`) is generated on the same path as ./tilt.sh

Please, tear down your environment and try with this change. If the warning messages still persist, please can you provide me with:

1. The result of 'ls'  under your 'peerdb' directory after having invoked `tilt.sh`
2. If `ancillary.env` is there, its contents.

:pray:

local_provision_scripts/mongodb.sh
pfcoperez: Is this still happening after having fixed the warnings you reported in https://github.com/PeerDB-io/peerdb/pull/4101#pullrequestreview-4034514451

ancillary-docker-compose.yml
pfcoperez: Good point, I did this for all ancillary services but ClickHouse in https://github.com/PeerDB-io/peerdb/pull/4101/changes/a237fbee77bd8c42662f846f059964365082734d 

I think all cases to use the environment require ClickHouse, right?

README.md
pfcoperez: @masterashu @ilidemi Did you experience this issue in your local envs?

local_provision_scripts/mongodb.sh
masterashu: Fixed: https://github.com/PeerDB-io/peerdb/pull/4101/commits/fd6b772c52edc2e38a4303dacbf39d63eeea7646

jgao54: confirm with the mongo credential pushed it now works for me 👍  

(not sure why admin account creation in `mongodb.sh`  didn't work for me)

jgao54: it was probably still related to the env vars not being consumed properly

jgao54: derp, spoke too soon. i am now seeing connection churning 
Screenshot 2026-04-01 at 08 19 12@2x


but the provisioning script works:
Screenshot 2026-04-01 at 08 18 17@2x

Tiltfile
```diff
@@ -58,3 +58,112 @@ dc_resource('flow-worker', resource_deps=['proto-gen'], labels=['PeerDB'])
...
+
+dc_resource('mongodb', labels=['Ancillary', 'DataStore'], links=[
+    link('http://localhost:11017', 'MongoDB'),
+], auto_init=False)
+
+dc_resource('mysql-gtid', labels=['Ancillary', 'DataStore'], links=[
+    link('http://localhost:3306', 'MySQL GTID'),
+], auto_init=False)
+
+dc_resource('mysql-pos', labels=['Ancillary', 'DataStore'], links=[
```
jgao54, line 133: not able to build this image on macos:

> Error response from daemon: no matching manifest for linux/arm64/v8 in the manifest list entries: no match for platform in manifest: not found

Tiltfile
```diff
@@ -58,3 +58,112 @@ dc_resource('flow-worker', resource_deps=['proto-gen'], labels=['PeerDB'])
...
+    link('http://localhost:3307', 'MySQL File-Pos'),
+], auto_init=False)
+
+dc_resource('mariadb', labels=['Ancillary', 'DataStore'], links=[
```
jgao54, line 137: Seeing this in the logs for MariaDB:

> 2026-04-01 18:45:35 100 [Warning] Access denied for user 'mysql'@'localhost' (using password: NO)

local_provision_scripts/mongodb.sh
jgao54: ohhh,  MongoDB is working and what I thought is "connection churn" is just the added healthy check with `ping` 🤦‍♀️  .  since ping doesn't require authentication we are seeing the log "connection not authenticating"

jgao54 (approved):
Looks great overall! The remaining comments can be addressed as follow-ups.

ancillary-docker-compose.yml
```diff
@@ -0,0 +1,127 @@
...
```
jgao54, line 1: Just noticed that we are not using named volumes for ancillary-docker-compose.yml

this means when tilt is shut down, we would actually lose all the test data (but the metadata in the catalog is persisted)

It would be good to add named volume, similar to what we have in the main docker-compose files.

ancillary-docker-compose.yml
```diff
@@ -0,0 +1,127 @@
...
+    environment:
+      MONGO_INITDB_ROOT_USERNAME: ${CI_MONGO_ADMIN_USERNAME}
+      MONGO_INITDB_ROOT_PASSWORD: ${CI_MONGO_ADMIN_PASSWORD}
+
+    ports:
+      - "${CI_MONGO_PORT}:27017"
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    healthcheck:
+      test: ["CMD", "mongosh", "--eval", "db.runCommand({ ping: 1 })"]
```
jgao54, line 18: nit: to make mongodb logs less noisy, could we execute this with "--quiet"

Tiltfile
```diff
@@ -58,3 +58,112 @@ dc_resource('flow-worker', resource_deps=['proto-gen'], labels=['PeerDB'])
...
+)
+
+# This is not defined as a resource as we need the file to be present
+# when `docker_compose` loads the configuration (next line).
+local('./generate-test-environment.sh')
+
+tiltfile_dir = config.main_path.replace('/Tiltfile', '')
+docker_compose('./ancillary-docker-compose.yml', env_file=tiltfile_dir + '/ancillary.env')
+
+# ClickHouse is the only datastore that is automatically started as the common destination for all pipes.
```
ilidemi, line 114: Let's make this one on-demand too, the other e2e environment has its own ClickHouse so optional is better

Tiltfile
```diff
@@ -58,3 +58,112 @@ dc_resource('flow-worker', resource_deps=['proto-gen'], labels=['PeerDB'])
...
+    resource_deps=['mariadb']
+)
+
+local_resource(
+    'provision-postgres',
+    cmd='./local_provision_scripts/postgres.sh',
+    labels=['Ancillary', 'Provisioning'],
+    resource_deps=['postgres']
+)
+
```
ilidemi, line 106:
```suggestion

allow_k8s_contexts(k8s_context()) # to unblock local(), k8s is not used in local dev

```

Tiltfile
```diff
@@ -58,3 +58,112 @@ dc_resource('flow-worker', resource_deps=['proto-gen'], labels=['PeerDB'])
...
+# This is not defined as a resource as we need the file to be present
+# when `docker_compose` loads the configuration (next line).
+local('./generate-test-environment.sh')
+
+tiltfile_dir = config.main_path.replace('/Tiltfile', '')
+docker_compose('./ancillary-docker-compose.yml', env_file=tiltfile_dir + '/ancillary.env')
+
+# ClickHouse is the only datastore that is automatically started as the common destination for all pipes.
+dc_resource('clickhouse', labels=['Ancillary', 'DataStore'], links=[
```
ilidemi, line 115: When a resource has two labels, clicking on it results in unexpected UI jumps for me. Any hurt in removing Ancillary from resources?

ancillary-docker-compose.yml
```diff
@@ -0,0 +1,127 @@
...
+      - "host.docker.internal:host-gateway"
+    healthcheck:
+      test: ["CMD", "mysqladmin", "ping", "-h", "localhost"]
+      interval: 2s
+      timeout: 10s
+      retries: 5
+
+  mysql-pos:
+    container_name: peerdb-mysql-pos
+    image: mysql:5.7
```
ilidemi, line 64: `biarms/mysql` works on arm64

ancillary-docker-compose.yml
```diff
@@ -0,0 +1,127 @@
...
+    command: ["--log_bin=mysql-bin", "--server-id=1", "--bind-address=::"]
+    restart: unless-stopped
+    ports:
+      - "${CI_MYSQL_POS_PORT}:3306"
+    environment:
+      MYSQL_ROOT_PASSWORD: ${CI_MYSQL_ROOT_PASSWORD}
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    healthcheck:
+      test: ["CMD", "mysqladmin", "ping", "-h", "localhost"]
```
ilidemi, line 74: Getting
`2026-04-03T06:07:35.938063Z 5 [Note] Access denied for user 'root'@'localhost' (using password: NO)`
with the binarms image, works with a password var

```suggestion
      test: ["CMD", "mysqladmin", "ping", "-h", "localhost", "-p${CI_MYSQL_ROOT_PASSWORD}"]
```

ancillary-docker-compose.yml
```diff
@@ -0,0 +1,127 @@
...
+    command: ["--log-bin=maria"]
+    restart: unless-stopped
+    ports:
+      - "${CI_MARIADB_PORT}:3306"
+    environment:
+      MARIADB_ROOT_PASSWORD: ${CI_MYSQL_ROOT_PASSWORD}
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    healthcheck:
+      test: ["CMD", "mariadb-admin", "ping", "-h", "localhost"]
```
ilidemi, line 91:
```suggestion
      test: ["CMD", "mariadb-admin", "ping", "-h", "localhost", "-u", "root", "-p${CI_MYSQL_ROOT_PASSWORD}"]
```

.env.example
```diff
@@ -0,0 +1,42 @@
...
+CI_CLICKHOUSE_HOST=host.docker.internal
+CI_CLICKHOUSE_HTTP_PORT=11123
+CI_CLICKHOUSE_NATIVE_PORT=11000
+
+CI_MYSQL_HOST=host.docker.internal
+CI_MYSQL_GTID_PORT=3306
+CI_MYSQL_POS_PORT=3307
+CI_MARIADB_PORT=3308
+CI_MYSQL_ROOT_PASSWORD=cipass
+CI_MYSQL_VERSION=mysql-gtid
```
ilidemi, line 19: ~~Add default versions for PG/Mongo/CH so it's clear how to override it at a glance~~
Nevermind, actually read the readme just now

Tiltfile
```diff
@@ -58,3 +58,112 @@ dc_resource('flow-worker', resource_deps=['proto-gen'], labels=['PeerDB'])
...
+    link('http://localhost:8118', 'Dozzle Container Monitor'),
+])
+
+# Tests launchers
+
+def e2e_test(name, test_run, extra_deps=[]):
+    local_resource(
+        'e2e_' + name,
+        cmd='cd flow && go clean -cache && env -f ../.env go test -v -run %s ./e2e/' % test_run,
+        labels=['e2e', 'Test'],
```
ilidemi, line 157: Same, let's keep one label

Tiltfile
```diff
@@ -58,3 +58,112 @@ dc_resource('flow-worker', resource_deps=['proto-gen'], labels=['PeerDB'])
...
+    )
+
+# Postgres to ClickHouse generic tests
+e2e_test('postgres', 'TestGenericCH_PG', ['provision-postgres'])
+
+# MySQL to ClickHouse generic tests
+e2e_test('mysql', 'TestGenericCH_MySQL', ['provision-mysql-gtid', 'provision-mysql-pos', 'provision-mariadb'])
+
+# MongoDB to ClickHouse test suite
+e2e_test('mongodb', 'TestMongoClickhouseSuite', ['provision-mongodb'])
```
ilidemi, line 169: As we make the whitelist of tests working, this would be a good place to list what they are, so people don't have to type them out

ilidemi (approved):
Works on my machine 😄

README.md
ilidemi: With review suggestions and clearing the ancillary env, everything works for me

pfcoperez:
Thanks for the last comments @ilidemi . I was a bit trigger happy and merged the PR before seeing them. I am following up with a PR to address them,

Tiltfile
pfcoperez: Done in follow-up PR: https://github.com/PeerDB-io/peerdb/pull/4147

Tiltfile
pfcoperez: No hurt at all, It was a personal preference to have some resources classified in different ways. No strong feelings about it so I changed it in https://github.com/PeerDB-io/peerdb/pull/4147

Tiltfile
pfcoperez: This is not needed in a native Docker installation. Are you using the embedded k8s cluster option in Docker desktop?

ancillary-docker-compose.yml
pfcoperez: Moved to this image in https://github.com/PeerDB-io/peerdb/pull/4147

ancillary-docker-compose.yml
pfcoperez: Addressed in https://github.com/PeerDB-io/peerdb/pull/4147

ancillary-docker-compose.yml
pfcoperez: Addressed in https://github.com/PeerDB-io/peerdb/pull/4147

Tiltfile
pfcoperez: Makes sense, if in the future we also want to have a Tilt independent test manifest file, we can use Tiltfile's Starlark language to read from it.

Tiltfile
pfcoperez: Some research shows this might be the case so I've added this at https://github.com/PeerDB-io/peerdb/pull/4147

ancillary-docker-compose.yml
pfcoperez: Great point, added in follow-up PR https://github.com/PeerDB-io/peerdb/pull/4147

ancillary-docker-compose.yml
pfcoperez: I tend to prefer to default to the stateless approach, new environment run fresh data. However, since the PeerDB services preserve metadata I've followed your advice in https://github.com/PeerDB-io/peerdb/pull/4147 and added a button to clean-up volumes on demand.

Tiltfile
ilidemi: Tilt is being paranoid here and errors out if any k8s context active at all, which outside of PeerDB we do have

PR #4155, opened 2026-04-08
skip pubviaroot query if postgres version is less than 13

### LINKED ISSUE #3544: ClickPipe Fails When Partition Column Order Differs from Parent (PostgreSQL)
**In PostgreSQL, a partition table can have a different column ordinal_position than its parent. PostgreSQL itself handles this correctly by mapping values by name. However, ClickPipe incorrectly assumes partitions share the same column order as the parent. This leads to wrong value-to-column mapping and runtime errors.**

**Does it reproduce on the most recent release?**
Yes

**How to reproduce**
PostgreSQL allows partitions and parent tables to have different column orders.

ClickPipe maps values to columns by ordinal position instead of by name/type.

When orders differ, a string can be interpreted as an integer or vice versa, causing replication failures.

Observed Error
```
error processing message: error converting tuple to map:
error decoding text column data:
strconv.ParseInt: parsing "IA": invalid syntax.
```
**Interpretation: ClickPipe attempted to parse "IA" (a VARCHAR ip_state) as an INT http_code because of mismatched column ordering.**

Steps to Reproduce (PostgreSQL)

Create an old table that will later become a partition (columns ordered ip_state before http_code):

```
CREATE TABLE payment.payment_orders_clickpipe_test_partition (
    id BIGSERIAL,
    at DATE NOT NULL,
    ip_state VARCHAR(10),
    http_code INT,
    PRIMARY KEY (id, at)
);

INSERT INTO payment.payment_orders_clickpipe_test_partition (id, at, ip_state, http_code)
VALUES (1, '2025-01-01', 'AI', 5);
```

Create a parent partitioned table with a different column order (http_code before ip_state):

```
CREATE TABLE payment.payment_orders_clickpipe_test (
    id BIGSERIAL,
    at DATE NOT NULL,
    http_code INT,
    ip_state VARCHAR(10),
    PRIMARY KEY (id, at)
) PARTITION BY RANGE (at);
```

Attach the existing table as a partition:

```
ALTER TABLE payment.payment_orders_clickpipe_test
ATTACH PARTITION payment.payment_orders_clickpipe_test_partition
FOR VALUES FROM ('2025-01-01') TO ('2025-12-31');
```

Insert into the partition:

```
INSERT INTO payment.payment_orders_clickpipe_test_partition (id, at, ip_state, http_code)
VALUES (6, '2025-01-01', 'IA', 5);
```
Observe ClickPipe failure during replication with the error above.

Diagnostic Query (compare parent vs partition column order)

```
WITH parent AS (
  SELECT ordinal_position, column_name, data_type
  FROM information_schema.columns
  WHERE table_schema = 'payment' AND table_name = 'payment_orders'
),
part AS (
  SELECT ordinal_position, column_name, data_type
  FROM information_schema.columns
  WHERE table_schema = 'payment' AND table_name = 'payment_orders_clickpipe_test_partition'
)
SELECT p.ordinal_position AS parent_pos,
       p.column_name AS parent_col,
       p.data_type AS parent_type,
       c.ordinal_position AS part_pos,
       c.column_name AS part_col,
       c.data_type AS part_type
FROM parent p
FULL JOIN part c ON p.ordinal_position = c.ordinal_position
WHERE COALESCE(p.column_name,'') <> COALESCE(c.column_name,'')
ORDER BY parent_pos;
```
This query highlights mismatched column orders.

**Expected Behavior**

ClickPipe should handle PostgreSQL partitions with different column orders correctly.

Values must be mapped to columns by name and type, not ordinal position.

Inserts into partitions should replicate successfully without type conversion errors.

Expected behavior
No response

Error message and/or stacktrace
```
error processing message: error converting tuple to map:
error decoding text column data:
strconv.ParseInt: parsing "IA": invalid syntax.
```

jgao54:
- Bring back https://github.com/PeerDB-io/peerdb/pull/4035, optionally query for pubviaroot only if pg version >= 13, otherwise default to false since pubviaroot is not supported anyways
- Add e2e coverage to check that a dynamically added partition with out-of-order columns works
- Validate that cdc continue to work on pg 12

Fixes: https://github.com/PeerDB-io/peerdb/issues/3544

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -93,12 +93,18 @@ func (c *PostgresConnector) NewPostgresCDCSource(ctx context.Context, cdcConfig 
 	}
 
 	var publishViaPartitionRoot bool
-	if err := c.conn.QueryRow(ctx,
-		"SELECT COALESCE(pubviaroot, false) FROM pg_publication WHERE pubname=$1",
-		cdcConfig.Publication,
-	).Scan(&publishViaPartitionRoot); err != nil {
-		return nil, fmt.Errorf("error checking publish_via_partition_root for publication %s: %w",
-			cdcConfig.Publication, err)
+	majorVersion, err := c.MajorVersion(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("error getting major version: %w", err)
+	}
+	if majorVersion >= shared.POSTGRES_13 {
+		if err := c.conn.QueryRow(ctx,
+			"SELECT COALESCE(pubviaroot, false) FROM pg_publication WHERE pubname=$1",
+			cdcConfig.Publication,
+		).Scan(&publishViaPartitionRoot); err != nil {
+			return nil, fmt.Errorf("error checking publish_via_partition_root for publication %s: %w",
+				cdcConfig.Publication, err)
+		}
 	}
 
 	var schemaNameForRelID map[uint32]string
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -38,6 +38,7 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
 
+//nolint:govet // fieldalignment: fields grouped by purpose for readability
 type PostgresCDCSource struct {
 	*PostgresConnector
 	srcTableIDNameMapping  map[uint32]string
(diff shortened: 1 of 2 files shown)


ilidemi (approved):
Thanks for the separated out commits!

vadymKovtunP:
@jgao54 Could you help me check how we can determine when this fix will be available on ClickPipe for testing?

jgao54:
@vadymKovtunP  this change has been rolled out!

PR #4161, opened 2026-04-09
SQL Server connector

serprex:
Doesn't handle ADD COLUMN


serprex:
#4052 would help include SQL Server as a generic source

Jeremyyang920:
@serprex We're not going to be able to pick this up at this time. Internal discussions are being made about how to best go about this.

PR #4163, opened 2026-04-10
Don't retry DropFlowSource on auth errors

ilidemi:
DropFlow retries source cleanup for many hours if it can't connect, in case the db is just turned off for the night and cleanup can still be done after. As these attempts and also ScheduledTasks are also running for a long time, it looks like an attack to external security tools. If our user was revoked, we shouldn't expect to successfully clean up anyway.

Closes DBI-670

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -17,6 +17,7 @@ import (
 	"github.com/jackc/pgerrcode"
 	"github.com/jackc/pglogrepl"
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/jackc/pgx/v5/pgtype"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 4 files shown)


flow/connectors/postgres/postgres.go
```diff
@@ -100,6 +101,15 @@ func NewPostgresConnector(ctx context.Context, env map[string]string, pgConfig *
 	conn, err := NewPostgresConnFromConfig(ctx, connConfig, pgConfig.TlsHost, rdsAuth, tunnel)
 	if err != nil {
 		tunnel.Close()
+
+		if pgErr, ok := errors.AsType[*pgconn.PgError](err); ok {
```
jgao54, line 105: ah, nice to have type safety with errors.AsType

jgao54 (approved):

flow/connectors/postgres/postgres.go
ilidemi: Go 1.26 is nice

PR #4165, opened 2026-04-11
Always chunk Avro on uncompressed bytes

ilidemi:
It's been running for some time without any issues

Closes DBI-581


jgao54 (approved):
nice!

PR #4167, opened 2026-04-13
feat(pg-pg): introduce new MERGE statement leveraging jsonb_to_record

Amogh-Bharadwaj:
> [!NOTE]
> This feature is only for mirrors with the PG type system

## Why
For workloads on source involving very wide tables and data, normalize times for PG to PG mirrors were seen to be considerably high, causing data staleness on target.

Research showed that the ->> JSON access we do for the `_peerdb_data` column in our raw table in Postgres is slow. 

This PR is an optimization we found where instead of having to access every field in `_peerdb_data`, we can use [jsonb_to_record](https://www.postgresql.org/docs/9.5/functions-json.html) to first convert the entire JSON value to a row tuple and access its fields, which is much faster.

For a 170GB table with 158 columns on Postgres receiving updates, this new MERGE statement was seen to be ~10x faster.

## What
- In `client.go`, declares a template for a new MERGE statement using `jsonb_to_record`.
- Adapts `normalize.go` to use this new MERGE
- Adapts unit tests
- Adds type test in E2E for PG type system

### DIFF
--- a/flow/e2e/postgres_test.go
+++ b/flow/e2e/postgres_test.go
@@ -85,11 +85,11 @@ func (s PeerFlowE2ETestSuitePG) Test_Geospatial_PG() {
 	RequireEnvCanceled(s.t, env)
 }
 
-func (s PeerFlowE2ETestSuitePG) Test_Types() {
+func (s PeerFlowE2ETestSuitePG) testTypes(suffix string, system protos.TypeSystem) {
 	tc := NewTemporalClient(s.t)
 
-	srcTableName := s.attachSchemaSuffix("test_types_pg")
-	dstTableName := s.attachSchemaSuffix("test_types_pg_dst")
+	srcTableName := s.attachSchemaSuffix("test_types_" + suffix)
+	dstTableName := s.attachSchemaSuffix("test_types_" + suffix + "_dst")
 
 	_, err := s.Conn().Exec(s.t.Context(), fmt.Sprintf(`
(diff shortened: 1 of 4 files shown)


flow/internal/dynamicconf.go
```diff
@@ -605,6 +613,10 @@ func PeerDBGroupNormalize(ctx context.Context, env map[string]string) (int64, er
 	return dynamicConfSigned[int64](ctx, env, "PEERDB_GROUP_NORMALIZE")
 }
 
+func PeerDBPGPGJSONBRecordMerge(ctx context.Context, env map[string]string) (bool, error) {
```
serprex: this doesn't need to be behind FF, change is solid, you tested, it has no regression, it has no compatibility issue

flow/e2e/postgres_test.go
```diff
@@ -154,15 +154,6 @@ func (s PeerFlowE2ETestSuitePG) testTypes(suffix string, system protos.TypeSyste
 		}
 		return err == nil
 	})
-	// Q type system converts timetz to UTC, PG type system preserves original offset
-	var c36 string
-	require.NoError(s.t, s.Conn().QueryRow(s.t.Context(), "select c36 from "+dstTableName).Scan(&c36))
-	if system == protos.TypeSystem_PG {
```
serprex: why remove timetz? this is used as an example of PGPG fidelity on weird types

serprex (approved):

PR #4168, opened 2026-04-13
NormalizeLoop: fix normalize race

Amogh-Bharadwaj:
There's a race condition in normalizeLoop with the LastChan signaling mechanism.

The race:

- `normalizeLoop` is inside `startNormalize` processing batch 346
- Sync thread completes batch 347 and calls `normRequests.Update(347)` — this stores value 347, creates a new channel, and closes the old channel
- `startNormalize` for 346 returns, normalizeLoop goes back to the top
- `normalizeRequests.Wait()` returns the new (unclosed) channel created in step 2
- `normalizeLoop` blocks on `<-ch` — but this channel will only be closed by the next `Update()` (batch 348)
- So the signal for batch 347 is missed. Normalize 347 won't start until batch 348 is synced (which may never happen if traffic is low)

Check for pending work via Load() before calling Wait(). Only block on the channel if there's genuinely nothing to process.

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -728,58 +728,64 @@ func (a *FlowableActivity) normalizeLoop(
 	defer normalizeWaiting.Store(false)
 
 	for {
-		normalizeWaiting.Store(true)
-		ch := normalizeRequests.Wait()
-		if ch == nil {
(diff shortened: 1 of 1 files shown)


serprex (approved):

ilidemi:
LGTM

PR #4174, opened 2026-04-14
retry transient tcp error

jgao54:
This PR fix two things:
- handle error `code: 1001, message: std::__1::ios_base::failure: ios_base::clear: unspecified iostream_category error`. This is a transient networking error that can happen in s3 during read and usually recovers on single retry. We want to also categorize it as retryable to avoid starting over the snapshot after a single error.
- when checking for VM substring, use error.Message() so we can search substring in all the wrapped error messages as well. The current behavior is we would classify as Normalization Error which would also notify, but ViewError is more direct here.

### DIFF
--- a/flow/otel_metrics/otel_manager.go
+++ b/flow/otel_metrics/otel_manager.go
@@ -80,7 +80,6 @@ const (
 	UnchangedToastValuesCounterName      = "unchanged_toast_values"
 	CodeNotificationCounterName          = "code_notification"
 	ServerWalEndLagGaugeName             = "wal_end_lag"
-	ClickHouseDDLExecutionGaugeName      = "clickhouse_ddl_execution_duration"
 )
 
 type Metrics struct {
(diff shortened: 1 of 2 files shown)


masterashu (approved):

flow/otel_metrics/otel_manager.go
```diff
@@ -80,6 +80,7 @@ const (
 	UnchangedToastValuesCounterName      = "unchanged_toast_values"
 	CodeNotificationCounterName          = "code_notification"
 	ServerWalEndLagGaugeName             = "wal_end_lag"
+	ClickHouseDDLExecutionGaugeName      = "clickhouse_ddl_execution_duration"
```
jgao54: oops, this was intended for another PR.

flow/pkg/clickhouse/query_retry.go
```diff
@@ -36,13 +36,30 @@ var retryableExceptions = map[chproto.Error]struct{}{
 	chproto.ErrKeeperException:            {},
 }
 
+// conditionallyRetryableExceptions are error codes that are only retryable
+// when the error message contains one of the specified substrings
+var conditionallyRetryableExceptions = map[chproto.Error][]string{
```
ilidemi: Just an idea, how about `retryableExceptionSubstrings`? Wouldn't need an explanation this way

ilidemi (approved):

PR #4178, opened 2026-04-14
Write permissions to claude action workflow

pfcoperez:
Note that, besides the safeguards using `author_association` the action has guardrails against external users invoking it 

https://github.com/PeerDB-io/peerdb/actions/runs/24410304258/job/71304922306#step:3:198


ilidemi (approved):
Tightening the permissions sounds great

PR #4180, opened 2026-04-15
Fix rerun of local tests

ilidemi:
- Cleanup of a toxi test to avoid port conflict on the next run
- Limit PG slot names as they silently get truncated and chop off the suffix, enforce that for future tests
- QoL improvement to not rebuild flow services on test file changes


pfcoperez (approved):
Nice!

Amogh-Bharadwaj (approved):

PR #4183, opened 2026-04-16
fix(e2e): use CI_MYSQL_VERSION as MySQL peer name

ilidemi:
## Summary
- All MySQL e2e tests shared a hardcoded peer name `"mysql"`. With `ON CONFLICT DO NOTHING` on peer creation, whichever flavor ran first (mysql-gtid, mysql-pos, maria) won and subsequent flavors silently used the wrong connection config.
- This caused schema fetches against the wrong MySQL instance, resulting in empty table schemas and `ORDER BY tuple()` errors on ClickHouse `ReplacingMergeTree` tables.
- Fix: use `CI_MYSQL_VERSION` env var as the peer name so each flavor gets a unique peer.

### DIFF
--- a/flow/e2e/mysql_rds_binlog_test.go
+++ b/flow/e2e/mysql_rds_binlog_test.go
@@ -72,6 +72,11 @@ func TestMySQLRDSBinlog(t *testing.T) {
 }
 
 func (s MySQLRDSBinlogAPITestSuite) TestMySQLRDSBinlogValidation() {
+	// Register cleanup before creating the table so it is always dropped even if the test fails early.
+	s.t.Cleanup(func() {
+		_ = s.source.Exec(context.Background(), "DROP TABLE IF EXISTS mysql.rds_configuration")
+	})
+
 	require.NoError(s.t, s.source.Exec(s.t.Context(),
 		fmt.Sprintf("CREATE TABLE %s(id int primary key, val text)", AttachSchema(s, "valid"))))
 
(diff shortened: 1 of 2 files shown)


pfcoperez (approved):
I was bitten by this once! Didn't happen again because I had cleared the catalog data volume in my local dev machine.

pfcoperez:
@claude https://github.com/PeerDB-io/peerdb/actions/runs/24497122116/job/71594583521?pr=4183 is failing because it seems that some tests match errors against hardcoded message samples "failed to validate source connector mysql". Please update these tests so the expected message is computed dynamically depending on the flow configuration

pfcoperez (approved):
AI changes LGTM too

pfcoperez:
@claude can you debug the latest failures?

```
        	Error:      	Received unexpected error:
        	            	rpc error: code = Internal desc = invalid mirror: rpc error: code = FailedPrecondition desc = failed to validate source connector mysql-gtid: binlog configuration error: RDS/Aurora setting 'binlog retention hours' should be at least 24, currently unset
```

e.g: https://github.com/PeerDB-io/peerdb/actions/runs/24497122116/job/71620555107#step:28:2869

flow/e2e/mysql_rds_binlog_test.go
```diff
@@ -72,6 +72,11 @@ func TestMySQLRDSBinlog(t *testing.T) {
 }
 
 func (s MySQLRDSBinlogAPITestSuite) TestMySQLRDSBinlogValidation() {
+	// Register cleanup before creating the table so it is always dropped even if the test fails early.
```
pfcoperez, line 75: I haven't debugged the issues myself, I just played the meat bot a bit. I don't think the AI fixes are enough, this requires human intervention. Leaving the two claude comments in the branch from now for @ilidemi to take a peek at its proposals.

pfcoperez (changes requested):

PR #4188, opened 2026-04-16
fix(helm-chart-bump): change semver bump version for helm chart version bump.

masterashu:
The original version bump action was considering the peerdb repo's ref for bumping and was deciding to create a pre-release version instead of normal 
patch version. Its now replaced with a different action which should work better.


masterashu:
Fixes the issue which occured in https://github.com/PeerDB-io/peerdb-enterprise/pull/65

ilidemi (approved):

PR #4192, opened 2026-04-17
Tests reliability improvements: Connection parameters extraction logic unification

pfcoperez:
This PR unifies test DB connection parameters acquisition into common functions used across all test suites.

## Ready when

- [x] Tested
- [x] Exhaustivity is confirmed.

## What does this PR resolve?

Closes: https://linear.app/clickhouse/issue/DBI-674

### DIFF
--- a/flow/connectors/clickhouse/s3_iam_role_test.go
+++ b/flow/connectors/clickhouse/s3_iam_role_test.go
@@ -3,7 +3,6 @@ package connclickhouse
 import (
 	"fmt"
 	"os"
-	"strconv"
 	"testing"
 	"time"
 
@@ -11,6 +10,7 @@ import (
 
 	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/internal"
 )
 
 func TestIAMRoleCanIssueSelectFromS3(t *testing.T) {
@@ -43,23 +43,10 @@ func TestIAMRoleCanIssueSelectFromS3(t *testing.T) {
 	t.Setenv("PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME", os.Getenv(bucketNameEnvVar))
 	ctx := t.Context()
 
-	chHost := "localhost"
-	if host := os.Getenv("CI_CLICKHOUSE_HOST"); host != "" {
-		chHost = host
-	}
-	var chPort uint32 = 9000
-	if port := os.Getenv("CI_CLICKHOUSE_NATIVE_PORT"); port != "" {
-		p, err := strconv.ParseUint(port, 10, 32)
-		if err != nil {
-			t.Fatalf("invalid CI_CLICKHOUSE_NATIVE_PORT %q: %v", port, err)
-		}
-		chPort = uint32(p)
-	}
-
 	conn, err := NewClickHouseConnector(ctx, nil,
 		&protos.ClickhouseConfig{
-			Host:       chHost,
-			Port:       chPort,
+			Host:       internal.ClickHouseTestHost(),
+			Port:       internal.ClickHouseTestPort(),
 			Database:   "default",
 			DisableTls: true,
 		})
--- a/flow/connectors/clickhouse/s3_iam_role_test.go
+++ b/flow/connectors/clickhouse/s3_iam_role_test.go
@@ -33,13 +33,7 @@ func TestIAMRoleCanIssueSelectFromS3(t *testing.T) {
 	} {
 		t.Setenv(envVar, "")
 	}
-	for _, envVar := range []string{
-		"AWS_ACCESS_KEY_ID",
-		"AWS_SECRET_ACCESS_KEY",
-		"AWS_SESSION_TOKEN",
-	} {
-		t.Setenv(envVar, os.Getenv("FLOW_TESTS_"+envVar))
-	}
+	internal.SetupFlowAWSCredentialsFromEnv(t)
 	t.Setenv("PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME", os.Getenv(bucketNameEnvVar))
 	ctx := t.Context()
 
--- a/flow/connectors/mysql/ssh_keepalive_test.go
+++ b/flow/connectors/mysql/ssh_keepalive_test.go
@@ -2,7 +2,6 @@ package connmysql
 
 import (
 	"context"
-	"os"
 	"strconv"
 	"sync/atomic"
 	"testing"
@@ -31,13 +30,9 @@ const (
 
 func resolveMySQL(t *testing.T) (string, uint32, string) {
 	t.Helper()
-
-	mysqlRootPass := "cipass"
-	if envPass := os.Getenv("CI_MYSQL_ROOT_PASSWORD"); envPass != "" {
-		mysqlRootPass = envPass
-	}
-
-	return internal.MySQLTestHostWithFallback("mysql"), internal.MySQLTestPortWithFallback(3306), mysqlRootPass
+	return internal.MySQLTestHostWithFallback("mysql"),
+		internal.MySQLTestPortWithFallback(3306),
+		internal.MySQLTestRootPasswordWithFallback("cipass")
 }
 
 // Connector -> Toxi -> SSH -> MySQL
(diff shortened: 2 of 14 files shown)


flow/e2e/mongo.go
```diff
@@ -99,34 +99,30 @@ func GetTestDatabase(suffix string) string {
...
-	mongoAdminUri := os.Getenv("CI_MONGO_ADMIN_URI")
-	require.NotEmpty(t, mongoAdminUri, "missing CI_MONGO_ADMIN_URI env var")
-	mongoAdminUsername := os.Getenv("CI_MONGO_ADMIN_USERNAME")
-	require.NotEmpty(t, mongoAdminUsername, "missing CI_MONGO_ADMIN_USERNAME env var")
-	mongoAdminPassword := os.Getenv("CI_MONGO_ADMIN_PASSWORD")
-	require.NotEmpty(t, mongoAdminPassword, "missing CI_MONGO_ADMIN_PASSWORD env var")
+	admin := internal.MongoAdminTestCredentials()
+	require.NotEmpty(t, admin.URI, "missing CI_MONGO_ADMIN_URI env var")
+	require.NotEmpty(t, admin.Username, "missing CI_MONGO_ADMIN_USERNAME env var")
+	require.NotEmpty(t, admin.Password, "missing CI_MONGO_ADMIN_PASSWORD env var")
```
ilidemi: Should these validations (+ below) move into the helper? Or is the situation messy enough that only this test would actually require them

flow/connectors/clickhouse/s3_iam_role_test.go
ilidemi, line 41: These could use the new helper too?

flow/internal/test_env.go
```diff
@@ -0,0 +1,118 @@
...
+
+func MySQLTestVersion() string {
+	return os.Getenv("CI_MYSQL_VERSION")
+}
+
+func MySQLTestVersionIsMaria() bool {
+	return MySQLTestVersion() == "maria"
+}
+
+func MySQLTestVersionIsMysqlPos() bool {
```
ilidemi: Nit: capitalization should be consistent

flow/e2e/mysql.go
```diff
@@ -25,7 +24,7 @@ type MySqlSource struct {
 
 func SetupMySQL(t *testing.T, suffix string) (*MySqlSource, error) {
 	t.Helper()
-	myVersion := os.Getenv("CI_MYSQL_VERSION")
+	myVersion := internal.MySQLTestVersion()
 	if myVersion == "" {
 		t.Error("Expected CI_MYSQL_VERSION to be set")
```
ilidemi, line 29: Would be good to do a scrub that the env variables are validated in the same place they're read

flow/connectors/postgres/rds_test.go
```diff
@@ -1,39 +1,28 @@
...
 	"testing"
 
 	"github.com/stretchr/testify/require"
 
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/internal"
 )
 
 func TestAwsRDSIAMAuthConnectForPostgres(t *testing.T) {
 	t.Skip("flaky")
```
ilidemi: Would be good to validate that this and the other one still pass when not skipped (maybe it was, just going off of PR commits)

flow/connectors/mysql/ssh_keepalive_test.go
```diff
@@ -35,26 +34,14 @@ func setupMySQLConnectorWithProxy(ctx context.Context, t *testing.T, proxyName s
...
 	mysqlRootPass := "cipass"
 	if envPass := os.Getenv("CI_MYSQL_ROOT_PASSWORD"); envPass != "" {
 		mysqlRootPass = envPass
 	}
 
 	connector, err := NewMySqlConnector(ctx, &protos.MySqlConfig{
-		Host:     mysqlHost,
-		Port:     mysqlPort,
+		Host:     internal.MySQLTestHostWithFallback("mysql"),
+		Port:     internal.MySQLTestPortWithFallback(3306),
```
ilidemi: I don't think we need fallback, let's just require the env as it's all automagic now

ilidemi (approved):

ilidemi:
Should catalog.go GetAncillaryPostgresConfigFromEnv also be moved to test_env.go?

flow/connectors/mysql/ssh_keepalive_test.go
pfcoperez: This fallback is for CI, I can make it explicit at flow.yml but that expands the scope of the PR even further.

pfcoperez:
> Should catalog.go GetAncillaryPostgresConfigFromEnv also be moved to test_env.go?

I was tempted to do this in the same PR too but finally decided not to given that this is production code. This PR factors out everything that load tests env vars not production code. `GetAncillaryPostgresConfigFromEnv` is a method used multiple times in business logic.

flow/connectors/mysql/ssh_keepalive_test.go
ilidemi: Ahhh right. Got confused

ilidemi:
> GetAncillaryPostgresConfigFromEnv is a method used multiple times in business logic.

Production postgres would be peers created at runtime, right? I'm only seeing test usages
Screenshot 2026-04-21 at 3 28 17 PM

Keeping Get**Catalog**PostgresConfigFromEnv as is makes sense

flow/connectors/postgres/rds_test.go
pfcoperez: That's a good point. One green is enough, added an ephemeral commit to try it out: 84f4ae34a602ccc367cd6781267629f01a26a46d

flow/e2e/mysql.go
pfcoperez: Indeed, I did a quick pattern search like `find flow -type f -name '*.go' -exec grep -F -H 'require.NotEmpty' {} \;` and followed up with a Claude Driven one.

I found some side cases that I've changed so, with the current change set, there are no more obvious things to move in. 

If I've left something behind we can always keep moving things.

pfcoperez:
> GetAncillaryPostgresConfigFromEnv



> > GetAncillaryPostgresConfigFromEnv is a method used multiple times in business logic.
> 
> Production postgres would be peers created at runtime, right? I'm only seeing test usages Screenshot 2026-04-21 at 3 28 17 PM
> 
> Keeping Get**Catalog**PostgresConfigFromEnv as is makes sense

@ilidemi  You are absolutely right! I did run the search for the method above it! That's embarrassing, moved in 6058d5c61b158ac2e2c1a711891e1ce74f108f02

flow/internal/test_env.go
pfcoperez: Eagle eyes! Addressed in 552396bc067173c404a00cf8eea7e0b8bc632106

flow/internal/test_env.go
```diff
@@ -0,0 +1,118 @@
...
+	t.Setenv("AWS_SESSION_TOKEN", os.Getenv("FLOW_TESTS_RDS_IAM_AUTH_AWS_SESSION_TOKEN"))
+}
+
+func RDSIAMAuthAssumeRoleConfig() *protos.AwsAuthenticationConfig {
+	return &protos.AwsAuthenticationConfig{
+		AuthType: protos.AwsIAMAuthConfigType_IAM_AUTH_ASSUME_ROLE,
+		AuthConfig: &protos.AwsAuthenticationConfig_Role{
+			Role: &protos.AWSAuthAssumeRoleConfig{
+				AssumeRoleArn:  os.Getenv("FLOW_TESTS_RDS_IAM_AUTH_ASSUME_ROLE"),
+				ChainedRoleArn: new(os.Getenv("FLOW_TESTS_RDS_IAM_AUTH_CHAINED_ROLE")),
```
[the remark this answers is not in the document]
pfcoperez: Addressed manually in recent changes.

flow/connectors/postgres/rds_test.go
pfcoperez: The test still occasionally pass:

image

The query compares it with a control PR https://github.com/PeerDB-io/peerdb/pull/4214 created just for this purpose.

flow/connectors/clickhouse/s3_iam_role_test.go
pfcoperez: Using just one of these values as an example, the concatenation results in `[FLOW_TESTS_AWS_ACCESS_KEY_ID](https://github.com/PeerDB-io/peerdb/blob/1a2dfd10b9f142fec6f970828b10f512fd95a757/.github/workflows/flow.yml#L534)`, the helper `SetupRDSIAMAuthAWSCredentials` reads `[FLOW_TESTS_RDS_IAM_AUTH_AWS_ACCESS_KEY_ID](https://github.com/PeerDB-io/peerdb/blob/1a2dfd10b9f142fec6f970828b10f512fd95a757/.github/workflows/flow.yml#L521)`.

In the current CI setup, according to `flow.yml` both values will be the same. I still feel reluctant to effectively change the source variables used in this test (even if they have the same value).   But the load logic is the same,  I've also factored out this behaviour and implemented specific helpers using it at 10c9fdbba7c3480950337b6f9b76a1aa0a9b9b34

PR #4194, opened 2026-04-18
Return structured gRPC errors on source table missing

ilidemi:
Helps to unify the handling across connectors and consume on the caller without string matching.

Also clean up unused api error and move string constants to pkg. Would be nicer to have everything in protos but string constants are not supported there and not all our constants are uppercase to be enums.

### DIFF
--- a/flow/connectors/core.go
+++ b/flow/connectors/core.go
@@ -45,7 +45,7 @@ type MirrorSourceValidationConnector interface {
 	GetTableSchemaConnector
 
 	// ValidateMirrorSource checks that the source is ready to replicate the configured tables.
-	// MUST return *common.SourceTableMissingError when a mapped source table is absent.
+	// MUST return *common.SourceTablesMissingError when a mapped source table is absent.
 	ValidateMirrorSource(context.Context, *protos.FlowConnectionConfigsCore) error
 }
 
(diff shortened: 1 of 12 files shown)


ilidemi:
Will add a change to return all missing tables at once, resulting contract might be different

flow/connectors/core.go
```diff
@@ -44,6 +44,8 @@ type ValidationConnector interface {
 type MirrorSourceValidationConnector interface {
 	GetTableSchemaConnector
 
+	// ValidateMirrorSource checks that the source is ready to replicate the configured tables.
+	// MUST return *common.SourceTablesMissingError when a mapped source table is absent.
 	ValidateMirrorSource(context.Context, *protos.FlowConnectionConfigsCore) error
```
jgao54, line 49: for my understanding, what's the reason for not just changing the return type here to `*common.SourceTablesMissingError`

jgao54: nvm, i see that the function can return other types of errors as well.

jgao54: eventually we will get to a place where we can replace all the `error.fmt` with structured errors 💪

jgao54 (approved):

PR #4195, opened 2026-04-18
fix mysql stuck on syncer.Close()

jgao54:
When mysql connector is set up with ssh tunnel, there is a possible deadlock observed where the syncer is never closed and PullRecords hangs indefinitely. 

```

  | 2026-04-10 23:09:01.794180854 | INFO | [mysql] PullRecords finished streaming |
 
  | 2026-04-10 23:09:01.794210647 | INFO | syncer is closing... |  
 
  | 2026-04-10 23:09:01.794213666 | WARN | could not set read deadline |  
 
  | 2026-04-10 23:09:01.835046706 | INFO | written records to OCF |
 
  | 2026-04-10 23:09:01.868201938 | ERROR | kill connection |  
 
  | 2026-04-10 23:09:01.868229754 | INFO | kill last connection |  
 
  | 2026-04-10 23:10:01.834992414 | INFO | written records to OCF | 
 
  | 2026-04-10 23:11:01.835059135 | INFO | written records to OCF | 
 
  | 2026-04-10 23:12:01.835527845 | INFO | written records to OCF | 
 
  | 2026-04-10 23:13:01.834876775 | INFO | written records to OCF | 
 
  | 2026-04-10 23:14:01.835584592 | INFO | written records to OCF | 
...
```

Notice the `could not set read deadline` warning, and the `kill connection` error, both coming from mysql binlogSyncer:
- the `could not set read deadline` is failing with error message "ssh: tcpChan: deadline not supported"
- the `kill connection` is failing with error message ""error":"ERROR 1094 (HY000): Unknown thread id: 743468""

What happened here is the library we are using for processing mysql replication, go-msyql, has a bug that can manifest if it's using SSH tunnel and if the thread is already reaped by the MySQL server. This let the `syncer.Close()` to hang indefinitely waiting because go-mysql thinks the connection is still alive (and the ssh tunneling connection very much is) so it gets stuck on `wg.Wait()` inside `BinlogSyncer.close()`

This is a different case than https://github.com/PeerDB-io/peerdb/pull/4067, because SSH tunnel itself is healthy.


flow/connectors/mysql/ssh_keepalive_test.go
```diff
@@ -270,3 +271,49 @@ func TestMySQLSSHKeepaliveCDCCloseHang(t *testing.T) {
...
+
+	pos, err := connector.GetMasterPos(ctx)
+	require.NoError(t, err)
+	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos)
+	require.NoError(t, err)
+
+	// Let CDC streaming establish before blocking the tunnel.
+	time.Sleep(2 * time.Second)
+
+	// Black-hole the SSH transport so syncer.Close() hangs
```
ilidemi: If SSH is healthy in this case, should toxi be declared on top of MySQL?

jgao54: that would be more direct, in this case was trying to reuse existing code for simplicity given what we are testing here is kind of trivial which is the watchdog works. maybe it's not worth a whole e2e test at all and better to test something abstract.

jgao54: let me do that.

flow/connectors/mysql/cdc.go
```diff
@@ -318,6 +319,19 @@ func (c *MySqlConnector) startCdcStreamingGtid(
...
+// closeSyncerWithTimeout closes the syncer with a 10s timeout and, on timeout,
+// force-closes the SSH tunnel if SSH tunnel exists. go-mysql's BinlogSyncer.close()
+// can hang indefinitely when its two unblock mechanisms fail (SetReadDeadline fails
+// for ssh tunnel, and killing a connection that has already been reaped by the server
+// but is not propagated to the client). This can lead to syncer.Close() stuck indefinitely.
+// TODO: better to fix properly upstream
+func (c *MySqlConnector) closeSyncer(syncer *replication.BinlogSyncer, deadline time.Duration) {
+	timeout.CloseWithTimeout(syncer.Close, deadline, func() {
+		c.logger.Error("[mysql] syncer.Close hung, force-closing SSH tunnel")
+		_ = c.ssh.Close()
```
ilidemi, line 331: Out of curiosity, what happens when the ssh tunnel is not closed here? Is the tunnel preserved between batches?

flow/connectors/mysql/cdc.go
```diff
@@ -318,6 +319,19 @@ func (c *MySqlConnector) startCdcStreamingGtid(
 	return syncer, stream, gset, mysql.Position{}, nil
 }
 
+// closeSyncerWithTimeout closes the syncer with a 10s timeout and, on timeout,
+// force-closes the SSH tunnel if SSH tunnel exists. go-mysql's BinlogSyncer.close()
+// can hang indefinitely when its two unblock mechanisms fail (SetReadDeadline fails
```
ilidemi, line 324: What is the default SetReadDeadline, is it under 10 seconds?

jgao54: default is 100ms.  We could tighten the feedback loop if 10s feels too generous of a wait here.  

from the upstream code:
```
func (b *BinlogSyncer) close() {
	if b.isClosed() {
		return
	}

	b.cfg.Logger.Info("syncer is closing...")

	b.running = false
	b.cancel()

	if b.c != nil {
		err := b.c.SetReadDeadline(utils.Now().Add(100 * time.Millisecond))
		if err != nil {
			b.cfg.Logger.Warn("could not set read deadline", slog.Any("error", err))
		}
	}

	// kill last connection id
	if b.lastConnectionID > 0 {
		// Use a new connection to kill the binlog syncer
		// because calling KILL from the same connection
		// doesn't actually disconnect it.
		c, err := b.newConnection(context.Background())
		if err == nil {
			b.killConnection(c, b.lastConnectionID)
			c.Close()
		}
	}

	b.wg.Wait()

	if b.c != nil {
		b.c.Close()
	}

	b.cfg.Logger.Info("syncer is closed")
}

```
seems like it's really intended to close immediately -- either triggered by deadline or by killing the connection, both of which would unblock `readPacket()`  (and unblock `b.wg.Wait()`) which is where it is hanging before this PR.

flow/connectors/mysql/cdc.go
jgao54: that's a good question. each batch should open a new ssh tunnel (since we are creating a new mysql connector), so seems like it would result in a leaked connection (and ssh keepalive would still think everything is healthy so would not help in this case)

flow/connectors/mysql/cdc.go
ilidemi: 10 seconds is good, given it's only when the connection is actually dead

flow/connectors/mysql/ssh_keepalive_test.go
ilidemi: I'm not sure the abstraction is needed here, given how short it is, and it comes with two callbacks. The scenario we want to test is being able to finish the batch when MySQL connection is down, and if the infra doesn't allow it, the infra needs to be better.

Better infra at https://github.com/PeerDB-io/peerdb/pull/4207

flow/connectors/mysql/cdc.go
ilidemi: Not leaking SSH connections sounds great

ilidemi (approved):

PR #4196, opened 2026-04-18
bugfix to ensure resync cannot fail due to table size too large

jgao54:
Scenario this change fixes:

1. User resynced table - a typical resync writes to a _resync table and at the end exchange it with the original table, and drop the old table.
2. Before the resync fully complete, user triggered resync again, this can leave the previous *_resync table with 1TB of data
3. during the second resync, the CREATE OR REPLACE table during setup fails. This under the hood does the following:
    a. create a new table with name _tmp_replace_*
    b. exchange new table with old table
    c.  drop old table
4. if the old table exceeds `max_table_size_to_drop`, 3c fails and the _tmp_replace_* table is orphaned. 
5. on retry, the swapped *_resync table is empty and CREATE OR REPLACE succeeds

The fix ensures drop would not fail because table is too large. Arguably cleaner if we make this a client-level setting, but safer to assess it on the per-query level since we are authorizing dropping 1TB+ table. 

Fixes: DBI-698

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -73,8 +73,16 @@ func (c *ClickHouseConnector) SetupNormalizedTable(
 		return false, fmt.Errorf("error while generating create table sql for destination ClickHouse table: %w", err)
 	}
 
+	qryCtx := ctx
+	if config.IsResync {
+		// CREATE OR REPLACE TABLE (used by resync) internally would create a NEW table with name
+		// <_tmp_replace_*>, atomic swap with the existing table, and then drop the existing table.
+		// Setting max_table_size_to_drop = 0 ensure the implicit drop can't fail, since it can be
+		// non-empty if a resync is triggered on top of another resync that is mid-snapshot.
+		qryCtx = clickhouse.Context(ctx, clickhouse.WithSettings(clickhouse.Settings{string(chinternal.SettingMaxTableSizeToDrop): 0}))
+	}
 	for _, sql := range normalizedTableCreateSQL {
-		if err := c.execWithLogging(ctx, sql); err != nil {
+		if err := c.execWithLogging(qryCtx, sql); err != nil {
 			return false, fmt.Errorf("[clickhouse] error while creating destination ClickHouse table: %w", err)
 		}
 	}
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -73,16 +73,8 @@ func (c *ClickHouseConnector) SetupNormalizedTable(
 		return false, fmt.Errorf("error while generating create table sql for destination ClickHouse table: %w", err)
 	}
 
-	qryCtx := ctx
-	if config.IsResync {
-		// CREATE OR REPLACE TABLE (used by resync) internally would create a NEW table with name
-		// <_tmp_replace_*>, atomic swap with the existing table, and then drop the existing table.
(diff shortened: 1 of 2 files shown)


masterashu (approved):

flow/connectors/clickhouse/normalize.go
```diff
@@ -73,8 +73,16 @@ func (c *ClickHouseConnector) SetupNormalizedTable(
 		return false, fmt.Errorf("error while generating create table sql for destination ClickHouse table: %w", err)
 	}
 
+	qryCtx := ctx
+	if config.IsResync {
+		// CREATE OR REPLACE TABLE (used by resync) internally would create a NEW table with name
+		// <_tmp_replace_*>, atomic swap with the existing table, and then drop the existing table.
+		// Setting max_table_size_to_drop = 0 ensure the implicit drop can't fail, since it can be
+		// non-empty if a resync is triggered on top of another resync that is mid-snapshot.
+		qryCtx = clickhouse.Context(ctx, clickhouse.WithSettings(clickhouse.Settings{string(chinternal.SettingMaxTableSizeToDrop): 0}))
```
pfcoperez: I was researching on this option and found that overriding `max_table_size_to_drop` setting at query time is available only after ClickHouse  `v23.12`  --> https://clickhouse.com/docs/whats-new/changelog/2023#2312

I understand that someone using such an ancient version is a low-probability corner case. Anyway, even if that happens, the setting will be just ignored and they would not get any worse than the current situation, right?

pfcoperez (approved):
LGTM. Would it make sense to put together a small test case where the ClickHouse target is set to use a ridiculously small `max_table_size_to_drop` value by default and where we exercise this change?

flow/connectors/clickhouse/normalize.go
pfcoperez: Actually, I was wrong in my assumption, I did run a little experiment:

```sql
SELECT COUNT(*) FROM table SETTINGS i_dont_exist=0;
```

```
Setting i_dont_exist is neither a builtin setting nor started with the prefix 'custom_' registered for user-defined settings.
``` 

So I think this would indeed fail on older versions. Maybe we can add a version check to the condition?

ilidemi: We have a whole framework for version-aware CH settings under https://github.com/PeerDB-io/peerdb/blob/main/flow/internal/clickhouse/settings.go. I think this logic should move to generateCreateTableSQLForNormalizedTable and be appended to SQL like the other cases we use settings

jgao54: @pfcoperez  good call on checking ch version before adding it to the context. 

@ilidemi i was worried that `SETTINGS max_table_size_to_drop = 0` get applied to the CREATE TABLE as a table-level storage setting, but looks like clickhouse handles this automagically [here](https://github.com/ClickHouse/ClickHouse/blob/6ff9d72dbf72c3b5261b3e84d70687e8159b7d7c/src/Interpreters/InterpreterSetQuery.cpp#L101-L104)

ilidemi (approved):

jgao54:
@pfcoperez  re: testing

I ran this in clickhouse on a non-empty table with  max_table_size_to_drop = 1 which repro-ed the error, and with max_table_size_to_drop = 0 it succeeded.

For testing, this feels more like a _clickhouse_ functionality test than peerdb test, but i realized we we don't have a unit test coverage for `generateCreateTableSQLForNormalizedTable` so creating a basic one to avoid regressions.

PR #4198, opened 2026-04-19
classify changestream creation error

jgao54:
Can happen during change stream creation (i.e. connection pool exhausted on the server). Notify customer when this happens.

Fixes: DBI-691


flow/alerting/classifier.go
```diff
@@ -221,6 +222,10 @@ var (
 	ErrorNotifyChangeStreamHistoryLost = ErrorClass{
 		Class: "NOTIFY_CHANGE_STREAM_HISTORY_LOST", action: NotifyUser,
 	}
+	// ErrorNotifyMongoServerSideTimeout fires when the Mongo driver rejects Watch because the computed maxTimeMS is <= 0
```
ilidemi, line 225: In which cases can this happen? What makes it auto-recover?

jgao54: originally i thought it had to do with server-side incoming connection exhausted, but this was not the case, new findings tracked [here](https://github.com/PeerDB-io/peerdb/pull/4213). 

will close this PR.

PR #4200, opened 2026-04-20
feat(clickhouse): add StagingStore abstraction for GCS staging support

### LINKED ISSUE #2769: GCP: Support GCS natively in PeerDB
Instead of supporting via S3 compatible endpoint, support GCS more natively.
Also support GCS function in ClickHouse

### LINKED ISSUE #2770: GCP: Support Getting Creds from GCP/GKE Workload Identity
Uses native GCP Auth for 1:1 compatibility with AWS

whites11:
## Summary

Resolves https://linear.app/clickhouse/issue/CLP-629/add-gcs-support-for-staging-bucket

Adds GCS support for ClickHouse CDC/snapshot staging, alongside the existing S3 path.

### What changed

- **`StagingStore` interface** (`utils/staging.go`) — abstracts Upload, TableFunctionExpr, DeletePrefix, Validate across cloud providers
- **`S3StagingStore`** (`utils/staging_s3.go`) — extracts existing S3 logic, zero behavior change
- **`GCSStagingStore`** (`utils/staging_gcs.go`) — native GCS SDK with Workload Identity auth; ClickHouse reads via `url()` table function with signed URLs
- **`staging.go`** (clickhouse connector) — `createStagingStore` routes to S3 or GCS based on `PEERDB_CLICKHOUSE_STAGING_PROVIDER` env var
- **`clickhouse.go`** — minimal diff: adds `staging` field, replaces inline S3 setup with `createStagingStore()` call, wraps session-token check in nil guard for non-S3 providers
- **`avro_sync.go`** — calls `store.Upload()` directly instead of legacy `WriteRecordsToS3`
- **`dynamicconf.go`** — `PeerDBClickHouseStagingBucketName` falls back to legacy `PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME`; adds `PEERDB_CLICKHOUSE_STAGING_PROVIDER` config

### How GCS staging works

1. PeerDB writes Avro files to GCS using the native Go SDK (Workload Identity on GKE)
2. PeerDB generates a 15-minute signed URL for each staged file
3. ClickHouse reads the file via `url('https://storage.googleapis.com/...?X-Goog-Signature=...', 'Avro')`

### Env vars

| Variable | Default | Description |
|----------|---------|-------------|
| `PEERDB_CLICKHOUSE_STAGING_PROVIDER` | `s3` | `s3` or `gcs` |
| `PEERDB_CLICKHOUSE_STAGING_BUCKET_NAME` | (empty) | Unified bucket name; falls back to `PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME` for S3 |

## Test plan

- [x] All existing unit tests pass (zero regression)
- [x] New unit tests for `S3StagingStore` (construction, table function expr, endpoint override, session tokens)
- [x] Tested on GCP dev: PG CDC pipe with GCS staging + directory TLS + GCP KMS — 5/5 CDC rows synced
- [x] Tested on AWS dev: PG CDC pipe with S3 staging + Lambda certs — confirmed no regression on existing AWS path


flow/internal/dynamicconf.go
```diff
@@ -710,6 +726,14 @@ func PeerDBClickHouseAWSS3BucketName(ctx context.Context, env map[string]string)
 	return dynLookup(ctx, env, "PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME")
 }
 
+func PeerDBClickHouseStagingProvider(ctx context.Context, env map[string]string) (string, error) {
+	return dynLookup(ctx, env, "PEERDB_CLICKHOUSE_STAGING_PROVIDER")
+}
+
+func PeerDBClickHouseStagingBucketName(ctx context.Context, env map[string]string) (string, error) {
+	return dynLookup(ctx, env, "PEERDB_CLICKHOUSE_STAGING_BUCKET_NAME")
```
whites11, line 730: does it make sense to move the fallback logic in this function? meaning we check PEERDB_CLICKHOUSE_STAGING_BUCKET_NAME and default to PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME if empty?

flow/connectors/clickhouse/staging.go
```diff
@@ -0,0 +1,137 @@
...
+		return createS3StagingStore(ctx, config, "", chVersion)
+	}
+
+	// Check environment-based staging provider config.
+	provider, err := internal.PeerDBClickHouseStagingProvider(ctx, env)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get staging provider: %w", err)
+	}
+
+	// Prefer unified PEERDB_CLICKHOUSE_STAGING_BUCKET_NAME, fall back to provider-specific env vars.
```
whites11: move this comment inside the function if missing

flow/connectors/clickhouse/staging.go
```diff
@@ -0,0 +1,137 @@
...
+// The credsProvider return is non-nil only for S3 and is kept for backward compatibility
+// with the session-token version check in createS3StagingStore.
+func createStagingStore(
+	ctx context.Context,
+	env map[string]string,
+	config *protos.ClickhouseConfig,
+	chVersion clickhouseproto.Version,
+) (utils.StagingStore, *utils.ClickHouseS3Credentials, error) {
+	// If user provided explicit S3 config, always use S3 staging.
+	if config.S3 != nil || config.S3Path != "" || config.AccessKeyId != "" {
```
whites11: i feel like this branch does not belong here. it should be in the createS3StaginStore function somehow am I wrong?

flow/connectors/clickhouse/staging.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func createS3StagingStore(
+	ctx context.Context,
+	config *protos.ClickhouseConfig,
+	unifiedBucketName string,
+	chVersion clickhouseproto.Version,
+) (utils.StagingStore, *utils.ClickHouseS3Credentials, error) {
+	var awsConfig utils.PeerAWSCredentials
```
whites11: i feel like the whole content of this func should be moved in the utils.NewS3StagingStore func. and actually this method should be deleted and just the utils one be called directly. WDYT?

flow/connectors/utils/staging.go
```diff
@@ -0,0 +1,36 @@
+package utils
+
+import (
+	"context"
+	"io"
+)
+
+var (
+	_ StagingStore = (*S3StagingStore)(nil)
```
whites11: this is ugly and suggests we have a wrong separation of concerns/interface definition

flow/connectors/clickhouse/staging_gcs.go
```diff
@@ -0,0 +1,150 @@
...
+	client   *storage.Client
+	bucket   string
+	prefix   string
+	fullPath string
+}
+
+func NewGCSStagingStore(ctx context.Context, bucketPath string) (*GCSStagingStore, error) {
+	// bucketPath is "gcs://bucket/prefix" or just "bucket/prefix"
+	path := strings.TrimPrefix(bucketPath, "gcs://")
+	path = strings.TrimPrefix(path, "gs://")
```
whites11, line 46: what is gs:// about?

flow/internal/dynamicconf.go
whites11: Done — moved the fallback into this function and added a doc comment. Also deleted PeerDBClickHouseAWSS3BucketName since it has no callers now.

flow/connectors/clickhouse/staging.go
whites11: Done — removed inline comment, added doc comment on PeerDBClickHouseStagingBucketName explaining the fallback.

flow/connectors/clickhouse/staging.go
whites11: Good call — removed the explicit S3 config check. The default provider is s3 (set in dynconf), so the switch default handles it.

flow/connectors/clickhouse/staging.go
whites11: I'd keep them separate. `createS3StagingStore` does ClickHouse-specific work that doesn't belong in `utils`:

1. Parses `protos.ClickhouseConfig` fields (S3, AccessKeyId, Endpoint, etc.) — this is ClickHouse connector config, not generic S3
2. Builds the bucket path from deployment UID + flow name — ClickPipes-specific convention
3. Validates session token vs ClickHouse version — ClickHouse version check

`NewS3StagingStore` is a clean constructor that takes `(bucketPath, creds)` — it doesn't know about ClickHouse config. Moving the config parsing into `utils` would couple the generic staging layer to ClickHouse protos.

flow/connectors/utils/staging.go
whites11: Agreed it's not pretty — it's a workaround for the iface linter which complains if an exported interface isn't used within its declaring package. The alternative was returning the interface from constructors, but the opaque linter rejected that. Happy to remove these if we can nolint the iface rule on the interface declaration instead.

flow/connectors/clickhouse/staging_gcs.go
whites11: gs:// is the standard Google Cloud Storage URI scheme (like s3:// for AWS). Added a clarifying comment.

flow/connectors/utils/staging.go
ilidemi: I would move the staging*.go files under clickhouse/ as they only have a single user

flow/connectors/clickhouse/avro_sync.go
```diff
@@ -38,38 +39,7 @@ func NewClickHouseAvroSyncMethod(
...
-	expr.WriteString(peerdb_clickhouse.QuoteLiteral(creds.AWS.AccessKeyID))
-	expr.WriteByte(',')
-	expr.WriteString(peerdb_clickhouse.QuoteLiteral(creds.AWS.SecretAccessKey))
-	if creds.AWS.SessionToken != "" {
-		expr.WriteByte(',')
-		expr.WriteString(peerdb_clickhouse.QuoteLiteral(creds.AWS.SessionToken))
-	}
-	expr.WriteString(",'Avro')")
-	return expr.String(), nil
+	return s.staging.TableFunctionExpr(ctx, avroFilePath, "Avro")
```
ilidemi: As we're only using Avro, can make it a constant in staging.go (or here), reference from the leaf funcs and inline this one

ilidemi: Inlining "Avro" into both leafs is ok too

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -57,95 +56,25 @@ func NewClickHouseConnector(
...
-		return fmt.Errorf("failed to create S3 client: %w", err)
-	}
-
-	object, err := utils.NewS3BucketAndPrefix(creds.BucketPath)
-	if err != nil {
-		return fmt.Errorf("failed to create S3 bucket and prefix: %w", err)
-	}
-
-	return utils.PutAndRemoveS3(ctx, s3Client, object.Bucket, object.Prefix)
+		credsProvider:    credsProvider,
```
ilidemi: I don't think we need credsProvider anymore

flow/connectors/clickhouse/staging.go
```diff
@@ -0,0 +1,130 @@
...
+	bucketName, err := internal.PeerDBClickHouseStagingBucketName(ctx, env)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get staging bucket name: %w", err)
+	}
+
+	switch strings.ToLower(provider) {
+	case "gcs":
+		store, err := createGCSStagingStore(ctx, bucketName)
+		return store, nil, err
+	default:
```
ilidemi, line 42: Let's rely on default `"s3"` from dynconf and do an explicit branch here and an exhaustiveness check

flow/connectors/clickhouse/staging_gcs.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	logger.Info("Deleted objects from GCS",
+		slog.String("bucket", g.bucket), slog.String("prefix", prefix))
+	return nil
+}
+
+func (g *GCSStagingStore) Validate(ctx context.Context) error {
+	testKey := g.prefix + "/_peerdb_check_" + time.Now().Format("20060102150405")
```
ilidemi: Let's use a guid as source of uniqueness, we can hit two mirror/pipe creations the same second

flow/connectors/utils/staging_s3.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return &S3StagingStore{
+		bucket:   s3o.Bucket,
+		prefix:   s3o.Prefix,
+		fullPath: bucketPath,
+		creds:    creds,
+	}, nil
+}
+
+func (s *S3StagingStore) Upload(ctx context.Context, env map[string]string, key string, body io.Reader) error {
```
ilidemi: A lot of this is a copy-paste from peerDBOCFWriter.WriteRecordsToS3 so that should be calling this

flow/connectors/clickhouse/avro_sync.go
```diff
@@ -407,19 +373,34 @@ func (s *ClickHouseAvroSyncMethod) writeToAvroFile(
...
-		return utils.AvroFile{}, fmt.Errorf("failed to write records to S3: %w", err)
+
+	r, w := io.Pipe()
+	defer r.Close()
+
+	var writeOcfError error
+	var numRows int64
+	go func() {
+		defer w.Close()
+		numRows, writeOcfError = ocfWriter.WriteOCF(ctx, env, w, typeConversions, numericTruncator)
```
ilidemi, line 391: peerDBOCFWriter.WriteRecordsToS3 has a panic handler that disappeared here? Code duplication bites us immediately

Let's try and refactor so that it's one implementation

flow/connectors/clickhouse/staging_gcs.go
```diff
@@ -0,0 +1,150 @@
...
+	"time"
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+
+	"github.com/PeerDB-io/peerdb/flow/internal"
+	peerdb_clickhouse "github.com/PeerDB-io/peerdb/flow/pkg/clickhouse"
+)
+
+const gcsSignedURLExpiry = 15 * time.Minute
```
ilidemi: AWS ones are 1 hour by default, let's give CH more time

flow/connectors/clickhouse/staging_gcs.go
ilidemi: nit: s3 supports empty prefix here with `strings.TrimPrefix(temporaryObjectPath, "/")`

flow/connectors/clickhouse/avro_sync.go
```diff
@@ -407,19 +373,34 @@ func (s *ClickHouseAvroSyncMethod) writeToAvroFile(
...
+	if err := s.staging.Upload(ctx, env, s3AvroFileKey, r); err != nil {
+		return utils.AvroFile{}, fmt.Errorf("failed to upload to staging: %w", err)
+	}
+	if writeOcfError != nil {
+		return utils.AvroFile{}, writeOcfError
 	}
 
-	return avroFile, nil
+	return utils.AvroFile{
+		StorageLocation: utils.AvroS3Storage,
```
ilidemi, line 402: This isn't always S3 right?

flow/connectors/clickhouse/avro_sync.go
```diff
@@ -407,19 +373,34 @@ func (s *ClickHouseAvroSyncMethod) writeToAvroFile(
 
 	var s3AvroFileKey string
```
ilidemi: stagingAvroFileKey

flow/connectors/clickhouse/avro_sync.go
ilidemi, line 1: pushDataToS3ForSnapshot, pushS3DataToClickHouseForSnapshot can be renamed. There is some sprawl in logs and configs too, they shouldn't mention S3 anymore

flow/connectors/utils/staging_gcs.go
```diff
@@ -0,0 +1,150 @@
...
+			return fmt.Errorf("failed to delete GCS object %s: %w", attrs.Name, err)
+		}
+	}
+
+	logger.Info("Deleted objects from GCS",
+		slog.String("bucket", g.bucket), slog.String("prefix", prefix))
+	return nil
+}
+
+func (g *GCSStagingStore) Validate(ctx context.Context) error {
```
ilidemi: Codex is flagging that one can have permissions for creating objects but not for signing - do we need to validate that too?

flow/internal/dynamicconf.go
```diff
@@ -706,7 +722,22 @@ func PeerDBSnowflakeAutoCompress(ctx context.Context, env map[string]string) (bo
...
+func PeerDBClickHouseStagingBucketName(ctx context.Context, env map[string]string) (string, error) {
+	name, err := dynLookup(ctx, env, "PEERDB_CLICKHOUSE_STAGING_BUCKET_NAME")
+	if err != nil {
+		return "", err
+	}
+	if name != "" {
+		return name, nil
+	}
+	// Fall back to legacy S3-specific env var.
 	return dynLookup(ctx, env, "PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME")
```
ilidemi, line 734: Let's move the fallback into the caller, it's very convenient to do that there.

ilidemi (commented):
Refactors are hard. @jgao54 please give it a good look after the next iteration

flow/connectors/utils/staging.go
whites11: Done — moved StagingStore interface, s3StagingStore, and gcsStagingStore into the clickhouse connector package. Types are now unexported. Removed the var _ conformance checks since the interface is used directly in the same package.

flow/connectors/clickhouse/clickhouse.go
whites11: Good catch — removed credsProvider entirely. The StagingStore interface covers everything now. Also cleaned up the utils import from clickhouse.go.

flow/connectors/clickhouse/staging.go
whites11: Done — explicit `case "s3", "":` branch, and `default` now returns an error for unknown providers.

flow/connectors/clickhouse/staging_gcs.go
whites11: Bumped to 1 hour to match AWS STS defaults.

flow/connectors/clickhouse/staging_gcs.go
whites11: Done — using uuid.NewString() now.

flow/connectors/utils/staging_s3.go
whites11: Agreed this is worth doing but it touches the S3/Snowflake connectors which are outside the scope of this PR. I'd prefer to do that as a follow-up refactor — making WriteRecordsToS3 delegate to s3StagingStore.Upload would be a clean second step.

flow/connectors/clickhouse/avro_sync.go
whites11: Good catch — restored the panic handler in the goroutine. Agree on the refactor to unify with WriteRecordsToS3, tracking as follow-up.

flow/connectors/clickhouse/staging_gcs.go
whites11: Done — added strings.TrimPrefix for empty prefix case.

flow/connectors/clickhouse/avro_sync.go
whites11: Right — renamed to AvroStagingStorage would be ideal but that's a broader change. Left as-is for now since it's an existing constant.

flow/connectors/clickhouse/avro_sync.go
whites11: Done — renamed to stagingAvroFileKey.

flow/connectors/clickhouse/avro_sync.go
whites11: Done — renamed pushDataToS3ForSnapshot → pushDataToStagingForSnapshot, pushS3DataToClickHouseForSnapshot → pushStagingDataToClickHouseForSnapshot, s3TableFunctionBuilder → stagingTableFunctionBuilder, and updated error/log messages.

flow/connectors/utils/staging_gcs.go
whites11: Good point. On GKE with Workload Identity, the SA needs both storage.objects.create (for uploads) and iam.serviceAccounts.signBlob (for signing). The Validate() method covers the upload path. We could add a signing check in Validate too — generate a test signed URL and HEAD it. Worth adding as a follow-up, WDYT?

flow/internal/dynamicconf.go
whites11: Done — moved fallback back to createStagingStore, restored PeerDBClickHouseAWSS3BucketName as a simple lookup.

flow/connectors/clickhouse/avro_sync.go
whites11: Done — inlined the one-liner and deleted the wrapper function.

flow/connectors/clickhouse/avro_sync.go
whites11: You're right, this isn't always S3. The `StorageLocation` field is only checked for `== AvroLocalStorage` (to delete temp files) and `!= AvroLocalStorage` (in Snowflake). S3 vs GCS doesn't matter functionally — it's just "remote, don't clean up locally".

`AvroGCSStorage` already exists in the enum (BigQuery connector uses it), but our ClickHouse code doesn't know which staging backend it's using since it goes through the `StagingStore` interface. I'd rather not add a `StorageLocation()` method to the interface just for this.

Happy to rename the constant to something like `AvroRemoteStorage` in a follow-up if you prefer, but it would touch BigQuery/Snowflake connectors too.

flow/connectors/clickhouse/avro_sync.go
```diff
@@ -393,33 +359,49 @@ func (s *ClickHouseAvroSyncMethod) writeToAvroFile(
...
+
+	r, w := io.Pipe()
+	defer r.Close()
+
+	var writeOcfError error
+	var numRows int64
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				writeOcfError = fmt.Errorf("panic occurred during WriteOCF: %v", r)
```
jgao54: nit: looks like the panic stacktrace was lost here, may be good idea to keep it for debuggability (despite duplicating the error logs)

jgao54:
```suggestion

				writeOcfError = fmt.Errorf("panic occurred during WriteOCF: %v", r)
				stack := string(debug.Stack())
				s.logger.Error("panic during WriteOCF", slog.Any("error", writeOcfError), slog.String("stack", stack))
```

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -33,11 +30,11 @@ import (
 
 type ClickHouseConnector struct {
 	*metadataStore.PostgresMetadata
-	database      clickhouse.Conn
-	logger        log.Logger
-	Config        *protos.ClickhouseConfig
-	credsProvider *utils.ClickHouseS3Credentials
```
jgao54, line 39: `ClickHouseS3Credentials` struct can be deleted altogether now that it's no longer used

jgao54 (approved):
lgtm

PR #4205, opened 2026-04-20
Update dependencies

ilidemi:


ilidemi:
Renovate seems to be handling at least the go part pretty well

PR #4216, opened 2026-04-22
Add go remote debugging support in TILT

masterashu:
Summary of Changes:
* Added go compiler flags to flow build command in flow.dockerfile (disabled optimizations, recommended for debugging)
* Added a new stage in flow.dockerfile with dlv (delve debugger) entrypoint which exposes endpoint to attach debugger to and updated the same in Tiltfile
* VSCode launch config which attaches the debugger and resolved breakpoints etc.


stacks/flow.Dockerfile
```diff
@@ -21,7 +25,8 @@ ENV CGO_ENABLED=1
 # Generate the typed handler wrapper
 RUN go generate
 ENV GOCACHE=/root/.cache/go-build
-RUN --mount=type=cache,target="/root/.cache/go-build" go build -o /root/peer-flow
+RUN --mount=type=cache,target="/root/.cache/go-build" go build ${DEBUG_BUILD:+-gcflags} ${DEBUG_BUILD:+"all=-N -l"} -o /root/peer-flow
+RUN --mount=type=cache,target="/root/.cache/go-build" go install github.com/go-delve/delve/cmd/dlv@latest
```
pfcoperez: Shouldn't executing this command be conditioned on `DEBUG_BUILD` too?

stacks/flow.Dockerfile
```diff
@@ -58,6 +63,17 @@ ENV PEERDB_VERSION_SHA_SHORT=${PEERDB_VERSION_SHA_SHORT}
...
+FROM flow-base AS flow-worker-debug
+ENV OTEL_METRIC_EXPORT_INTERVAL=10000
+ENV OTEL_EXPORTER_OTLP_COMPRESSION=gzip
+ARG PEERDB_VERSION_SHA_SHORT
+ENV PEERDB_VERSION_SHA_SHORT=${PEERDB_VERSION_SHA_SHORT}
+USER root
+COPY --from=builder /go/bin/dlv /usr/local/bin/dlv
+USER peerdb
+EXPOSE 40000
+ENTRYPOINT ["dlv", "--headless", "--continue", "--accept-multiclient", "--listen=:40000", "--api-version=2", "exec", "./peer-flow", "--", "worker"]
```
pfcoperez: Nit: Would it make sense to make the hard-coded port number '40000' configurable through an env var?

docker-compose-dev.yml
```diff
@@ -161,6 +161,8 @@ services:
       context: .
       dockerfile: stacks/flow.Dockerfile
       target: flow-worker
+    ports:
+      - 40000:40000
```
pfcoperez: Same comment about the port being hard coded.

stacks/flow.Dockerfile
```diff
@@ -58,6 +63,17 @@ ENV PEERDB_VERSION_SHA_SHORT=${PEERDB_VERSION_SHA_SHORT}
 ENTRYPOINT ["./peer-flow", "snapshot-worker"]
 
 
+FROM flow-base AS flow-worker-debug
```
pfcoperez: I am wondering if the CI steps building and pushing images can inadvertently get added a new image to build. I tend to think that's not the case because CI steps building images such us `customer-docker.yml` use https://github.com/PeerDB-io/peerdb/blob/main/docker-bake.hcl . 

`flow-worker-debug` is not in that build definition file so, by default, we are not going to build it.

If you can confirm, this is a no-comment. Otherwise, we might want to move `flow-worker-debug` and other future debug images to a separate Dockerfile.

Tiltfile
pfcoperez, line 24: Is there any reason to stop at flow-worker-debug? Why not including `flow-api` and others too? Is it because you want to start with one?

pfcoperez (approved):
This is fantastic! 
Tbh, I think this opens the doors for debugging beyond "just" vs code. For example, we can add the instructions to attach the debugger at https://github.com/PeerDB-io/peerdb/blob/main/.claude/skills/skills/local-debug/SKILL.md . This way we'd close the loop on automatic debugging which so far was limited to test code

stacks/flow.Dockerfile
masterashu: it won't happen unless we want to build & publish debug images

masterashu:
@pfcoperez For now, changed the ports in docker-compose for now. seeing how to make it dynamic based on env, also refactored dockerfile and added the debug for flow-api and flow-snapshot worker

Tiltfile
```diff
@@ -4,27 +4,35 @@ allow_k8s_contexts(k8s_context()) # to unblock local() in local set-ups with a K
...
     only=['flow/', 'stacks/flow.Dockerfile'],
     ignore=flow_ignore,
-    build_args={'PEERDB_VERSION_SHA_SHORT': os.getenv('PEERDB_VERSION_SHA_SHORT', 'unknown')},
+    build_args={'DEBUG_BUILD': resolve_env('DOCKER_GO_DEBUG_FLOW_API',''),'PEERDB_VERSION_SHA_SHORT': os.getenv('PEERDB_VERSION_SHA_SHORT', 'unknown')},
 )
 
 docker_build('flow-worker', '.',
     dockerfile='stacks/flow.Dockerfile',
-    target='flow-worker',
+    target='flow-worker-debug' if resolve_env('DOCKER_GO_DEBUG_FLOW_WORKER') == '1' else 'flow-worker',
```
pfcoperez: Surely, someone will try to use value `true`  just from pure inertia. I think we can protect a bit by allowing `1` or `true` .

stacks/flow.Dockerfile
pfcoperez: @masterashu you added a 👍  to this comment but the command still runs unconditionally, did you forget to add that or you think we can skip?

masterashu: Yeah, i was thinking of moving this in the flow-base-debug stage, but its better to have it in build stage itself, added a condition now.

masterashu:
hey @pfcoperez made some improvements.

Tiltfile
masterashu: Added for flow-snapshot worker and flow-api as well,

PR #4234, opened 2026-04-24
fix(ctid-load): fix for mixed case partitioned tables

Amogh-Bharadwaj:
### Why
When the CTID based partitioning initial load setting is enabled, initial loads for mixed case table names do not sync data to target, because the input we pass to `to_regclass` in queries was not quoted, and Postgres assumes lowercase, making those queries not return anything

### What
- Fix `to_regclass` usage in estimated table size function
- Make `classifyTableName` return OID of the table as well
- Replace usage of `to_regclass` with comparing with OID directly
- Adapt E2E test to cover mixed case usage

Fixes: DBI-718

### DIFF
--- a/flow/connectors/postgres/qrep_partition.go
+++ b/flow/connectors/postgres/qrep_partition.go
@@ -303,9 +303,9 @@ func getTableBlockCount(ctx context.Context, tx pgx.Tx, table string) (int64, er
 }
 
 type tableClassification struct {
-	oid            uint32
 	qualifiedName  string
 	relkind        string
+	oid            uint32
 	relhassubclass bool
 }
 
@@ -349,10 +349,10 @@ func getPartitionedTables(ctx context.Context, tx pgx.Tx, parentOID uint32) (map
 	defer rows.Close()
 
 	type tableInfo struct {
-		oid    uint32
 		name   string
 		kind   string
 		blocks int64
+		oid    uint32
 	}
 	var tableInfos []tableInfo
 	for rows.Next() {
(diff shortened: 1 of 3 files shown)


flow/connectors/postgres/qrep_partition.go
```diff
@@ -303,6 +303,7 @@ func getTableBlockCount(ctx context.Context, tx pgx.Tx, table string) (int64, er
 }
 
 type tableClassification struct {
+	oid            uint32
 	qualifiedName  string
```
jgao54: the only time qualifiedName is needed is is for this query, since we are now using oid (which is a great idea), let's just remove qualifiedName altogether from here (and use oid as key for blocksPerTable as well)

jgao54: nvm ignore, we do need the table names to pass into `ChildTableRange`

jgao54 (commented):
change looks good to me, just need to add/modify test case for inherited table

flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -348,8 +348,9 @@ func TestCTIDPartitioningOnMultiLevelPartitionedTable(t *testing.T) {
 	t.Parallel()
 	schemaName, conn, connStr := setupTestSchema(t)
 
-	// Root table partitioned by region
-	rootTable := schemaName + ".multi_level"
+	// Root table partitioned by region — uses mixed-case names to verify
+	// that OID-based lookups work correctly (to_regclass lowercases unquoted identifiers)
+	rootTable := fmt.Sprintf(`%s."MultiLevel"`, schemaName)
```
jgao54: looks like test is failing on this because parsing the qualified table name results in double-quoting with `"db".""table""`

jgao54: also for completion of coverage, could we also test for schema name that also has mixed case which I think is also possible

Amogh-Bharadwaj: Included

jgao54 (approved):

PR #4243, opened 2026-04-26
feat(catalog): add GCP Cloud KMS support for secret decryption

whites11:
## Summary
- Add `PEERDB_KMS_PROVIDER` env var to dispatch between AWS KMS and GCP Cloud KMS for catalog secret decryption
- **Rust** (`nexus/catalog`): GCP Cloud KMS REST API via `reqwest`, using a Workload Identity token fetched from the GKE metadata server
- **Go** (`flow/internal/env.go`): GCP Cloud KMS via the official `cloud.google.com/go/kms/apiv1` client (uses Application Default Credentials, which resolves to Workload Identity on GKE)
- Provider switch on both sides enumerates `aws` (and empty for back-compat), `gcp`, and returns an explicit error for unknown values; constants `KmsProviderAWS` / `KmsProviderGCP` introduced in Go to mirror the Rust pattern
- Defaults to `aws` everywhere — no behavior change for existing AWS deployments

## Test plan
- [ ] End-to-end test on GCP GKE cluster with Workload Identity, decrypting a catalog secret encrypted with a GCP KMS key
- [ ] Verify `PEERDB_KMS_PROVIDER=` (empty), unset, and `PEERDB_KMS_PROVIDER=aws` all still hit the AWS path
- [ ] Verify `PEERDB_KMS_PROVIDER=invalid` returns the new error rather than silently falling through to AWS

### DIFF
--- a/flow/internal/env.go
+++ b/flow/internal/env.go
@@ -3,7 +3,10 @@ package internal
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"reflect"
 	"strconv"
@@ -17,7 +20,8 @@ import (
 )
 
 const (
-	KmsKeyIDEnvVar = "PEERDB_KMS_KEY_ID"
+	KmsKeyIDEnvVar    = "PEERDB_KMS_KEY_ID"
+	KmsProviderEnvVar = "PEERDB_KMS_PROVIDER"
 )
 
 // getEnvUint returns the value of the environment variable with the given name
@@ -71,6 +75,16 @@ func decryptWithKms(ctx context.Context, data []byte) ([]byte, error) {
 		return data, nil
 	}
 
+	provider := GetEnvString(KmsProviderEnvVar, "aws")
+	switch provider {
+	case "gcp":
+		return decryptWithGcpKms(ctx, data, keyID)
+	default:
+		return decryptWithAwsKms(ctx, data, keyID)
+	}
+}
+
+func decryptWithAwsKms(ctx context.Context, data []byte, keyID string) ([]byte, error) {
 	cfg, err := config.LoadDefaultConfig(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load AWS config: %w", err)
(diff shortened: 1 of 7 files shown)


flow/internal/env.go
```diff
@@ -71,6 +74,16 @@ func decryptWithKms(ctx context.Context, data []byte) ([]byte, error) {
 		return data, nil
 	}
 
+	provider := GetEnvString(KmsProviderEnvVar, "aws")
+	switch provider {
```
whites11, line 81: make this switch contain all the possible options and the default branch end up in an error. use constants for the provider strings (reuse if possible)

masterashu (approved):
Looks good.
Any reason to use api calls to use http calls instead of google_cloud_kms library?

whites11:
> Looks good. Any reason to use api calls to use http calls instead of google_cloud_kms library?

Let me give you claude's explanation, althought I have no rust experience to evaluate it

```
The reasoning is dependency-footprint pragmatism:                                                                                  
  
  1. The whole "gcp" path is feature-gated (gcp = ["dep:reqwest"]). The catalog crate compiles without any GCP deps when the feature 
  is off; turning it on adds only reqwest. A real Google Cloud SDK crate (e.g. google-cloud-kms/gcloud-sdk) would pull in tonic,
  prost, generated protobuf modules, h2, and a chain of transitive crates — even when feature-gated, just enabling it would balloon  
  compile time and binary size.                               
  2. The surface needed is tiny — two HTTP calls:
    - GET metadata.google.internal/.../token for a Workload Identity bearer token
    - POST cloudkms.googleapis.com/v1/{key}:decrypt                                                                                  
                                                                                                                                     
  That's ~30 LOC with reqwest. An SDK would be more code via builder objects with no functional gain.                                
  3. The Rust GCP-SDK ecosystem isn't as canonical as Go's. In Go there's exactly one obvious choice (cloud.google.com/go/kms/apiv1),
   and that's why the Go side of this same PR was refactored to it (commit acf7c23a). In Rust, the options (gcloud-sdk,              
  google-cloud-kms, google-cloud-rust) are competing third-party crates with different trade-offs — committing to one is a
  non-trivial choice.                                                                                                                
  4. Workload Identity on GKE is intentionally trivial over HTTP — the metadata server is designed to be hit with curl-equivalent
  calls. Using it that way is idiomatic, not a hack.                                                                                 
  
  Should it be changed? Probably not for this PR. If/when the catalog crate grows more GCP touchpoints (Pub/Sub source on the catalog
   side, etc.), promoting to a real SDK becomes worth the dep cost. For one decrypt call, raw reqwest is the right call.

```

PR #4250, opened 2026-04-29
Automatic dependency updates: Stubbed generated modules

pfcoperez:
This PR adds stubs for the packages to be generated by `generate-protos.sh` . 
This way the script is not required for  `go get -t ./...` finish successfully.

That command is used by Renovate to explore the dependencies it needs to update. Prior to this PR, it was failing leaving the update PRs in inconsistent states and crashing before tidy was even invoked.
Any dependency update bot invoking `go mod tidy` must detect theses packages in order to avoid removing entries.

The stubs are transparent to the proto generated files with no collisions and imports replicating their dependencies. 
Not running the protocol generation script will lead to failed compilations even with the stubs in.

Follows up: https://github.com/PeerDB-io/peerdb/pull/4246 and https://github.com/PeerDB-io/peerdb/pull/4248 in the series of changes fixing automatic updates.

### DIFF
--- a/.github/actions/genprotos/action.yml
+++ b/.github/actions/genprotos/action.yml
@@ -21,6 +21,4 @@ runs:
         github_token: ${{ github.token }}
     - if: steps.cache.outputs.cache-hit != 'true'
       shell: sh
-      run: |
-        buf generate protos
-        cd flow && go generate
+      run: ./generate-protos.sh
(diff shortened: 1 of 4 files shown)


jgao54 (commented):
one question, this could lead to drifts over time iiuc right? 

i am wondering if it would make sense to have some kind of drift detection script in CI
(i.e. generate the file, compare imports, and surface the missing imports from the stub files).

manually adding the missing imports is probably sufficient but a script works too.

jgao54 (approved):

PR #4253, opened 2026-04-30
Renovate: Do not attempt to upgrade pinned versions

pfcoperez:
This PR adds the list of pinned versions at

https://github.com/PeerDB-io/peerdb/blob/7ef375c5e8f88df847bd5f22786d02313f584b16/flow/cmd/check_pinned_versions/main.go#L18-L24 

to a renovate package rule that disables updates on them.

Follow-up work will add mechanisms to keep these in sync once we complete Renovate infra fixes.

### DIFF
--- a/flow/cmd/check_pinned_versions/main.go
+++ b/flow/cmd/check_pinned_versions/main.go
@@ -14,7 +14,8 @@ import (
 // accidentally upgraded. Each entry maps a module path to the exact version
 // that should appear in go.mod.
 //
-// To update a pinned dependency, change both go.mod AND this map.
+// To update a pinned dependency, change both go.mod, this map AND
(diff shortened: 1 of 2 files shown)


masterashu (approved):

jgao54 (approved):
could we add a comment in check_pinned_versions/main.go that when we removed the pinned version, also remove it from here. So we don't accidentally skip package updates in the future once we fix the DBIs

pfcoperez:
> could we add a comment in check_pinned_versions/main.go that when we removed the pinned version, also remove it from here. So we don't accidentally skip package updates in the future once we fix the DBIs

Good idea, at least until we add a step to keep them in sync automatically. Added in: 5d03e91da49f8947c401fd55917e8b454a312744

Also created https://linear.app/clickhouse/issue/DBI-726

PR #4254, opened 2026-04-30
mongodb: improving idle change stream resume

jgao54:
This PR:
- revert the previous 5 minute context timeout [change](https://github.com/PeerDB-io/peerdb/pull/4166/changes) (this was [causing an issue](https://github.com/PeerDB-io/peerdb/pull/4213) where change stream that takes a long time to resume can stuck and retry loop).
- setting `client.Watch` with batchSize = 0, allowing change stream to return immediately, and the "catch-up" can be blocked on `changestream.Next` instead of `client.Watch`
- improve handle of idle stream by:
  - always checkpoint in-memory when context timed out with records accumulated
  - always persist offset to catalog when context timed out without records accumulated, so that if pipe is paused or an error occurred, we can resume from a recent offset
- add unit test with a mock change stream to avoid having to slow down CI; this will also make it easier to extend unit tests for other mongodb cdc behavior

Fixes: DBI-691

### DIFF
--- a/flow/connectors/mongo/cdc.go
+++ b/flow/connectors/mongo/cdc.go
@@ -36,6 +36,27 @@ type ChangeEvent struct {
 	ClusterTime   bson.Timestamp `bson:"clusterTime"`
 }
 
+type createChangeStreamFn func(
+	ctx context.Context, pipeline mongo.Pipeline, opts ...options.Lister[options.ChangeStreamOptions],
+) (ChangeStream, error)
+
+// ChangeStream is defined as an interface, allowing tests inject mock change stream.
+type ChangeStream interface {
+	Next(ctx context.Context) bool
+	ResumeToken() bson.Raw
+	Err() error
+	Close(ctx context.Context) error
+	Current() bson.Raw
+}
+
+type changeStreamWrapper struct {
+	*mongo.ChangeStream
+}
+
+func (w *changeStreamWrapper) Current() bson.Raw {
+	return w.ChangeStream.Current
+}
+
 func (c *MongoConnector) GetTableSchema(
 	ctx context.Context,
 	_ map[string]string,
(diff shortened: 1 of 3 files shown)


flow/connectors/mongo/cdc.go
```diff
@@ -36,6 +36,23 @@ type ChangeEvent struct {
 	ClusterTime   bson.Timestamp `bson:"clusterTime"`
 }
 
+// ChangeStream is defined as an interface, allowing tests inject mock change stream.
```
pfcoperez, line 39: ❤️

flow/connectors/mongo/cdc_test.go
```diff
@@ -0,0 +1,180 @@
...
+		return false
+	}
+}
+
+func (cs *mockChangeStream) ResumeToken() bson.Raw       { return cs.resumeToken }
+func (cs *mockChangeStream) Err() error                  { return cs.err }
+func (cs *mockChangeStream) Current() bson.Raw           { return cs.current }
+func (cs *mockChangeStream) Close(context.Context) error { return nil }
+
+var _ ChangeStream = (*mockChangeStream)(nil)
```
ilidemi, line 80: nit: this would be asserted by the test compiling, right?

ilidemi (approved):

flow/connectors/mongo/cdc_test.go
jgao54: ah yeah, technically already re-enforced by:
```
		createChangeStream: func(
			context.Context, mongo.Pipeline, ...options.Lister[options.ChangeStreamOptions],
		) (ChangeStream, error) {
			return mockCS, nil
		},
```
below, but it is not very obvious where as this makes it more explicit and future-proof test refactoring.

PR #4256, opened 2026-04-30
Load dynamic settings at start

serprex:
Have query at top of activity load most settings up front, rather than querying for each setting individually

This is designed to be refactor/optimization, there is incompatibility when invalid setting now makes LoadSettings fail on code paths that wouldn't've loaded that setting before. This could be fixed by making value parsing lazy. But means making all accessor methods for every field


ilidemi:
The goal is great but the reflection approach complicates the setup even further and long term we'd like to go towards some bog standard configuration approach. It may take a bit of time and untangling wrt different sources and `env` semantics but I'd prefer not to take steps further away from that.

serprex:
the goal here consolidates settings so that most code just gets Settings struct, how that struct gets created/declared over time can change

PR #4274, opened 2026-05-05
Classify MySQL LinearRing WKB parse error as unsupported datatype

Jeremyyang920:
Match the specific geom-library failure "Points of LinearRing do not form a closed linestring" produced via processGeometryData and surface it to the user as ErrorUnsupportedDatatype with code UNSUPPORTED_GEOMETRY_LINEAR_RING_NOT_CLOSED. Other "failed to parse geometry WKB" errors continue to fall through so we can spot unclassified cases.

Resolves DBI-732

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -42,6 +42,10 @@ const (
 	MongoShutdownInProgress              = "(ShutdownInProgress) The server is in quiesce mode and will shut down"
 	MongoInterruptedDueToReplStateChange = "(InterruptedDueToReplStateChange) operation was interrupted"
 	MongoIncompleteReadOfMessageHeader   = "incomplete read of message header"
+
+	// MySQLGeometryLinearRingNotClosedError is the specific WKB parse failure raised by the geom
+	// library when a LinearRing's points do not close.
+	MySQLGeometryLinearRingNotClosedError = "Points of LinearRing do not form a closed linestring"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


flow/alerting/classifier.go
```diff
@@ -1101,6 +1105,13 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
+	if strings.Contains(err.Error(), MySQLGeometryLinearRingNotClosedError) {
+		return ErrorUnsupportedDatatype, ErrorInfo{
```
Jeremyyang920, line 1112: I wasn't sure if we wanted to create a different classification since rather than Unsupported datatype, this one is more so malformed, but figured the category is close enough.

flow/alerting/classifier.go
```diff
@@ -1101,6 +1105,13 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
...
+			Source: ErrorSourceMySQL,
```
ilidemi, line 1113: To say for sure it's MySQL, we need to wrap it at the source into a separate exception type, like we have in `flow/shared/exceptions/mysql.go`. Message itself is coming from C code of go-geos, theoretically could be Postgres too. I lean towards going MySQL-specific for clarity.

flow/alerting/classifier.go
ilidemi: Unsupported datatype is conceptually fine. Noticed the link we have in clickpipes customer-facing notifications is [postgres-specific](https://clickhouse.com/docs/integrations/clickpipes/postgres/faq#what-if-i-am-seeing-unexpected-datatype-errors) but that is a pre-existing issue

flow/alerting/classifier.go
```diff
@@ -1101,6 +1105,13 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
+	if strings.Contains(err.Error(), MySQLGeometryLinearRingNotClosedError) {
```
ilidemi: nit: if we were to not wrap into a MySQL-specific error, would need to match on `geos.Error` for clarity

flow/alerting/classifier.go
Jeremyyang920: I agree. Updating it to wrap as a new MySQL Exception

ilidemi (approved):
Nicely done with the negative unit test

PR #4276, opened 2026-05-06
feat(resync): Add support to update snapshot config (max_parallel_workers, num_tables, rows_per_partition) during resync signal

masterashu:
### Initial + CDC Pipes:
* Running pipe -> Resync (working)
* Paused Pipe -> Resync (working)
* Snapshot -> Resync (working)
### Initial only Pipes:
* Completed -> Resync (working)
* Failed -> Resync (working)
* Snapshot -> Resync (working)
### CDC Only Pipes:
* Running -> Resync (working) [Does snapshot]
* Paused -> Resync (working) [Does snapshot]

### DIFF
--- a/flow/cmd/handler.go
+++ b/flow/cmd/handler.go
@@ -503,10 +503,9 @@ func (h *FlowRequestHandler) FlowStateChange(
 
 				config.Resync = true
 				config.DoInitialSnapshot = true
-				// Override Snapshot Parameters
-				overrideSnapshotParametersInResync(req, config)
 
 				// validate mirror first because once the mirror is dropped, there's no going back
+				// Note: We do not validate snapshot parameters (SnapshotNumRowsPerPartition, etc)
 				if _, err := h.ValidateCDCMirror(ctx, &protos.CreateCDCFlowRequest{
 					ConnectionConfigs: config,
 				}); err != nil {
--- a/flow/cmd/handler.go
+++ b/flow/cmd/handler.go
@@ -422,7 +422,6 @@ func (h *FlowRequestHandler) FlowStateChange(
 	ctx context.Context,
 	req *protos.FlowStateChangeRequest,
 ) (*protos.FlowStateChangeResponse, APIError) {
-	// Flow-API
 	logs := slog.String(string(shared.FlowNameKey), req.FlowJobName)
 	slog.InfoContext(ctx, "FlowStateChange called", logs, slog.Any("req", req))
 	if underMaintenance, err := internal.PeerDBMaintenanceModeEnabled(ctx, nil); err != nil {
@@ -554,16 +553,15 @@ func (h *FlowRequestHandler) FlowStateChange(
 }
 
 func overrideSnapshotParametersInResync(req *protos.FlowStateChangeRequest, configs *protos.FlowConnectionConfigs) {
-	if req.FlowConfigUpdate != nil && req.FlowConfigUpdate.GetCdcFlowConfigUpdate() != nil {
-		cdcConfigUpdate := req.FlowConfigUpdate.GetCdcFlowConfigUpdate()
-		if cdcConfigUpdate.SnapshotMaxParallelWorkers > 0 {
-			configs.SnapshotMaxParallelWorkers = cdcConfigUpdate.SnapshotMaxParallelWorkers
+	if u := req.GetFlowConfigUpdate().GetCdcFlowConfigUpdate(); u != nil {
+		if u.SnapshotMaxParallelWorkers > 0 {
+			configs.SnapshotMaxParallelWorkers = u.SnapshotMaxParallelWorkers
 		}
-		if cdcConfigUpdate.SnapshotNumTablesInParallel > 0 {
-			configs.SnapshotNumTablesInParallel = cdcConfigUpdate.SnapshotNumTablesInParallel
+		if u.SnapshotNumTablesInParallel > 0 {
+			configs.SnapshotNumTablesInParallel = u.SnapshotNumTablesInParallel
 		}
-		if cdcConfigUpdate.SnapshotNumRowsPerPartition > 0 {
-			configs.SnapshotNumRowsPerPartition = cdcConfigUpdate.SnapshotNumRowsPerPartition
+		if u.SnapshotNumRowsPerPartition > 0 {
+			configs.SnapshotNumRowsPerPartition = u.SnapshotNumRowsPerPartition
 		}
 	}
 }
--- a/flow/cmd/handler.go
+++ b/flow/cmd/handler.go
@@ -447,8 +447,12 @@ func (h *FlowRequestHandler) FlowStateChange(
 	}
 
 	if req.FlowConfigUpdate != nil && req.FlowConfigUpdate.GetCdcFlowConfigUpdate() != nil &&
-		// Don't allow config updates if the flow is already in a terminal state since it can lead to confusion where the config is updated but the flow is not reflecting those changes since it's already completed/failed
-		(currState != protos.FlowStatus_STATUS_TERMINATED && currState != protos.FlowStatus_STATUS_TERMINATING && currState != protos.FlowStatus_STATUS_FAILED && currState != protos.FlowStatus_STATUS_COMPLETED) {
+		// Don't allow config updates if the flow is already in a terminal state since it can lead to confusion
+		//  where the config is updated but the flow is not reflecting those changes since it's already completed/failed
+		(currState != protos.FlowStatus_STATUS_TERMINATED &&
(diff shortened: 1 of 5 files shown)


flow/shared/telemetry/activity_logging.go
```diff
@@ -93,6 +93,16 @@ func LogActivityStartFlowConfigUpdate(ctx context.Context, flowName string, upda
 		changes = append(changes, fmt.Sprintf("tables removed: %v", removedTables))
 	}
 
+	if update.SnapshotMaxParallelWorkers > 0 {
```
jgao54: thanks for adding this to activity logging as well!

jgao54: nit: for same of completion, let's include batchSize, idleTimeout, snapshotNumPartitionsOverride to match  `LogActivityUpdateFlowConfig`

flow/cmd/handler.go
```diff
@@ -422,6 +422,7 @@ func (h *FlowRequestHandler) FlowStateChange(
 	ctx context.Context,
 	req *protos.FlowStateChangeRequest,
 ) (*protos.FlowStateChangeResponse, APIError) {
+	// Flow-API
```
jgao54: nit: don't think this comment add much valuem can delete

flow/cmd/handler.go
```diff
@@ -549,6 +553,21 @@ func (h *FlowRequestHandler) FlowStateChange(
 	return &protos.FlowStateChangeResponse{}, nil
 }
 
+func overrideSnapshotParametersInResync(req *protos.FlowStateChangeRequest, configs *protos.FlowConnectionConfigs) {
+	if req.FlowConfigUpdate != nil && req.FlowConfigUpdate.GetCdcFlowConfigUpdate() != nil {
```
jgao54: nit: 
```
if u := req.GetFlowConfigUpdate().GetCdcFlowConfigUpdate(); u != nil {
    if u.SnapshotMaxParallelWorkers > 0 {...}
}
```

flow/workflows/cdc_flow.go
```diff
@@ -627,6 +662,9 @@ func CDCFlowWorkflow(
 				cfg.TableMappings = originalTableMappings
 				// this is the only place where we can have a resync during a resync
 				// so we need to NOT sync the tableMappings to catalog to preserve original names
+
+				// We still override the snapshot parameters (when resync with updated values)
+				overrideSnapshotParametersInConfig(val, cfg)
```
jgao54: for consistency, i think we should update state here like the other callsites:

```
overrideSnapshotParametersInState(val, state)
syncStateToConfigProtoInCatalog(ctx, cfg, state)
```

instead of directly calling `uploadConfigToCatalog`

it's not clear to me why we added these snapshot configs to `CDCFlowWorkflowState`, but to be safe and to keep the PR scoped, we can just use the same pattern. and revisit if we can remove these configs from state. this means you can delete the duplicated `overrideSnapshotParametersInConfig`

flow/workflows/cdc_flow.go
```diff
@@ -206,6 +209,36 @@ func handleFlowSignalStateChange(
 	}
 }
 
+func overrideSnapshotParametersInState(req *protos.FlowStateChangeRequest, state *cdc_state.CDCFlowWorkflowState) {
+	if req.FlowConfigUpdate != nil && req.FlowConfigUpdate.GetCdcFlowConfigUpdate() != nil {
```
jgao54: similar to above, use `req.GetFlowConfigUpdate().GetCdcFlowConfigUpdate()`

also, a bit cleaner to make this a struct method,  something like:

```
func (s *CDCFlowWorkflowState) MergeSnapshotConfigUpdate(u *protos.CDCFlowConfigUpdate) {
	if u == nil {
		return
	}
	if u.SnapshotMaxParallelWorkers > 0 {
		s.SnapshotMaxParallelWorkers = u.SnapshotMaxParallelWorkers
	}
	if u.SnapshotNumTablesInParallel > 0 {
		s.SnapshotNumTablesInParallel = u.SnapshotNumTablesInParallel
	}
	if u.SnapshotNumRowsPerPartition > 0 {
		s.SnapshotNumRowsPerPartition = u.SnapshotNumRowsPerPartition
	}
	if u.SnapshotNumPartitionsOverride > 0 {
		s.SnapshotNumPartitionsOverride = u.SnapshotNumPartitionsOverride
	}
}

````

Comment (jgao54) [reply flow/workflows/cdc_flow.go]: (`SnapshotNumPartitionsOverride` is exposed in OSS, so we can add it everywhere for completion)

File: flow/cmd/handler.go [hunk 6]
```diff
@@ -503,6 +504,9 @@ func (h *FlowRequestHandler) FlowStateChange(
 
 				config.Resync = true
 				config.DoInitialSnapshot = true
+				// Override Snapshot Parameters
+				overrideSnapshotParametersInResync(req, config)
```
jgao54: nit: i may regret saying this (lol) but for now we are not validating these snapshot parameters so we could just remove it altogether. 

maybe can add a comment here mentioning that. this avoids having update these config fields.

jgao54:
missing e2e test for this to ensure that resync propagates snapshot config changes

flow/workflows/cdc_flow.go
```diff
@@ -627,6 +651,10 @@ func CDCFlowWorkflow(
 				cfg.TableMappings = originalTableMappings
 				// this is the only place where we can have a resync during a resync
 				// so we need to NOT sync the tableMappings to catalog to preserve original names
+
+				// We still override the snapshot parameters (when resync with updated values)
+				overrideSnapshotParametersInState(val, state)
+				syncStateToConfigProtoInCatalog(ctx, cfg, state)
 				uploadConfigToCatalog(ctx, cfg)
```
jgao54: this line should be removed, since `syncStateToConfigProtoInCatalog` already call it.

masterashu: my bad. got it

flow/cmd/handler.go
```diff
@@ -549,6 +551,20 @@ func (h *FlowRequestHandler) FlowStateChange(
 	return &protos.FlowStateChangeResponse{}, nil
 }
 
+func overrideSnapshotParametersInResync(req *protos.FlowStateChangeRequest, configs *protos.FlowConnectionConfigs) {
```
pfcoperez: Actually, as the linter found, this method doesn't seem to be used.

pfcoperez (commented):
Changes LGTM except for the seemingly unused method. I also wonder if we can unit test the config extraction, basically checking the right fields are extracted from a test.

jgao54 (approved):
lgtm as well once @pfcoperez 's comment is addressed :) !

flow/shared/telemetry/activity_logging.go
ilidemi: > // Only logging fields that don't need old values to compare to

The caller here will actually pass all values, changed and unchanged, and the flow-api can't distinguish which ones to log. Logging the list of tables has to happen here because they're not applied to the config until snapshotting is done, so keeping just tables and leaving the rest to the worker is better.

flow/workflows/cdc_flow.go
```diff
@@ -627,10 +651,13 @@ func CDCFlowWorkflow(
 				cfg.TableMappings = originalTableMappings
 				// this is the only place where we can have a resync during a resync
 				// so we need to NOT sync the tableMappings to catalog to preserve original names
-				uploadConfigToCatalog(ctx, cfg)
+
+				// We still override the snapshot parameters (when resync with updated values)
+				overrideSnapshotParametersInState(val, state)
+				resyncCfg := syncStateToConfigProtoInCatalog(ctx, cfg, state)
```
ilidemi: Comments just above mention needing to not sync the tableMappings to catalog because they have `_resync` at the end, and `syncStateToConfigProtoInCatalog` does exactly that

ilidemi: To avoid without code duplication, can move `applySnapshotConfigOverrides` to internal, change it to ConfigsCore, and inline the `h.ValidateCDCMirror` call to be able to pass ConfigsCore earlier. Then this one can be
```go
internal.ApplySnapshotConfigOverrides(cfg, val.GetFlowConfigUpdate().GetCdcFlowConfigUpdate())
uploadConfigToCatalog(ctx, cfg)
```

ilidemi: The whole signal + mutable state thing is a mess so I understand that it's very confusing to deal with

flow/shared/telemetry/activity_logging.go
masterashu: Okay

PR #4280, opened 2026-05-06
Fix postgres18 new volume path in docker images

masterashu:


masterashu:
This got created, with the recent postgres image bump in https://github.com/PeerDB-io/peerdb/pull/4245.
The new version fails to start with following error
>  Error: in 18+, these Docker images are configured to store database data in a
       format which is compatible with "pg_ctlcluster" (specifically, using
       major-version-specific directory names).  This better reflects how
       PostgreSQL itself works, and how upgrades are to be performed.
>  See also https://github.com/docker-library/postgres/pull/1259
>  Counter to that, there appears to be PostgreSQL data in:
         /var/lib/postgresql/data (unused mount/volume)
>  This is usually the result of upgrading the Docker image without
       upgrading the underlying database using "pg_upgrade" (which requires both
       versions).
>  The suggested container configuration for 18+ is to place a single mount
       at /var/lib/postgresql which will then place PostgreSQL data in a
       subdirectory, allowing usage of "pg_upgrade --link" without mount point
       boundary issues.
>  See https://github.com/docker-library/postgres/issues/37 for a (long)
       discussion around this process, and suggestions for how to do so.

This update fixes it

pfcoperez (approved):
I got bitten by this in my local dev env! Thanks for fixing 👏

PR #4281, opened 2026-05-06
feat: add second postgres instance for local testing

Amogh-Bharadwaj:
For testing Postgres to Postgres mirrors, it would be good to have a second Postgres instance rather than have source and destination be same, for testing things like roles being copied over

### DIFF
--- a/ancillary-docker-compose.yml
+++ b/ancillary-docker-compose.yml
@@ -125,6 +125,27 @@ services:
       timeout: 10s
       retries: 5
 
+  postgres2:
+    container_name: peerdb-postgres2
+    image: ${POSTGRES_IMAGE}
+    restart: unless-stopped
+    volumes:
+      - postgres2_data:/var/lib/postgresql/data
+    ports:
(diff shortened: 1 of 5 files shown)


ancillary-docker-compose.yml
```diff
@@ -125,6 +125,27 @@ services:
       timeout: 10s
       retries: 5
 
+  postgres2:
+    container_name: peerdb-postgres2
+    image: ${POSTGRES_IMAGE}
```
pfcoperez, line 130: This is not a change request, I'd rather keep it simpler, but I wonder if you foresee needing to test different PG versions in source and destination.

pfcoperez (approved):
LGTM!

ancillary-docker-compose.yml
Amogh-Bharadwaj: Yup that's a great shout @pfcoperez, would probably need to test PG16 to PG18 type mirrors

PR #4283, opened 2026-05-06
feat(pg-pg): Automated schema dump mode

Amogh-Bharadwaj:
### Why
- For Postgres to Postgres migration use-cases, users would prefer an apples-to-apples mirroring of their source setup on the target instance.
- A well-recognized method to migrate schema only is pg_dump.

### What
- This PR adds an activity in `SetupFlowWorkflow` which runs pg_dump on the source database and pipes its output to the target database via PSQL. 
- It adds a **new dynamic flag setting to gate the above**. Also the above is gated to PG type system mirrors.
- When the above flag is set, destination validation and destination create normalized tables steps are skipped.
- It sets **--no-owners and --no-privileges**, leaving it to the user to add the desired roles later. `pg_dumpall` requires exact major version matching between pg_dumpall version and target version, which we cannot guarantee.
- E2E tests added

### DIFF
--- a/flow/connectors/postgres/pgdump_schema.go
+++ b/flow/connectors/postgres/pgdump_schema.go
@@ -1,0 +1,165 @@
+package connpostgres
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/exec"
+	"strconv"
+
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+)
+
+// RunPgDumpSchema first migrates roles via pg_dumpall --roles-only, then streams
+// a schema-only pg_dump from source directly into psql on the destination,
+// piping stdout into stdin without intermediate files.
+func RunPgDumpSchema(ctx context.Context, srcConfig *protos.PostgresConfig, dstConfig *protos.PostgresConfig) error {
+	// Step 1: migrate roles from source to destination
+	if err := pipeCommand(ctx, srcConfig, dstConfig, "pg_dumpall", buildPgDumpAllArgs(srcConfig)); err != nil {
+		return fmt.Errorf("pg_dumpall roles migration failed: %w", err)
+	}
+
+	// Step 2: migrate schema from source to destination
+	if err := pipeCommand(ctx, srcConfig, dstConfig, "pg_dump", buildPgDumpArgs(srcConfig)); err != nil {
+		return fmt.Errorf("pg_dump schema migration failed: %w", err)
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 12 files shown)


flow/workflows/setup_flow.go
```diff
@@ -251,9 +310,25 @@ func (s *SetupFlowExecution) executeSetupFlow(
 		}
 	}
 
-	// then fetch the table schema and setup the normalized tables
-	if err := s.setupNormalizedTables(ctx, config); err != nil {
-		return nil, fmt.Errorf("failed to fetch table schema and setup normalized tables: %w", err)
+	if err := s.setupTableSchema(ctx, config); err != nil {
+		return nil, fmt.Errorf("failed to fetch table schema: %w", err)
+	}
```
Amogh-Bharadwaj, line 330: Split out here: https://github.com/PeerDB-io/peerdb/pull/4290
Will rebase once that lands

serprex (approved):

flow/connectors/postgres/pgdump_schema.go
```diff
@@ -0,0 +1,143 @@
...
+func pipeCommand(
+	ctx context.Context,
+	srcConfig *protos.PostgresConfig,
+	dstConfig *protos.PostgresConfig,
+	srcBinary string,
+	srcArgs []string,
+) error {
+	psqlArgs := buildPsqlArgs(dstConfig)
+
+	srcCmd := exec.CommandContext(ctx, srcBinary, srcArgs...)
```
Jeremyyang920, line 49: I believe this can't come up since its installed via apt and should be in PATH, but is there any need to be defensive here with an absolute path to pg_dump.

flow/connectors/postgres/pgdump_schema.go
```diff
@@ -0,0 +1,143 @@
...
+	pipe, err := srcCmd.StdoutPipe()
+	if err != nil {
+		return fmt.Errorf("failed to create %s stdout pipe: %w", srcBinary, err)
+	}
+	psqlCmd.Stdin = pipe
+
+	var srcStderr, psqlStderr bytes.Buffer
+	srcCmd.Stderr = &srcStderr
+	psqlCmd.Stderr = &psqlStderr
+
```
Jeremyyang920: Any reason here to not just use os.Pipe?
```
     pr, pw, err := os.Pipe()                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                               
        srcCmd.Stdout = pw   
        psqlCmd.Stdin = pr
```
This should handle any error cases more gracefully.  From the current pipe usage, I think if there is an error half way with the psql command, the srcCmd pipe will keep trying to send to nothing and the kernel buffer will fill up and could lead to a deadlock

PR #4288, opened 2026-05-08
Provide ClickHouse Peers validations as a library

pfcoperez:
This Pull Request introduces the possibility of performing ClickHouse peers validations outside PeerDB while 
offering maximum re-utilization of the validation logic.

To achieve this, it offers four main changes:

1. Staging bucket validation logic is abstracted out behind the `StagingValidator` type which is implemented as a function taking a context and potentially returning an error. The logic used at PeerDB to implement this function is shared at `flow/pkg` for external code to use through  `NewS3StagingValidator` and `NewGCSStagingValidator` factory methods.
 
2. In a similar vein, accepted ClickHouse hosts becomes a parameter for the ClickHouse PeerValidation logic.

3. Which in turn is shared at `flow/pkg` through the `ValidateClickHousePeer` method. This method is also used within PeerDB when peer validation is not disabled.

4. ~Hence, peer validation can be now disabled for any peer in the gRPC methods  `CreatePeer` and `ValidatePeer`  using a new optional parameter: `disable_connector_validation`~ We assume validation happens both upstream and at PeerDB provision side

5. We have disabled bucket checks at ClickHouse validation time, this was actually testing static infrastructure.  

Part of DBI-65


flow/pkg/clickhouse/validation.go
```diff
@@ -95,6 +99,123 @@ func CheckIfTablesEmptyAndEngine(ctx context.Context, logger log.Logger, conn cl
...
+	if err := Exec(ctx, logger, conn, fmt.Sprintf("INSERT INTO %s VALUES (1, now64())", validateDummyTableNameRenamed)); err != nil {
+		return fmt.Errorf("failed to insert into validation table %s: %w", validateDummyTableNameRenamed, err)
+	}
+
+	// drop the table
+	if err := Exec(ctx, logger, conn, "DROP TABLE IF EXISTS "+validateDummyTableNameRenamed); err != nil {
+		return fmt.Errorf("failed to drop validation table %s: %w", validateDummyTableNameRenamed, err)
+	}
+
+	// validate staging storage
```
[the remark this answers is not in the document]
pfcoperez: No, this is a serious fault. I'd rather have the process panicking. Optionality should be implemented by passing a no check function like ` func (ctc Context) error { return nil }`

flow/cmd/validate_peer.go
```diff
@@ -64,9 +70,14 @@ func (h *FlowRequestHandler) ValidatePeer(
 		}, NewFailedPreconditionApiError(displayErr)
 	}
 
+	validationMsg := fmt.Sprintf("%s peer %s is valid", req.Peer.Type, req.Peer.Name)
+
+	if shouldSkipConnectorValidation {
+		validationMsg += " (connector validation skipped)"
+	}
```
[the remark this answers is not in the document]
pfcoperez: overkill

flow/pkg/common/utils.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"crypto/rand"
+)
+
+func RandomString(n int) string {
+	const alphanum = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+	bytes := make([]byte, n)
+	if _, err := rand.Read(bytes); err != nil {
+		return "temp"
```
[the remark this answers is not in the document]
pfcoperez: Outside the scope of this PR

flow/pkg/clickhouse/validation.go
```diff
@@ -95,6 +99,123 @@ func CheckIfTablesEmptyAndEngine(ctx context.Context, logger log.Logger, conn cl
...
+func ValidateClickHousePeer(
+	ctx context.Context,
+	logger log.Logger,
+	allowedDomains string,
+	peerName string,
+	conn clickhouse.Conn,
+	stagingValidator StagingValidator,
+) error {
+	// Validate ClickHouse host domain
+	if err := ValidateClickHouseHost(ctx, peerName, allowedDomains); err != nil {
```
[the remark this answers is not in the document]
pfcoperez: Resolved

pfcoperez:
@copilot resolve the merge conflicts in this pull request

flow/shared/random.go
```diff
@@ -1,17 +1,9 @@
 package shared
 
 import (
-	"crypto/rand"
+	"github.com/PeerDB-io/peerdb/flow/pkg/common"
 )
 
 func RandomString(n int) string {
```
jgao54: nit: can remove this altogether and use the function in pkg instead

flow/pkg/objectstore/validation.go
```diff
@@ -0,0 +1,69 @@
...
+
+// stagingCheckObjectPrefix is the basename used for temporary objects written
+// during a staging-bucket smoke test. Kept short and unambiguous so it's easy
+// to spot in bucket listings if a test object is ever leaked.
+const stagingCheckObjectPrefix = "_peerdb_check_"
+
+// NewS3StagingValidator returns a StagingValidator that smoke-tests an S3
+// bucket by writing a small object under prefix and then deleting it. Callers
+// supply a constructed *s3.Client and the bucket+prefix to test.
+func NewS3StagingValidator(client *s3.Client, bucket, prefix string) StagingValidator {
```
jgao54, line 29: NIT: can move this to `flow/pkg/objectstore/validation.go` then have it used by both clickhouse and the legacy s3 connector to get better decoupling of s3 check, and less awkward for s3 connector to import it (since s3 connector import from pkg/clickhouse would be awkward)

protos/route.proto
```diff
@@ -66,12 +66,17 @@ message PostScriptResponse { int32 id = 1; }
...
+  peerdb_peers.Peer peer = 1;
+  bool disable_connector_validation = 2;
+}
 
 message CreatePeerRequest {
   peerdb_peers.Peer peer = 1;
   bool allow_update = 2;
   bool disable_validation = 3;
+  // Only disable connector validation, keep payload and connectivity checks.
+  bool disable_connector_validation = 4;
```
jgao54: Having `ValidatePeerRequest` perform a subset of the validations seems a bit odd here, and `CreatePeerRequest` having two distinct types of validation is a bit confusing here.  Would be cleaner to just leverage the existing `disable_validation` in discovery while validating ClickHouse. 

ooc what's the motivation for this design? maybe I am missing something here.

pfcoperez: I've removed the conditional de-activation from this PR so this is not longer.

The reason behind adding a second parameter was that  `disable_validation` had a lot of more scope: For example, when set to `true` it would even avoid checking if the request included the Peer details. That is, it would have skipped all these:

https://github.com/PeerDB-io/peerdb/blob/1ed5a427b157deefc87cbb7109e6cc670d2cddd2/flow/cmd/validate_peer.go#L16-L50

So I wanted to add a soft skip level that wouldn't change the contract much.

Anyway, this is gone now :slightly_smiling_face:

flow/shared/random.go
pfcoperez: Good point, I wanted to avoid changing all the files importing the original package but that's really unnecessary. Done in:    fed4fae1e8837939d8b354b6986de093fac3bec3

flow/pkg/objectstore/validation.go
pfcoperez: Good point! Done in cb2d73ca1a460018013466829b6f3ad3adfe57c8

flow/pkg/go.mod
```diff
@@ -62,4 +62,4 @@ require (
 	google.golang.org/protobuf v1.36.6 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-)
+)
```
serprex, line 65: ?

serprex (approved):

jgao54 (approved):
lgtm!

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -74,87 +74,18 @@ func NewClickHouseConnector(
...
 
-	// rename the table
-	if err := c.exec(ctx,
-		fmt.Sprintf("RENAME TABLE %s TO %s", validateDummyTableName, validateDummyTableNameRenamed),
+	// NOTE: We are skipping staging validation, there is no need to
+	// confirm infrastructure is working at this stage again.
+	// To enable it back, just replace `objectstore.NoStagingValidator`
+	// with `c.staging.Validate`.
+	if err := peerdb_clickhouse.ValidateClickHousePeer(
+		ctx, c.logger, allowedDomains, c.Config.Host, c.database, objectstore.NoStagingValidator,
```
jgao54: actually, because of OSS validation flow via /api/v1/peers/validate, we probably want to set this to c.staging.Validate still; since with OSS user set up staging themselves and and validation is still useful.
(the side-effect of this is that CreatePeer will re-validate object store, which is fine for now as we have been running this pattern for a while).

The alternative is passing in an option to CreatePeerRequest to skip object storage validation and thread it thru, but for now I think it's okay to just validate during peer creation

pfcoperez: Good point, restored.

pfcoperez:
Manual validation: https://linear.app/clickhouse/issue/DBI-65/move-ch-validations-to-discovery#comment-b9f89773

protos/route.proto
ilidemi: @jgao54 @pfcoperez I'd prefer to add a partial skip as a part of this change, for the sake of never ever having provisioning failures from peer validation in managed envs. Overloaded CH can let the first validation through then drag the second one and page us for no reason

ilidemi: On a second look and to elaborate a bit:
- We probably still want to validate contracts but not connect to the peer
- The other path that's passing disable_validation in managed env is also intending to pass the correct contract but not hit CH
- Handler calling a handler is a little weird, so can extract all the CH-talking logic into a helper called by CreatePeer and ValidatePeer, always do contract validations in both and skip just the helper call on disable_validation=true. This way there's no need to change data contracts

PR #4296, opened 2026-05-12
add index creation to migration scripts

jgao54:
and remove the one-off `run_non_transactional_migrations`.

For existing prod DBs this is a no-op; for fresh DBs the table is empty, so a non-concurrent build is fine.

Will add a comment in the next release note about this so anyone using this in prod can also create this index before performing the upgrade.


masterashu (approved):

jgao54:
test failure is unrelated.

PR #4299, opened 2026-05-12
CI: Bump e2e timeout to 20'

pfcoperez:
Most of our successful CI e2e executions finish in the last 10'' before the 900'' (15') timeout.
This means that  most of the runs are close to hit the timeout and the slightest delay leads to a failed e2e run which.

image

This process repeats  several times until a run is fast enough per commit, per PR.

This PR bumps the timeout to 20' with the expectation of ending up with faster (no retries) and more predictable builds.


whites11 (approved):

masterashu (approved):

jgao54 (approved):

pfcoperez:
@copilot Merge main into this branch

PR #4311, opened 2026-05-15
use heartbeat to detect stale mysql connection 

jgao54:
When connecting to a MySQL server via SSH, the existing SSH keepalive can detect connectivity issues if connectivity drops between peerdb <-> bastion server; however, we have observed scenarios where the the peerdb <-> bastion host ssh tunnel is healthy (keepalive loop did not detect any issues), but events are not received even though there are writes to the replicated table; the connector waits on `mystream.GetEvent` (not receiving error or EOF). Then upon pause/resume, the pipe resumes processing from the last recorded offset. 

This fix enables `HeartbeatPeriod` (passed in as a client-side config). Every 30s, we ask MySQL server to send  a HeartbeatEvent if it's idle. This allow us to detect connectivity issue end-to-end if we don't receive heartbeat within the configured time (set as 3*heartbeat period, so 90s). Note that heartbeat events are only submitted when service is idle, tracking the time delta from the last event rather than from the last heartbeat event for detection. 

An additional benefit of heartbeat event is it prevents the connection from being stale and getting terminated somewhere upstream, especially with ssh tunnel where we don't directly connect to the server.

Fixes: DBI-750

### DIFF
--- a/flow/connectors/mysql/cdc.go
+++ b/flow/connectors/mysql/cdc.go
@@ -32,6 +32,15 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
 
+const (
+	defaultBinlogHeartbeatPeriod = 30 * time.Second
+	binlogStalenessMultiplier    = 3
+)
+
+func (c *MySqlConnector) binlogStalenessThreshold() time.Duration {
+	return binlogStalenessMultiplier * c.binlogHeartbeatPeriod
+}
+
 func (c *MySqlConnector) GetTableSchema(
 	ctx context.Context,
 	env map[string]string,
@@ -264,6 +273,7 @@ func (c *MySqlConnector) startSyncer(ctx context.Context) (*replication.BinlogSy
 		UseDecimal:       true,
 		ParseTime:        true,
 		TLSConfig:        tlsConfig,
+		HeartbeatPeriod:  c.binlogHeartbeatPeriod,
 	}), nil
 }
 
@@ -413,7 +423,7 @@ func (c *MySqlConnector) PullRecords(
 	})
 	defer shutdown()
 
-	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, time.Hour)
+	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, c.binlogStalenessThreshold())
 	//nolint:gocritic // cancelTimeout is rebound, do not defer cancelTimeout()
 	defer func() {
 		cancelTimeout()
@@ -441,6 +451,7 @@ func (c *MySqlConnector) PullRecords(
 		return nil
 	}
 
+	lastEventAt := time.Now()
 	var mysqlParser *parser.Parser
 	for inTx || (!overtime && recordCount < req.MaxBatchSize) {
 		var event *replication.BinlogEvent
@@ -456,6 +467,10 @@ func (c *MySqlConnector) PullRecords(
 				return ctxErr
 			} else if errors.Is(err, context.DeadlineExceeded) {
 				if recordCount == 0 {
+					if since := time.Since(lastEventAt); since > c.binlogStalenessThreshold() {
+						return exceptions.NewMySQLStaleConnectionError(since, c.binlogHeartbeatPeriod)
+					}
+
 					// progress offset while no records read to avoid falling behind when all tables inactive
 					if updatedOffset != "" {
 						c.logger.Info("[mysql] updating inactive offset", slog.Any("offset", updatedOffset))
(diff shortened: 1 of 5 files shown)


flow/connectors/mysql/cdc.go
```diff
@@ -413,7 +423,7 @@ func (c *MySqlConnector) PullRecords(
 	})
 	defer shutdown()
 
-	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, time.Hour)
```
jgao54, line 414: A side-effect of changing the context timeout for idle connection from 1 hour to now effectively 90s, is we are checkpointing to disk more frequently and potential more disk writes across services; but this only affect idle pipes and 90s would not introduce a high write pressure on the db.

flow/connectors/mysql/cdc.go
```diff
@@ -32,6 +32,15 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
 
+const (
+	defaultBinlogHeartbeatPeriod = 30 * time.Second
```
pfcoperez: Assuming MySQL idle connection timeout is configurable, would it make sense to make of this constant a factor of such timeout?

In other words, how do we know 30'' is the right number for all/most cases?

flow/connectors/mysql/cdc.go
```diff
@@ -491,6 +506,8 @@ func (c *MySqlConnector) PullRecords(
 			return exceptions.NewMySQLStreamingError(err)
 		}
 
+		lastEventAt = time.Now()
```
pfcoperez, line 509: Operations in error handling above such as:

```gi
if err := c.SetLastOffset(ctx, req.FlowJobName, 
```

are potentially blocking and can take a long time, would it make sense to move `		lastEventAt = time.Now()` to line 463, right before the `if err ...` block ?

pfcoperez (approved):
LGTM, left a couple of minor comments. 

It's fun how this feels like in the same category as https://github.com/PeerDB-io/peerdb/pull/4324

flow/connectors/mysql/cdc.go
jgao54: ah, if we set lastEventAt at line 463 we are no longer comparing with the last event but are comparing with the current event, and would never enter the NewMySQLStaleConnectionError block. 

in general `SetLastOffset` should be fast (it's a db write to catalog), but if for whatever reason it hangs, returning a NewMySQLStaleConnectionError and retry is still safe.

flow/connectors/mysql/cdc.go
jgao54: Ah good question. The 30s is a bit arbitrary, and is intended to allow us to notice stale connection in a reasonable amount of time. We could be a bit conservative with it, and start at a higher number, like 5 minutes (and would take 15 minutes to notice). 

The default `[wait_timeout](https://dev.mysql.com/doc/refman/9.7/en/server-system-variables.html#sysvar_wait_timeout)` is 28800 seconds (8 hours), which is too long to base heartbeat off of it.

So a side effect of this change is that if user configured a small wait_timeout (say 20s), instead of close/reopen a connection, the connection would stay established/open; given it's just 1 connection i think the impact should be benign.

pfcoperez: Makes sense, it seems to me that it is not worth the effort trying to compute it dynamically in function of source mysql configuration.

flow/connectors/mysql/cdc.go
```diff
@@ -450,45 +466,48 @@ func (c *MySqlConnector) PullRecords(
...
-							c.logger.Error("[mysql] failed to update offset, ignoring", slog.Any("error", err))
-						} else {
-							updatedOffset = ""
+			}
+
+			if errors.Is(err, context.DeadlineExceeded) {
+				if recordCount == 0 || inTx {
+					if since := time.Since(lastEventAt); since > c.binlogStalenessThreshold() {
+						return exceptions.NewMySQLStaleConnectionError(since, c.binlogHeartbeatPeriod)
+					}
```
jgao54, line 476: previously was only checking this in the `recordCount == 0` branch, but if a connection becomes stale mid-txn, we'd want to abort as well so we don't reset timeout indefinitely

flow/connectors/mysql/cdc.go
```diff
@@ -450,45 +466,48 @@ func (c *MySqlConnector) PullRecords(
...
+					continue
 				}
 
-				continue
-			} else {
-				c.logger.Error("[mysql] PullRecords failed to get event", slog.Any("error", err))
+				return nil
 			}
+
+			c.logger.Error("[mysql] PullRecords failed to get event", slog.Any("error", err))
```
jgao54, line 505: didn't change the logic here, just makes it a bit easier to follow

jgao54:
Lint error is unrelated to this PR

PR #4312, opened 2026-05-15
Add 90-day TTL on ClickHouse _peerdb_raw table

Jeremyyang920:
## Summary
- Adds `TTL fromUnixTimestamp64Nano(_peerdb_timestamp) + INTERVAL 90 DAY` to the `_peerdb_raw_*` MergeTree CREATE statement so raw-table rows expire after 90 days.
- Addresses customer reports of the raw table growing unbounded over time.

There will be a followup to expose this via a setting in both clickpipes UI and PeerDB UI

Resolves DBI-149

## Test plan
- [ ] Create a new ClickHouse mirror and confirm `SHOW CREATE TABLE _peerdb_raw_*` shows the TTL clause

### DIFF
--- a/flow/connectors/clickhouse/cdc.go
+++ b/flow/connectors/clickhouse/cdc.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"strconv"
 	"strings"
 
 	chproto "github.com/ClickHouse/ch-go/proto"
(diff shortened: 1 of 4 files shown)


serprex:
this should be configurable setting

Jeremyyang920:
> this should be configurable setting

@serprex Yes, this is discussed internally as well. For now we are hard coding this in at 90 days just to have soemthing and then we can work on exposing it via a config to users.

jgao54 (approved):
LGTM, would be good to add a simple integration/ e2e tests to validate TTL is added for ClickHouse, can be validated with something like:
```
SELECT count() > 0 AS has_ttl
FROM system.tables
WHERE name = '<table_name>' AND engine_full LIKE '%TTL%';
```

flow/connectors/clickhouse/cdc.go
```diff
@@ -34,9 +36,20 @@ const (
 		_peerdb_batch_id Int64,
 		_peerdb_unchanged_toast_columns String
 	)`
-	zooPathPrefix = "/clickhouse/tables/{uuid}/{shard}/{database}/"
+	zooPathPrefix          = "/clickhouse/tables/{uuid}/{shard}/{database}/"
+	defaultRawTableTTLDays = 90
+	// rawTableTTLDaysOverrideEnv lets tests bypass the 90-day TTL when fixtures contain stale
+	// _peerdb_timestamp values that would otherwise be evicted immediately on insert.
+	rawTableTTLDaysOverrideEnv = "PEERDB_CLICKHOUSE_RAW_TABLE_TTL_DAYS_OVERRIDE"
```
jgao54: nit: since we plan to make this configurable, we can set it up as a dynamic config and just call it `PEERDB_CLICKHOUSE_RAW_TABLE_TTL_DAYS` with 90 as the default value.

jgao54 (commented):
changes lgtm, just one nit

PR #4314, opened 2026-05-16
otel spans for cdc.batch: pull/sync/normalize

serprex:


flow/activities/flowable_core.go
```diff
@@ -247,6 +264,11 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 
 	var res *model.SyncResponse
 	errGroup.Go(func() error {
+		syncCtx, syncSpan := a.OtelManager.Tracer.Start(errCtx, "cdc.sync", trace.WithAttributes(
+			attribute.String(otel_metrics.FlowNameKey, flowName),
+			attribute.Int(otel_metrics.TableCountKey, len(options.TableMappings)),
+		))
+		defer syncSpan.End()
 		syncBatchID, err := func() (int64, error) {
```
ilidemi: Can lift this and the +1 outside of the error group and also have batch id on the pull span - would allow for a nice grafana chart correlating pull-sync-normalize

ilidemi (approved):
Great change, covers a good amount of [DBI-575](https://linear.app/clickhouse/issue/DBI-575/manual-peerdb-traces-for-cdc-batches) which we can add on top

flow/activities/flowable_core.go
ilidemi: (not blocking)

serprex: done

PR #4317, opened 2026-05-17
fix case sensitivity behavior in MariaDB when fetching schema

jgao54:
In MySQL, collation for information_schema.* tables/views are decided by `lower_case_table_names`:

```
mysql> SELECT @@lower_case_table_names;
+--------------------------+
| @@lower_case_table_names |
+--------------------------+
|                        0 |
+--------------------------+
1 row in set (0.001 sec)

mysql> SELECT COLLATION(table_name) FROM information_schema.columns LIMIT 1;
+-----------------------+
| COLLATION(table_name) |
+-----------------------+
| utf8mb3_bin           |
+-----------------------+
1 row in set (0.003 sec)
```

In MySQL 5.X, this was not the case:
```
mysql> SELECT @@lower_case_table_names;
+--------------------------+
| @@lower_case_table_names |
+--------------------------+
|                        0 |
+--------------------------+
1 row in set (0.00 sec)

mysql> SELECT COLLATION(table_name) FROM information_schema.columns LIMIT 1;
+-----------------------+
| COLLATION(table_name) |
+-----------------------+
| utf8_general_ci       |
+-----------------------+
1 row in set (0.09 sec)
```
And In MariaDB, they are hard-coded to `utf8mb3_general_ci`:

```
MariaDB [test]> SELECT @@lower_case_table_names;
+--------------------------+
| @@lower_case_table_names |
+--------------------------+
|                        0 |
+--------------------------+
1 row in set (0.000 sec)

MariaDB [test]> SELECT COLLATION(table_name) FROM information_schema.columns LIMIT 1;
+-----------------------+
| COLLATION(table_name) |
+-----------------------+
| utf8mb3_general_ci    |
+-----------------------+
1 row in set (0.088 sec)
```

And when creating a table `A` and `a`, our query returns primary key from both tables:
```
MariaDB [test]> create table A (name int primary key);
Query OK, 0 rows affected (0.006 sec)

MariaDB [test]> create table a (name int primary key);
Query OK, 0 rows affected (0.145 sec)

MariaDB [test]> select c.column_name, c.column_type, c.is_nullable, c.numeric_precision, c.numeric_scale, s.seq_in_index from information_schema.columns c left join information_schema.statistics s on c.table_schema = s.table_schema and c.table_name = s.table_name and c.column_name = s.column_name and s.index_name = 'PRIMARY' where c.table_schema = 'test' and c.table_name = 'a' order by c.ordinal_position;
+-------------+-------------+-------------+-------------------+---------------+--------------+
| column_name | column_type | is_nullable | numeric_precision | numeric_scale | seq_in_index |
+-------------+-------------+-------------+-------------------+---------------+--------------+
| name        | int(11)     | NO          |                10 |             0 |            1 |
| name        | int(11)     | NO          |                10 |             0 |            1 |
+-------------+-------------+-------------+-------------------+---------------+--------------+
```

(same issue with mysql 5.X)

The fix is based on [[ref](https://dev.mysql.com/doc/refman/9.1/en/charset-collation-information-schema.html)]:
> String columns in INFORMATION_SCHEMA tables have a collation of utf8mb3_general_ci, which is case-insensitive. However, for values that correspond to objects that are represented in the file system, such as databases and tables, searches in INFORMATION_SCHEMA string columns can be case-sensitive or case-insensitive, depending on the characteristics of the underlying file system and the [lower_case_table_names](https://dev.mysql.com/doc/refman/9.1/en/server-system-variables.html#sysvar_lower_case_table_names) system variable setting. For example, searches may be case-sensitive if the file system is case-sensitive. 

Instead of JOIN, use WHERE which matches the table using file-system semantics directly, honoring the `lower_case_table_names` system variable setting and bypasses the view's `utf8mb3_general_ci` setting.

For MariaDBs that have lower_case_table_names > 0 set, duplicated table name is not allowed, so this bug is not possible. 

Test: 
- validated behavior locally
- added e2e test, failed before code change, succeeded after. 

Fixes DBI-753

### DIFF
--- a/flow/connectors/mysql/cdc.go
+++ b/flow/connectors/mysql/cdc.go
@@ -70,14 +70,32 @@ func (c *MySqlConnector) getTableSchemaForTable(
 		return nil, err
 	}
 
+	pkSeqRs, err := c.Execute(ctx, fmt.Sprintf(`
+		select column_name, seq_in_index
+		from information_schema.statistics
+		where table_schema = '%s' and table_name = '%s' and index_name = 'PRIMARY'`,
+		mysql.Escape(qualifiedTable.Namespace), mysql.Escape(qualifiedTable.Table)))
+	if err != nil {
+		return nil, err
+	}
+	pkSeqMap := make(map[string]int64, pkSeqRs.RowNumber())
+	for idx := range pkSeqRs.RowNumber() {
+		name, err := pkSeqRs.GetString(idx, 0)
+		if err != nil {
+			return nil, err
+		}
+		seq, err := pkSeqRs.GetInt(idx, 1)
+		if err != nil {
+			return nil, err
+		}
+		pkSeqMap[name] = seq
+	}
+
 	rs, err := c.Execute(ctx, fmt.Sprintf(`
-		select c.column_name, c.column_type, c.is_nullable, c.numeric_precision, c.numeric_scale, s.seq_in_index
-		from information_schema.columns c
-		left join information_schema.statistics s
-			on c.table_schema = s.table_schema and c.table_name = s.table_name
-			and c.column_name = s.column_name and s.index_name = 'PRIMARY'
-		where c.table_schema = '%s' and c.table_name = '%s'
-		order by c.ordinal_position`,
+		select column_name, column_type, is_nullable, numeric_precision, numeric_scale
+		from information_schema.columns
+		where table_schema = '%s' and table_name = '%s'
+		order by ordinal_position`,
 		mysql.Escape(qualifiedTable.Namespace), mysql.Escape(qualifiedTable.Table)))
 	if err != nil {
 		return nil, err
@@ -132,15 +150,7 @@ func (c *MySqlConnector) getTableSchemaForTable(
 		}
 		columns = append(columns, column)
 
-		seqIsNull, err := rs.IsNull(idx, 5)
-		if err != nil {
-			return nil, err
-		}
-		if !seqIsNull {
-			seq, err := rs.GetInt(idx, 5)
-			if err != nil {
-				return nil, err
-			}
+		if seq, ok := pkSeqMap[columnName]; ok {
 			primaryEntries = append(primaryEntries, pkEntry{name: columnName, seqInIndex: seq})
 		}
 	}
--- a/flow/connectors/mysql/cdc_test.go
+++ b/flow/connectors/mysql/cdc_test.go
@@ -50,16 +50,6 @@ func createTestDB(t *testing.T, ctx context.Context, c *MySqlConnector, dbName s
 	})
 }
 
-func assertSchema(t *testing.T, schema *protos.TableSchema, pk []string, cols ...string) {
-	t.Helper()
-	require.NotNil(t, schema)
-	require.Equal(t, pk, schema.PrimaryKeyColumns)
-	require.Len(t, schema.Columns, len(cols))
-	for i, name := range cols {
-		require.Equal(t, name, schema.Columns[i].Name)
-	}
-}
-
 func TestGetTableSchemaCaseSensitiveIdentifiers(t *testing.T) {
 	t.Parallel()
 	ctx := t.Context()
@@ -88,6 +78,16 @@ func TestGetTableSchemaCaseSensitiveIdentifiers(t *testing.T) {
 		require.NoError(t, err)
 	}
 
+	assertSchema := func(schema *protos.TableSchema, pk []string, cols ...string) {
+		t.Helper()
+		require.NotNil(t, schema)
+		require.Equal(t, pk, schema.PrimaryKeyColumns)
+		require.Len(t, schema.Columns, len(cols))
+		for i, name := range cols {
+			require.Equal(t, name, schema.Columns[i].Name)
+		}
+	}
+
 	exec(fmt.Sprintf("CREATE TABLE %s (id INT PRIMARY KEY, ll TEXT)", llTable))
 	exec(fmt.Sprintf("CREATE TABLE %s (id INT PRIMARY KEY, lu TEXT)", luTable))
 	exec(fmt.Sprintf("CREATE TABLE %s (id INT PRIMARY KEY, ul TEXT)", ulTable))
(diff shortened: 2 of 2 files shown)


flow/connectors/mysql/cdc.go
```diff
@@ -70,14 +70,32 @@ func (c *MySqlConnector) getTableSchemaForTable(
...
-		left join information_schema.statistics s
-			on c.table_schema = s.table_schema and c.table_name = s.table_name
-			and c.column_name = s.column_name and s.index_name = 'PRIMARY'
-		where c.table_schema = '%s' and c.table_name = '%s'
-		order by c.ordinal_position`,
+		select column_name, column_type, is_nullable, numeric_precision, numeric_scale
+		from information_schema.columns
+		where table_schema = '%s' and table_name = '%s'
+		order by ordinal_position`,
 		mysql.Escape(qualifiedTable.Namespace), mysql.Escape(qualifiedTable.Table)))
```
ilidemi, line 90: Can also do something like
```go
rs, err := c.Execute(ctx, fmt.Sprintf(`
        select c.column_name, c.column_type, c.is_nullable, c.numeric_precision, c.numeric_scale, s.seq_in_index
        from information_schema.columns c
        left join (
                select column_name, seq_in_index
                from information_schema.statistics
                where table_schema = '%s' and table_name = '%s' and index_name = 'PRIMARY'
        ) s on c.column_name = s.column_name
        where c.table_schema = '%s' and c.table_name = '%s'
        order by c.ordinal_position`,
        mysql.Escape(qualifiedTable.Namespace), mysql.Escape(qualifiedTable.Table),
        mysql.Escape(qualifiedTable.Namespace), mysql.Escape(qualifiedTable.Table)))
```
to avoid 2x queries/network roundtrips per table. Not blocking, your choice

ilidemi (approved):

flow/connectors/mysql/cdc.go
jgao54: taking the idea, but required a bit of tweaking; query planner doing query planner things i guess  🤷‍♀️ 

```
MariaDB [test]> select c.column_name, c.column_type, c.is_nullable, c.numeric_precision, c.numeric_scale, s.seq_in_index from information_schema.columns c left join ( select column_name, seq_in_index from information_schema.statistics where table_schema = 'test' and table_name = 'A' and index_name = 'PRIMARY' ) s on c.column_name = s.column_name where c.table_schema = 'test' and c.table_name = 'A' order by c.ordinal_position;
+-------------+-------------+-------------+-------------------+---------------+--------------+
| column_name | column_type | is_nullable | numeric_precision | numeric_scale | seq_in_index |
+-------------+-------------+-------------+-------------------+---------------+--------------+
| name        | int(11)     | NO          |                10 |             0 |            1 |
| name        | int(11)     | NO          |                10 |             0 |            1 |
| foo2        | text        | YES         |              NULL |          NULL |         NULL |
+-------------+-------------+-------------+-------------------+---------------+--------------+
3 rows in set (0.009 sec)

MariaDB [test]> select column_name, column_type, is_nullable, numeric_precision, numeric_scale, (select seq_in_index from information_schema.statistics where table_schema = 'test' and table_name = 'A' and index_name = 'PRIMARY') seq_in_index from information_schema.columns where table_schema = 'test' and table_name = 'A' order by ordinal_position;
+-------------+-------------+-------------+-------------------+---------------+--------------+
| column_name | column_type | is_nullable | numeric_precision | numeric_scale | seq_in_index |
+-------------+-------------+-------------+-------------------+---------------+--------------+
| name        | int(11)     | NO          |                10 |             0 |            1 |
| foo2        | text        | YES         |              NULL |          NULL |            1 |
+-------------+-------------+-------------+-------------------+---------------+--------------+
2 rows in set (0.002 sec)

MariaDB [test]> EXPLAIN select c.column_name, c.column_type, c.is_nullable, c.numeric_precision, c.numeric_scale, s.seq_in_index from information_schema.columns c left join ( select column_name, seq_in_index from information_schema.statistics where table_schema = 'test' and table_name = 'A' and index_name = 'PRIMARY' ) s on c.column_name = s.column_name where c.table_schema = 'test' and c.table_name = 'A' order by c.ordinal_position;
+------+-------------+------------+------+---------------+-------------------------+---------+------+------+---------------------------------------------------------------------------------------+
| id   | select_type | table      | type | possible_keys | key                     | key_len | ref  | rows | Extra                                                                                 |
+------+-------------+------------+------+---------------+-------------------------+---------+------+------+---------------------------------------------------------------------------------------+
|    1 | SIMPLE      | c          | ALL  | NULL          | TABLE_SCHEMA,TABLE_NAME | NULL    | NULL | NULL | Using where; Open_frm_only; Scanned 0 databases; Using temporary; Using filesort      |
|    1 | SIMPLE      | statistics | ALL  | NULL          | NULL                    | NULL    | NULL | NULL | Using where; Open_frm_only; Scanned all databases; Using join buffer (flat, BNL join) |
+------+-------------+------------+------+---------------+-------------------------+---------+------+------+---------------------------------------------------------------------------------------+
2 rows in set (0.001 sec)

MariaDB [test]> EXPLAIN select column_name, column_type, is_nullable, numeric_precision, numeric_scale, (select seq_in_index from information_schema.statistics where table_schema = 'test' and table_name = 'A' and index_name = 'PRIMARY') seq_in_index from information_schema.columns where table_schema = 'test' and table_name = 'A' order by ordinal_position;
+------+-------------+------------+------+---------------+-------------------------+---------+------+------+-----------------------------------------------------------------+
| id   | select_type | table      | type | possible_keys | key                     | key_len | ref  | rows | Extra                                                           |
+------+-------------+------------+------+---------------+-------------------------+---------+------+------+-----------------------------------------------------------------+
|    1 | PRIMARY     | columns    | ALL  | NULL          | TABLE_SCHEMA,TABLE_NAME | NULL    | NULL | NULL | Using where; Open_frm_only; Scanned 0 databases; Using filesort |
|    2 | SUBQUERY    | statistics | ALL  | NULL          | TABLE_SCHEMA,TABLE_NAME | NULL    | NULL | NULL | Using where; Open_frm_only; Scanned 0 databases                 |
+------+-------------+------------+------+---------------+-------------------------+---------+------+------+-----------------------------------------------------------------+
2 rows in set (0.001 sec)

```

jgao54: edit: introduced a bug, updating code to support composite PK

ilidemi: Oof. Doesn't seem like we're protected from the planner in the general case here, so actually makes sense to revert to two queries. If we want to optimize later on, can still do it via batching all query 1 then all query 2

PR #4324, opened 2026-05-19
PSQL CDC: PeerDB driven WAL replication connection activity keeping it alive

pfcoperez:
WAL Replication connection in PSQL needs activity so it doesn't get closed. 
Usually, PSQL itself sends keepalive messages to prevent disconnections but there are cases where a busy server fail to do so in a timely fashion, thus leading to disconnections.

To avoid that, PeerDB set a ticket in a separate go-routine to send ping messages over this connection. However, this mechanism was not working as the connection resource was never available for this go-routine to use.

This PR:
- Moves the ping logic form the ticker, which is removed, to the `PullCdcRecords` packages loop.
- The loop lock period is reduced to `min(idle_timeout, wal_sender_timeout/")` fit in the pings required frequency to avoid disconnections.

Closes: https://linear.app/clickhouse/issue/DBI-721

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -20,7 +20,6 @@ import (
 	"github.com/jackc/pgx/v5/pgtype"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pgvector/pgvector-go"
-	"github.com/pingcap/tidb/pkg/parser/duration"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 	"go.temporal.io/sdk/log"
@@ -480,16 +479,6 @@ func PullCdcRecords[Items model.Items](
 ) error {
 	logger := internal.LoggerFromCtx(ctx)
 
-	walSenderTimeout, err := duration.ParseDuration(
-		p.PostgresConnector.conn.Config().RuntimeParams["wal_sender_timeout"],
-	)
-	if err != nil {
-		return fmt.Errorf("failed to get WAL sender timeout: %w", err)
-	}
-
-	// This value controls for how long the main message loop is blocked waiting for new messages from Postgres.
-	messageWaitPeriod := min(req.IdleTimeout, walSenderTimeout/2)
-
 	// use only with taking replLock
 	conn := p.replConn.PgConn()
 	sendStandbyAfterReplLock := func(updateType string) error {
@@ -502,6 +491,26 @@ func PullCdcRecords[Items model.Items](
 		}
 		return nil
 	}
+
+	// determine message wait period in function of idle and wal_sender timeouts
+	var walSenderTimeout time.Duration
+	if walSenderTimeoutStr, err := internal.PeerDBPostgresWalSenderTimeout(ctx, req.Env); err != nil {
+		return fmt.Errorf("could't get wal_sender_timeout parameter: %w", err)
+	} else {
+		if walSenderTimeout, err = time.ParseDuration(walSenderTimeoutStr); err != nil {
+			return fmt.Errorf("failed to parse wal_sender_timeout value: %w", err)
+		} else if walSenderTimeout <= 0 {
+			return fmt.Errorf("invalid wal_sender_timeout value: %s", walSenderTimeout)
+		}
+	}
+	// this value controls for how long the main message loop is blocked waiting for new messages from Postgres.
+	messageWaitPeriod := min(req.IdleTimeout, walSenderTimeout/2)
+
+	logger.Debug("Message wait period determined",
+		slog.Duration("messageWaitPeriod", messageWaitPeriod),
+		slog.Duration("wal_sender_timeout", walSenderTimeout),
+		slog.Duration("req.IdleTimeout", req.IdleTimeout),
+	)
 
 	records := req.RecordStream
 	var totalRecords int64
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -488,7 +488,7 @@ func PullCdcRecords[Items model.Items](
 	}
 
 	// This value controls for how long the main message loop is blocked waiting for new messages from Postgres.
-	var messageWaitPeriod time.Duration = min(req.IdleTimeout, walSenderTimeout/2)
+	messageWaitPeriod := min(req.IdleTimeout, walSenderTimeout/2)
 
 	// use only with taking replLock
 	conn := p.replConn.PgConn()
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -601,7 +601,7 @@ func PullCdcRecords[Items model.Items](
 	defer shutdown()
 
 	var standByLastLogged time.Time
-	nextStandbyMessageDeadline := time.Now().Add(messageWaitPeriod)
+	nextStandbyMessageDeadline := time.Now().Add(req.IdleTimeout)
 	pkmRequiresResponse := false
 
 	addRecordWithKey := func(key model.TableWithPkey, rec model.Record[Items]) error {
@@ -618,7 +618,7 @@ func PullCdcRecords[Items model.Items](
 
 		if totalRecords == 1 {
 			records.SignalAsNotEmpty()
-			nextStandbyMessageDeadline = time.Now().Add(messageWaitPeriod)
+			nextStandbyMessageDeadline = time.Now().Add(req.IdleTimeout)
 			logger.Info(fmt.Sprintf("pushing the standby deadline to %s", nextStandbyMessageDeadline))
 		}
 		if totalRecords%50000 == 0 {
@@ -706,14 +706,17 @@ func PullCdcRecords[Items model.Items](
 			} else {
 				logger.Info(("standby deadline reached, no records accumulated, continuing to wait"))
 			}
-			nextStandbyMessageDeadline = time.Now().Add(messageWaitPeriod)
+			nextStandbyMessageDeadline = time.Now().Add(req.IdleTimeout)
 		}
 
-		var receiveCtx context.Context
-		var cancel context.CancelFunc
-
-		receiveCtx, cancel = context.WithDeadline(ctx, nextStandbyMessageDeadline)
-
+		// Since we are interrupting the receive message call as soon as `messageWaitPeriod` is over
+		// to send a ping to Postgres, now()+messageWaitPeriod might overshoot `nextStandbyMessageDeadline`
+		// this check prevents that situation.
+		receiveDeadline := time.Now().Add(messageWaitPeriod)
+		if receiveDeadline.After(nextStandbyMessageDeadline) {
+			receiveDeadline = nextStandbyMessageDeadline
+		}
+		receiveCtx, cancel := context.WithDeadline(ctx, receiveDeadline)
 		rawMsg, err := func() (pgproto3.BackendMessage, error) {
 			replLock.Lock()
 			defer replLock.Unlock()
@@ -726,12 +729,21 @@ func PullCdcRecords[Items model.Items](
 		}
 
 		if err != nil && pgconn.Timeout(err) {
-			// Send ping to make sure the connection is held alive.
-			if err := p.ReplPing(ctx); err != nil {
-				return fmt.Errorf("ReplPing failed: %w", err)
+			// If we have hit timeout, either we just need to ping to keep the connection alive
+			// or we are actually hitting `nextStandbyMessageDeadline`.
+			if !time.Now().After(nextStandbyMessageDeadline) {
+				// The timeout is not hit, hence we just send a ping before moving on to read more packages.
+				if err := p.ReplPing(ctx); err != nil {
+					return fmt.Errorf("ReplPing failed: %w", err)
+				}
 			}
+			// After that, we let the condition checks at the beginging of the loop,
+			// specifically "if we are past the next standby deadline (?)" to
+			// handle having reached the standby deadline.
 			continue
-		} else if err != nil {
+		}
+
+		if err != nil {
 			return fmt.Errorf("ReceiveMessage failed: %w", err)
 		}
 
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log/slog"
 	"maps"
+	"regexp"
 	"slices"
 	"strings"
 	"sync"
@@ -469,6 +470,12 @@ func (p *PostgresCDCSource) decodeColumnData(
 	return types.QValueString{Val: string(data)}, nil
 }
 
+func clientSidePingPeriod(walSenderTimeout time.Duration) time.Duration {
+	return 3 * walSenderTimeout / 4
+}
+
+const NoWALSenderTimeoutSetting = "NONE"
+
 // PullCdcRecords pulls records from req's cdc stream
 func PullCdcRecords[Items model.Items](
 	ctx context.Context,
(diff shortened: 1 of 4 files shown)


pfcoperez:
## QA Validation

### Tooling

To reproduce the issue, a customized version of PG18 was used: https://github.com/pfcoperez/postgres-lab/tree/keepalive-deactivation-18

In this version, it is possible to disable WAL keep alive outbound messages from Postgres using at any time with a couple of queries:

```sql
ALTER SYSTEM SET wal_sender_keepalive_enabled = off;
select pg_reload_conf();
```

A Docker image for this special PG edition can be obtained as follows:

1. Clone the repository and check out the branch at  https://github.com/pfcoperez/postgres-lab/tree/keepalive-deactivation-18
2. Get into the source directory and run:

```bash
./configure # Install required dependencies until this step succeeds 
make docker-image
```

3. Confirm the image has been generated and tagged locally as `postgres-custom:latest`:

```bash
docker ps | grep postgres-custom
```

With the image at hand, after having run the local Tilt environment at least once (so the `ancillary.env` file is generated), edit the  environment env file for Tilt and make sure the crafted image is used:

```
➜  peerdb-alt git:(DBI-721/keep-wal-sender-alive) ✗ grep POST ancillary.env 
#POSTGRES_IMAGE=imresamu/postgis:18-3.5-alpine
POSTGRES_IMAGE=postgres-custom:latest 
```

Now start the Tilt environment from scratch and enable Postgres Ancillary resource:

Screenshot From 2026-05-18 14-53-09

Additionally, it makes the debugging session easier to add some telltales to the `ReplPing` method for type `PostgresConnector`:

Screenshot From 2026-05-18 15-21-59


### Checks

These are the scenarios validating the change.

All of them use the same test source table: `events`.

- Created with: `CREATE TABLE events (id serial primary key, t timestamp);`
- Populated with events with: `insert into events (t) SELECT now();`
- Checked with: `select count(*) from events;`

Replication is verified using the following SQL query in target ClikcHouse:

```sql
select * from public_events order by t desc limit 1;
```    

#### Scenario 0: Reproduce the issue without any change

1. Create a pipe in peerdb-ui with PG using our crafted image as source and CH as destination.
2. Insert some events and confirm replication.
3. Disable WAL keepalive messages in PG with the alter system query indicated above.
4. Wait as much as `wal_sender_timeout`
5. Insert some more events.
6. Observe lack of replication and failed workflow in temporal ui with `unexpected EOF` at the time of trying to use the now expired connection:

Screenshot From 2026-05-18 12-33-47

Screenshot From 2026-05-18 12-35-57

:memo: Observe the lack of telltale messages in `flow-worker` container logs.

#### Scenario 1: Changes are present and the problem is fixed

1. Repeat `Scenario 0`'s steps 1-5 with the changes in this PR.
2. Observe the telltales printing: 
image
3. The flow doesn't longer crash.
4. New events are replicated.

flow/connectors/postgres/cdc.go
```diff
@@ -492,6 +492,26 @@ func PullCdcRecords[Items model.Items](
...
 
+	// determine message wait period in function of idle and wal_sender timeouts
+	var walSenderTimeout time.Duration
+	if walSenderTimeoutStr, err := internal.PeerDBPostgresWalSenderTimeout(ctx, req.Env); err != nil {
+		return fmt.Errorf("could't get wal_sender_timeout parameter: %w", err)
+	} else {
+		if walSenderTimeout, err = time.ParseDuration(walSenderTimeoutStr); err != nil {
+			return fmt.Errorf("failed to parse wal_sender_timeout value: %w", err)
+		} else if walSenderTimeout <= 0 {
+			return fmt.Errorf("invalid wal_sender_timeout value: %s", walSenderTimeout)
```
jgao54, line 515: nit: `wal_sender_timeout` supports a value with or without a unit (in ms if without), but ParseDuration will error if it gets a non-zero integer. To not shoot ourselves in the foot, may be good to to support non-unit values as well. 

similarly, accepting 0 seems useful as it's supported by pg. and i can see a scenario where we need to set this to 0 when we can't get the estimate correct.  So would be good for messageWaitPeriod to not depend on this value when it's 0.

jgao54: actually, we also support `NONE` as a valid value [ref](https://github.com/PeerDB-io/peerdb/blob/71d6cdd13ccc3048c3563eaee06895c7d43ef68f/flow/connectors/postgres/postgres_source.go#L61), which would end up using pg server's `wal_sender_timeout` and is something we'd want to respect here too.

flow/connectors/postgres/cdc.go
```diff
@@ -562,7 +582,7 @@ func PullCdcRecords[Items model.Items](
 	defer shutdown()
 
 	var standByLastLogged time.Time
-	nextStandbyMessageDeadline := time.Now().Add(req.IdleTimeout)
+	nextStandbyMessageDeadline := time.Now().Add(messageWaitPeriod)
```
jgao54: this area of the code is quite complex and has non-obvious footguns.  The current behavior we want to maintain is:
1. before we receive the first record, a PullRecord cycle runs infinitely
2. after we receive the first record, a PullRecord cycle ends at the first of "sync interval is reached" or "max records is processed"
3. _unless_, if we are mid-transaction when 2 happens, we process all the records until we hit commit 

So when `nextStandbyMessageDeadline` is modified, if idle timeout > walSenderTimeout / 2, it introduces a subtle but important behavior change which is a PullRecords cycle will no longer respect `IdleTimeout`.

flow/connectors/postgres/cdc.go
```diff
@@ -689,13 +707,17 @@ func PullCdcRecords[Items model.Items](
...
-			if pgconn.Timeout(err) {
+			if totalRecords != 0 && pgconn.Timeout(err) {
 				logger.Info("Stand-by deadline reached, returning currently accumulated records",
 					slog.Int64("records", totalRecords),
 					slog.Int64("bytes", totalFetchedBytes.Load()),
 					slog.Int("channelLen", records.ChannelLen()),
 					slog.Float64("elapsedMinutes", time.Since(pullStart).Minutes()))
 				return nil
+			} else if pgconn.Timeout(err) {
+				if err := p.ReplPing(ctx); err != nil {
```
jgao54, line 736: ReplPing here is in the `p.commitLock == nil` block, which means this won't run during large txns, which is the particular scenario where ReplPing is the most important

Edit: actually, I am wrong. if commitLock !=nil, that means we are processing records already so we don't actually need the ping. it's the "idle" pre-arrival of the large txn that is returning EOF. So this is reasonable and doesn't have to be moved out side of the if condition. but maybe worth a comment.

flow/connectors/postgres/cdc.go
pfcoperez: Think out loud mode:

The way I see this function is as a state machine where messages generate transitions and outputs.
Being these messages waited for a maximum period of time and, deadlines, treated as yet another message.

With this idea, I observed that all  `nextStandbyMessageDeadline = time.Now().Add(...)` updates in the loop used `req.IdleTimeout`. I considered that as the beat-drum of the loop waiting for messages. 

With this change, the idea here is to reduce the loop wait cycles to a fraction of `walSenderTimeout` trying not to affect the logic of the loop.

The way I see it, this block:

```go
               rawMsg, err := func() (pgproto3.BackendMessage, error) {
			replLock.Lock()
			defer replLock.Unlock()
			return conn.ReceiveMessage(receiveCtx)
		}()
```

Is the one waiting for new messages from PQ-WAL connection. It was while this call was blocked when the connection became stale as PKMs were delayed because PSQL was busy. This is the critical point to interrupt before the wal_sender_timeout expires to send ping upstream.

`messageWaitPeriod` is computed as a minimum:

```go
messageWaitPeriod := min(req.IdleTimeout, walSenderTimeout/2)
```

So the potential change in behviour occurs when `walSenderTimeout/2 < req.IdleTimeout`  as you rightfully highlighted. In this case, we are effectively reducing the wait time.

Does this affect the logic in the block dealing with `rawMsg` reception?

```go
if err != nil && p.commitLock == nil {
			if totalRecords != 0 && pgconn.Timeout(err) {
				logger.Info("Stand-by deadline reached, returning currently accumulated records",
					slog.Int64("records", totalRecords),
					slog.Int64("bytes", totalFetchedBytes.Load()),
					slog.Int("channelLen", records.ChannelLen()),
				return nil
			} else if pgconn.Timeout(err) {
				if err := p.ReplPing(ctx); err != nil {
					return fmt.Errorf("ReplPing failed: %w", err)
				}
			} else {
				return fmt.Errorf("ReceiveMessage failed: %w", err)
			}
		}
```

Only on the first and second branches in the if block:

- For (2), we are sending the ping as we want and then moving on the loop, the switch does nothing and we start the loop again with expiration times that might deal to expiration actions.
- For (1), we are are reaching the deadline earlier.

I think you are right in that it leads to noticeable changes.

I think the way of hardening is: For (1): Use Afer checks to do no-op when the time is not ready.

jgao54: I think we can delete the `totalRecords != 0 && pgconn.Timeout(err)` branch altogether, and simplify it to:
```
  if err != nil {
      if pgconn.Timeout(err) {
          if err := p.ReplPing(ctx); err != nil {
              return fmt.Errorf("ReplPing failed: %w", err)
          }
          continue
      }
      return fmt.Errorf("ReceiveMessage failed: %w", err)
  }
```
Then we just have a single place (where "no commit lock, returning currently accumulated records" is logged) that handles the "returning on idle timeout" in the next iteration.

flow/connectors/postgres/cdc.go
```diff
@@ -492,6 +492,26 @@ func PullCdcRecords[Items model.Items](
...
+		return fmt.Errorf("could't get wal_sender_timeout parameter: %w", err)
+	} else {
+		if walSenderTimeout, err = time.ParseDuration(walSenderTimeoutStr); err != nil {
+			return fmt.Errorf("failed to parse wal_sender_timeout value: %w", err)
+		} else if walSenderTimeout <= 0 {
+			return fmt.Errorf("invalid wal_sender_timeout value: %s", walSenderTimeout)
+		}
+	}
+	// this value controls for how long the main message loop is blocked waiting for new messages from Postgres.
+	messageWaitPeriod := min(req.IdleTimeout, walSenderTimeout/2)
```
ilidemi: As PG would normally send its own keepalive at sender_timeout/2, maybe do sender_timeout*0.75?

flow/connectors/postgres/cdc.go
pfcoperez: Because it is earlier in the loop, that makes sense.
It introduces the change that now we'll ping (which can fail) even in the midst of a transaction though.

The ping is implemented as `pglogrepl.SendStandbyStatusUpdate`, this goes back to @jgao54 's previous comment: https://github.com/PeerDB-io/peerdb/pull/4324#pullrequestreview-4343423701

Which I think is actually accurate: We might be processing events but that doesn't mean that the source is not stuck busy not sending WAL connection keep alive messages.

I've followed this simplification path at: 93f501b7c801462947afdc218be7f0851167e8ef

flow/connectors/postgres/cdc.go
pfcoperez: Resolved with https://github.com/PeerDB-io/peerdb/pull/4324#discussion_r3296978299

flow/connectors/postgres/cdc.go
pfcoperez: Good point! addressed in a34f327082c761efd6aba884a686ebfe1df393b8

flow/connectors/postgres/cdc.go
pfcoperez: I chose 1/2  to allow for a ping miss without hitting the  timeout but, this is TCP and a ping is unlikely to be delayed as much as timeout*2, bumped in a34f327082c761efd6aba884a686ebfe1df393b8 where I also extracted the magic number into a function and improved value extraction reliability after @jgao54 's comments at https://github.com/PeerDB-io/peerdb/pull/4324#pullrequestreview-4342918595

pfcoperez:
@jgao54 @ilidemi Thanks a lot for the thorough reviews and suggestions, I think I've addressed them all.
After these changes, I've followed the same manual validation as described at https://github.com/PeerDB-io/peerdb/pull/4324#issuecomment-4499108985

with successful outcome.

flow/connectors/postgres/cdc.go
```diff
@@ -586,7 +618,7 @@ func PullCdcRecords[Items model.Items](
 
 		if totalRecords == 1 {
 			records.SignalAsNotEmpty()
-			nextStandbyMessageDeadline = time.Now().Add(req.IdleTimeout)
+			nextStandbyMessageDeadline = time.Now().Add(messageWaitPeriod)
```
jgao54: this needs to be idleTimeout when first record is received,  to ensure that the PullRecords method terminate on idle timeout [here](https://github.com/PeerDB-io/peerdb/blob/6b8b0da4864e141ac57b824169d0af8e9513d902/flow/connectors/postgres/cdc.go#L687-L697)
 
exception is 0 record or mid-txn, in which case we extend the timeout.

jgao54: i think we should keep `nextStandbyMessageDeadline` unchanged to represent the deadline for the PullRecord function, but make sure `conn.ReceiveMessage(receiveCtx)` wait for no more than `messageWaitPeriod`, instead of setting `messageWaitPeriod` to `nextStandbyMessageDeadline`

pfcoperez: As discussed, yes you are right! In general we should keep the `nextStandbyMessageDeadline` time completely separated from the period used at the time cadence.

That's done in 127f7f1cba7a9c27a945db12c96aeff03423c1ab

In that commit, we also adjust the receive wait time at each iteration so `nextStandbyMessageDeadline` is not overshot as much as `messageWaitPeriod`.

flow/connectors/postgres/cdc.go
```diff
@@ -726,12 +729,21 @@ func PullCdcRecords[Items model.Items](
 		}
 
 		if err != nil && pgconn.Timeout(err) {
-			// Send ping to make sure the connection is held alive.
-			if err := p.ReplPing(ctx); err != nil {
-				return fmt.Errorf("ReplPing failed: %w", err)
+			// If we have hit timeout, either we just need to ping to keep the connection alive
+			// or we are actually hitting `nextStandbyMessageDeadline`.
+			if !time.Now().After(nextStandbyMessageDeadline) {
+				// The timeout is not hit, hence we just send a ping before moving on to read more packages.
```
jgao54: nit: read more messages(?)

jgao54 (approved):

flow/connectors/postgres/cdc.go
jgao54: nit: should we rename this to `pullRecordsDeadline` (or something else that represent what the variable is actually for. `nextStandbyMessageDeadline` used to make sense but naming is just misleading now

pfcoperez: This is a great point, actually a great deal of the inertia that led to the original proposal to mix `nextStandbyMessageDeadline` and the message check cadence came from assuming the semantic of the name.

@claude Please rename  `nextStandbyMessageDeadline` to `nextRecordDeadline`.

flow/connectors/postgres/cdc.go
pfcoperez: @claude Apply this substitution 's/on to read more packages/on to read more messages/' in this comment following @jgao54 's suggestion.

flow/connectors/postgres/cdc.go
pfcoperez: Yeah, AI ignored me because my org membership was not public

pfcoperez: @claude Please rename  `nextStandbyMessageDeadline` to `nextRecordDeadline`.

flow/connectors/postgres/cdc.go
pfcoperez: @claude Apply this substitution 's/on to read more packages/on to read more messages/' in this code comment.

PR #4326, opened 2026-05-20
AdditionalTablesHasOverlap: don't check destination tables

serprex:
We key on source tables as a feature, allowing multiple sources to the same destination

This is particularly important for MySQL when partitioning is be ad-hoc


Amogh-Bharadwaj (approved):

jgao54 (changes requested):
Discussed with @serprex as this has a user-facing behavior change (i.e. allow table addition to be written to the same destination), and we should put this behind `PEERDB_CLICKHOUSE_INITIAL_LOAD_ALLOW_NON_EMPTY_TABLES`

jgao54 (approved):

jgao54 (commented):
nit: shared.ArraysHaveOverlap can be deleted

PR #4329, opened 2026-05-20
Expose environment variables to adjust workflow timeouts

serprex:
These can timeout on requests involving 1000+ tables

Only implementing as environment variable for now, could be expanded to dynamic conf in future


ilidemi (approved):

ilidemi:
Lint failure unrelated #4341

PR #4340, opened 2026-05-23
fix: avoid duplicate normalize error logs

ilidemi:
## Summary
- `startNormalize` was calling `Alerter.LogFlowError` on normalize failures, but the outer activity error handler already logs the returned error. Result: every normalize failure showed up 2× in the mirror logs view.
- Return the wrapped/`NormalizationError` directly and let the activity handler do the single log.

Low pri but had an active repro during local testing.

Screenshot 2026-05-22 at 11 06 36 PM

Closes DBI-402


ilidemi:
Linter error is from main, fix in #4341

jgao54 (approved):

PR #4349, opened 2026-05-25
feat: add testing for more types for PG type system mirrors, split test

Amogh-Bharadwaj:
- Adds more types testing for PG type system 
- Splits type test for PG destination into two


serprex:
In this case, should make sure to test values pgpg preserves over qvalue: nans & infinities. Thought that was already covered

should also be making every pgpg test both normalization methods until we have confidence to use just the one

serprex (approved):

PR #4358, opened 2026-05-26
DBI-707/sub/DBI-763: Enable TLS in local Tilt environment

pfcoperez:
This PR enables TLS on PG instances managed as ancillary services in local Tilt environment.

To do so, two long lived certs (`Not After : Dec 31 23:59:59 9999 GMT`)  are added to the repository and used in the entry point for these instances.

Additionally, two env vars have been added to the e2e test code for each one of the PG instances:

- Primary PG source: 
  - `PG_REQUIRE_TLS`: Boolean (default: `false`) controlling `RequireTls` in PG connector config used in the tests. 
  - `PG_ROOT_CA_PATH`: String containing the path of the CA certificate, we can re-use the server certificate for this purpose. e.g: `$(pwd)/../volumes/pg-tls/server.crt`. If present, the test reads the file and pass it to  `RootCa` connector configuration
- Second PG source:
  - `PG2_REQUIRE_TLS`: Same as `PG_REQUIRE_TLS` for the second PG instance.
  - `PG2_ROOT_CA_PATH`: Same as `PG2_ROOT_CA_PATH` for the seconf PG instance.

Finally, a test resource has been added to the Tilt file to exercise `TestGenericCH_PG` e2e tests using TLS connections.

:warning: TLS and plaintext connections are accepted with this change in PG instances regardless of the envars passed to the tests. These envars only control whether the tests  will set-up TLS in their mirrors. 

Closes: https://linear.app/clickhouse/issue/DBI-763


pfcoperez:
cc @Amogh-Bharadwaj This might be useful for you too.

masterashu (approved):

PR #4362, opened 2026-05-28
error classification: classify a transient mysql error

jgao54:
- Rename to from `MySQLStreamingError` to `MySQLExecuteError` to be more general (this covers `Execute`, `ExecuteSelectStreaming`, as well errors coming from `syncer.StartSync` and ` syncer.StartSyncGTID`), basically upstream execution protocol related errors
- Make the `invalid compressed sequence` error retryable (both at the query level, and at the error classification level, in case the states in the existing conn is broken and recreating the connector is required)
- Make functions like `Execute` and `ExecuteSelectStreaming` in mysql.go return`MySQLExecuteError`, instead of having to wrap it at every caller. 

Fixes: DBI-769

### DIFF
--- a/flow/connectors/mysql/mysql.go
+++ b/flow/connectors/mysql/mysql.go
@@ -302,20 +302,25 @@ func (c *MySqlConnector) Execute(ctx context.Context, cmd string, args ...any) (
 	var retryableErr error
 	for conn, err := range c.withRetries(ctx) {
 		if err != nil {
-			return nil, err
+			return nil, exceptions.NewMySQLExecuteError(err)
 		}
 
 		if rs, err := conn.Execute(cmd, args...); err != nil {
-			if mysql.ErrorEqual(err, mysql.ErrBadConn) || exceptions.InvalidSequenceRe.MatchString(err.Error()) {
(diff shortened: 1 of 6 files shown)


flow/connectors/mysql/mysql.go
```diff
@@ -302,20 +302,25 @@ func (c *MySqlConnector) Execute(ctx context.Context, cmd string, args ...any) (
...
 		if err != nil {
-			return nil, err
+			return nil, exceptions.NewMySQLExecuteError(err)
 		}
 
 		if rs, err := conn.Execute(cmd, args...); err != nil {
-			if mysql.ErrorEqual(err, mysql.ErrBadConn) || exceptions.InvalidSequenceRe.MatchString(err.Error()) {
+			if mysql.ErrorEqual(err, mysql.ErrBadConn) ||
+				exceptions.InvalidSequenceRe.MatchString(err.Error()) ||
+				exceptions.InvalidCompressedSequenceRe.MatchString(err.Error()) {
```
ilidemi: Just from the verbosity standpoint, should we merge the two regexes?

ilidemi (approved):

PR #4363, opened 2026-05-28
PG: clean up replication pings, unify wal_sender_timeout handling

ilidemi:
* Make ReplPing private as it's only called within the PG connector
* Unify wal_sender_timeout handling - started as deduping the constant
but we'd really want to reduce catalog accesses on every batch and
sticking to a canoncial value seems good. The struct thing still seems
messy but that's the least messy shape I could think of without losing
nuance

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"log/slog"
 	"maps"
-	"regexp"
 	"slices"
 	"strings"
 	"sync"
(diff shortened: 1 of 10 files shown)


flow/connectors/postgres/postgres_source.go
```diff
@@ -55,47 +94,32 @@ func (c *PostgresConnector) CreateReplConn(ctx context.Context, env map[string]s
...
 	if err != nil {
 		return err
 	}
 	c.replConn = conn
-	return nil
-}
-
-// To keep connection alive between sync batches.
-// By default postgres drops connection after 1 minute of inactivity.
-func (c *PostgresConnector) ReplPing(ctx context.Context) error {
```
jgao54, line 115: nit: Looks like `ReplPing` loads StandbyStatusUpdate from `c.replState.LastOffset.Load()`, while the `sendStandbyAfterReplLock`/`sendStandbyStatusUpdate` method loads it from `req.ConsumedOffset.Load()`.

Not sure if the c.replState.LastOffset field still provides any value (it looks like the only place that reads from the value is here); if not, maybe we could delete that too

jgao54 (approved):

jgao54 (commented):
nice cleanup!

flow/connectors/postgres/postgres_source.go
serprex: `LastOffset` is used for streaming connectors which update their LSN midbatch, since they don't really have batches to begin with

serprex (approved):
I see #4324 was where this optimization became possible, which itself opened up when we stopped splitting up sync/normalize phases

flow/connectors/postgres/cdc.go
```diff
@@ -476,23 +474,17 @@ func clientSidePingPeriod(walSenderTimeout time.Duration) time.Duration {
...
 	ctx context.Context,
 	p *PostgresCDCSource,
 	req *model.PullRecordsRequest[Items],
 	processor replProcessor[Items],
-	replLock *sync.Mutex,
 ) error {
 	logger := internal.LoggerFromCtx(ctx)
 
-	// use only with taking replLock
 	conn := p.replConn.PgConn()
```
pfcoperez, line 489: not blocking: As someone who as been diving into this code lately, I think it wouldn't harm to add a comment   

```suggestion
    // keep in mind that `PgConn` doesn't offer safe concurrent access, do not use in goroutines without protection. 
	conn := p.replConn.PgConn()
```

flow/connectors/postgres/postgres_source.go
```diff
@@ -55,47 +94,32 @@ func (c *PostgresConnector) CreateReplConn(ctx context.Context, env map[string]s
...
-		return nil, fmt.Errorf("failed to create replication connection: %w", err)
+		return nil, walSenderTimeout{}, fmt.Errorf("failed to create replication connection: %w", err)
 	}
-	return conn, nil
+	return conn, wst, nil
 }
 
 func (c *PostgresConnector) SetupReplConn(ctx context.Context, env map[string]string) error {
-	conn, err := c.CreateReplConn(ctx, env)
+	conn, wst, err := c.CreateReplConn(ctx, env)
```
pfcoperez, line 117: Since walSenderTimeout is only used here as `wst`, and since its constructor function is self contained, what's the reason for for `CreateReplConn` return it? 

Would it make sense just to have:

```go
conn, err := c.CreateReplConn(ctx, env)

wst := WalSenderTimeoutOverride(ctx, env) 
```

pfcoperez (approved):
Nice clean-up! Just left a couple of non-blocking comments.

pfcoperez:
> Nice clean-up! Just left a couple of non-blocking comments.

Another point @ilidemi , I am 99% sure that your changes don't change the behaviour of the message loop. Yet I wonder if you validated using a wal-PKM suppressed PG instance to confirm empirically.

flow/connectors/postgres/postgres_source.go
jgao54: in the case for Postgres, it seem to be updating `req.ConsumedOffset` mid-batch, replState.LastOffset is only being read by the already-deleted ping so i think is safe to delete here.  (EDIT: ah, this is only if there are 0 records so this is safe to send PKM response)

(and at the end of a PullRecords, `c.replState.Offset` is updated with the latest checkpoint from the record stream)

jgao54:
since we brought back the async keepalive goroutine. we'll still need:
- replLock
- replState.LastOffset

to bring this PR up-to-date.

flow/connectors/postgres/cdc.go
ilidemi: Keeping the lock instead

flow/connectors/postgres/postgres_source.go
ilidemi: CreateReplConn needs to apply it to Postgres, so the options are either reading before and passing as an optional arg (then CreateReplConn mutates the connector) or adding an optional return value. Right now SetupReplConn is the one mutating the connector so seems more uniform to treat the new field the same way.

ilidemi:
> I am 99% sure that your changes don't change the behaviour of the message loop. Yet I wonder if you validated using a wal-PKM suppressed PG instance to confirm empirically.

Now it for sure doesn't modify the keepalive behavior

PR #4365, opened 2026-05-29
fix(clickhouse): normalization when `_peerdb_is_deleted` has different column casing

isakruas:
### Summary

This change fixes a ClickHouse normalization failure caused by assuming a fixed lowercase column name for `_peerdb_is_deleted`.

### Problem

Replication could fail during normalization with an error like:

`Normalization Error: failed to normalize records: ... code: 16, message: No such column _peerdb_is_deleted in table tenant_x.sample_table (...)`

### Root Cause

Normalization logic referenced `_peerdb_is_deleted` using a hardcoded identifier.
In some ClickHouse tables, the physical column exists with different casing (for example `_PEERDB_IS_DELETED`), so the generated query referenced a non-existent identifier.

### Fix Implemented

- Updated normalization/validation logic to resolve the actual soft-delete column name from the destination schema in a case-insensitive way.
- Ensured the query uses the resolved identifier instead of a fixed lowercase literal.

### Expected Impact

- Prevents `code: 16` normalization failures due to casing mismatch.
- Improves compatibility across ClickHouse schemas where `_peerdb_is_deleted` may be created with different letter casing.
- Reduces replication interruptions caused by column-name assumptions.

### Files Changed

- `flow/connectors/clickhouse/normalize_query.go`
- `flow/connectors/clickhouse/validate.go`

### DIFF
--- a/flow/connectors/clickhouse/normalize_query.go
+++ b/flow/connectors/clickhouse/normalize_query.go
@@ -283,8 +283,8 @@ func (t *NormalizeQueryGenerator) BuildQuery(ctx context.Context) (string, error
 	}
 
 	// add _peerdb_sign as _peerdb_record_type / 2
-	fmt.Fprintf(&projection, "intDiv(_peerdb_record_type, 2) AS %s,", peerdb_clickhouse.QuoteIdentifier(isDeletedColName))
-	fmt.Fprintf(&colSelector, "%s,", peerdb_clickhouse.QuoteIdentifier(isDeletedColName))
+	fmt.Fprintf(&projection, "intDiv(_peerdb_record_type, 2) AS %s,", peerdb_clickhouse.QuoteIdentifier(t.isDeletedColName))
+	fmt.Fprintf(&colSelector, "%s,", peerdb_clickhouse.QuoteIdentifier(t.isDeletedColName))
 
 	// add _peerdb_timestamp as _peerdb_version
 	fmt.Fprintf(&projection, "_peerdb_timestamp AS %s", peerdb_clickhouse.QuoteIdentifier(versionColName))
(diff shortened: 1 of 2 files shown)


ilidemi (approved):
Thanks for the contribution!

nastynaz:
@ilidemi I think this broke my setup. I upgraded from v0.27.1 to v0.37.1 and I'm getting these errors in logs for _all_ of by postgres -> clickhouse mirrors:
```
Normalization Error: failed to normalize records: error while inserting into target clickhouse table simulation_jobs: code: 16, message: No such column _PEERDB_IS_DELETED in table tdb.simulation_jobs (b6ee3e4b-4fd2-449f-9172-82a7fcfb818c)
```
Under the old default mirror configuration they were using `_peerdb_is_deleted` as the column name.
How do I fix it without renaming all my columns?

I tried to recreate the mirror using `CREATE MIRROR` using these settings but it also fails:
```sql
CREATE MIRROR <mirror_name>
FROM <postgres peer>
TO <clickhouse peer>
WITH TABLE MAPPING
(
  -- mappings
)
WITH (
  sync_interval = 60,
  do_initial_copy = false,
  publication_name = 'peerdb_publication',
  snapshot_max_parallel_workers = 2,
  soft_delete = true,
  soft_delete_col_name = '_peerdb_is_deleted',
  synced_at_col_name = '_peerdb_synced_at '
);
```
However this also gives me errors:
```shell
ERROR:  unable to submit job: "code: 'Internal error', message: \"invalid mirror: rpc error: code = FailedPrecondition desc = failed to validate destination connector <clickhouse peer>: not all PeerDB columns found in destination table <table name>\""
```

PR #4366, opened 2026-05-30
Seperately configure enabling CDC store for ClickHouse destination

jgao54:
- Introduce `PEERDB_CLICKHOUSE_CDC_STORE_ENABLED` to allow separating handling of cdc store enablement for ClickHouse (since other destinations like PG-to-PG and PG-to-Snowflake requires this to be enabled). 
- Still honor existing services with CDC store disabled, to avoid OOM issues resurfacing
- Currently `PEERDB_CLICKHOUSE_CDC_STORE_ENABLED` is default to `true`, can be updated to `false` once we are ready to flip the switch (PR is currently a no-op)

---

UPDATE: 
- consolidated to just `PEERDB_CDC_STORE_ENABLED`, but with default value destination-specific. 
- `PEERDB_CDC_STORE_ENABLED` default to false for ClickHouse as destination. So we can hold off merging this PR for now until we are ready for this change.

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -356,6 +356,22 @@ func (a *FlowableActivity) SyncFlow(
 	ctx = internal.WithOperationContext(ctx, protos.FlowOperation_FLOW_OPERATION_SYNC)
 	logger := internal.LoggerFromCtx(ctx)
 
+	destinationType, err := connectors.LoadPeerType(ctx, a.CatalogPool, config.DestinationName)
+	if err != nil {
+		return a.Alerter.LogFlowError(ctx, config.FlowJobName, fmt.Errorf("failed to load destination peer type: %w", err))
+	}
+
+	cdcStoreEnabled, err := internal.PeerDBCDCStoreEnabled(ctx, config.Env, destinationType)
+	if err != nil {
+		return fmt.Errorf("failed to fetch PEERDB_CDC_STORE_ENABLED setting: %w", err)
+	}
+
+	// Postgres connector reads this at construction to populate its cdcStoreEnabled
+	// attribute; only set here because cdcStoreEnabled is only used by SyncFlow
+	if config.Env == nil {
+		config.Env = make(map[string]string)
+	}
+	config.Env["PEERDB_CDC_STORE_ENABLED"] = strconv.FormatBool(cdcStoreEnabled)
 	srcConn, srcClose, err := connectors.GetByNameAs[connectors.CDCPullConnectorCore](ctx, config.Env, a.CatalogPool, config.SourceName)
 	if err != nil {
 		return a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
@@ -369,16 +385,6 @@ func (a *FlowableActivity) SyncFlow(
 	reconnectAfterBatches, err := internal.PeerDBReconnectAfterBatches(ctx, config.Env)
 	if err != nil {
 		return a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
-	}
-
-	destinationType, err := connectors.LoadPeerType(ctx, a.CatalogPool, config.DestinationName)
-	if err != nil {
-		return a.Alerter.LogFlowError(ctx, config.FlowJobName, fmt.Errorf("failed to load destination peer type: %w", err))
-	}
-
-	cdcStoreEnabled, err := internal.PeerDBCDCStoreEnabledForDestination(ctx, config.Env, destinationType)
-	if err != nil {
-		return a.Alerter.LogFlowError(ctx, config.FlowJobName, fmt.Errorf("failed to resolve CDC store enabled: %w", err))
 	}
 
 	// syncDone will be closed by SyncFlow,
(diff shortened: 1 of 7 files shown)


flow/activities/flowable_core.go
```diff
@@ -200,6 +200,15 @@ func syncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDCSyncCon
 		return nil, err
 	}
 
+	dstPeerType, err := connectors.LoadPeerType(ctx, a.CatalogPool, config.DestinationName)
```
jgao54: this introduces an additional query per batch, which is a bit wasteful given destination type does not change during the course of a CDCFlowWorkFlow. but the query is relatively cheap.

Alternatively we can pass in DestinationType to FlowConnectionConfigCore, but it is a bit redundant and we also have to worry about backward compatibility handling for existing workflows.

flow/model/model.go
```diff
@@ -90,6 +90,8 @@ type PullRecordsRequest[T Items] struct {
 	InternalVersion uint32
 	// IdleTimeout is the timeout to wait for new records.
 	IdleTimeout time.Duration
+	// CDCStoreEnabled determines whether the CDC store for hydrating TOAST values should be enabled
+	CDCStoreEnabled bool
```
jgao54: not a huge fan of adding this to the `PullRecordsRequests` since it's specific to PG, but this is the cleaning way to pass it in that I came up with (other connectors just ignore this field). The alternative of passing it as an argument to `PullCdcRecords`, means threading it thru `pullCore` and `PullRecords` which would require changing `CDCPullConnector` interface, which gets uglier

flow/activities/flowable_core.go
ilidemi: Query per batch seems like a high cost to me for what it does (we're already hitting pool exhaustion on overloaded instances and https://github.com/PeerDB-io/peerdb/pull/4256 is helping us counteract that). Maybe a flow env could be a place that can be set once then queried from pg connector creation and remembered as a connector field?

Overall my opinion is still that overriding the existing setting statically depending on the deployment is the lowest-code solution.

flow/internal/dynamicconf.go
```diff
@@ -68,15 +68,15 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		Description:      "Controls whether to enable the store for recovering unchanged Postgres TOAST values within a CDC batch",
 		DefaultValue:     "true",
 		ValueType:        protos.DynconfValueType_BOOL,
-		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_IMMEDIATE,
+		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_AFTER_RESUME,
```
jgao54, line 71: since we are fetching this value on the activity level, no need for apply immediate as it would not take effect anyways

ilidemi: This one would be fetched on the next batch? Although I'm all for moving it to activity-level

flow/activities/flowable.go
```diff
@@ -375,6 +371,16 @@ func (a *FlowableActivity) SyncFlow(
 		return a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
 	}
 
+	destinationType, err := connectors.LoadPeerType(ctx, a.CatalogPool, config.DestinationName)
+	if err != nil {
+		return a.Alerter.LogFlowError(ctx, config.FlowJobName, fmt.Errorf("failed to load destination peer type: %w", err))
+	}
+
+	cdcStoreEnabled, err := internal.PeerDBCDCStoreEnabledForDestination(ctx, config.Env, destinationType)
```
ilidemi: Maybe I'm missing something, but if the CDC store is PG-only concern, why not read the config on the connector construction and store as a field on the connector?

ilidemi: (both configs)

flow/internal/dynamicconf.go
jgao54: IMMEDIATE skips setting it in the env map so it would always fetch from db, but with AFTER_RESUME it will just return what's in the env (after it gets fetched/set the first time).

that env gets "reset" on a new activity, which can be achieved with pause/resume.

i guess another way to handle this is just directly populate it into config.Env ourselves, also a bit hacky

flow/activities/flowable.go
jgao54: hmmm, taking a look..

jgao54: updated.

I also adjusted the code to determine the destination type and update the config.Env in the SyncFlow itself, such that the default value is destination-specific, but explicitly setting the env can still be honored. This is also given PG-to-PG and PG-to-CH will be different services so a single env is cleaner than having a separate one for PG-to-CH.

ilidemi: Was thinking of something like this https://github.com/PeerDB-io/peerdb/compare/cdc-store-maybe-cleaner?expand=1

jgao54: okay, let's go with that, it's got a bit less magic which seems like a good thing.

ilidemi: https://github.com/PeerDB-io/peerdb/pull/4438

PR #4370, opened 2026-06-01
feat(api): add endpoint for resetting sequences for PG PG mirrors

Amogh-Bharadwaj:
For Postgres to Postgres migrations, PeerDB ingests to target tables in a way where sequences do not get incremented. If a future insert uses DEFAULT, it can cause violations.

This PR adds an API endpoint to reset sequences in a target Postgres peer for a mirror's destination table set. This endpoint is gated to PG type system PG to PG mirrors.

A couple of tests also added.

The query is quite cheap:
```
 Index Scan using pg_attribute_relid_attnum_index on pg_attribute a  (cost=0.28..12.18 rows=3 width=96) (actual t
ime=0.046..0.046 rows=0.00 loops=1)
   Index Cond: ((attrelid = '21074'::oid) AND (attnum > 0))
   Filter: ((NOT attisdropped) AND (pg_get_serial_sequence('public.t1'::text, (attname)::text) IS NOT NULL))
   Rows Removed by Filter: 1
   Index Searches: 1
   Buffers: shared hit=6
 Planning Time: 0.071 ms
 Execution Time: 0.064 ms
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -45,6 +45,7 @@ flow/connectors/postgres/pgdump_schema_test.go          @PeerDB-io/ch-postgres
 flow/connectors/postgres/sink_pg.go                     @PeerDB-io/ch-postgres
 flow/model/pg_items.go                                  @PeerDB-io/ch-postgres
 flow/pkg/postgres/dest_validation.go                    @PeerDB-io/ch-postgres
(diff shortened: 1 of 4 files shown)


ilidemi (approved):
Can do codeowners for the new file too

PR #4371, opened 2026-06-02
add 'log_replica_updates' validation for mysql replica sources

dtunikov:
This PR resolves DBI-214.
It adds additional validation for mysql replica sources (either `SHOW REPLICA STATUS` or `SHOW SLAVE STATUS` returns not empty result).
New validation checks whether replica writes cdc records to binlog or not. If not - we return an error.
The plan is to merge this PR and use the same validation helper in the caller service.

### DIFF
--- a/flow/connectors/mysql/validate.go
+++ b/flow/connectors/mysql/validate.go
@@ -111,7 +111,7 @@ func (c *MySqlConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.F
 
 	conn, err := c.connect(ctx)
 	if err != nil {
-		return fmt.Errorf("unable to set connect: %w", err)
+		return fmt.Errorf("unable to connect: %w", err)
 	}
 
 	if isVitess, err := mysql_validation.IsVitess(conn); err != nil {
--- a/flow/connectors/mysql/validate.go
+++ b/flow/connectors/mysql/validate.go
@@ -53,38 +53,36 @@ func (c *MySqlConnector) CheckReplicationConnectivity(ctx context.Context) error
 }
 
 func (c *MySqlConnector) CheckBinlogSettings(ctx context.Context, requireRowMetadata bool) error {
-	for conn, err := range c.withRetries(ctx) {
-		if err != nil {
-			return err
-		}
-
-		switch c.config.Flavor {
-		case protos.MySqlFlavor_MYSQL_MARIA:
-			// check if binlog_row_metadata is supported is done inside this function
-			return mysql_validation.CheckMariaDBBinlogSettings(conn, c.logger, requireRowMetadata)
-		case protos.MySqlFlavor_MYSQL_MYSQL:
-			cmp, err := c.CompareServerVersion(ctx, mysql_validation.MySQLMinVersionForBinlogRowMetadata)
-			if err != nil {
-				return fmt.Errorf("failed to get server version: %w", err)
-			}
-			if cmp < 0 {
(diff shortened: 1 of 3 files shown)


flow/connectors/mysql/validate.go
```diff
@@ -109,16 +109,21 @@ func (c *MySqlConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.F
 		return fmt.Errorf("unable to establish replication connectivity: %w", err)
 	}
 
-	for conn, err := range c.withRetries(ctx) {
-		if err != nil {
-			return err
-		}
+	conn, err := c.connect(ctx)
```
dtunikov, line 110: c.withRetries was retrying success path 3 times
I think we don't need to call it here, do we?

flow/connectors/mysql/validate.go
```diff
@@ -109,16 +109,21 @@ func (c *MySqlConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.F
...
+	if isVitess, err := mysql_validation.IsVitess(conn); err != nil {
+		return err
+	} else if isVitess && !(cfg.DoInitialSnapshot && cfg.InitialSnapshotOnly) {
+		return fmt.Errorf("vitess is currently not supported for MySQL mirrors in CDC")
+	}
+	if err := mysql_validation.CheckRDSBinlogSettings(conn, c.logger); err != nil {
+		return fmt.Errorf("binlog configuration error: %w", err)
+	}
+	if err := mysql_validation.CheckLogReplicaUpdates(conn); err != nil {
+		return fmt.Errorf("binlog configuration error: %w", err)
```
ilidemi: Needs to be updated

flow/connectors/mysql/validate.go
ilidemi: Yup looks like a misuse of withRetries (which also looks like an easy to misuse API to me). Makes sense to just connect()

ilidemi (approved):
Please make sure to test, with RDS instances or local containers

flow/connectors/mysql/validate.go
ilidemi: Can do the same for CheckBinlogSettings as well, to reuse the connection

dtunikov: @ilidemi removed withRetries wrapper for CheckBinlogSettings and GetVersion methods as well
we still use it inside Execute and ExecuteSelectStreaming methods, but there the code seems correct. 
Even though I would replace this withRetries helper with smth less confusing in future. The API is quite unintuitive for the caller (+ c.connect itself acts as a singleton and the caller of withRetries mutates the connection now).

dtunikov:
> Please make sure to test, with RDS instances or local containers

with our current e2e tests setup it's not that easy to test this change
we would need to run a separate mysql container with log_replica_updates=ON
if we were using smth like testcontainers, then we could add a unit test that runs mysql container with required options
but with the current approach it basically means re-running the whole e2e suite with a different mysql configuration - https://github.com/PeerDB-io/peerdb/blob/main/.github/workflows/flow.yml#L196

or you meant checking that `ValidateMirrorSource` doesn't break on RDS instances after applying this change?

flow/connectors/mysql/validate.go
ilidemi: I'd undo GetVersion just in case as it's not called from validation. Even though the current caller in the worker does that call first so the connection probably isn't bad, I wouldn't want to have to think too much about random connection errors in possibly long running methods

PR #4373, opened 2026-06-02
Deflake pg_dump tests

ilidemi:
Finds from #4634.
* Processes might exit by themselves while/right before they're being killed and that should count as success
* A small race between short-lived test commands

Codex says there's still a race where destination can read all output from the source and exit gracefully, and we try to kill source while it could just need another second to also finish.


Amogh-Bharadwaj (approved):

Amogh-Bharadwaj:
Thanks 🙏

PR #4378, opened 2026-06-02
fix(clickhouse): exact microsecond->Time64 conversion on snapshot path

ilidemi:
## Problem

On the snapshot/QRep path, TIME/TIMETZ columns are converted from Avro time-micros (int64 microseconds) to ClickHouse `Time64(6)` via:

```
toTime64(<micros> / 1000000.0, 6)
```

The integer microsecond count is divided by a **Float64** literal. Float64 is binary, so many decimal microsecond fractions (e.g. `0.000249`) have no exact representation and land just *below* the true value; `toTime64` then **truncates**, dropping 1µs. ~11,549 of 1,000,000 sub-second values are affected — value-dependent, which is why `now()`-based tests flake intermittently.

Affects Postgres `TIME`/`TIMETZ` and MySQL `TIME` on the initial-snapshot/QRep load, only with `Flag_ClickHouseTime64Enabled`. The CDC path parses the value from a string (`toTime64OrNull`) and is lossless.

Failure repro (new test without a fix): https://github.com/PeerDB-io/peerdb/actions/runs/26849534704/job/79177892814?pr=4378

## Fix

`toTime64(toDecimal64(<micros>, 6) / 1000000, 6)` — exact base-10 division

Also fix a test helper that was handling nulls inconsistently.

### DIFF
--- a/flow/connectors/clickhouse/table_function.go
+++ b/flow/connectors/clickhouse/table_function.go
@@ -76,8 +76,8 @@ func timeFieldExpressionConverter(
 	}
 
 	// QValueTime is stored as time-micro logical type in Avro, toTime64 accepts a
-	// fractional second, so conversion is necessary
-	return fmt.Sprintf("toTime64(%s / 1000000.0, 6)", sourceFieldIdentifier), nil
+	// fractional second, so conversion is necessary. Trip through Decimal64 to preserve precision.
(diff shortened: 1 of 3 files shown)


jgao54 (approved):

jgao54:
good find!

PR #4380, opened 2026-06-03
Ensure ssh honors context deadlines

jgao54:
Introduce a `DeadlineCapableConn` which creates connection that always respect context deadline. Used for ssh connection where deadline is not supported and an error is returned when attempting to set deadline. 

The previous implementation with NoDeadlineConn swallows the upstream error but does not honor context deadline, which Postgres depends on when reading messages, so `conn.ReceiveMessage` could block indefinitely when network hiccups upstream. 

The fix here is introducing a deadline-capable in-memory transport between the connector and SSH using net.Pipe(), which respect deadlines. When deadline fires, any chunk the io.Copy goroutine already read from SSH is served on next read, avoiding data loss.

This PR also:
- update PG/MySQL/Mongo to use DeadlineCapableConn for ssh tunneling:
- ~remove the `closeSyncerWithTimeout` logic now that deadline is respected on close~ will do this as a follow-up

Testing:
- add unit & e2e tests. test failing before this change https://github.com/PeerDB-io/peerdb/pull/4380/commits/a9615f91677020f65b16e614171bb4be04bc6ed7

Fixes: DBI-629


flow/connectors/utils/metered_dialer.go
```diff
@@ -22,20 +22,16 @@ func (mc *MeteredConn) Read(b []byte) (int, error) {
...
-	deltaBytesRead     *atomic.Int64
-	innerDialer        innerDialer
-	noDeadlineRequired bool
+	totalBytesRead *atomic.Int64
+	deltaBytesRead *atomic.Int64
+	innerDialer    innerDialer
 }
 
-func NewMeteredDialer(totalBytesRead *atomic.Int64, deltaBytesRead *atomic.Int64,
-	innerDialer innerDialer, noDeadlineRequired bool,
```
jgao54, line 32: removing the encapsulation of SSH wrapped conn handling here; instead make it more explicit / consistent by passing the dialContext with deadline support

flow/connectors/utils/deadline_capable_conn.go
```diff
@@ -0,0 +1,46 @@
...
+func NewDeadlineCapableConn(sshConn net.Conn) net.Conn {
+	localConn, proxyConn := net.Pipe()
+
+	var closeOnce sync.Once
+	closeAll := func() {
+		closeOnce.Do(func() {
+			// we do not explicitly close localConn here
+			// as that is the responsibility of the caller
+			_ = proxyConn.Close()
+			_ = sshConn.Close()
```
jgao54: make it more explicit that if connection from one direction dies, connection for both direction should be closed.

flow/connectors/postgres/ssh_wrapped_conn_test.go
```diff
@@ -0,0 +1,75 @@
...
+	t.Helper()
+
+	sshPort, err := strconv.ParseUint(utils.SSHServerPort, 10, 32)
+	require.NoError(t, err)
+
+	pgConfig := internal.GetAncillaryPostgresConfigFromEnv()
+	// The database connection is opened from inside the SSH server container; In CI,
+	// PG_HOST=localhost only works for runner-to-PG connections; but from inside the
+	// SSH server container, host must be resolvable by the SSH container.
+	pgConfig.Host = internal.PostgresToxiproxyUpstreamHostWithFallback(pgConfig.Host)
```
ilidemi: Should the helper and the env var be renamed as it's more of an SSH upstream host?

flow/connectors/postgres/ssh_wrapped_conn.go
```diff
@@ -21,13 +20,7 @@ func NewPostgresConnFromConfig(
 	tunnel *utils.SSHTunnel,
 ) (*pgx.Conn, error) {
 	if tunnel != nil && tunnel.Client != nil {
-		connConfig.DialFunc = func(ctx context.Context, network, addr string) (net.Conn, error) {
-			conn, err := tunnel.Client.DialContext(ctx, network, addr)
-			if err != nil {
-				return nil, err
-			}
-			return &utils.NoDeadlineConn{Conn: conn}, nil
```
ilidemi, line 29: Is NoDeadlineConn used anymore? Can be cleaned up if not

flow/connectors/utils/deadline_capable_conn.go
```diff
@@ -0,0 +1,46 @@
...
+		})
+	}
+
+	go func() {
+		_, _ = io.Copy(proxyConn, sshConn)
+		closeAll()
+	}()
+	go func() {
+		_, _ = io.Copy(sshConn, proxyConn)
```
ilidemi: Claude is flagging that if read from SSH hangs, Close won't affect it and there will be a goroutine leak (but the caller DeadlineCapableConn will be ok). I'm too braindead today to offer suggestions so just flagging

flow/connectors/utils/deadline_capable_conn.go
```diff
@@ -0,0 +1,46 @@
...
+// https://github.com/jackc/pgx/issues/382#issuecomment-1496586216
+// which silently ignores deadline instead of erroring. However, for connectors
+// that depend on context deadline (e.g. Postgres), this means reading from
+// ssh channel blocks indefinitely when no message arrives.
+//
+// The fix here is introducing a deadline-capable in-memory transport between
+// the connector and SSH using net.Pipe(). When deadline fires, any chunk the
+// io.Copy goroutine already read from SSH is served on next read, avoiding
+// data loss.
+func NewDeadlineCapableConn(sshConn net.Conn) net.Conn {
```
ilidemi, line 31: There is a possibility of unit testing by passing a conn that misbehaves in every possible way (like read/write/close hanging) and checking the wrapped conn behavior along with goroutine leaks

flow/connectors/utils/deadline_capable_conn.go
ilidemi: Another flag is that EOF errors will become ErrClosedPipe and drivers may or may not handle those differently, plus classification is not there

ilidemi (approved):
This will definitely help with receive hangs, but with our current level of understanding of drivers+go net behavior I have a feeling that this will come back. Not blocking the release as I'm out tomorrow but would suggest to have bits without this change ready as a backup as well.

flow/connectors/postgres/ssh_wrapped_conn_test.go
jgao54: Yeah good point. Can rename.

flow/connectors/utils/deadline_capable_conn.go
jgao54: great callout. switched to wrapper (which is what we actually had before with NoDeadlineConn) that updated the SSHCapableConn's Close() to also close proxyConn/sshConn.

flow/connectors/utils/deadline_capable_conn.go
jgao54: added more unit tests

flow/connectors/utils/deadline_capable_conn.go
jgao54: Added unit tests, we should be okay as it only happens if Close is triggered from localConn

jgao54: one exception would be if we try to write to a closed Conn, i think in this case if it happens, we probably want to get alerted anyways as it indicates a bug on our side. but it's a good callout!

PR #4385, opened 2026-06-04
fix(ci): isolate dev docker PR builds from secrets

Jeremyyang920:
Remove pull_request_target from the dev Docker workflow so fork PR code no longer runs in a privileged base-repository context. Split PR validation from push publishing: PRs build cache-only with read-only permissions and no GHCR or Depot credentials, while main pushes retain package publishing.

### DIFF
--- a/.github/workflows/dev-docker.yml
+++ b/.github/workflows/dev-docker.yml
@@ -7,8 +7,10 @@ on:
     branches: [main]
 
 jobs:
-  docker-build:
-    if: github.event_name == 'pull_request'
+  docker-build-external:
+    if: |
+      github.event_name == 'pull_request' &&
+      github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       matrix:
         # ubuntu-latest leverages larger GH runner pool & completes in ~30s instead of ~3m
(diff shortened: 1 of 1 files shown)


.github/workflows/dev-docker.yml
```diff
@@ -5,16 +5,40 @@ on:
...
+
+      - name: Set Short Commit Hash
+        id: vars
+        run: echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+
+      - name: Build PeerDB Images
+        run: |
+          docker buildx bake \
+            --file ./docker-bake.hcl \
+            --set '*.platform=linux/amd64' \
```
ilidemi, line 36: We're generally building for both x64 and ARM, if ARM build breaks it would be nice to still catch it for internal PRs. How about https://github.com/PeerDB-io/peerdb/compare/jyang/ci-patch...ci-patch-alt?expand=1

ilidemi (approved):

.github/workflows/dev-docker.yml
Jeremyyang920: That looks good to me. Let me cherry pick it onto this branch.

PR #4386, opened 2026-06-05
fix: emit Slack alert members as <@USERID> mentions

### LINKED ISSUE #4353: Slack alerts: per-user mentions in "members" field never trigger notifications
## Bug Description

Slack alerts configured with one or more entries in the `members` field
never produce real Slack mentions. The names are rendered as plain text
(e.g. `@<username>`) rather than as proper Slack user mentions, so no
notification is delivered to the intended recipients.

Image

The `<!channel>` fallback (when `members` is empty) does work correctly.

## Steps to Reproduce

1. Configure a Slack alert via the UI (`/alert-config`).
2. Set the `members` field to any value — a Slack display name, a username,
   or even a Slack user ID such as `U01ABC234`.
3. Trigger any alert (e.g. set `slot_lag_mb_alert_threshold` very low).
4. The Slack message contains literal text `cc: @<value>` with no
   highlighting and no push notification — the user is not actually
   mentioned.

## Root Cause

`flow/alerting/slack_alert_sender.go:53-56`:

```go
ccMembersPart.WriteString(" @")
ccMembersPart.WriteString(member)
```

This emits the literal characters `@` + the member string. Slack mrkdwn
only renders a real user mention when the syntax is `<@USERID>` (see
https://api.slack.com/reference/surfaces/formatting#mentioning-users).
The empty-members branch (line 50) already uses the correct syntax
(`<!channel>`).

The UI hint at `ui/app/alert-config/new.tsx` describing this field as
"Slack usernames" is also misleading, since even a correct username would
not produce a mention with the current emit format.

## Suggested Fix

Emit `<@member>` instead of `@member`, and document in the UI that the
field expects Slack user IDs (e.g. `U01ABC234`).

## Environment

- Reproduced on current `main` (around commit `f64f92f3`).

kang8:
## Problem

Slack alert cc lines were built with `@<name>` syntax, which Slack renders as **plain text** and never turns into an actual mention — so the tagged users were never notified.

## Fix

- Wrap each member in Slack's mrkdwn mention syntax `<@USERID>` so the cc line produces real, clickable mentions.
- Keep the `<!channel>` fallback when no members are configured.
- Update the alert-config UI hint to clarify the field expects Slack **user IDs** (e.g. `U01ABC234`), not display names.
- Extract the cc-line construction into `formatCCMembers`, covered by a table-driven unit test.

### Before:
image

### After:
image

## Testing

- `go test ./alerting/ -run TestFormatCCMembers` — passes (empty / single / multiple members, `<!channel>` fallback).
- End-to-end verified against a live Slack workspace using the local dev stack: triggered an alert and confirmed the cc line renders as a real highlighted `@mention` (previously plain text).

Close https://github.com/PeerDB-io/peerdb/issues/4353

### DIFF
--- a/flow/alerting/slack_alert_sender_test.go
+++ b/flow/alerting/slack_alert_sender_test.go
@@ -1,0 +1,44 @@
+package alerting
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFormatCCMembers(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		members []string
+		want    string
+	}{
+		{
+			name:    "empty members pings whole channel",
+			members: nil,
+			want:    "cc: <!channel>",
+		},
+		{
+			name:    "empty slice pings whole channel",
+			members: []string{},
+			want:    "cc: <!channel>",
+		},
+		{
+			name:    "single member wrapped in Slack mention syntax",
+			members: []string{"U01ABC234"},
+			want:    "cc: <@U01ABC234>",
+		},
+		{
+			name:    "multiple members each wrapped and space-separated",
+			members: []string{"U01ABC234", "U05XYZ999"},
+			want:    "cc: <@U01ABC234> <@U05XYZ999>",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tt.want, formatCCMembers(tt.members))
+		})
+	}
+}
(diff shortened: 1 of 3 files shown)


ilidemi (approved):
Thanks!

flow/alerting/slack_alert_sender_test.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFormatCCMembers(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
```
ilidemi, line 13:
```suggestion
	//nolint:govet // it's a test, no need for fieldalignment
	tests := []struct {
```

PR #4387, opened 2026-06-05
fix: redact Slack auth_token in alert config responses

### LINKED ISSUE #4345: Alert Configurations API returns Slack auth_token in plaintext to UI
## Bug Description

`GET /api/v1/alerts/config` returns the full decrypted `service_config` JSON
to any user with UI access, including the Slack `auth_token` (bot token).
The UI then renders it as visible JSON, and the same plaintext is also
delivered over the network in the API response body.

While the token is stored encrypted at rest (XChaCha20-Poly1305, via
`PEERDB_ENC_KEYS`), it is decrypted on read and handed back to the client
unredacted, defeating the at-rest protection for anyone who can reach the
UI / API.

Image
*Screenshot: the `/alert-config` page rendering `auth_token` ("xoxb-...") in plaintext (to be attached).*

## Steps to Reproduce

1. Configure a Slack alert via the UI (`/alert-config`) with any `auth_token`.
2. Open browser DevTools and inspect the response of `GET /api/v1/alerts/config`
   — the body contains `"auth_token":"xoxb-..."` in plaintext.
3. The same value is also visibly rendered in the table on the alert config
   page (`ConfigJSONView`), and re-populated into the edit form when the
   config is opened for editing.

## Root Cause

- Backend handler returns the decrypted JSON to the client as-is:
  `flow/cmd/alerts.go:30-34` (`GetAlertConfigs`).
- UI renders the full JSON including `auth_token`:
  `ui/app/alert-config/page.tsx:75-82, 142`.
- Edit flow prepopulates the token into the form by parsing the same JSON:
  `ui/app/alert-config/page.tsx:63-73` → `ui/app/alert-config/new.tsx`.

The same pattern applies to any future secret-bearing alert service type
(e.g. email SMTP credentials), since the dispatch is generic and only
handles encryption/decryption — not redaction.

## Suggested Fix

The API should not return secret fields (e.g. `auth_token`) in responses.

## Environment

- Reproduced on current `main` (around commit `f64f92f3`).

kang8:
## What

`GetAlertConfigs` returned the decrypted Slack `auth_token` in plaintext, exposing the credential to anyone with UI access.

This PR:
- Redacts secret fields (Slack `auth_token`) on read, so the API never returns the value.
- On update, preserves the stored secret when the client submits an empty value, so the UI can edit other fields without ever seeing the token.
- Renders `auth_token` as a password field with a "leave blank to keep current token" hint and uses a relaxed edit-time validation schema.
- Masks redacted secrets in the read-only config view (`••••••• (hidden)`) so a configured token no longer looks unset.
- Adds `asChild` to the `AlertDropdown` trigger so Radix no longer wraps the `Button` in its own `button`, removing a `<button>` in `<button>` hydration error in the alert config page.

## Example

Google Chrome 2026-06-05 18 27 15

## Why

The Slack auth token is a sensitive credential and should never be exposed to UI clients. See #4345.

Closes #4345

### DIFF
--- a/flow/cmd/alerts.go
+++ b/flow/cmd/alerts.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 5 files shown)


ilidemi (commented):
Thanks for the nudge and inspiration, opened #4391 with a slightly different approach

kang8:
Closing in favor of #4391, where the maintainer took over with a slightly different approach. Thanks!

PR #4388, opened 2026-06-05
DBI-782: PG Connector TLS settings unification

pfcoperez:
This Pull Request aligns PG connector TLS configuration parameters with other connectors that use `DisableTLS`.
To keep backward compatibility, this new config parameter is optional and is always overridden by the existing`RequireTLS`.

Additionally, it adds `SkipCertVerification`  (default `false` in accordance to the current hard-coded behaviour) thus allowing for encrypted connection to unverified hosts. 

Closes: DBI-782

### DIFF
--- a/flow/connectors/postgres/postgres.go
+++ b/flow/connectors/postgres/postgres.go
@@ -146,7 +146,10 @@ func ParseConfig(connectionString string, pgConfig *protos.PostgresConfig) (*pgx
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse connection string: %w", err)
 	}
-	if pgConfig.RequireTls || pgConfig.RootCa != nil {
+
+	shouldUseTls := internal.PGMustUseTlsConnection(pgConfig)
+
+	if shouldUseTls || pgConfig.RootCa != nil {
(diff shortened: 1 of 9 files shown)


flow/connectors/postgres/postgres.go
```diff
@@ -146,8 +146,11 @@ func ParseConfig(connectionString string, pgConfig *protos.PostgresConfig) (*pgx
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse connection string: %w", err)
 	}
-	if pgConfig.RequireTls || pgConfig.RootCa != nil {
-		tlsConfig, err := common.CreateTlsConfig(tls.VersionTLS12, pgConfig.RootCa, connConfig.Host, pgConfig.TlsHost, false)
+
+	shouldUseTls := internal.PGMustUseTlsConnection(pgConfig)
+
+	if shouldUseTls || pgConfig.RootCa != nil {
```
ilidemi, line 152: nit
```suggestion
	mustUseTls := internal.PGMustUseTlsConnection(pgConfig)

	if mustUseTls || pgConfig.RootCa != nil {
```

ilidemi (approved):

PR #4393, opened 2026-06-06
use exact count when estimates fail

jgao54:
Encountered a scenario where the query count estimate led to `ERROR: bigint out of range (SQLSTATE 22003)`; and the while the exact row count for the table is very low.

The query is:

```
SELECT CASE
WHEN c.relhassubclass THEN 0
WHEN c.reltuples >= 0 AND c.relpages > 0 THEN
	(c.reltuples / c.relpages * (pg_relation_size(c.oid) / current_setting('block_size')::integer))::bigint
ELSE c.reltuples::bigint
END
FROM pg_class c WHERE c.oid = to_regclass($1)
```

When stats are not populated due to a relatively new table, reltuples could be 0 or -1, and in both cases the result would not overflow.  Given the actual row count is small, the most likely scenario seem to be table bloat with dead rows that led to both reltuples and pg_relation_size to be very high. 

When we hit this error, we can conservatively fall back to actual table count. 

Fixes: DBI-766

### DIFF
--- a/flow/connectors/postgres/qrep_partition.go
+++ b/flow/connectors/postgres/qrep_partition.go
@@ -486,12 +486,12 @@ func ComputeNumPartitions(ctx context.Context, pp PartitionParams, numRowsPerPar
 		); err != nil {
 			return 0, fmt.Errorf("failed to query for estimated row count: %w", err)
 		}
+
 		pp.logger.Info("estimated row count",
 			slog.String("query", estimatedCountTemplate),
 			slog.String("watermarkTable", pp.watermarkTable),
 			slog.Bool("relhassubclass", relhassubclass),
-			slog.String("formula", fmt.Sprintf("%v / %d * (%d / %d)",
-				reltuples, relpages, relationSize, blockSize)))
+			slog.String("formula", fmt.Sprintf("%v / %d * (%d / %d)", reltuples, relpages, relationSize, blockSize)))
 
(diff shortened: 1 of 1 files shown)


masterashu (approved):

ilidemi (approved):

flow/connectors/postgres/qrep_partition.go
```diff
@@ -456,48 +456,55 @@ func ComputeNumPartitions(ctx context.Context, pp PartitionParams, numRowsPerPar
...
 	var queryArgs []any
 
 	if pp.lastRangeEnd == nil {
 		pp.logger.Info("fetch estimated row count", slog.String("query", estimatedCountTemplate))
-		if err := pp.tx.QueryRow(ctx, estimatedCountTemplate, pp.watermarkTable).Scan(&totalRows); err != nil {
+		var n pgtype.Numeric
+		if err := pp.tx.QueryRow(ctx, estimatedCountTemplate, pp.watermarkTable).Scan(&n); err != nil {
 			return 0, fmt.Errorf("failed to query for estimated row count: %w", err)
+		} else if v, err := n.Int64Value(); err != nil {
+			pp.logger.Warn("estimated row count outside int64 range, falling back to precise count",
```
ilidemi, line 497: Just a thought, could also query the inputs into the formula and log them for debugging purposes

PR #4403, opened 2026-06-09
add skip_validation flag to create cdc flow request

dtunikov:
this will allow the caller service to skip validation when it's unnecesary

### DIFF
--- a/protos/flow.proto
+++ b/protos/flow.proto
@@ -96,7 +96,7 @@ message FlowConnectionConfigs {
   // Used to control type mapping backwards compatibility.
   repeated string flags = 27;
 
-  bool skip_validation = 28;
+  optional bool skip_validation = 28;
 }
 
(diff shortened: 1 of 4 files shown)


ilidemi (approved):
Just need to pass false in UI and Rust code

protos/flow.proto
```diff
@@ -95,6 +95,8 @@ message FlowConnectionConfigs {
   // Connector capabilities detected at flow creation time.
   // Used to control type mapping backwards compatibility.
   repeated string flags = 27;
+
+  bool skip_validation = 28;
```
ilidemi: Actually, forgot `optional` is a thing. It would be more appropriate here as OSS users shouldn't need to bother with choosing true or false

dtunikov: made the field optional
i guess in such case we don't need to pass 'false' explicitly from ui/rust

dtunikov: nvm, rust forced me to pass it anyways

PR #4405, opened 2026-06-10
restore async replication connection keepalive when PullRecords is not active

jgao54:
Prior to https://github.com/PeerDB-io/peerdb/pull/4324, we had an async `maintainReplConn` goroutine that was running on a ticker every 15 seconds.  When we moved the keepalive inside of PullRecords,  we introduced an edge case where there can be _a large time gap_ between PullRecords where we are no longer keeping the connection alive(e.g. when normalize backpressure kicks in).  This led to a new failure mode when PullRecords resumes and tries to read from PG, it hit wal_sender_timeout error. 

This PR  brings back the async goroutine ~but only run the keepalive when PullRecords is not active; otherwise, it let the PullRecords handle the keepalive logic in the loop.~

The keepalive is hard-coded to 15 seconds. When we merge https://github.com/PeerDB-io/peerdb/pull/4363 later, we can use the computed keepalive instead. ~Also note it's still safe to remove `replLock` because this ReplPing only runs when PullRecords is inactive, so there should be no lock contention (but doesn't hurt if we wish to keep it either for safety).~ There can be edge cases (see discussion below) where a race condition can happen, so we will want to keep the replLock around when.  

Also note that https://github.com/PeerDB-io/peerdb/pull/4404  addresses slightly separate edge case, with the goal of ensuring ReplPing has full coverage _inside_ a PullRecord cycle.


flow/connectors/postgres/postgres_source.go
```diff
@@ -80,11 +80,40 @@ func (c *PostgresConnector) SetupReplConn(ctx context.Context, env map[string]st
...
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			if c.pullActive.Load() {
+				continue
+			}
```
jgao54: The alternative would be to fully revert to the previous change where we always run keepalive in the async loop and not in the PullRecords loop; but then we do have to worry about potentially delayed Ping when there is a replLock contention. So decided to go with this approach as it's more predictable.

ilidemi (approved):
@pfcoperez please review as well

flow/connectors/postgres/postgres_source.go
```diff
@@ -80,11 +80,40 @@ func (c *PostgresConnector) SetupReplConn(ctx context.Context, env map[string]st
 		return err
 	}
 	c.replConn = conn
+	go c.runReplConnKeepalive(ctx, 15*time.Second, c.ReplPing)
```
pfcoperez: nit: I think we should isolate the 15'' value as a constant.

flow/connectors/postgres/postgres_source.go
```diff
@@ -80,11 +80,40 @@ func (c *PostgresConnector) SetupReplConn(ctx context.Context, env map[string]st
...
+	keepalive func(context.Context) error,
+) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			if c.pullActive.Load() {
```
pfcoperez: (non blocking comment)
 
Thinking about the following interweave:

A -> One go-routine
B -> Another go-routine

1. A: Reaches this line [reads `false`] 
2. B: `c.pullActive.Store(true)` [withing pull-core]
3. A: Wrongly assumes pull is not active and does the ping but `replLock` protects the connection and the ping is skipped.

The reason to have pings as part of the loop is that, most of the time while pulling records, the lock is captured not letting the ticket ping.   

With this in mind,  `pullActive` mechanism might save some pings, is it worth of the extra elements in the state? 

If so, I'd add a comment stating that this aims for optimization and that real protection happens in the lock.

pfcoperez (approved):
My main, non-blocking, observation is the need to expand the state with `pullActive`

flow/connectors/postgres/postgres_source.go
pfcoperez: > Also note it's still safe to remove replLock because this ReplPing only runs when PullRecords is inactive, so there should be no lock contention (but doesn't hurt if we wish to keep it either for safety).

Is it? If my observation above is not wrong, I think it is not safe to remove the lock. If you agree I'd edit that in the PR description so future commit inspection doesn't lead to confusion :pray:  (also maybe adding some comments about this reasoning)

flow/connectors/postgres/postgres_source.go
jgao54: @pfcoperez  was thinking that once https://github.com/PeerDB-io/peerdb/pull/4363 is in we can use the computed  value(3 * walSenderTimeout / 4) instead

flow/connectors/postgres/postgres_source.go
jgao54: @pfcoperez  yeah that's a good point! we should keep the replLock around for the interweave scenario you described. 

> With this in mind, pullActive mechanism might save some pings, is it worth of the extra elements in the state?

Since we are keeping the lock for thread-safety, I agree we can keep it simple and just have the goroutine always ping() with TryLock.

jgao54: Note: this also has the added benefit of not having to worry about AddRecord blocking, closing https://github.com/PeerDB-io/peerdb/pull/4409

PR #4409, opened 2026-06-10
make sure blocking AddRecord also handles keepalive

jgao54:
PullRecords has 2 blocking sites that requires extra keepalive handling:
1. waiting for message to arrive
2. waiting for processed event to be send to the record stream

1 is already handled by https://github.com/PeerDB-io/peerdb/pull/4324, this PR handles 2.

This can happen if there is downstream backpressure while we are trying to add a record into the record stream, the channel itself will block until backpressure is cleared. 

This PR adds an additional keepalive handling when backpressure is detected. 

Note: long-term, i think we still want a single place to handle keepalive, and not having to worry about every blocking callsite inside PullRecord needing to manage it. What's preventing us from doing this today is indeterminism caused by lock contention on the replication connection, making it hard to reason about if keepalive has fired or not. However that fix requires a bigger refactor, so scoping PR to the current approach where PullRecords handle its own keepalives while running.

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -611,13 +611,36 @@ func PullCdcRecords[Items model.Items](
 	nextRecordDeadline := time.Now().Add(req.IdleTimeout)
 	pkmRequiresResponse := false
 
+	pushToStream := func(rec model.Record[Items]) error {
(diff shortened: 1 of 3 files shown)


jgao54:
No longer needed, see https://github.com/PeerDB-io/peerdb/pull/4405#discussion_r3397929778

PR #4411, opened 2026-06-11
Use unique staging paths for BigQuery source

ilidemi:
A very good write-up of the situation is in [DBI-790](https://linear.app/clickhouse/issue/DBI-790/).

To resolve the conflicts between the previous snapshot cleanup and the new snapshot sync, add a subfolder with the workflow's run id. Use the same temporal session mechanism that's used to pass PG snapshot name to the QRep flows, it's great that we have it. It won't help us on the future snapshot+CDC path but we'll get there when we get there, the whole "return string as an `any` transaction" situation is already a mess.

Validated locally and as a hotfix on the customer service that hit this.

Closes DBI-790

### DIFF
--- a/flow/activities/snapshot_activity.go
+++ b/flow/activities/snapshot_activity.go
@@ -25,7 +25,8 @@ type SlotSnapshotState struct {
 }
 
 type TxSnapshotState struct {
-	SnapshotName string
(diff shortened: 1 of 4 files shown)


jgao54 (approved):
nice!

PR #4413, opened 2026-06-11
mysql: add syncer.EventCacheCount setting

dtunikov:
We saw OOM errors on mysql pipes with large rows
go-mysql has internal events channel where it stores binlog events, that is where this allocation is coming from
It reads packet header that contains info about payload length in bytes and allocates memory for reading next events from the binlog
I think it's expected driver behavior and i'm not sure we should fix smth in that regard
But
There is one thing we could configure additionally to our PEERDB_CDC_CHANNEL_BUFFER_SIZE, we could set EventCacheCount on mysql syncer level.
This way we can cap how many events mysql driver loads in memory. (we can't control events' size though, so if the events themselves are big -> it won't really save us from OOM), by default it equals 10240

mysql go driver simply read packets over network and tries to push events to it's internal channel (which is capped by EventCacheCount):
```
loop:
  data := b.c.ReadPacket()        // blocking socket read
  event := parse(data)
  s.AddEventToStreamer(event)     // blocking channel send → s.ch (cap = EventCacheCount)
```

### DIFF
--- a/flow/connectors/mysql/ssh_keepalive_test.go
+++ b/flow/connectors/mysql/ssh_keepalive_test.go
@@ -316,7 +316,7 @@ func TestMySQLCloseSyncerWithTimeout(t *testing.T) {
 
 	pos, err := connector.GetMasterPos(ctx)
 	require.NoError(t, err)
-	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos, map[string]string{}) //nolint:dogsled
+	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos, nil) //nolint:dogsled
 	require.NoError(t, err)
 
 	// Let CDC streaming establish before blocking the MySQL server.
--- a/flow/connectors/mysql/ssh_keepalive_test.go
+++ b/flow/connectors/mysql/ssh_keepalive_test.go
@@ -316,7 +316,7 @@ func TestMySQLCloseSyncerWithTimeout(t *testing.T) {
 
 	pos, err := connector.GetMasterPos(ctx)
 	require.NoError(t, err)
-	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos) //nolint:dogsled
+	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos, map[string]string{}) //nolint:dogsled
 	require.NoError(t, err)
 
 	// Let CDC streaming establish before blocking the MySQL server.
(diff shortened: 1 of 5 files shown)


jgao54 (approved):

flow/connectors/mysql/ssh_keepalive_test.go
```diff
@@ -316,7 +316,7 @@ func TestMySQLCloseSyncerWithTimeout(t *testing.T) {
 
 	pos, err := connector.GetMasterPos(ctx)
 	require.NoError(t, err)
-	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos) //nolint:dogsled
+	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos, map[string]string{}) //nolint:dogsled
```
ilidemi:
```suggestion
	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos, nil) //nolint:dogsled
```

flow/connectors/mysql/ssh_test.go
```diff
@@ -53,7 +53,7 @@ func TestMySQLSyncerClose(t *testing.T) {
 
 	pos, err := connector.GetMasterPos(ctx)
 	require.NoError(t, err)
-	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos) //nolint:dogsled
+	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos, map[string]string{}) //nolint:dogsled
```
ilidemi:
```suggestion
	syncer, _, _, _, err := connector.startCdcStreamingFilePos(ctx, pos, nil) //nolint:dogsled
```

flow/internal/dynamicconf.go
```diff
@@ -460,6 +460,14 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_AFTER_RESUME,
 		TargetForSetting: protos.DynconfTarget_POSTGRES,
 	},
+	{
+		Name:             "PEERDB_MYSQL_EVENT_CACHE_COUNT",
+		Description:      "Maximum number of events the MySQL Go driver loads in memory at once, useful when pushing large rows",
+		DefaultValue:     "10240",
+		ValueType:        protos.DynconfValueType_INT,
+		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_IMMEDIATE,
+		TargetForSetting: protos.DynconfTarget_MYSQL,
```
ilidemi: MySQL is not a target in PeerDB, let's use `DynconfTarget_ALL`

ilidemi (approved):

PR #4414, opened 2026-06-11
Edit Mirror: don't strip shared destination on table removal

serprex:
Multiple source tables mapping to one destination works by keying on source. In raw table we only have destination table. Removing one such source previously cleaned based on destination, breaking surviving source

cdc_flow: compute exclusivelyRemovedTables, removed mappings whose destination is not fed by any surviving mapping, & scope cleanup to those. Publication alteration keeps full RemovedTables set: keyed on source, so every removed source must leave regardless of shared destination

RemoveTablesFromRawTable: sort+compact destination names & skip when none remain or syncBatchID <= normBatchID. The DELETE predicate is _peerdb_batch_id in (normBatchID, syncBatchID]; an empty range makes it a no-op, so skipping avoids connecting to destination for nothing

clickhouse: delete in 100 table batches to group mutations

Adds e2e Test_Removal_Shared_Destination covering surviving-source replication into a shared destination after removal. This also adds coverage for N:1 mirrors

### DIFF
--- a/flow/e2e/clickhouse_test.go
+++ b/flow/e2e/clickhouse_test.go
@@ -207,6 +207,98 @@ func (s ClickHouseSuite) Test_Addition_Removal() {
 	RequireEnvCanceled(s.t, env)
 }
 
+// Removing one of several source tables feeding a shared destination must keep
+// that destination's raw rows & catalog schema mapping intact, otherwise the
+// surviving source can no longer normalize into it. Disjoint id ranges keep the
+// two sources' rows distinct under the destination's ReplacingMergeTree.
+func (s ClickHouseSuite) Test_Removal_Shared_Destination() {
+	tc := NewTemporalClient(s.t)
+
+	srcTableA := s.attachSchemaSuffix("test_shared_dst_a")
+	srcTableB := s.attachSchemaSuffix("test_shared_dst_b")
+	dstTableName := "test_shared_dst_target"
+
+	for _, srcTableName := range []string{srcTableA, srcTableB} {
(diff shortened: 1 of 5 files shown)


flow/e2e/clickhouse_test.go
```diff
@@ -207,6 +207,98 @@ func (s ClickHouseSuite) Test_Addition_Removal() {
 	RequireEnvCanceled(s.t, env)
 }
 
+// Removing one of several source tables feeding a shared destination must keep
+// that destination's raw rows & catalog schema mapping intact, otherwise the
+// surviving source can no longer normalize into it. Disjoint id ranges keep the
+// two sources' rows distinct under the destination's ReplacingMergeTree.
+func (s ClickHouseSuite) Test_Removal_Shared_Destination() {
```
ilidemi, line 214: TIL we unconditionally allow sharing a destination table within a mirror and the [UI check](https://github.com/PeerDB-io/peerdb/blob/0709374ba06182e9c9df6806bf402215713afdc8/ui/app/mirrors/create/schema.ts#L32-L33) around this is broken. Can this use `PEERDB_SOURCE_SCHEMA_AS_DESTINATION_COLUMN` please

ilidemi (approved):

flow/e2e/clickhouse_test.go
ilidemi: Or is the use case getting fixed actually relying on this functionality without the schema thing?

serprex: Yes, usecase is partitioned table in mysql (read: add/remove date suffix table everyday)

PR #4422, opened 2026-06-15
fix(clickhouse): fix avro substream send

Amogh-Bharadwaj:
Replaces substream.Records <- record with substream.Send which is context-aware, so that when the consumer is down, we do not leak goroutines here in the avro writer.
Otherwise this has seen to cause initial load failures due to resource exhaustion.

Also adds more logging in this area


serprex (approved):

ilidemi (approved):
Thanks for the bugfix!

Issue #4412, opened 2026-06-11
Creation of raw tables is broken on older CH clusters


eskarimov:
Hi team,

facing an issue after https://github.com/PeerDB-io/peerdb/pull/4312, creation of `_peerdb_raw_*` tables fails with `TTL expression result column should have DateTime or Date type, but has DateTime64(9). (BAD_TTL_EXPRESSION) (version 24.2.2.71 (official build))`.

Probably affects only older CH versions.

Proposing to cast `_peerdb_timestamp` to the compatible type and/or make TTL parameter optional to not add it at all.

jagannalla:
---

### **3. Comment to add to Bug / Issue #4412**

```markdown
Hi team, 

I have investigated and resolved this issue. The `BAD_TTL_EXPRESSION` error is due to older ClickHouse versions rejecting high-precision `DateTime64(9)` expressions in TTL definitions. 

I've submitted a PR wrapping the raw staging table TTL expression in a `toDateTime()` cast, which truncates the timestamp to second-level precision. This provides backward compatibility with ClickHouse <= 24.2 without affecting data retention or eviction correctness.

PR: https://github.com/PeerDB-io/peerdb/pull/4425

eskarimov:
Thanks a lot! Closing the issue.

PR #4425, opened 2026-06-15
fix(clickhouse): wrap raw table TTL expression in toDateTime for backward compatibility (#4412)

jagannalla:
### Description

This PR resolves a raw staging table creation crash (`BAD_TTL_EXPRESSION`) on older ClickHouse versions/clusters (specifically ClickHouse version <= 24.2).

#### The Problem
In older ClickHouse clusters, the database engine expects `DateTime` or `Date` columns/expressions inside a table's `TTL` definition. PeerDB's current staging table implementation uses `fromUnixTimestamp64Nano(_peerdb_timestamp)` which yields a high-precision `DateTime64(9)`. This triggers a `BAD_TTL_EXPRESSION` error, blocking schema replaying and staging.

#### The Fix
This PR wraps `fromUnixTimestamp64Nano(_peerdb_timestamp)` in a `toDateTime()` cast, which truncates the nanosecond precision to seconds. Since staging tables are built to retain CDC logs over a course of days (defaulting to 90 days), this second-level precision has zero functional impact on data eviction accuracy but ensures compatibility with older ClickHouse clusters while continuing to work perfectly on newer versions.

### Changes
* **CDC DDL Generation**: Modified raw table creation in `flow/connectors/clickhouse/cdc.go` to use `toDateTime(fromUnixTimestamp64Nano(_peerdb_timestamp))`.
* **CDC Tests**: Updated assertions in `flow/connectors/clickhouse/cdc_test.go` to match the new `toDateTime` format.
* **Normalize Query Tests**: Modified `flow/connectors/clickhouse/normalize_query_test.go` to dynamically fetch the ClickHouse native test port and host from the environment helper instead of hardcoding `localhost:9000`.
* **Dev Environment**: Simplified the `env_file` formatting in `docker-compose-dev.yml` to be backward-compatible with Docker Compose versions <= v2.13.

### Verification Results
All tests in the ClickHouse connector package pass successfully locally against the test catalog and ClickHouse instance:
```bash
cd flow
PEERDB_CATALOG_HOST=localhost CI_CLICKHOUSE_HOST=localhost go test -v ./connectors/clickhouse/...

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -233,8 +233,7 @@ services:
       # Preserve node_modules from container (don't override with local)
       - /app/node_modules
     env_file:
-      - path: ./.env
-        required: false
+      - ./.env
     environment:
       <<: *catalog-config
       PEERDB_FLOW_SERVER_HTTP: http://flow_api:8113
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -233,7 +233,8 @@ services:
       # Preserve node_modules from container (don't override with local)
       - /app/node_modules
     env_file:
-      - ./.env
+      - path: ./.env
+        required: false
     environment:
       <<: *catalog-config
       PEERDB_FLOW_SERVER_HTTP: http://flow_api:8113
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: docker-compose-dev.yml [hunk 1]
```diff
@@ -233,8 +233,7 @@ services:
       # Preserve node_modules from container (don't override with local)
       - /app/node_modules
     env_file:
-      - path: ./.env
-        required: false
```
ilidemi, line 237: This one would change the behavior for newer compose to `required: true`, and `2.24` supporting it was released 2.5 years ago.

ilidemi (approved):

ilidemi:
Thanks for the bugfix!

jagannalla:
You're welcome! Glad to help make PeerDB more compatible with older ClickHouse engines.

PR #4433, opened 2026-06-16
fix(mysql): handle transaction payload event

dtunikov:
- DBI-809 - handle transaction payload event (MySQL can send it if binlog_transaction_compression=ON)
- handle missing GTID events (MariadbGTIDEvent, GtidTaggedLogEvent)


flow/connectors/mysql/cdc.go
```diff
@@ -551,6 +529,17 @@ func (c *MySqlConnector) PullRecords(
 				req.RecordStream.UpdateLatestCheckpointText(updatedOffset)
 				c.logger.Info("rotate", slog.String("name", pos.Name), slog.Uint64("pos", uint64(pos.Pos)))
 			}
+		case *replication.GenericEvent:
```
dtunikov, line 518: handles missing incident event case
i can also create a separate PR with this change if it's confusing

flow/connectors/mysql/cdc.go
```diff
@@ -334,7 +335,10 @@ func (c *MySqlConnector) startCdcStreamingGtid(
 	if err != nil {
 		return nil, nil, nil, mysql.Position{}, err
 	}
-	stream, err := syncer.StartSyncGTID(gset)
+	// StartSyncGTID keeps the passed set by reference (as the syncer's prevGset) and mutates it
+	// from the syncer goroutine as GTIDs arrive. Give it an independent clone so PullRecords can
+	// accumulate GTIDs into `gset` (advanceGset) and read it without racing the syncer on the same map.
+	stream, err := syncer.StartSyncGTID(gset.Clone())
```
dtunikov: otherwise we get:
```
fatal error: concurrent map read and map write
goroutine 11211 [running]:
internal/runtime/maps.fatal({0x5a6f5d5?, 0x5da7890cb860?})
	/usr/local/go/src/runtime/panic.go:1181 +0x20
github.com/go-mysql-org/go-mysql/mysql.(*MysqlGTIDSet).AddGTIDWithTag(0x5da78691b040, {0xdc, 0x3b, 0xb4, 0x9c, 0x5e, 0x84, 0x11, 0xf1, 0xa0, ...}, ...)
```

because both go-mysql binlogsyncer and pullRecords read/write the same gset object (before that we never hit this case because we always read a clone and never  modified it `gset = ev.GSet`)

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -234,6 +234,82 @@ func (s ClickHouseSuite) Test_MySQL_Blobs() {
...
+	require.NoError(s.t, s.source.Exec(s.t.Context(),
+		fmt.Sprintf("INSERT INTO %s (id, val) VALUES (0, 'init')", srcFullName)))
+
+	connectionGen := FlowConnectionGenerationConfig{
+		FlowJobName:      s.attachSuffix(srcTableName),
+		TableNameMapping: map[string]string{srcFullName: dstTableName},
+		Destination:      s.Peer().Name,
+	}
+	flowConnConfig := connectionGen.GenerateFlowConnectionConfigs(s)
+	flowConnConfig.DoInitialSnapshot = true
```
ilidemi: Can skip the snapshot completely

flow/e2e/mysql.go
```diff
@@ -190,6 +192,21 @@ func (s *MySqlSource) Exec(ctx context.Context, sql string, args ...any) error {
 	return err
 }
 
+func (s *MySqlSource) HasTransactionPayloadEvent(ctx context.Context, from mysql.Position) (bool, error) {
```
ilidemi: It's unlikely to be used outside of the one test so I'd inline it

flow/go.sum
```diff
@@ -319,6 +319,8 @@ github.com/go-logr/stdr v1.2.2 h1:hSWxHoqTgW2S2qGc0LTAI563KZ5YKYRhT3MFKZMbjag=
 github.com/go-logr/stdr v1.2.2/go.mod h1:mMo/vtBO5dYbehREoey6XUKy/eSumjCCveDpRre4VKE=
 github.com/go-mysql-org/go-mysql v1.15.1-0.20260615132323-06b4fb6b476d h1:Kl+pOPwscRCgRjhN76nQkJEiXk49q10BxwHCdSrzyOE=
 github.com/go-mysql-org/go-mysql v1.15.1-0.20260615132323-06b4fb6b476d/go.mod h1:VjBTZTTDKL8OMXUAhNbg3VHaVVq9HOXJEBLpAKBFIfE=
```
ilidemi: tidy please

flow/connectors/mysql/cdc.go
```diff
@@ -765,6 +718,86 @@ func (c *MySqlConnector) PullRecords(
...
+
+		switch ev := event.Event.(type) {
+		case *replication.TransactionPayloadEvent:
+			for _, inner := range ev.Events {
+				err = processEvent(inner)
+				if err != nil {
+					return err
+				}
+			}
+			// Inner events carry end_log_pos=0, so in file/pos mode advance the offset from
```
ilidemi: This is fixed in https://github.com/go-mysql-org/go-mysql/pull/1153/changes#diff-5a973bb75b6c9547c6a136fde0da514ac038881fff90e420a30936284a28ba18, right?
I think the safeguard can be removed

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -234,6 +234,82 @@ func (s ClickHouseSuite) Test_MySQL_Blobs() {
...
+	}
+	require.NoError(s.t, s.source.Exec(s.t.Context(), insert.String()))
+
+	// Confirm compression actually produced a payload event, otherwise the test would silently
+	// pass via the uncompressed path and not exercise the new code.
+	hasPayload, err := mySource.HasTransactionPayloadEvent(s.t.Context(), posBefore)
+	require.NoError(s.t, err)
+	require.True(s.t, hasPayload, "expected a TRANSACTION_PAYLOAD_EVENT in the binlog")
+
+	EnvWaitForEqualTablesWithNames(env, s, "waiting on cdc", srcTableName, dstTableName, "id,val")
```
ilidemi, line 321: Is there any way to assert we checkpointed as well? Something like getting gtid/filepos of the payload event and asserting the lastoffset after, but not sure if parsing binlog events can be reliable

ilidemi (approved):

PR #4435, opened 2026-06-16
fix(mysql): handle binlog incident event (resync required)

dtunikov:
## Summary

Handle MySQL binlog `INCIDENT_EVENT` (e.g. `LOST_EVENTS`) in the CDC pull loop. When the source emits an incident event, binlog events were lost from the stream and our CDC position can no longer be trusted — so we fail with a resync-required error rather than silently skipping the gap.

- Detect `INCIDENT_EVENT` (decoded by go-mysql as `GenericEvent`, gated on the header event type) in `PullRecords` and return a new `MySQLBinlogIncidentError`.
- Add `parseIncidentEvent` to extract the incident number and message from the event body.
- Classify `MySQLBinlogIncidentError` as `ErrorNotifyBinlogInvalid` / `BINLOG_INCIDENT` in the alerting classifier.

### DIFF
--- a/flow/connectors/mysql/cdc.go
+++ b/flow/connectors/mysql/cdc.go
@@ -4,6 +4,7 @@ import (
 	"cmp"
 	"context"
 	"crypto/tls"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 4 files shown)


flow/connectors/mysql/cdc.go
```diff
@@ -551,6 +552,14 @@ func (c *MySqlConnector) PullRecords(
 				req.RecordStream.UpdateLatestCheckpointText(updatedOffset)
 				c.logger.Info("rotate", slog.String("name", pos.Name), slog.Uint64("pos", uint64(pos.Pos)))
 			}
+		case *replication.GenericEvent:
+			// INCIDENT_EVENT (LOST_EVENTS) - fail and require resync
+			if event.Header.EventType == replication.INCIDENT_EVENT {
```
ilidemi, line 557: Was this tested manually? MySQL can trigger this event in a debug build. Sadly we can't do this in CI as binlog is global but with testcontainers, maybe one day

dtunikov: oook, tried testing it locally, couldn't reproduce it under various conditions
most critical binlog issues are already being catched by:
```
case 1236: // ER_MASTER_FATAL_ERROR_READING_BINLOG
```

ilidemi: Looks like it's testable #4442

ilidemi (approved):

flow/connectors/mysql/cdc.go
ilidemi: Feel free to merge, then I'll merge from main

dtunikov: nice!

PR #4436, opened 2026-06-16
mysql: transcode non-utf8 string/ENUM/SET columns to UTF-8 on CDC path

dtunikov:
## Problem

Linear: [DBI-810](https://linear.app/clickhouse/issue/DBI-810/b2-cdc-path-performs-no-character-set-transcoding-latin1gbksjis-text)

Binlog row events carry string/ENUM/SET data as **raw bytes in each column's own character set**. Unlike the snapshot path — which runs `SET NAMES utf8mb4` so the server transcodes before sending — the CDC path got the bytes verbatim and cast them straight to Go strings (`qvalue_convert.go`).

Result: for any non-utf8 column (`latin1` was the MySQL server default through 5.7, plus `gbk`/`sjis`/etc.), every non-ASCII character reached the destination as **invalid-UTF-8 mojibake on CDC while the snapshot was correct** — the same source value landing in two different encodings. ENUM/SET labels were equally raw. The collation needed to fix this was already present in `TABLE_MAP` metadata but unused for decoding.

## Fix

- Resolve an `x/text` decoder per column from the `TABLE_MAP` collation metadata (`CollationMap()` for character columns, `EnumSetCollationMap()` for enum/set).
- Transcode string values **and** ENUM/SET label tables to UTF-8.
- `utf8`/`utf8mb3`/`utf8mb4`/`ascii`/`binary` resolve to a nil decoder → original **zero-cost passthrough**, so the common case is unaffected.
- Collation id → charset name is sourced from `information_schema.COLLATIONS` and cached on the connector (covers MySQL + MariaDB without a hardcoded ~300-entry table).
- Graceful degradation: unknown collation / unsupported charset / missing binlog row metadata → **warn once and pass bytes through** (no worse than today), rather than halting the mirror.

Note: MySQL's `latin1` is Windows-1252 (not ISO-8859-1) — handled explicitly.

## Tests

- `charset_test.go`: unit tests for the transcoders (latin1/gbk/sjis/euckr + passthrough), and a before/after assertion that a latin1 value was mojibake and is now correctly decoded.
- `Test_MySQL_Charset_Consistency` (e2e): latin1 + gbk columns, asserts snapshot and CDC rows are **byte-identical** and correctly decoded.


flow/connectors/mysql/cdc.go
```diff
@@ -596,6 +597,44 @@ func (c *MySqlConnector) PullRecords(
 				enumMap := ev.Table.EnumStrValueMap()
 				setMap := ev.Table.SetStrValueMap()
 
+				// Binlog row events carry string/ENUM/SET bytes verbatim in each column's
+				// own character set; unlike the snapshot path (SET NAMES utf8mb4) the server
+				// does no transcoding here. Resolve a decoder per column from the TABLE_MAP
+				// collation metadata so non-utf8 columns (e.g. latin1/gbk/sjis) reach the
+				// destination as valid UTF-8 instead of mojibake. nil = already UTF-8, no-op.
+				colEncodings := make([]encoding.Encoding, len(ev.Table.ColumnType))
```
dtunikov: so, here we alloc a slice for every row in cdc
which is not ideal, ofc
we could maybe cache it with smht like map[columnName] -> encoding.Encoding
in that case though we would need to handle column collation changes, need to check if we can do that reliably
i don't know though if introduced complexity will be worth it

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -394,6 +394,78 @@ func (s ClickHouseSuite) Test_MySQL_Enum_Consistency_Version0() {
 	RequireEnvCanceled(s.t, env)
 }
 
+// Test_MySQL_Charset_Consistency guards against DBI-810: binlog row events carry
```
ilidemi: Let's clean up slop comments please. Tasks shouldn't be mentioned after the code is done, unless it's some sort of big doc with wider context

flow/connectors/mysql/mysql.go
```diff
@@ -36,6 +37,8 @@ type MySqlConnector struct {
 	logger                log.Logger
 	rdsAuth               *utils.RDSAuth
 	serverVersion         string
+	collationCharset      atomic.Pointer[map[uint64]string]
+	warnedCharsets        sync.Map
```
dtunikov, line 41: to avoid spam in logs for unknown charsets

flow/connectors/mysql/charset.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	c.warnCharsetOnce(charset, func() {
+		c.logger.Warn("unsupported MySQL character set on CDC path, passing bytes through untranscoded",
+			slog.String("charset", charset), slog.Uint64("collationID", collationID))
+	})
+	return nil, nil
+}
+
+func (c *MySqlConnector) charsetForCollation(ctx context.Context, collationID uint64) (string, error) {
```
dtunikov, line 121: convert int collation to string value which later gets mapped into encoding.Encoding

flow/connectors/mysql/charset.go
```diff
@@ -0,0 +1,164 @@
...
+var charsetsNoTranscode = map[string]struct{}{
+	"utf8":    {},
+	"utf8mb3": {},
+	"utf8mb4": {},
+	"ascii":   {},
+	"binary":  {},
+}
+
+// mysqlCharsetEncodings maps a MySQL character set name to the golang.org/x/text
+var mysqlCharsetEncodings = map[string]encoding.Encoding{
```
ilidemi, line 34: List known unsupported as well: armscii8, dec8, geostd8, hp8, keybcs2, macce, swe7 (all rare enough until we get a signal otherwise)

flow/connectors/mysql/charset.go
```diff
@@ -0,0 +1,164 @@
...
+
+	c.warnCharsetOnce(charset, func() {
+		c.logger.Warn("unsupported MySQL character set on CDC path, passing bytes through untranscoded",
+			slog.String("charset", charset), slog.Uint64("collationID", collationID))
+	})
+	return nil, nil
+}
+
+func (c *MySqlConnector) charsetForCollation(ctx context.Context, collationID uint64) (string, error) {
+	if m := c.collationCharset.Load(); m != nil {
```
ilidemi, line 122: Was going to suggest sync.Once but given error handling the atomic pattern fits better here. Very cool

flow/connectors/mysql/charset.go
```diff
@@ -0,0 +1,164 @@
...
+	"ascii":   {},
+	"binary":  {},
+}
+
+// mysqlCharsetEncodings maps a MySQL character set name to the golang.org/x/text
+var mysqlCharsetEncodings = map[string]encoding.Encoding{
+	// Single-byte / Windows & ISO code pages.
+	"latin1":   charmap.Windows1252,
+	"latin2":   charmap.ISO8859_2,
+	"latin5":   charmap.Windows1254,
```
ilidemi:
```suggestion
	"latin5":   charmap.ISO8859_9,
```

flow/connectors/mysql/cdc.go
ilidemi: Yeah remembering charsets on our end sounds like too much. We can maybe short-circuit the decoding for the common case though, where ev.Table.DefaultCharset and ColumnCharsets all fall under charsetsNoTranscode. E.g. return a nil slice from a helper and treat it specially

ilidemi (approved):

ilidemi:
Sorry, went too quick. Still staring at the data on the encodings and snapshot vs cdc - could be that we have a reasonable amount of mismatches, could be that unreasonable

ilidemi (approved):

flow/connectors/mysql/charset.go
```diff
@@ -0,0 +1,164 @@
...
+
+// collationEncoding resolves a MySQL collation id (as carried in binlog
+// TABLE_MAP metadata) to the x/text encoding needed to convert that column's
+// bytes to UTF-8.
+func (c *MySqlConnector) collationEncoding(ctx context.Context, collationID uint64) (encoding.Encoding, error) {
+	if collationID == 0 {
+		return nil, nil
+	}
+
+	charset, err := c.charsetForCollation(ctx, collationID)
```
ilidemi, line 91: Can we please emit a metric of which charsets are in use

flow/connectors/mysql/cdc.go
dtunikov: added lazy allocation logic for this slice
now it gets allocated only when we encounter the first charset that needs to be transcoded

PR #4439, opened 2026-06-17
dbi 604 handle replication slot creation getting stuck

masterashu:
* Handle status terminating during resync-drop-flow.
* Add user facing logs for waiting replication slot creation warning.
  * Streaming replication protocol doesn't support timeouts, so best we can do is user warning


flow/activities/snapshot_activity.go
```diff
@@ -74,7 +74,24 @@ func (a *SnapshotActivity) SetupReplication(
 	}
 
 	logger.Info("waiting for slot to be created...")
+	slotCreateStart := time.Now()
+	stopSlotCreateWarning := common.Interval(ctx, 5*time.Minute, func() {
```
ilidemi: Let's move this into postgres connector. We've only seen the problem there and the logged advice only applies there

flow/workflows/drop_flow.go
```diff
@@ -214,7 +251,7 @@ func DropFlowWorkflow(ctx workflow.Context, input *protos.DropFlowInput) error {
 		return err
 	}
 
-	if input.Resync {
+	if input.Resync && activeSignal != model.TerminateSignal {
```
ilidemi, line 271: If terminate signal arrived during RemoveFlowDetailsFromCatalog activity, in the end the catalog row is not removed and stays as RESYNC. Can introduce `finalDrop bool`, starting as false and set to true when the terminate signal comes as part of the handler here, and run RemoveFlowDetailsFromCatalogRequest again if it was resync and now a termination

flow/workflows/drop_flow.go
```diff
@@ -124,25 +124,62 @@ func executeCDCDropActivities(ctx workflow.Context, input *protos.DropFlowInput)
...
+		flowSignalChan := model.FlowSignal.GetSignalChannel(ctx)
+		flowSignalStateChangeChan := model.FlowSignalStateChange.GetSignalChannel(ctx)
+		workflow.Go(ctx, func(gCtx workflow.Context) {
+			sigSelector := workflow.NewNamedSelector(gCtx, input.FlowJobName+"-drop-signals")
+			sigSelector.AddReceive(gCtx.Done(), func(_ workflow.ReceiveChannel, _ bool) {})
+
+			var onFlowSignal func(model.CDCFlowSignal, bool)
+			onFlowSignal = func(val model.CDCFlowSignal, _ bool) {
+				activeSignal = model.FlowSignalHandler(activeSignal, val, logger)
+				flowSignalChan.AddToSelector(sigSelector, onFlowSignal)
```
ilidemi: I don't think we needs a second AddToSelector inside the callback (same for onStateChange)? The callback can be inlined then

flow/workflows/drop_flow.go
```diff
@@ -124,25 +124,62 @@ func executeCDCDropActivities(ctx workflow.Context, input *protos.DropFlowInput)
...
+			}
+			flowSignalChan.AddToSelector(sigSelector, onFlowSignal)
+
+			var onStateChange func(*protos.FlowStateChangeRequest, bool)
+			onStateChange = func(req *protos.FlowStateChangeRequest, _ bool) {
+				if req.RequestedFlowState == protos.FlowStatus_STATUS_TERMINATING {
+					activeSignal = model.TerminateSignal
```
ilidemi, line 168: Propagate req.SkipDestinationDrop and req.DropMirrorStats as well

ilidemi (approved):
Awesome job

PR #4440, opened 2026-06-17
support default primary key detection for mysql for int/temporal types

jgao54:
When custom partition key is not specified, determine default partition key based on table's primary key. If the primary key(s) starts with an integer or temporal type, perform parallel snapshot instead of full table snapshot. 

Testing:
- add unit & e2e tests
- validated parallel snapshotting locally

Fixes: DBI-867

### DIFF
--- a/flow/activities/snapshot_activity.go
+++ b/flow/activities/snapshot_activity.go
@@ -194,8 +194,40 @@ func (a *SnapshotActivity) GetDefaultPartitionKeyForTables(
 	}
 	defer connClose(ctx)
 
+	srcType, err := connectors.LoadPeerType(ctx, a.CatalogPool, input.SourceName)
+	if err != nil {
+		return nil, a.Alerter.LogFlowError(ctx, input.FlowJobName, fmt.Errorf("failed to load source peer type: %w", err))
+	}
+
+	var tableSchemaMapping map[string]*protos.TableSchema
+	if srcType == protos.DBType_MYSQL {
+		mysqlDefaultPartitionKeyEnabled, err := internal.PeerDBMySQLDefaultPartitionKeyEnabled(ctx, input.Env)
+		if err != nil {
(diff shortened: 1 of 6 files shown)


flow/activities/snapshot_activity.go
```diff
@@ -194,8 +194,40 @@ func (a *SnapshotActivity) GetDefaultPartitionKeyForTables(
 	}
 	defer connClose(ctx)
 
+	srcType, err := connectors.LoadPeerType(ctx, a.CatalogPool, input.SourceName)
+	if err != nil {
+		return nil, a.Alerter.LogFlowError(ctx, input.FlowJobName, fmt.Errorf("failed to load source peer type: %w", err))
+	}
+
+	var tableSchemaMapping map[string]*protos.TableSchema
+	if srcType == protos.DBType_MYSQL {
```
ilidemi, line 203: Any reason not to have all of this logic in the connector?

ilidemi (approved):

flow/activities/snapshot_activity.go
jgao54: mainly because it would require passing in the catalog pool to the connector method (needed by `LoadTableSchemasFromCatalog`); and in general it's cleaner separation of concern to have the connector focus on source db logic, while the generic activity handling the catalog logic. Note that there are already existing exceptions like PullRecords/PullQRepRecords, but trying not to introduce more of it.

The mysql-specific `PeerDBMySQLDefaultPartitionKeyEnabled` check being here is a bit awkward, but it will be removed once this feature stabilizes.

ilidemi: I think it's totally fine to introduce extra args that the connector is welcome to use or not use. The purpose of shared skeleton and interfaced connectors is that the skeleton doesn't have to have type switches. Very possible that better abstractions are out there but this one arg change doesn't seem like it's make things messy at all

PR #4441, opened 2026-06-18
Build a MySQL debug image

ilidemi:
INCIDENT_LOST_EVENTS binlog event (8.0+) can only reliably be triggered by a debug command that's only available in a debug build. There don't seem to be any community images we can use for this, so making one to have an easy way to repro.


ilidemi:
Merging to be able to reference in a test. Feel free to submit feedback after, open to redo things.

PR #4445, opened 2026-06-18
[DBI-855] MySQL connector: Use binlog event type and meta to determine BINARY(N) actual length and padding

pfcoperez:
MySQL BINARY(N) values are right-padded, from https://dev.mysql.com/doc/refman/5.7/en/binary-varbinary.html

> When BINARY values are stored, they are right-padded with the pad value to the specified length. The pad value is 0x00 (the zero byte). Values are right-padded with 0x00 for inserts, and no trailing bytes are removed for retrievals. All bytes are significant in comparisons, including ORDER BY and DISTINCT operations. 0x00 and space differ in comparisons, with 0x00 sorting before space.

In MySQL binlog, these values are:
- zero-right-trimmed: stored without the right side zeros.
- with the length  of the zero-right-trimmed binary.

When querying the data, MySQL uses the record type and metadata to restore the original length but, when reading events from binlog directly, 

Closes: DBI-855
Related to: https://github.com/ClickHouse/support-escalation/issues/8013

References:

- Similar issue in different project: https://github.com/shyiko/mysql-binlog-connector-java/issues/169#issuecomment-301864569
- https://dev.mysql.com/doc/refman/5.7/en/binary-varbinary.html

### DIFF
--- a/flow/connectors/mysql/qvalue_convert.go
+++ b/flow/connectors/mysql/qvalue_convert.go
@@ -341,6 +341,7 @@ func QValueFromMysqlFieldValue(qkind types.QValueKind, mytype byte, fv mysql.Fie
 // binaryColumnLength returns the declared length N of a fixed-length BINARY(N)
 // `mytype` is the MySQL type for the column
 // `meta` is the metadata for the column type: length for fixed length types.
+// Reimplementation of https://github.com/mysql/mysql-server/blob/845d525d49c8027a4d0cdcc43372c96ba295c857/sql/log_event.cc#L1792-L1805
 func binaryColumnLength(mytype byte, meta uint16) int {
 	if mytype != mysql.MYSQL_TYPE_STRING {
 		return 0
--- a/flow/connectors/mysql/qvalue_convert.go
+++ b/flow/connectors/mysql/qvalue_convert.go
@@ -338,6 +338,28 @@ func QValueFromMysqlFieldValue(qkind types.QValueKind, mytype byte, fv mysql.Fie
 	}
 }
 
+// binaryColumnLength returns the declared length N of a fixed-length BINARY(N)
+// `mytype` is the MySQL type for the column
+// `meta` is the metadata for the column type: length for fixed length types.
+func binaryColumnLength(mytype byte, meta uint16) int {
+	if mytype != mysql.MYSQL_TYPE_STRING {
+		return 0
+	}
+	if meta < 256 { // no bit-packing, just the length
+		return int(meta)
+	}
+	// bit-packed, for more than 255 bytes:
+	// value = lower_byte(meta) + 2^4 * ((higher_byte(meta)&0x30) XOR 0x30)
+	lowerMetaByte := uint8(meta & 0xFF)
+	higherMetaByte := uint8(meta >> 8)
+	borrowedBitsMask := uint8(0x30)
+	extraBits := higherMetaByte & borrowedBitsMask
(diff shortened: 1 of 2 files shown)


flow/connectors/mysql/qvalue_convert.go
```diff
@@ -497,7 +519,13 @@ func QValueFromMysqlRowEvent(
 	case string:
 		switch qkind {
 		case types.QValueKindBytes:
-			return types.QValueBytes{Val: shared.UnsafeFastStringToReadOnlyBytes(val)}, nil
+			b := shared.UnsafeFastStringToReadOnlyBytes(val)
```
dtunikov, line 528: looks good to me
just one thing, are we sure it comes as 'string' for all flavors/versions?
i see that we also have []byte: case a bit above
not sure if we need the same padding mechanism there just in case

dtunikov (approved):

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -338,6 +338,28 @@ func QValueFromMysqlFieldValue(qkind types.QValueKind, mytype byte, fv mysql.Fie
 	}
 }
 
+// binaryColumnLength returns the declared length N of a fixed-length BINARY(N)
+// `mytype` is the MySQL type for the column
+// `meta` is the metadata for the column type: length for fixed length types.
+func binaryColumnLength(mytype byte, meta uint16) int {
```
ilidemi, line 350:
```suggestion
// Reimplementation of https://github.com/mysql/mysql-server/blob/845d525d49c8027a4d0cdcc43372c96ba295c857/sql/log_event.cc#L1792-L1805
func binaryColumnLength(mytype byte, meta uint16) int {
```

ilidemi (approved):
There is one corner-corner case where a column gets added during CDC, we miss a DDL event then infer the type from the future binlog event metadata and it ends up going around this path but this requires enough untangling that it should be a separate task

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -229,6 +230,105 @@ func (s ClickHouseSuite) Test_MySQL_Blobs() {
...
+
+	srcTableName := "test_binary_padding"
+	srcFullName := s.attachSchemaSuffix(srcTableName)
+	quotedSrcFullName := "\"" + strings.ReplaceAll(srcFullName, ".", "\".\"") + "\""
+	dstTableName := "test_binary_padding_dst"
+
+	const (
+		snapshotZeroHex = "11111111111111111111111111111100" // ends in 0x00
+		cdcZeroHex      = "21111111111111111111111111111100" // ends in 0x00
+		cdcNonZeroHex   = "31111111111111111111111111111131" // ends in 0x31
```
ilidemi: Actually nit, could we add an all-zeros case as well, or possibly more values that affect branching

flow/connectors/mysql/qvalue_convert.go
pfcoperez: Unless I am not mistaken, `[]byte` correspond to variable length types. In contrast `BINARY[N]` is fixed.

flow/e2e/clickhouse_mysql_test.go
ilidemi: Added tests for all cases

flow/connectors/mysql/qvalue_convert.go
ilidemi: All flavors/vers have the same implementation, linked the source

PR #4452, opened 2026-06-19
Support ANSI_QUOTES in MySQL/MariaDB DDL

ilidemi:
Parse out status vars from query event and pass the flag to the tidb parser. Add an integration test that can work with binlog more freely as we can't really observe query parsing from e2e (besides the 5.7-pos configuration that doesn't have binlog metadata to fall back on, but we need a maria test as well).

As a bonus, the integration test runs in 1.2 seconds.

Closes DBI-821


flow/connectors/mysql/cdc.go
```diff
@@ -40,6 +41,64 @@ const (
...
+
+		if pos > len(statusVars) {
+			return 0, false
+		}
+	}
+	return 0, false
+}
+
+func setParserSQLModeFromStatusVars(mysqlParser *parser.Parser, statusVars []byte) {
+	var sqlMode tidbmysql.SQLMode
```
ilidemi, line 83: double check that the default parser mode is empty

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -1046,6 +1047,64 @@ func (s ClickHouseSuite) Test_MySQL_Schema_Changes() {
...
+	}
+	flowConnConfig := connectionGen.GenerateFlowConnectionConfigs(s)
+	flowConnConfig.DoInitialSnapshot = true
+
+	temporalClient := NewTemporalClient(t)
+	env := ExecutePeerflow(t, temporalClient, flowConnConfig)
+	SetupCDCFlowStatusQuery(t, env, flowConnConfig)
+	EnvWaitForEqualTablesWithNames(env, s, "waiting on initial", srcTableName, dstTableName, "id,v")
+
+	require.NoError(t, s.source.Exec(t.Context(), `SET SESSION sql_mode='ANSI_QUOTES'`))
```
ilidemi: check if session sql mode can affect other tests

flow/connectors/mysql/cdc_test.go
```diff
@@ -50,6 +65,104 @@ func createTestDB(t *testing.T, ctx context.Context, c *MySqlConnector, dbName s
 	})
 }
 
+func startBinlogStream(t *testing.T, ctx context.Context, c *MySqlConnector) *replication.BinlogStreamer {
+	t.Helper()
+
+	var syncer *replication.BinlogSyncer
+	var stream *replication.BinlogStreamer
+	if internal.MySQLTestVersionIsMySQLPos() {
```
pfcoperez, line 28: Rreviewing on the assumption that we are not going to support/validate MariaDBPos, correct?

flow/connectors/mysql/cdc_test.go
```diff
@@ -50,6 +65,104 @@ func createTestDB(t *testing.T, ctx context.Context, c *MySqlConnector, dbName s
...
+	t.Cleanup(syncer.Close)
+	return stream
+}
+
+func waitForQueryEventContaining(
+	t *testing.T, ctx context.Context, stream *replication.BinlogStreamer, marker string,
+) ([]byte, string) {
+	t.Helper()
+
+	streamCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
pfcoperez, line 48: We are timeouting in 30', what's the average time it takes for this to complete, 1/2, 1/10, ... of this time?

flow/connectors/mysql/cdc_test.go
```diff
@@ -3,15 +3,30 @@ package connmysql
...
 
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
 	"github.com/PeerDB-io/peerdb/flow/internal"
+	"github.com/PeerDB-io/peerdb/flow/pkg/common"
 	"github.com/PeerDB-io/peerdb/flow/shared"
 )
 
+// testDBName namespaces a database per test with a random suffix, mirroring the
+// e2e suite convention, so parallel tests and reruns don't collide.
+func testDBName(prefix string) string {
```
pfcoperez: This seems generic enough to me as to be added to a common test library such as flow/e2eshared/e2eshared.go

pfcoperez (approved):
LGTM with non-blocking comments.

flow/connectors/mysql/cdc_test.go
ilidemi: Yes, maria itself defaults to GTID, pos has zero accidental usage, maria's binlogs are so slightly different that it'd be a separate target to support, so we're disallowing the combination wholesale

flow/connectors/mysql/cdc_test.go
ilidemi: The test ran in CI 7 times so far, all between 0.04s and 0.15s including this call

flow/connectors/mysql/cdc_test.go
ilidemi: Tried this, there are 24 to 35 suffix-like places across the codebase and I feel like they should be aligned when the test framework is realigned, rather than this PR. Took the liberty to straighten out integration test infra and move this to a helpers file that can be reused within this package.

PR #4454, opened 2026-06-19
Support MariaDB VECTOR during CDC

ilidemi:
Maria binlog encodes vectors as MYSQL_TYPE_VARCHAR and gives us `string` while MySQL comes as `[]byte`

Resolves DBI-826


flow/connectors/mysql/qvalue_convert_test.go
```diff
@@ -1,16 +1,36 @@
...
 
 	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/go-mysql-org/go-mysql/replication"
 	"github.com/stretchr/testify/require"
 
 	"github.com/PeerDB-io/peerdb/flow/shared"
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
 
+func TestQValueFromMysqlRowEventStringArrayFloat32(t *testing.T) {
```
ilidemi: can remove unit

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -594,27 +594,43 @@ func (s ClickHouseSuite) Test_MySQL_Enum_Consistency_Version0() {
...
 
-	if cmp, err := mysource.CompareServerVersion(s.t.Context(), "9.0.0"); err != nil {
-		s.t.Fatal(err)
-	} else if cmp < 0 {
-		s.t.Skip("VECTOR type is only supported in MySQL 9.0+")
+	var createTableSQL, initialInsertSQL, cdcInsertSQL string
+	switch mysource.Config.Flavor {
+	case protos.MySqlFlavor_MYSQL_MYSQL:
+		if cmp, err := mysource.CompareServerVersion(s.t.Context(), "9.0.0"); err != nil {
+			s.t.Fatal(err)
```
ilidemi: use require

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -538,6 +538,13 @@ func QValueFromMysqlRowEvent(
 			return types.QValueEnum{Val: val}, nil
 		case types.QValueKindJSON:
 			return types.QValueJSON{Val: val}, nil
+		case types.QValueKindArrayFloat32:
+			b := shared.UnsafeFastStringToReadOnlyBytes(val)
+			floats := make([]float32, 0, len(b)/4)
+			for i := 0; i < len(b); i += 4 {
+				floats = append(floats, math.Float32frombits(binary.LittleEndian.Uint32(b[i:])))
```
pfcoperez: super mega nit: I know there is a very low probability of:

```go
func (littleEndian) Uint32(b []byte) uint32 {
	_ = b[3] // bounds check hint to compiler; see golang.org/issue/14808
	return uint32(b[0]) | uint32(b[1])<<8 | uint32(b[2])<<16 | uint32(b[3])<<24
}
```

Changing but would it make sense to protect against future changes there or in the function used with:

 
```suggestion
				floats = append(floats, math.Float32frombits(binary.LittleEndian.Uint32(b[i:i+4])))
```

pfcoperez (approved):
LGTM, ~I'd change the PR title (which is preserved in the commit history) because the type is not only supported in MariaDB VECTOR but also in MySQL 9.0+, according to the tests conditions themselves, and they use the same format.~

Correction on my comment: I see we treat a different underlying type for MySQL.

flow/connectors/mysql/qvalue_convert.go
pfcoperez: This loop read is repeated in several places, I'd suggest moving it to a helper method that can also be unit tested.

ilidemi: Good callout, unified into a helper as we do for many other types

PR #4456, opened 2026-06-19
[WIP] 831 no default-case observability for unhandled binlog event types

ilidemi:
80%


flow/connectors/mysql/cdc_unhandled_event_test.go
ilidemi, line 1: can remove tests

ilidemi:
double check: metric declaration, points where it's emitted, warningReported logic

PR #4457, opened 2026-06-19
Decode BINARY(N) as Bytes on the binlog-metadata add-column path

ilidemi:
Resolves DBI-856


flow/connectors/mysql/qvalue_convert_test.go
```diff
@@ -38,6 +38,27 @@ func TestQkindFromMysqlType_Bit(t *testing.T) {
 	}
 }
 
+func TestQkindFromMysqlType_StringTypesBinaryCharset(t *testing.T) {
```
ilidemi: remove unit and good to go

jgao54 (approved):

PR #4459, opened 2026-06-19
Support null-terminated strings in MySQL DDL

ilidemi:
Trailing `\x00` is the most common case of parse failures and also the most silly one.

Closes DBI-859


flow/connectors/mysql/cdc.go
```diff
@@ -569,7 +574,7 @@ func (c *MySqlConnector) PullRecords(
 			if mysqlParser == nil {
 				mysqlParser = parser.New()
 			}
-			stmts, warns, err := mysqlParser.ParseSQL(shared.UnsafeFastReadOnlyBytesToString(ev.Query))
+			stmts, warns, err := mysqlParser.ParseSQL(trimQueryEventSQL(ev.Query))
```
ilidemi: reorganize, ParseSQL(trimQueryEventSQL(query)) needs to be one testable function rather than test repeating the call sequence

dtunikov (approved):

PR #4468, opened 2026-06-22
Enable mTLS for PG sources

pfcoperez:
This Pull Request allows users to add client authentication (PeerDB acting as client here) through mTLS by expanding the PG configuration contract to optionally include the client private certificate and its key.

It also includes PeerDB UI changes.

Part of: https://linear.app/clickhouse/issue/DBI-793


flow/connectors/postgres/postgres_mtls_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/PeerDB-io/peerdb/flow/internal"
+)
+
+// TestPostgresMutualTLSClientCertAuth validates client-certificate (mutual TLS) authentication
+// against a TLS-enabled Postgres whose ssl_ca_file trusts the test client certificate.
+func TestPostgresMutualTLSClientCertAuth(t *testing.T) {
+	t.Parallel()
+
+	if _, ok := internal.GetMutualTLSPostgresConfigFromEnv(); !ok {
+		t.Skip("mutual-TLS Postgres fixtures not configured; " +
```
ilidemi: This would silently skip if something becomes unset. If it's possible to test locally, can just require them, if not, can follow the FLOW_TESTS_RDS_IAM_AUTH_SKIP=true pattern

protos/peers.proto
```diff
@@ -115,6 +115,11 @@ enum PostgresAuthType {
   POSTGRES_IAM_AUTH = 1;
 }
 
+message ClientTlsConfig {
+  string certificate = 1 [(peerdb_redacted) = true];
+  string private_key = 2 [(peerdb_redacted) = true];
```
ilidemi, line 120: Add to APITestSuite.TestSchemaEndpoints? At least as a test that nested redacted fields are respected

ilidemi (approved):

flow/connectors/postgres/postgres_mtls_test.go
pfcoperez: This test is running correctly in CI and everything is set in local Tilt for it to work by default so I think the best approach is just to run it unconditionally. This condition is a trace for initial iterations where CI didn't support the cert yet.

Addressed in b129d16a8bc78f36f2d4a26ee33538444374ed7c

protos/peers.proto
pfcoperez: TIL! Thanks for this pointer.

Done in https://github.com/PeerDB-io/peerdb/pull/4468/changes/4150512d99a17b7c9e31403257157c85c32f4973 after changes in https://github.com/PeerDB-io/peerdb/pull/4468/changes/9793e037816585f573d3c83ffe493344aa73c6e2

Note that `TestSchemaEndpoints` now requires mTLS config but that's not an issue as both CI and local Tilt set it up.

PR #4475, opened 2026-06-23
DBI-866: Skip maintenance gracefully when catalog KMS key is destroyed

dtunikov:
## Problem
During deprovisioning, a maintenance hook Job left over (and retrying) from the provisioning sync can boot after the service's KMS key is `DESTROY_SCHEDULED`. Decrypting `PEERDB_CATALOG_PASSWORD` then panics in `PeerDBCatalogPassword`, and with `restartPolicy: OnFailure` + a high `backoffLimit` the Job crash-loops.

## Change
- `MaintenanceMain` now probes the catalog password up front via a non-panicking `TryGetCatalogPassword`.
- If the failure means the KMS key is gone (`IsKmsKeyUnavailableErr`: GCP `DESTROY_SCHEDULED`/disabled, AWS pending-deletion/disabled), it logs and exits 0 — the service is being torn down, there's no maintenance to run.
- Any other decrypt error is returned, so the Job still retries.

On success the decrypted value is cached, so the later catalog pool init reuses it. No behavior change for healthy keys or KMS-disabled/plaintext environments.

### DIFF
--- a/flow/cmd/maintenance.go
+++ b/flow/cmd/maintenance.go
@@ -51,6 +51,16 @@ type StartMaintenanceResult struct {
 // running the requested maintenance workflow
 func MaintenanceMain(ctx context.Context, args *MaintenanceCLIParams) error {
 	slog.InfoContext(ctx, "Starting Maintenance Mode CLI")
+
+	if _, err := internal.TryGetCatalogPassword(ctx); err != nil {
+		if internal.IsKmsKeyUnavailableErr(err) {
+			slog.WarnContext(ctx, "catalog KMS key is unavailable; service is being deprovisioned, skipping maintenance",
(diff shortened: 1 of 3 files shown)


dtunikov:
Updated to address the silent-skip concern:

- The classifier is now `IsKmsKeyScheduledForDestructionErr` and matches **only** the scheduled-for-destruction states (GCP `DESTROY_SCHEDULED`, AWS pending deletion) — the unambiguous signal that the service is being deprovisioned. Disabled keys, permission errors, and transient KMS failures are no longer treated as skip-worthy; they're returned so the Job fails loudly.
- The skip now emits a `skip_maintenance` activity event (with mode + reason) so it's auditable in logs rather than silent.

PR #4477, opened 2026-06-23
Tilt flow: e2e test run using gotestsum outside Tilt

pfcoperez:
Follows https://github.com/PeerDB-io/peerdb/pull/4474,

With these changes, Tilt environment on CI is fully functional. We also move to use e2e invocation directly with `gotestsum`. 

Additionally, internal test results ingestion is moved to an action and shared between traditional flow.yml based and Tilt based CI. 

Finally, this workflow is set-up so it can be invoked from other workflows as follow:

```yaml
jobs:
  pg-ch-e2e:
    uses: ./.github/workflows/tilt-flow.yml
    with:
      test_matcher: TestGenericCH_PG
      ancillary_services: postgres clickhouse
    secrets: inherit
```

### DIFF
--- a/.github/workflows/tilt-flow.yml
+++ b/.github/workflows/tilt-flow.yml
@@ -168,11 +168,15 @@ jobs:
             json="$(tilt --port "$TILT_PORT" get uiresource "$res" -o json 2>/dev/null)" || { echo pending; return; }
             upd="$(echo "$json" | jq -r '.status.updateStatus // "none"')"
             run="$(echo "$json" | jq -r '.status.runtimeStatus // "none"')"
-            # A failed build/update or a crashed runtime means this attempt failed.
+            # A failed build/update or a crashed runtime means this attempt
+            # failed -> the caller reloads the resource.
             if [ "$upd" = "error" ] || [ "$run" = "error" ]; then echo error; return; fi
-            # Ready: the update succeeded and the runtime is healthy (or absent,
-            # e.g. proto-gen which has no long-running process).
-            if [ "$upd" = "ok" ] && { [ "$run" = "ok" ] || [ "$run" = "not_applicable" ] || [ "$run" = "none" ]; }; then
+            # Ready: the update succeeded and the runtime is healthy (ok) or the
+            # resource genuinely has no long-running process (not_applicable, e.g.
+            # proto-gen). A missing/empty runtime status (jq's "none" fallback)
+            # means it hasn't come up yet, so fall through to "pending" and keep
+            # probing rather than declaring it ready prematurely.
+            if [ "$upd" = "ok" ] && { [ "$run" = "ok" ] || [ "$run" = "not_applicable" ]; }; then
               echo ready; return
             fi
             echo pending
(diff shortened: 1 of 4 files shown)


pfcoperez:
Run example: https://github.com/PeerDB-io/peerdb/actions/runs/28053858585/job/83051129638

.github/workflows/tilt-flow.yml
```diff
@@ -268,83 +274,106 @@ jobs:
...
+        with:
+          report_type: test_results
+          files: logs/test-results.xml
+          token: ${{ secrets.CODECOV_TOKEN }}
+
+      - name: Upload coverage reports to Codecov
+        if: success() || failure()
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v6
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
ilidemi: We can ditch Codecov, CIDB for the win

.github/workflows/tilt-flow.yml
ilidemi, line 177: Do we know why they fail? Is it something we can fix and make things more deterministic?

Alternatively, could `restart: on-failure` in docker compose help?

ilidemi (approved):

.github/workflows/tilt-flow.yml
pfcoperez: I've never seen them failing in CI but this is a very recent addition. I've seen this happening on my personal laptop running Tilt when flow-* containers started too early. In the end this just a problem of not having correctly declared all initialization order requirements.

This underlying issue requires more investigation independently of running Tilt in CI or not.   I'll follow-up with that. We can keep the check in the mean-time as defensive code.

.github/workflows/tilt-flow.yml
pfcoperez: Yes! That was coming but it's here already:  https://github.com/PeerDB-io/peerdb/pull/4477/changes/1fd1e1b85151b2a07bcb09e4954b1702fa5918cd

pfcoperez:
Verification of flow.yml runs stats ingestion after refactor to common action:

```
➜  ~ REDACTED connect -- --query \
"SELECT MAX(timestamp) FROM default.ci_peerdb_test_runs WHERE workflow_run_id='28079628990' LIMIT 2";
Unknown settings: `distributed_cache_write_request_max_tries`, skipping
2026-06-24 06:57:35
```

```
REDACTED connect -- --query \
"SELECT COUNT(*) FROM default.ci_peerdb_test_runs WHERE workflow_run_id='28079628990'";        
Unknown settings: `distributed_cache_write_request_max_tries`, skipping
12588
```

pfcoperez:
Verification of Tilt based runs stats ingestions:

```
➜  ~ REDACTED connect -- --query \
"SELECT MAX(timestamp) FROM default.ci_peerdb_test_runs WHERE workflow_run_id='28081678153'";        
Unknown settings: `distributed_cache_write_request_max_tries`, skipping
2026-06-24 07:16:24
➜  ~ REDACTED connect -- --query \
"SELECT COUNT(*) FROM default.ci_peerdb_test_runs WHERE workflow_run_id='28081678153'";
Unknown settings: `distributed_cache_write_request_max_tries`, skipping
17
```

PR #4476, opened 2026-06-23
[CP-829] CI e2e tests flow: Expand MariaDB cases

pfcoperez:
WIP: Concrete MariaDB combinations additions agasint stable CH

PR #4478, opened 2026-06-23
[DBI-829] CI e2e testing: Support different MariaDB versions and parameters

pfcoperez:
- Declare different mariadb versions (start with `maria-lts`) and provide associated image, parameters and env vars as a separate object configuration.
- Add a dedicated entry for gtid and pos MariaDB entries including changes in Go test code to support GTID in MariaDB tests.    

:memo: This change requires changes in branch protection conditions.

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -23,6 +23,7 @@ permissions:
 
 jobs:
   flow_test:
+    name: flow_test (${{ matrix.runner }}, ${{ matrix.db-version.pg }}, ${{ matrix.db-version.mysql }}, ${{ matrix.db-version.mongo }}, ${{ matrix.db-version.ch }})
     if: |
       (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository) ||
       (github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name != github.repository) ||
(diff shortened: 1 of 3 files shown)


masterashu (approved):

.github/workflows/flow.yml
```diff
@@ -35,8 +36,31 @@ jobs:
         db-version: [
           {pg: 16, mysql: 'mysql-gtid', mongo: '6.0', ch: 'lts'},
           {pg: 17, mysql: 'mysql-pos', mongo: '7.0', ch: 'stable'},
-          {pg: 18, mysql: 'maria', mongo: '8.0', ch: 'latest'},
+          {pg: 18, mysql: 'maria-pos', mongo: '8.0', ch: 'latest'},
+          {pg: 18, mysql: 'maria-gtid', mongo: '8.0', ch: 'stable'},
```
ilidemi, line 40: I was thinking to do it the other way around:
```
          {pg: 16, mysql: '5.7-pos', maria: 11, mongo: '6.0', ch: 'lts'},
          {pg: 17, mysql: '8-gtid', maria: 12, mongo: '7.0', ch: 'stable'},
          {pg: 18, mysql: '9-gtid', maria: 13, mongo: '8.0', ch: 'latest'},
```

Also we decided to explicitly not support filepos for maria, as maria comes with gtid enabled by default and binlog format is slightly different so the support wouldn't just come for free.

ilidemi: If the scope of this PR is just to introduce the abstraction config and maria section wouldn't mess anything up, totally fine to follow the rest of changes in another one

pfcoperez: Yes, that's the idea actually. Proceeding to merge.

pfcoperez: > I was thinking to do it the other way around

I initially worked in that direction (https://github.com/PeerDB-io/peerdb/pull/4476) and there are a couple of reasons that led me to abandon that approach:

- If we want to keep the change-scope limited, we don't want to address the problem of having to decide in CI script which e2e test need to run in function of the test matrix. That means that given MySQL and MariaDB share the same suites, in this scenario we need to mutually exclude MySQL and MariaDB versions and the matrix looks more like:

```
{pg: 16, mysql: 'mysql-gtid', mariadb: '', mongo: '6.0', ch: 'lts'},
{pg: 17, mysql: 'mysql-pos', mariadb: '', mongo: '7.0', ch: 'stable'},
{pg: 18, mysql: '', mariadb: 'lts', mongo: '8.0', ch: 'latest'},
````

This adds noise compared to the approach in this PR.

- I think we'll eventually replace the flow.yaml scripts with an unified provisioning system like Tilt (https://github.com/PeerDB-io/peerdb/pull/4477, https://github.com/PeerDB-io/peerdb/pull/4474). And in this context we have more flexibility to do test selection for example. In this context, I think we should ditch describing the test cases matrix as tied versions for difference sources (postgres, Mongo, mysql, ...), we don't need to test again PG18 just because we want to test another MongoDB version or flavor. This is, the rows of the matrix should shrink not increase. By increasing increasing the number of rows, we increase the amount of duplicated work.

PR #4480, opened 2026-06-23
Log a warning when a replica idenity column changes

ilidemi:
Works for replica identity `default` and `index`. Intentionally excluding `full`. Handles the most common case but given false positives/negatives keeping it to internal logs.

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -513,6 +513,7 @@ func PullCdcRecords[Items model.Items](
 	)
 
 	records := req.RecordStream
+	warnedReplIdentTables := make(map[string]struct{})
 	var totalRecords int64
 	var fetchedBytes, totalFetchedBytes, allFetchedBytes atomic.Int64
 	// clientXLogPos is the last checkpoint id, we need to ack that we have processed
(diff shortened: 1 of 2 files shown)


jgao54 (approved):

Jeremyyang920 (approved):

flow/connectors/postgres/cdc.go
```diff
@@ -1082,6 +1085,20 @@ func processUpdateMessage[Items model.Items](
...
 
+	// By default, replica identity is the same as source PK/destination ordering key,
+	// so when it changes the destination can see duplicates or missed deletes.
+	// UpdateMessageTupleTypeKey fires when any replica identity column changes; if a replica identity is a superset
+	// of the dest ordering key (much rarer), the change to the extra columns is harmless and the warning doesn't apply.
+	// Replica identity full (OldTupleType=UpdateMessageTupleTypeOld) would require us to compare column values
+	// ourselves which we're avoiding for perf reasons, so absence of warning doesn't guarantee there wasn't a change.
+	if msg.OldTupleType == pglogrepl.UpdateMessageTupleTypeKey {
+		if _, warned := warnedReplIdentTables[tableName]; !warned {
+			warnedReplIdentTables[tableName] = struct{}{}
```
jgao54, line 1096: nit: maybe relId instead of tableName (Just in case there are two tables with same name in two different schemas)

jgao54: er, nvm, it is already full qualified

PR #4484, opened 2026-06-24
ci: drive stable Docker/deploy builds off release publish

dtunikov:
## What

Switch the **Stable Docker images** and **Update Deployment tags** workflows from the `push: tags` trigger to `release: published`.

## Why

Tagging and publishing the GitHub Release were decoupled. Tags were pushed independently, so the release-drafter draft could sit unpublished even though Docker images were already built:
- `v0.36.28` sat as a draft for 4 days (images built 06-15, published 06-19).
- `v0.36.29` was never published — images built 06-19, draft only published manually on 06-23.

Triggering off `release: published` makes **publishing the draft the single source of truth**: GitHub natively creates the `vX.Y.Z` tag at the release's target commit, which fires both workflows.

## Notes

- `stable-docker.yml` needs no other change — `github.ref_name` resolves to the tag name on both event types, so `SHA_SHORT: stable-${{ github.ref_name }}` still yields `stable-vX.Y.Z` and `actions/checkout` checks out the tag automatically.
- `update-docker-compose-stable.yaml` keeps `workflow_dispatch`; its `releases/latest` lookup returns the just-published release.
- **Behavior change:** bare `git push origin vX.Y.Z` no longer triggers builds — you must publish a release.
- **GITHUB_TOKEN caveat:** `release: published` fires on human publishes (the current flow). If release-drafter is later set to auto-publish via the default `GITHUB_TOKEN`, that bot event would not trigger these downstream workflows (GitHub's recursion guard).

### DIFF
--- a/.github/workflows/stable-docker.yml
+++ b/.github/workflows/stable-docker.yml
@@ -16,6 +16,8 @@ jobs:
       packages: write
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          ref: ${{ github.event.release.tag_name }}
 
       - name: generate or hydrate protos
         uses: ./.github/actions/genprotos
@@ -36,5 +38,5 @@ jobs:
           files: ./docker-bake.hcl
           push: true
         env:
-          SHA_SHORT: stable-${{ github.ref_name }}
+          SHA_SHORT: stable-${{ github.event.release.tag_name }}
           TAG: latest-stable
--- a/.github/workflows/stable-docker.yml
+++ b/.github/workflows/stable-docker.yml
@@ -1,9 +1,9 @@
 name: Stable Docker images
 
 on:
-  push:
-    tags:
-      - "v[0-9]+.[0-9]+.[0-9]+"
+  release:
+    types:
+      - published
 
 jobs:
   docker-build:
--- a/.github/workflows/stable-docker.yml
+++ b/.github/workflows/stable-docker.yml
@@ -16,8 +16,6 @@ jobs:
       packages: write
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
-        with:
-          ref: ${{ github.event.release.tag_name }}
 
       - name: generate or hydrate protos
         uses: ./.github/actions/genprotos
--- a/.github/workflows/update-docker-compose-stable.yaml
+++ b/.github/workflows/update-docker-compose-stable.yaml
@@ -3,9 +3,9 @@ name: Update Deployment tags (Docker Compose and Helm Charts)
 on:
   workflow_dispatch:
     inputs: {}
-  push:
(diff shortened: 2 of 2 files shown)


.github/workflows/stable-docker.yml
masterashu: Since now its a release event and not a tag event, not sure what the value of [github.ref_name](https://docs.github.com/en/actions/reference/workflows-and-actions/variables) will be, it can possibly be branch name (`main`) or empty. 
Better to use `github.event.release.tag_name` (https://docs.github.com/en/webhooks/webhook-events-and-payloads#release)

.github/workflows/update-docker-compose-stable.yaml
masterashu: This can be skipped and we can just use `${{ github.event.release.tag_name }}`

masterashu (changes requested):
Since you're changin

.github/workflows/stable-docker.yml
dtunikov: Good catch — switched to `github.event.release.tag_name` for `SHA_SHORT`. Also pinned the `actions/checkout` step to `ref: ${{ github.event.release.tag_name }}` so the image is built from the exact released commit rather than whatever `github.ref`/HEAD resolves to. Done in 34a2d23b.

.github/workflows/update-docker-compose-stable.yaml
dtunikov: Updated to prefer `github.event.release.tag_name`. I kept the `releases/latest` API call as a fallback only, since this workflow also has a `workflow_dispatch` trigger where `github.event.release` is empty. Done in 34a2d23b.

.github/workflows/update-docker-compose-stable.yaml
```diff
@@ -25,10 +25,15 @@ jobs:
...
-          latest_tag="$(gh api \
-            -H "Accept: application/vnd.github+json" \
-            -H "X-GitHub-Api-Version: 2022-11-28" \
-            /repos/${{ github.repository }}/releases/latest | jq -r '.tag_name')"
+          # On a release event use the published tag directly; fall back to
+          # the latest release for manual workflow_dispatch runs.
+          latest_tag="${{ github.event.release.tag_name }}"
+          if [ -z "${latest_tag}" ]; then
+            latest_tag="$(gh api \
+              -H "Accept: application/vnd.github+json" \
```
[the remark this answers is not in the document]
masterashu: Don't need we need a failback api call

.github/workflows/stable-docker.yml
```diff
@@ -16,6 +16,8 @@ jobs:
       packages: write
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          ref: ${{ github.event.release.tag_name }}
```
masterashu: This is also not required.

masterashu (approved):

PR #4487, opened 2026-06-24
mongo: partition snapshots for numeric and string _id collections

andreyzhelnin-st:
## Summary

Extends the MongoDB connector so initial snapshots **parallelize for numeric and string `_id` collections**, not just ObjectID. Previously any non-ObjectID `_id` fell back to a single serial full-table partition, making large naturally-keyed collections (app-store numeric IDs, package-name strings) too slow to snapshot inside the oplog retention window.

Based on `v0.36.29` (our deployed release).
It improves initial sync (MongoDb -> Clickhouse from 10 hours to 3 hours on 8mln collection).

## Approach

`GetQRepPartitions` dispatches on the collection's min/max `_id` BSON type:

- **ObjectID** — unchanged (uniform division of the 12-byte keyspace).
- **Int32/Int64** — uniform min/max division, reusing `utils.PartitionHelper` to emit non-overlapping `IntPartitionRange` partitions (`$gte`/`$lte`).
- **String** — sampling-based quantile boundaries via `$sample` (honours read preference), anchored by the real min/max, emitted as a new `StringPartitionRange` proto type with contiguous half-open ranges (`$gte`/`$lt`).
- **Mixed-type / Double / empty** — safe full-table fallback.

## Changes

- `protos/flow.proto` — add `StringPartitionRange` + `string_range` oneof field.
- `connectors/mongo/qrep_partition.go` — type dispatcher; numeric + string builders; pure, unit-tested `computeStringBoundaries`.
- `connectors/mongo/qrep.go` — `IntRange`/`StringRange` cases in `toRangeFilter`.
- `connectors/utils/monitoring/monitoring.go` — handle `StringRange` in `addPartitionToQRepRun` (avoids the `unknown range type` error path).

## Testing

- `go build ./...`, `go vet` clean.
- New unit tests for `computeStringBoundaries` (even/clustered/sparse/duplicate/out-of-range/empty + contiguity), `toRangeFilter` (Int/String/ObjectID/unsupported), and `_id` type detection — all passing.

## Notes / limitations

- Numeric uses uniform value division; monitor partition balance on first run for sparse ID spaces.
- Double `_id` and mixed-type collections fall back to full-table by design.
- Requires regenerating protobuf bindings (`./generate-protos.sh`) before building images, as `flow/generated` is gitignored.


flow/connectors/clickhouse/cdc.go
ilidemi, line 1: I see this is already included in #4490, please keep the changes separate

flow/connectors/mongo/qrep_partition.go
```diff
@@ -88,43 +147,205 @@ func (c *MongoConnector) minMaxPartitions(
...
+		sampleSize = stringSampleMaxSize
 	}
 
-	maxRaw, err := findBoundaryID(ctx, collection, Upper, readPreference)
+	aggCmd := bson.D{
+		{Key: "aggregate", Value: collection.Name()},
+		{Key: "pipeline", Value: bson.A{
+			bson.D{{Key: "$sample", Value: bson.D{{Key: "size", Value: int32(sampleSize)}}}},
+			bson.D{{Key: "$project", Value: bson.D{{Key: DefaultDocumentKeyColumnName, Value: 1}}}},
+		}},
```
jgao54, line 234:
```suggestion
{Key: "pipeline", Value: bson.A{
    bson.D{{Key: "$sample", Value: bson.D{{Key: "size", Value: int32(sampleSize)}}}},
    bson.D{{Key: "$project", Value: bson.D{{Key: DefaultDocumentKeyColumnName, Value: 1}}}},
    bson.D{{Key: "$sort", Value: bson.D{{Key: DefaultDocumentKeyColumnName, Value: 1}}}}, // <- add
}},
```
and drop
```
slices.Sort(interior)
```

This avoids needing to sort the sample records in Go later on and uses Mongo's collation setting instead; so the partitions are guaranteed to be monotonically increasing

protos/flow.proto
```diff
@@ -324,6 +324,11 @@ message ObjectIdPartitionRange {
   string end = 2;
 }
 
+message StringPartitionRange {
+  string start = 1;
+  string end = 2;
+}
```
jgao54, line 334: we are correctly working on StringPartitionRange for MySQL and decide to have StringPartitionRange introduced an `end_inclusive` boolean flag for String (see https://github.com/PeerDB-io/peerdb/pull/4482), bringing in a small optimization where we don't have to reserve one of the partitions for the last row/document.

flow/connectors/mongo/qrep.go
```diff
@@ -234,6 +234,26 @@ func toRangeFilter(watermarkColumn string, partitionRange *protos.PartitionRange
...
+	case *protos.PartitionRange_IntRange:
+		// Numeric _id: inclusive range. PartitionHelper builds non-overlapping
+		// [start, end] integer ranges. Mongo compares int32/int64 numerically, so
+		// int64 bounds match int32-typed _ids.
+		return bson.D{
+			bson.E{Key: watermarkColumn, Value: bson.D{
+				bson.E{Key: "$gte", Value: r.IntRange.Start},
+				bson.E{Key: "$lte", Value: r.IntRange.End},
+			}},
+		}, nil
```
jgao54, line 238: for mongo, using inclusive-inclusive int range [start, end] is prone to missing data in the rare scenario where user has doubles mixed in with integers. For example, if the `_id`s are 1, 2, 2.5, 3, 4; min/max will show up as 1 and 4, ended up with ranges [1, 2], [3, 4], we'd miss the 2.5 document.

we could introduce a `NumericPartitionRange` and use inclusive-exclusive [start, end),  similar to `StringPartitionRange`, to avoid any gaps.

(While Mongo sorts by type so we don't have to worry about this for ObjectIDs, all numeric types are in the same type bucket)

jgao54 (commented):
Thank you for your contribution. Using sampling for string is a great strategy for MongoDB.

Could you also add e2e tests in`flow/e2e/mongo_test.go`, similar to `Test_Simple_Flow_Partitioned`, but for string and numeric _ids for full test coverage.

flow/connectors/mongo/qrep_partition.go
```diff
@@ -88,43 +147,205 @@ func (c *MongoConnector) minMaxPartitions(
...
-) (bson.ObjectID, bson.ObjectID, bool, error) {
-	minRaw, err := findBoundaryID(ctx, collection, Lower, readPreference)
+	minVal string,
+	maxVal string,
+	numPartitions int64,
+) ([]*protos.QRepPartition, error) {
+	if minVal >= maxVal {
+		c.logger.Info("[mongo] string min/max range is non-positive, falling back to full table partition")
+		return fullTablePartitions(), nil
+	}
```
jgao54, line 175: this comparison uses golang instead of db collation, so is prone to false-positive. I think we can trust the db-fetched min/max using _its_ ordering, and just check for equality here.

flow/connectors/mongo/qrep_partition.go
```diff
@@ -16,37 +17,95 @@ import (
...
+// partition use uniform half-open [start, end) ($gte/$lt) semantics while still
+// capturing the maximum document in the last partition.
+const stringPartitionUpperSentinel = "\U0010FFFF"
+
+const (
+	// stringSampleOversample draws more samples than partitions so quantile
+	// boundaries are well distributed even with clustered keys.
+	stringSampleOversample = 20
+	// stringSampleMaxSize caps sampling cost on very large collections.
+	stringSampleMaxSize = 100000
```
jgao54, line 24: today we limit numPartitions to 1000, so technically the max possible sample size is 20*1000=20000; we'll never hit this limit, but fine to keep as an upper bound

flow/connectors/mongo/qrep_partition.go
andreyzhelnin-st: Done

andreyzhelnin-st:
@jgao54 @ilidemi Could you please check the updated code, I have addressed comments.

flow/connectors/mongo/qrep_partition.go
```diff
@@ -150,6 +142,12 @@ func (c *MongoConnector) objectIDPartitions(
 // numericPartitions divides a numeric (int32/int64) _id range uniformly, reusing
 // the shared PartitionHelper which builds non-overlapping IntPartitionRange
 // partitions with tested +1 boundary semantics covering [min, max] inclusive.
+//
+// Limitation: IntPartitionRange uses inclusive-inclusive [start, end] boundaries,
+// so documents with double _id values that fall between two adjacent integer
+// boundaries (e.g. _id=2.5 between [1,2] and [3,4]) would be missed. This is an
+// edge case for collections with mixed int/double _ids; a follow-up will introduce
+// a NumericPartitionRange with half-open [start, end) semantics to address it.
```
jgao54, line 138: 👍

jgao54:
CI is failing and the broken test can be deleted as it no longer applies: https://github.com/PeerDB-io/peerdb/actions/runs/28171385862/job/83568078090?pr=4487#step:31:1389

Also could you rebase your change on top of the latest main. Some of the other errors may be due to the branch base being too old. It will also make the diff clearer. 

We have to hold off merging this PR until https://github.com/PeerDB-io/peerdb/pull/4501 is merged. String boundaries may contain PII data so for privacy/security reasons we can't store arbitrary string ranges as is. 

Added a few nit comments, PR looks good overall!

flow/connectors/mongo/qrep_partition.go
```diff
@@ -88,43 +139,219 @@ func (c *MongoConnector) minMaxPartitions(
...
+	} else {
+		picked = make([]string, 0, desiredBoundaries)
+		for i := int64(1); i <= desiredBoundaries; i++ {
+			idx := i * int64(len(interior)) / numPartitions
+			if idx >= int64(len(interior)) {
+				idx = int64(len(interior)) - 1
+			}
+			picked = append(picked, interior[idx])
+		}
+		picked = slices.Compact(picked) // drop consecutive duplicate quantiles
```
jgao54: nit: this is not needed because of `seen` already dedupes

flow/connectors/mongo/qrep_partition.go
```diff
@@ -16,37 +17,87 @@ import (
...
+// fullTablePartitions returns the single-partition full-table fallback used when a
+// collection cannot be partitioned (unsupported/mixed _id type, empty collection,
+// or degenerate min/max range).
+func fullTablePartitions() []*protos.QRepPartition {
+	return []*protos.QRepPartition{{
 		PartitionId:        utils.FullTablePartitionID,
 		Range:              nil,
 		FullTablePartition: true,
 	}}
+}
```
jgao54: nit: dupe of `utils.FullTablePartition()`

flow/connectors/mongo/qrep_partition.go
jgao54: i see that it's there in case of upper bound reached:
```
if idx >= int64(len(interior)) {
	idx = int64(len(interior)) - 1
}
```
I don't think this is reachable today given len(interior) > desiredBoundaries and i <= numPartitions. but doesn't hurt to keep the code defensive.

flow/e2e/mongo_test.go
```diff
@@ -157,6 +157,120 @@ func (s MongoClickhouseSuite) Test_Simple_Flow_Partitioned() {
...
+
+	EnvWaitForEqualTablesWithNames(env, s, "initial load to match", srcTable, dstTable, "_id,doc")
+
+	catalogPool, err := internal.GetCatalogConnectionPoolFromEnv(t.Context())
+	require.NoError(t, err)
+	var partitionCount int
+	require.NoError(t, catalogPool.QueryRow(t.Context(),
+		`SELECT COUNT(*) FROM peerdb_stats.qrep_partitions WHERE parent_mirror_name = $1`,
+		flowConnConfig.FlowJobName).Scan(&partitionCount))
+	require.Greater(t, partitionCount, 1, "expected multiple partitions for 100 string-id rows")
```
jgao54: nit: could we use exact partition count here and in the next test as a sanity check. given SnapshotNumRowsPerPartition = 10, this should be 100/10=10.

andreyzhelnin-st:
@jgao54 Addressed comments & rebased
Also `go test -count=1 -v -timeout 30m -run 'TestMongoClickhouseSuite' ./e2e/` passes on my laptop.

jgao54 (approved):

PR #4488, opened 2026-06-24
observe mysql/postgres column type changes

dtunikov:
add metric to count column type change events for MySQL and postgres

### DIFF
--- a/flow/connectors/mysql/cdc.go
+++ b/flow/connectors/mysql/cdc.go
@@ -20,6 +20,7 @@ import (
 	"github.com/pingcap/tidb/pkg/parser/ast"
 	_ "github.com/pingcap/tidb/pkg/types/parser_driver"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/text/encoding"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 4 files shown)


flow/connectors/mysql/cdc.go
```diff
@@ -861,18 +871,33 @@ func (c *MySqlConnector) processAlterTableQuery(ctx context.Context, catalogPool
...
+				qkind, err := QkindFromMysqlColumnType(col.Tp.InfoSchemaStr(), binlogRowMetadataSupported, mirrorVersion)
+				if err != nil {
+					return err
+				}
+
+				// If the column already exists, this is a MODIFY COLUMN that may change
+				// its type rather than an ADD COLUMN. Record the type change for telemetry;
+				// we don't propagate column type changes yet.
+				if oldType, exists := existingColTypes[col.Name.OrigColName()]; exists {
+					if oldType != string(qkind) {
```
dtunikov: so, this might not always work as we want
claude noticed that we map some types differently for DDL and binlog paths 
the `oldType` comes from initial DDL types mapping (QkindFromMysqlColumnType)
and `qkind` comes from binlog types mapping (qkindFromMysqlType)
as an example DDL gives boolean for TINYINT(1), but binlog path gives QValueKindUInt8/QValueKindInt8

flow/otel_metrics/attributes.go
```diff
@@ -46,6 +46,9 @@ const (
 	EndBatchIDKey              = "endBatchID"
 	ResumeTokenKey             = "resumeToken"
 	GtidKey                    = "gtid"
+	TypeChangeSourceKey        = "source"
```
ilidemi: We have a SourcePeerType already stamped for this one

ilidemi: Can add a source=`ddl`/`event_metadata`

ilidemi (approved):

PR #4494, opened 2026-06-24
Optimize encoding maps

ilidemi:
Follow-up to #4436

Optimize building two extra maps per event on go-mysql side when re-encoding won't be needed (common case)

### DIFF
--- a/flow/connectors/mysql/charset.go
+++ b/flow/connectors/mysql/charset.go
@@ -133,25 +133,26 @@ func (c *MySqlConnector) tableMapColumnEncodings(
 	setMap map[int][]string,
 	otelManager *otel_metrics.OtelManager,
 ) ([]encoding.Encoding, error) {
-	var colEncodings []encoding.Encoding
-	if err := c.addTableMapColumnEncodings(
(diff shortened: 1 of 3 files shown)


flow/connectors/mysql/charset.go
```diff
@@ -118,6 +126,149 @@ func (c *MySqlConnector) collationEncoding(
...
+func (c *MySqlConnector) addTableMapColumnEncodings(
+	ctx context.Context,
+	otelManager *otel_metrics.OtelManager,
+	tableMap *replication.TableMapEvent,
+	collationType tableMapCollationType,
+	defaultCharset []uint64,
+	columnCharset []uint64,
+	enumMap map[int][]string,
+	setMap map[int][]string,
+	colEncodings *[]encoding.Encoding,
```
dtunikov: nit: I think standart go way would be to return []encoding.Encoding  instead of passing *[]encoding.Encoding (similar to how append function works)

dtunikov (approved):

flow/connectors/mysql/charset.go
ilidemi: nice, agreed

PR #4495, opened 2026-06-25
DBI-609: Add IAM auth support for CDC pipes with AWS Auruora'

masterashu:


Jeremyyang920:
Do we also need to update something for MySQL RDS/Auoura?

masterashu:
> Do we also need to update something for MySQL RDS/Auoura?

Not required, MySQL cdc was always supported using AWS IAM

flow/connectors/utils/rds.go
```diff
@@ -105,6 +105,24 @@ func GetRDSToken(ctx context.Context, connConfig RDSConnectionConfig, rdsAuth *R
...
+// ForceRefreshToken bypasses the cache TTL and immediately fetches a new RDS IAM token.
+// Use this when a connection break requires a guaranteed-fresh token for reconnect.
+func (r *RDSAuth) ForceRefreshToken(
+	ctx context.Context,
+	connConfig RDSConnectionConfig,
+	connectorName string,
+) (string, error) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	token, err := buildRdsToken(ctx, connConfig, BuildPeerAWSCredentials(r.AwsAuthConfig), connectorName)
```
dtunikov, line 102: would it make sense to call `ForceRefreshToken` from `GetRDSToken` to avoid duplication of lock/cache behavior?

flow/connectors/postgres/postgres_source.go
```diff
@@ -114,6 +114,22 @@ func (c *PostgresConnector) CreateReplConn(ctx context.Context, env map[string]s
 }
 
 func (c *PostgresConnector) SetupReplConn(ctx context.Context, env map[string]string) error {
+	// For IAM-authenticated connections, force a fresh token before connecting.
+	// This guarantees each activity retry uses a new token rather than a potentially
+	// stale cached one (cache TTL 10 min; token validity 15 min).
+	if c.rdsAuth != nil {
+		host := c.Config.Host
+		if c.Config.TlsHost != "" {
+			host = c.Config.TlsHost
```
dtunikov, line 123: Hm, maybe we could pass RDSConnectionConfig into RDSAuth constructor (when it gets initialized in `newPostgresConnector`)? This RDSAuth anyways seems to be scoped to connector instance.
Then we wouldn't need to pass host/port/user in each method and duplicate things like if tlsHost != "" -> host = tlsHost (we already do it inside `NewPostgresConnFromConfig`).

pfcoperez (approved):
LGTM, do we want to add an input in PeerDB UI to set this too?

masterashu:
> LGTM, do we want to add an input in PeerDB UI to set this too?

Nothing in the peerdb ui which prevented us from creating the mirror, since in peerdb ui you create the peers separately.

flow/connectors/utils/rds.go
masterashu: Yeah, correct

flow/connectors/postgres/postgres_source.go
masterashu: its was touching a lot of places so moved it into separate pr: https://github.com/PeerDB-io/peerdb/pull/4526

PR #4497, opened 2026-06-25
add online_schema_migrations metric

dtunikov:
emit metric for gh-ost MySQL migrations to track it's usage


flow/otel_metrics/attributes.go
```diff
@@ -71,3 +72,9 @@ const (
 	SourceEventTypeDDL           = "ddl"
 	SourceEventTypeEventMetadata = "eventMetadata"
 )
+
+const (
+	OnlineSchemaMigrationToolGhOst = "gh-ost"
+	OnlineSchemaMigrationToolPtOsc = "pt-online-schema-change"
+	OnlineSchemaMigrationToolOther = "other"
```
ilidemi: These can live in the connector package, we don't anticipate any other connector using them

flow/connectors/mysql/cdc_test.go
```diff
@@ -89,6 +96,155 @@ func TestParseSQLParsesTrailingNull(t *testing.T) {
...
+	// the rename that lands on the tracked table is the ghost table swap
+	swap := rename.TableToTables[1]
+	require.Equal(t, "users", swap.NewTable.Name.String())
+	require.Equal(t, "_users_gho", swap.OldTable.Name.String())
+	require.Equal(t, otel_metrics.OnlineSchemaMigrationToolGhOst,
+		classifyOnlineSchemaMigrationTool(swap.OldTable.Name.String(), swap.NewTable.Name.String()))
+}
+
+// newMetricsTestOtelManager builds an OtelManager backed by a ManualReader so tests
+// can collect and assert recorded metric values without a real exporter.
```
ilidemi, line 205: First one we have with metrics consumption, very cool. On the fence if we need it for this specific metric but there are definitely some critical ones that aren't tested

ilidemi (approved):

flow/connectors/mysql/cdc_test.go
dtunikov: yeah, this metric is not so critical
but it was the easiest way to exercise `c.processRenameTableQuery` method without an e2e test that renames a table..
just wanted to make sure that this code path is covered

PR #4498, opened 2026-06-25
feat(mysql): add counter metric for parseSQL errors

dtunikov:
## Summary

Adds a counter metric tracking failures parsing MySQL `QueryEvent` SQL on the CDC path.

- New `parse_sql_errors` Int64 counter in `otel_metrics`
- Incremented in the `parseSQL` error branch in `flow/connectors/mysql/cdc.go`
- Classify known error patterns and avoid reporting failures for such

Also partially resolves DBI-860.

The counter is intentionally **label-free** to avoid high cardinality from the varied SQL queries / error messages that can fail to parse.


ilidemi (approved):

flow/connectors/mysql/query_parser.go
```diff
@@ -0,0 +1,95 @@
...
+// in PullRecords). Every other statement — CREATE/DROP TABLE, any DATABASE / SCHEMA /
+// INDEX / VIEW / stored-routine / trigger / event / user DDL, SET, XA, GRANT/REVOKE,
+// etc. — is ignored even when it parses cleanly, so a parse failure on it cannot cause
+// schema divergence. The TiDB parser routinely rejects such statements, which RDS/
+// MariaDB emit constantly (e.g. SET STATEMENT ... FOR ... heartbeats), so they are
+// the bulk of the noise we want to drop.
+//
+// Only ALTER TABLE / RENAME TABLE failures are reported. If the handler ever starts
+// acting on more statement kinds, this allowlist must grow to match.
+func isBenignUnparsedStatement(query string) bool {
```
dtunikov: i think this approach is more reliable than regex, but happy to reconsider if you think otherwise

flow/connectors/mysql/query_parser.go
```diff
@@ -0,0 +1,95 @@
...
+	kw := leadingKeywords(query, 20)
+	if len(kw) == 0 {
+		return true
+	}
+
+	switch kw[0] {
+	case "ALTER", "RENAME":
+		// ALTER TABLE and RENAME TABLE are processed; ALTER/RENAME of any other object
+		// (DATABASE, SCHEMA, INDEX, VIEW, EVENT, USER, ...) is not.
+		return firstObjectKeyword(kw[1:]) != "TABLE"
```
ilidemi: We've seen `ALTER TABLE ... ADD UNIQUE INDEX ...` that TIDB choked on (same log link) and it's also benign

flow/connectors/mysql/query_parser.go
```diff
@@ -0,0 +1,95 @@
...
+// procedure body. Used to classify statements that failed to parse, where the AST
+// is unavailable and only the raw query text remains.
+func leadingKeywords(query string, limit int) []string {
+	out := make([]string, 0, limit)
+	for i, n := 0, len(query); i < n && len(out) < limit; {
+		switch c := query[i]; {
+		case c == ' ' || c == '\t' || c == '\n' || c == '\r':
+			i++
+		case c == '/' && i+1 < n && query[i+1] == '*':
+			end := strings.Index(query[i+2:], "*/")
```
ilidemi, line 39: I didn't realize how easy and small lexing is here, we might be able to get away with it.
For comments, there's also [executable](https://dev.mysql.com/doc/refman/9.7/en/comments.html#:~:text=MySQL%20Server%20supports%20certain%20variants%20of%20C%2Dstyle%20comments) [comments](https://mariadb.com/docs/server/reference/sql-statements/comment-syntax#executable-comments):
```
/*! not a comment */
/*!55000 not a comment */
/*!80000 not a comment in mysql but a comment in maria */
/*M! not a comment in maria but a comment in mysql */
/*M!80000 not a comment in maria but a comment in mysql */
```

"Adversarial" example for our scenario:
```
ALTER /*! TABLE db.t ALTER */ INDEX idx_c INVISIBLE
```

Closer to home:
```
ALTER TABLE db.t
    /*! ADD COLUMN mysql_only INT, */
    DROP COLUMN old_col;
```

Probably don't need to go all the way into version comparison but just handling them and distinguishing maria might be ok

flow/connectors/mysql/query_parser.go
ilidemi: Also Maria has `RENAME TABLES` ([synonym](https://mariadb.com/docs/server/reference/sql-statements/data-definition/rename-table#renaming-multiple-tables-at-once))

flow/connectors/mysql/query_parser_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsBenignUnparsedStatement(t *testing.T) {
+	for _, tc := range []struct {
+		name  string
+		query string
+		want  bool
+	}{
+		// --- benign: MariaDB/RDS "SET STATEMENT ... FOR ..." (the dominant prod noise) ---
```
ilidemi, line 16: Turns out SET STATEMENT can wrap an arbitrary query and we have some real (non-heartbeat) ones in logs
https://mariadb.com/docs/server/reference/sql-statements/administrative-sql-statements/set-commands/set-statement
Doesn't look impossible to parse though

flow/connectors/mysql/query_parser.go
```diff
@@ -0,0 +1,95 @@
...
+// leadingKeywords returns up to limit uppercased word tokens from the start of a
+// SQL statement, skipping leading/inline whitespace and -- , # and /* */ comments.
+// It only has to look at the head of the statement, so it never scans a whole
+// procedure body. Used to classify statements that failed to parse, where the AST
+// is unavailable and only the raw query text remains.
+func leadingKeywords(query string, limit int) []string {
+	out := make([]string, 0, limit)
+	for i, n := 0, len(query); i < n && len(out) < limit; {
+		switch c := query[i]; {
+		case c == ' ' || c == '\t' || c == '\n' || c == '\r':
```
ilidemi: nit: `\f` `\v`

flow/connectors/mysql/query_parser.go
```diff
@@ -0,0 +1,95 @@
...
+			i++
+		case c == '/' && i+1 < n && query[i+1] == '*':
+			end := strings.Index(query[i+2:], "*/")
+			if end < 0 {
+				return out
+			}
+			i += 2 + end + 2
+		case (c == '-' && i+1 < n && query[i+1] == '-') || c == '#':
+			nl := strings.IndexByte(query[i:], '\n')
```
ilidemi: nit: `\r`

flow/connectors/mysql/cdc.go
ilidemi, line 604: Not related to this change but looked into the logs and the warnings we get get serialized as `[{}]` because they look like structs to json
```suggestion
			if len(warns) > 0 {
				warnStrs := make([]string, len(warns))
				for i, w := range warns {
					warnStrs[i] = w.Error()
				}
				c.logger.Warn("processing QueryEvent with logged warnings", slog.Any("warns", warnStrs))
			}
```

ilidemi: Would be pretty fragile to have the two classifications completely separate in their own files. Best I could come up with in a few mins:
```
// -- query_parser.go

type ddlKind int

const (
	ddlKindIgnored ddlKind = iota
	ddlKindAlterTable
	ddlKindRenameTable
)

func classifyUnparsedStatement(query string) ddlKind {
	kw := leadingKeywords(query, 20)
	if len(kw) == 0 {
		return ddlKindIgnored
	}

	switch kw[0] {
	case "ALTER":
		if firstObjectKeyword(kw[1:]) == "TABLE" {
			return ddlKindAlterTable
		}
		return ddlKindIgnored
	case "RENAME":
		if firstObjectKeyword(kw[1:]) == "TABLE" {
			return ddlKindRenameTable
		}
		return ddlKindIgnored
	default:
		// CREATE/DROP (of any object, including TABLE) and everything else are not processed.
		return ddlKindIgnored
	}
}

// Explaining comment
func classifyParsedStatement(stmt ast.StmtNode) (ddlKind, *ast.AlterTableStmt, *ast.RenameTableStmt) {
	switch s := stmt.(type) {
	case *ast.AlterTableStmt:
		return ddlKindAlterTable, s, nil
	case *ast.RenameTableStmt:
		return ddlKindRenameTable, nil, s
	default:
		return ddlKindIgnored, nil, nil
	}
}

// -- cdc.go
	if err != nil {
		if classifyUnparsedStatement(string(ev.Query)) == ddlKindIgnored {
			...
		}
		...
	}
	...
	for _, stmt := range stmts {
		kind, alterStmt, renameStmt := classifyParsedStatement(stmt)
		switch kind {
		case ddlKindAlterTable:
			if err := c.processAlterTableQuery(..., alterStmt, ...); err != nil {
				return fmt.Errorf(...)
			}
		case ddlKindRenameTable:
			c.processRenameTableQuery(..., renameStmt, ...)
		case ddlKindIgnored:
		default:
			return fmt.Errorf("unknown stmt kind: %v", kind)
		}
	}
```
(multi-return kinda works for two but at 3/4 we'd look for another option)

flow/connectors/mysql/query_parser.go
dtunikov: yeah, thanks for pointing it out!
added support for executable comments

flow/connectors/mysql/query_parser.go
dtunikov: yeah, you're right, thanks for pointing it out!

flow/connectors/mysql/query_parser_test.go
dtunikov: yeah, makes sense
added logic to strip "set" -> "statement" -> "for" from the list of keywords

flow/connectors/mysql/query_parser.go
```diff
@@ -0,0 +1,255 @@
...
+// its leading keywords, the AST being unavailable. It mirrors classifyParsedStatement:
+// ddlKindIgnored means PeerDB would not have acted on the statement even had it parsed,
+// so the parse failure is benign noise; any other kind means a statement we care about
+// failed to parse and the failure should be reported.
+//
+// The TiDB parser routinely rejects statements RDS/MariaDB emit constantly (e.g.
+// SET STATEMENT ... FOR ... heartbeats, stored-routine bodies, MariaDB-only DDL),
+// which are the bulk of the noise this drops.
+func classifyUnparsedStatement(query string, isMariaDb bool) ddlKind {
+	kw := stripSetStatementPrefix(leadingKeywords(query, 24, isMariaDb))
```
ilidemi:
```suggestion
	kw := stripSetStatementPrefix(leadingKeywords(query, 100, isMariaDb))
```
just in case, especially with the `set statement` sequences

flow/connectors/mysql/query_parser.go
```diff
@@ -0,0 +1,255 @@
...
+	return ""
+}
+
+// indexConstraintKeywords are the keywords that, immediately following ADD or DROP
+// in an ALTER TABLE spec, name an index, key, or constraint rather than a column,
+// e.g. ADD UNIQUE INDEX, ADD FULLTEXT KEY, ADD PRIMARY KEY, ADD CONSTRAINT,
+// ADD FOREIGN KEY, DROP INDEX, DROP PRIMARY KEY.
+var indexConstraintKeywords = map[string]struct{}{
+	"INDEX": {}, "KEY": {}, "UNIQUE": {}, "FULLTEXT": {}, "SPATIAL": {},
+	"PRIMARY": {}, "FOREIGN": {}, "CONSTRAINT": {}, "CHECK": {},
```
ilidemi, line 158: There could be added columns named like this (in backticks/quotes). Fable tells me skipping quoted values from tokenization would do the job at +25 lines.

Query like
```
ALTER TABLE t ADD `key` VARCHAR(20)
```

ilidemi (approved):
Awesome job. There are always more corner cases but if it ever becomes important, I wonder if the next step would be an actual tidb replacement (hopefully not).

flow/connectors/mysql/query_parser.go
dtunikov: oh, yeah, very nice find!

PR #4500, opened 2026-06-26
Add a user-facing log when processing large txn

jgao54:
Add a user-facing log when a large txn result in a PullRecord sync interval to get extended.

Logged once per sync_interval, but throttled at 1 minute in case customer has a very short sync intervals that makes the log excessively noisy.

Fixes: DBI-681

### DIFF
--- a/flow/connectors/postgres/cdc.go
+++ b/flow/connectors/postgres/cdc.go
@@ -25,6 +25,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 	"go.temporal.io/sdk/log"
 
+	"github.com/PeerDB-io/peerdb/flow/alerting"
 	connmetadata "github.com/PeerDB-io/peerdb/flow/connectors/external_metadata"
 	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
 	"github.com/PeerDB-io/peerdb/flow/connectors/utils/monitoring"
@@ -59,6 +60,7 @@ type PostgresCDCSource struct {
 
 	// for storing schema delta audit logs to catalog
 	catalogPool                              shared.CatalogPool
+	alerter                                  *alerting.Alerter
 	otelManager                              *otel_metrics.OtelManager
 	hushWarnUnhandledMessageType             map[pglogrepl.MessageType]struct{}
 	hushWarnUnknownTableDetected             map[uint32]struct{}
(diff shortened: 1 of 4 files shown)


flow/connectors/postgres/cdc.go
```diff
@@ -679,11 +683,21 @@ func PullCdcRecords[Items model.Items](
...
 				} else {
+					accumulatedBytes := totalFetchedBytes.Load()
 					logger.Info("commit lock, waiting for commit to return records",
 						slog.Int64("records", totalRecords),
-						slog.Int64("bytes", totalFetchedBytes.Load()),
+						slog.Int64("bytes", accumulatedBytes),
 						slog.Int("channelLen", records.ChannelLen()),
 						slog.Float64("elapsedMinutes", time.Since(pullStart).Minutes()))
+
+					if time.Since(largeTxnLastLogged) > time.Minute {
```
ilidemi, line 693: This could also come up when there's a tiny tx right at the end of the batch, I've seen it extending the batch by <100ms. Maybe skip the first log?

flow/connectors/postgres/cdc.go
```diff
@@ -679,11 +683,21 @@ func PullCdcRecords[Items model.Items](
...
+						userMsg := fmt.Sprintf(
+							"Waiting for a large transaction to commit before syncing can continue (%d records, %s buffered so far).",
```
ilidemi: Have to be careful to avoid confusion with a long-running tx on the source. Maybe
```suggestion
							"Reading a large already-committed transaction off the replication slot; sync will continue once it's fully received (%d records, %s buffered so far)",
```

flow/connectors/postgres/cdc.go
```diff
@@ -679,11 +683,21 @@ func PullCdcRecords[Items model.Items](
...
+							totalRecords, utils.HumanReadableBytes(accumulatedBytes))
+						p.alerter.LogFlowInfo(ctx, p.flowJobName, userMsg)
```
ilidemi:
```suggestion
						p.alerter.LogFlowWarning(ctx, p.flowJobName, userMsg)
```
? Or depending on the time since batch was supposed to finish

ilidemi (approved):

flow/connectors/postgres/cdc.go
jgao54: good point, since commitLock is held we have already received at least a BEGIN record, so "waiting" is a bit ambiguous here.

flow/connectors/postgres/cdc.go
jgao54: this feels more informative than being a concern, but using warning is okay as well!

flow/connectors/postgres/cdc.go
jgao54: good idea

flow/connectors/postgres/cdc.go
jgao54: oh, our `alerter.LogFlowWarn` is designed to take in an error. So i'm inclined to keep as is instead of generating an error for logging purpose

ilidemi (approved):

PR #4501, opened 2026-06-26
treat string partition ranges as sensitive data

jgao54:
Before introducing parallel snapshotting for arbitrary string watermark column, in order to avoid sensitive data from being accidentally exposed in db/Temporal as plaintext, this PR will offload partition ranges to db (encrypted) before return the `GetQRepPartitions` Activity result, and then restored in the `ReplicateQRepPartitions` Activity, to avoid crossing Temporal boundary. Long term, offloading all partition ranges to DB is also ideal, in order to not having to worry about Temporal payload limit.

This PR introduces a new table `metadata_qrep_partition_ranges` for storing partition ranges. I also evaluated building on top of existing `peerdb_stats.qrep_partitions` but that table is intended for monitoring so adding operational logic to it is not ideal. This PR also makes sure that sensitive ranges are removed before writing to `peerdb_stats.qrep_partitions`, which are not being used afaik. 

Open to feedback on what we should deem as sensitive that should be encrypted (e.g. all string vs. non-UUIDs vs. only encrypt email-shaped string). Currently all string ranges are encrypted. 

Testing: add integration test coverage for offload/restore correctness logic; as well as e2e test for the workflow, and that table is cleanup when mirror is dropped. 

Fixes: DBI-869


flow/connectors/external_metadata/store.go
```diff
@@ -286,6 +288,148 @@ func (p *PostgresMetadata) IsQRepPartitionSynced(ctx context.Context, req *proto
...
+		if err != nil {
+			return fmt.Errorf("failed to marshal partition range: %w", err)
+		}
+		encrypted, err := key.Encrypt(payload)
+		if err != nil {
+			return fmt.Errorf("failed to encrypt partition range: %w", err)
+		}
+		insertRows = append(insertRows, []any{parentMirrorName, runUUID, partition.PartitionId, key.ID, encrypted})
+
+		partition.Range = &protos.PartitionRange{
```
ilidemi: Add an `Offloaded: true` somewhere? Would be ideal to also fail at consumption points if it's accidentally `true` but haven't checked how feasible that is.

ilidemi: Alternative: add offloaded_range to the oneof range proto and at least have to handle it in all the switches

flow/connectors/external_metadata/store.go
```diff
@@ -286,6 +288,148 @@ func (p *PostgresMetadata) IsQRepPartitionSynced(ctx context.Context, req *proto
...
+
+	return nil
+}
+
+func partitionsContainSensitiveRanges(partitions []*protos.QRepPartition) bool {
+	for _, partition := range partitions {
+		if partition == nil || partition.Range == nil {
+			continue
+		}
+		// String ranges may contain PII data like email, so deem it as sensitive.
```
ilidemi: I'm leaning towards doing this for all types, just because you can extract signal from things and we want to be able to snapshot all sorts of tables. Let's discuss with the team?

flow/connectors/external_metadata/store_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/internal"
+)
+
+func TestOffloadRestoreSensitivePartitionRanges(t *testing.T) {
+	const encKeyID = "test_enc_key"
+	t.Setenv("PEERDB_CURRENT_ENC_KEY_ID", encKeyID)
+	t.Setenv("PEERDB_ENC_KEYS", `[{"id":"`+encKeyID+`","value":"cGVlcmRiX2NpX3Rlc3RfZW5jX2tleV8zMl9ieXRlcyE="}]`)
```
ilidemi, line 20: Hilarious key :D

nexus/catalog/migrations/V54__qrep_partition_ranges.sql
```diff
@@ -0,0 +1,15 @@
+-- Stores QRep partition ranges offloaded to the catalog instead of being passed
+-- through Temporal. Today this is used only for sensitive partition ranges.
+-- The table name and columns are deliberately generic so a future change can
+-- offload all partition ranges to catalog to stay under Temporal's payload limit.
+CREATE TABLE IF NOT EXISTS metadata_qrep_partition_ranges (
```
ilidemi: Have you thought about limiting retention? I think something like indexed inserted_at and a scheduled task that wipes everything older than 30 days should be about it. Going too low would risk breaking unfortunate long snapshots or complicating the retention logic, blanket 30 seems like a reasonable default

flow/connectors/external_metadata/store.go
jgao54: discussed with team: conclusion is to encrypt this for all data types, as integer PKs may  potentially contain sensitive data as well like government-issued IDs; especially given that we also support secondary indices as partition keys.

nexus/catalog/migrations/V54__qrep_partition_ranges.sql
jgao54: Inserted_at makes a lot of sense.

flow/connectors/external_metadata/store.go
```diff
@@ -262,6 +264,15 @@ func (p *PostgresMetadata) FinishQRepPartition(
 	jobName string,
 	startTime time.Time,
 ) error {
+	if partition.RangeOffloaded {
```
jgao54, line 267: this was missed earlier but we also serialize the entire QRepPartition in metadata_qrep_partitions. 
so here we sanitize this as well.

jgao54:
addressed review feedback: 
- always encrypt non-nil ranges regardless of type
- added `range_offloaded` field to indicate whether a partition is offloaded
- update schema to include inserted_at timestamp

additional changes:
- make sure metadata_qrep_partitions also does not store serialized partition in plaintext
- add dynconf PEERDB_OFFLOAD_PARTITION_RANGES (allow disabling offload/encryption for OSS)
- update tests accordingly

what this PR doesn't cover:
- TTL implementation (there are a number of tables that we'd actually want to implement TTL for; so better to assess them holistically)
- support offload/encryption for stand-alone QRepFlowWorkflow (legacy feature)

flow/connectors/external_metadata/store.go
```diff
@@ -260,6 +264,15 @@ func (p *PostgresMetadata) FinishQRepPartition(
...
+		// The range was offloaded but has since been rehydrated, so drop it before persisting
+		// to metadata_qrep_partitions.
+		// Clone rather than mutate in place because this partition is shared with the pull
+		// goroutine, cloning keeps us from relying on implicit ordering to avoid a write race.
+		// TODO: drop the sync_partition column entirely; nothing reads it back.
+		partition = proto.Clone(partition).(*protos.QRepPartition)
```
ilidemi:
```suggestion
		partition = proto.CloneOf(partition)
```

ilidemi (approved):

PR #4503, opened 2026-06-26
refactor(mysql): move used_mysql_charsets metric into connector package

dtunikov:
## What

Moves the MySQL-only `used_mysql_charsets` counter out of the shared `flow/otel_metrics` package and into `flow/connectors/mysql`.

## Why

MySQL-specific metrics currently live in the common `otel_metrics` folder, which means adding/changing one touches a shared path and makes it hard to scope CI test selection by the path where the change happened. Keeping the metric definition in the connector package keeps such changes confined to `flow/connectors/mysql/`.

### DIFF
--- a/flow/connectors/mysql/cdc.go
+++ b/flow/connectors/mysql/cdc.go
@@ -432,6 +432,11 @@ func (c *MySqlConnector) PullRecords(
 ) error {
 	defer req.RecordStream.Close()
 
+	mysqlMetrics, err := newMetrics(otelManager)
+	if err != nil {
+		return err
+	}
+
 	sourceSchemaAsDestinationColumn, err := internal.PeerDBSourceSchemaAsDestinationColumn(ctx, req.Env)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


flow/connectors/mysql/cdc.go
```diff
@@ -438,6 +432,11 @@ func (c *MySqlConnector) PullRecords(
 ) error {
 	defer req.RecordStream.Close()
 
+	mysqlMetrics, err := newMetrics(otelManager)
```
dtunikov, line 435: same can be called on top of qrep if needed

pfcoperez (commented):
This is a new pattern for organizing metrics. The optimization to ease test selection upon code change kind of depends on the adoption in others part of the peerdb codebase.

I think we should document this in the README and point to this file as an example as well as opening tickets to follow up in other connectors. 

I also think we should consider if we have currently all the existing metrics described in otel_manager.go . If so, we are are not longer concentrating all metrics changes in a single file with your change.

This yields to the question: What's better, to describe all metrics in a single place or   distributed in their related components?

dtunikov:
> This is a new pattern for organizing metrics. The optimization to ease test selection upon code change kind of depends on the adoption in others part of the peerdb codebase.
> 
> I think we should document this in the README and point to this file as an example as well as opening tickets to follow up in other connectors.
> 
> I also think we should consider if we have currently all the existing metrics described in otel_manager.go . If so, we are are not longer concentrating all metrics changes in a single file with your change.
> 
> This yields to the question: What's better, to describe all metrics in a single place or distributed in their related components?

for the context - i'm not suggesting to merge it as is, just wanted to share a draft of how it could look like if we decide to make the code a bit more modular and maybe later get to the state where we wouldn't need to run pg/mongo/etc tests if we only changed mysql logic.

PR #4504, opened 2026-06-26
DBI-829: Expand tested MariaDB version range in CI

pfcoperez:
This PR expands the pseudo-DSL within flow.yml used to described test matrix cases by allowing an optional row field that overrides the  tag described by `version-configs` when present. e.g: 

- Traditional entry: `{pg: 18, mysql-flavor: 'maria-pos', mongo: '8.0', ch: 'latest'},`, falling back to using:

```yaml
              mysql-pos:
                img: 'mysql'
                tag: '5.7'
                env: ['MYSQL_ROOT_PASSWORD=cipass']
                parameters: ['--log_bin=mysql-bin', '--server-id=1', '--bind-address=::']
```

- Entry with parametrized tag:  `{pg: 18, mysql-flavor: 'maria-gtid', mysql-tag: '12-ubi', mongo: '8.0', ch: 'stable'},`

Then it uses this mechanism to add the cases for MariaDB 12 and 13:

```
          {pg: 18, mysql-flavor: 'maria-gtid', mysql-tag: '12-ubi', mongo: '8.0', ch: 'stable'},
          {pg: 18, mysql-flavor: 'maria-gtid', mysql-tag: '13-ubi', mongo: '8.0', ch: 'stable'},
```


.github/workflows/flow.yml
```diff
@@ -33,31 +33,46 @@ jobs:
...
+        # '12-ubi' -> 'mariadb:12-ubi'). The mysql-tag value also disambiguates the job name,
+        # artifact name and combination-id across those variants, so it must be filesystem-safe
+        # (no ':' or '/') — which is why the lts row uses the plain 'lts-ubi9' tag, not a digest.
         db-version: [
-          {pg: 16, mysql: 'mysql-gtid', mongo: '6.0', ch: 'lts'},
-          {pg: 17, mysql: 'mysql-pos', mongo: '7.0', ch: 'stable'},
-          {pg: 18, mysql: 'maria-gtid', mongo: '8.0', ch: 'latest'},
+          {pg: 16, mysql-flavor: 'mysql-gtid', mongo: '6.0', ch: 'lts'},
+          {pg: 17, mysql-flavor: 'mysql-pos', mongo: '7.0', ch: 'stable'},
+          {pg: 18, mysql-flavor: 'maria-pos', mongo: '8.0', ch: 'latest'},
```
dtunikov: hm
i think we don't allow mariadb with file pos, do we?

dtunikov: i mean, i see that we already have it in the code
but i'm not sure why

.github/workflows/flow.yml
```diff
@@ -33,31 +33,46 @@ jobs:
...
+          {pg: 18, mysql-flavor: 'maria-gtid', mysql-tag: 'lts-ubi9', mongo: '8.0', ch: 'stable'},
```
dtunikov: maybe stupid question, but
does it mean that we will run pg tests N times with different mysql tags? or do we have some guard later to prevent this?

ilidemi: Yeah, whyyyyyyy not do pg-mysql-maria-mongo-ch matrix, it's so easy to go 3-wide for everything

pfcoperez: That's right @dtunikov  that's what we have been doing so far but will change as soon as possible. The gold standard we want to achieve eventually is to list just (source, target) combinations thus removing wasteful runs. The linked ticket has a discussion about  this idea.

pfcoperez: > Yeah, whyyyyyyy not do pg-mysql-maria-mongo-ch matrix, it's so easy to go 3-wide for everything

maria+mysql in the same row moves us even further away from the gold standard goal. 

This PR starts introducing the split up between goal description and resource selection.

ilidemi: Is there a plan to merge the gold standard in a reasonable time?

Until then, it looks like this to me personally
image

pfcoperez: > Is there a plan to merge the gold standard in a reasonable time?

@ilidemi I think we should do these changes in incremental steps, specially when this one unblocks the needed description of jobs with different MariaDB versions without changing the existing strategy.

Yet, I am starting a separate PR where the requirement for (source, destination) combinations (like test selection) are satisfied. 

If that is finished in  a short time-box I'll just close this one. If things get hairy, I wouldn't block the expanding range of MariaDB versions for too long and merge this stop-gap solution.

pfcoperez: I personally find this analogy more accurate:

image

pfcoperez:
Superseded by https://github.com/PeerDB-io/peerdb/pull/4510

PR #4505, opened 2026-06-26
add more mysql/maria column type tests and support uuid/inet types

dtunikov:
- add more tests for mysql alter table (DBI-820)
- support UUID and INET MariaDB types (DBI-843)

Unfortunately, tidb parser doesn't recognize UUID/INET4/INET6 as they are mariadb specific
So, we can't handle `alter table add column smth UUID` without additional parsing.
But, at least we can handle these types on snapshot lvl now.


flow/connectors/mysql/cdc.go
```diff
@@ -1140,7 +1140,7 @@ func (c *MySqlConnector) processTableMapEventSchema(
 
 			if qkind, err := qkindFromMysqlType(
 				tableMap.ColumnType[idx], unsignedMap[idx], charset, req.InternalVersion,
-			); err == nil && qkind != types.QValueKind(fd.Type) {
+			); err == nil && qkind != types.QValueKind(fd.Type) && !isBinlogStringBackedType(types.QValueKind(fd.Type)) {
```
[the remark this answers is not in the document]
dtunikov: it's on purpose
because binlog gives us STRING for any of these types -> therefore we will always see type diff compared to original type

flow/connectors/mysql/cdc_test.go
```diff
@@ -162,6 +163,164 @@ func TestParseSQLParsesTrailingNull(t *testing.T) {
 	}
 }
 
+func TestParseSQLAlterTableAddColumnTypes(t *testing.T) {
+	// addColumnType parses `ALTER TABLE t ADD COLUMN c <colDef>` and returns the
+	// column's InfoSchemaStr, the same value processAlterTableQuery feeds to
+	// QkindFromMysqlColumnType.
+	addColumnType := func(t *testing.T, colDef string) string {
+		t.Helper()
+		stmts, warns, err := parseSQL(parser.New(), []byte("ALTER TABLE t ADD COLUMN c "+colDef))
```
dtunikov, line 183: we can also add some fuzzy tests for column names
though, i'd expect that it's covered on lib level

flow/connectors/mysql/cdc.go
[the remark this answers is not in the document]
ilidemi: Copilot is saying the exclusion will also exclude INTEGER + qkind UUID which we don't want, plus we know it's expected for Maria but not MySQL.

I would merge the equality checks into one helper that looks at two qkinds and the flavor and decides if a column type change should be reported.

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -536,6 +564,18 @@ func QValueFromMysqlRowEvent(
...
+			if err != nil {
+				return nil, err
+			}
+			return types.QValueUUID{Val: u}, nil
+		case types.QValueKindINET:
+			s, err := formatMariaDBInet(val)
+			if err != nil {
+				return nil, err
+			}
+			return types.QValueINET{Val: s}, nil
```
ilidemi: Each of these types should probably fall either []byte or string but not both, right? Or not necessarily?

flow/connectors/mysql/qvalue_convert_test.go
```diff
@@ -5,12 +5,77 @@ import (
...
+	got, err := decodeMariaDBUUID(stored)
+	require.NoError(t, err)
+	require.Equal(t, "6ccd780c-baba-1026-9564-5b8c656024db", got.String())
+	require.Equal(t, uuid.MustParse("6ccd780c-baba-1026-9564-5b8c656024db"), got)
+
+	_, err = decodeMariaDBUUID([]byte{0x00, 0x01})
+	require.Error(t, err)
+}
+
+func TestFormatMariaDBInet(t *testing.T) {
```
ilidemi: Let's stick to just e2e (but thorough) for both decode tests. Crafting bytes that Maria would send doesn't make sense to me when we have real Maria right there

flow/connectors/mysql/cdc_test.go
ilidemi: Yeah, for names let's trust the parser

flow/connectors/mysql/cdc_test.go
```diff
@@ -162,6 +163,164 @@ func TestParseSQLParsesTrailingNull(t *testing.T) {
...
+		col := alter.Specs[0].NewColumns[0]
+		require.NotNil(t, col.Tp)
+		return col.Tp.InfoSchemaStr()
+	}
+
+	for _, tc := range []struct {
+		name   string
+		colDef string
+		want   types.QValueKind
+	}{
```
ilidemi, line 204: Good selection so far, although would prefer a more complete coverage. From looking through mysql spec (didn't get to maria): 
- zerofill
- (M) for smallint/mediumint/int/integer/bigint
- float(m, d)
- float(p)
- double(m, d)
- double precision(m, d)
- unsigned for floats
- year(4)
- char/national char/nchar (without length)
- modifiers for ..char(N)/..text/enum/set types: charset, collate, binary (deprecated one)
- binary without length
- blob(M)

Didn't check what exactly comes through col.Tp.InfoSchemaStr() - it's a bit silly that we go string->tidb parse->slightly different string->our parse, but the snapshot path just gives us the last string for full info, and more of those cases will be there and we'll also need to add tests for those. So maybe the list of types can be somehow aligned between ddl add, binlog metadata add and snapshot.

Also we want to be comparing the complete FieldDescription here, with type modifier for decimals and nullability

flow/connectors/mysql/qvalue_convert.go
dtunikov: yes, had the same concern, but decided to follow existing pattern 
now checked in go-mysql code, it must be always "string" for these new types

flow/connectors/mysql/cdc_test.go
dtunikov: added all of the above + all custom mariadb types (mostly synonyms) I could find
i collected all maria types from their docs:
```
{
  "source": "https://mariadb.com/docs/server/reference/data-types",
  "categories": {
    "numeric": {
      "source": "https://mariadb.com/docs/server/reference/data-types/numeric-data-types",
      "types": [
        { "name": "TINYINT", "description": "Very small integer; 1 byte, -128 to 127 signed or 0 to 255 unsigned." },
        { "name": "SMALLINT", "description": "Small integer; 2 bytes, -32768 to 32767 signed or 0 to 65535 unsigned." },
        { "name": "MEDIUMINT", "description": "Medium-sized integer; 3 bytes, -8388608 to 8388607 signed or 0 to 16777215 unsigned." },
        { "name": "INT", "description": "Standard 4-byte integer; -2147483648 to 2147483647 signed or 0 to 4294967295 unsigned." },
        { "name": "BIGINT", "description": "Large integer; 8 bytes, -9223372036854775808 to 9223372036854775807 signed or 0 to 18446744073709551615 unsigned." },
        { "name": "DECIMAL", "description": "Fixed-point number with defined precision and scale (M≤65, D≤38)." },
        { "name": "NUMERIC", "description": "Synonym for DECIMAL; stores exact numeric data values." },
        { "name": "NUMBER", "description": "Oracle-compatible synonym for DECIMAL." },
        { "name": "FLOAT", "description": "Single-precision floating-point number; 4 bytes, approximate values." },
        { "name": "DOUBLE", "description": "Double-precision floating-point number; 8 bytes, approximate values." },
        { "name": "REAL", "description": "Synonym for DOUBLE in standard SQL mode." },
        { "name": "BIT", "description": "Bit-field type storing M bits per value (1 to 64 bits)." },
        { "name": "BOOL / BOOLEAN", "description": "Synonym for TINYINT(1); 0 is false, non-zero is true." },
        { "name": "Numeric aliases", "description": "Additional synonyms: INT1, INT2, INT3, INT4, INT8, FLOAT4, FLOAT8, FIXED, DEC, MIDDLEINT." }
      ]
    },
    "date_and_time": {
      "source": "https://mariadb.com/docs/server/reference/data-types/date-and-time-data-types",
      "types": [
        { "name": "DATE", "description": "Date in YYYY-MM-DD format, range 1000-01-01 to 9999-12-31." },
        { "name": "DATETIME", "description": "Combined date and time value." },
        { "name": "TIME", "description": "Time duration or time of day, range '-838:59:59' to '838:59:59' with optional microsecond precision." },
        { "name": "TIMESTAMP", "description": "Timestamp value with syntax, valid values, storage requirements, and range limits." },
        { "name": "YEAR", "description": "Year value in 2- or 4-digit format, 1901 to 2155 plus 0000." },
        { "name": "SQL_TSI_YEAR", "description": "Synonym for YEAR." }
      ]
    },
    "string": {
      "source": "https://mariadb.com/docs/server/reference/data-types/string-data-types",
      "types": [
        { "name": "BLOB", "description": "Variable-length binary large object storing up to 65,535 bytes." },
        { "name": "BINARY", "description": "Fixed-length binary string type with zero-byte padding." },
        { "name": "CHAR", "description": "Fixed-length character string type (0-255 characters)." },
        { "name": "CHARACTER", "description": "Equivalent to CHAR; stores fixed-length strings with space padding." },
        { "name": "CHAR BYTE", "description": "Binary variant storing fixed bytes with zero-padding." },
        { "name": "CHAR VARYING", "description": "Variable-length character strings." },
        { "name": "CLOB", "description": "Oracle mode alias for LONGTEXT." },
        { "name": "ENUM", "description": "String object with a value chosen from a predefined list." },
        { "name": "INET4", "description": "IPv4 address storage as 4-byte binary." },
        { "name": "INET6", "description": "IPv6 address storage as 16-byte binary." },
        { "name": "JSON", "description": "JSON document type with defined valid values, storage requirements, and range." },
        { "name": "LONG / LONG VARCHAR", "description": "Compatibility synonym for MEDIUMTEXT." },
        { "name": "LONGBLOB", "description": "Very large binary object up to 4GB." },
        { "name": "LONGTEXT", "description": "Text type with a 4GB (2^32-1) storage limit." },
        { "name": "MEDIUMBLOB", "description": "Medium-sized binary object up to 16MB." },
        { "name": "MEDIUMTEXT", "description": "Medium-sized character string up to 16MB." },
        { "name": "NATIONAL CHAR / VARCHAR variants", "description": "UTF-8 versions of standard character types." },
        { "name": "NCHAR variants", "description": "UTF-8 synonyms for fixed/variable character types." },
        { "name": "RAW", "description": "Oracle mode variable-length binary type." },
        { "name": "ROW", "description": "Data type used in stored programs for cursor rows." },
        { "name": "SET", "description": "Zero or more values from a predefined list." },
        { "name": "TEXT", "description": "Variable-length text with a 65,535 byte maximum." },
        { "name": "TINYBLOB", "description": "Up to 255 bytes of binary data." },
        { "name": "TINYTEXT", "description": "Up to 255 characters of text." },
        { "name": "UUID", "description": "128-bit storage-optimized type for UUIDs." },
        { "name": "VARBINARY", "description": "Variable-length binary strings." },
        { "name": "VARCHAR", "description": "Variable-length character string type with a specified maximum." },
        { "name": "VARCHAR2", "description": "Up to 65,535 variable-length characters." },
        { "name": "VARCHARACTER", "description": "Equivalent to VARCHAR." },
        { "name": "XMLTYPE", "description": "XML data storage." }
      ]
    }
  },
  "additional_features": {
    "source": "https://mariadb.com/docs/server/reference/data-types",
    "items": [
      { "name": "AUTO_INCREMENT", "description": "Attribute that automatically generates sequential integer values for a column." },
      { "name": "NULL Values", "description": "Representation and handling of NULL/missing values." },
      { "name": "SERIAL", "description": "Alias for BIGINT UNSIGNED NOT NULL AUTO_INCREMENT UNIQUE." },
      { "name": "TYPE OF", "description": "Declares a variable/column with the type of another column." },
      { "name": "ROW TYPE OF", "description": "Declares a row variable with the type of a table row or cursor." }
    ]
  }
}

```
and wrote a script to cross check with mysql/cdc_test.go file
we have a much better coverage now:
image

flow/connectors/mysql/cdc.go
```diff
@@ -1140,7 +1140,7 @@ func (c *MySqlConnector) processTableMapEventSchema(
 
 			if qkind, err := qkindFromMysqlType(
 				tableMap.ColumnType[idx], unsignedMap[idx], charset, req.InternalVersion,
-			); err == nil && qkind != types.QValueKind(fd.Type) {
+			); err == nil && shouldReportColumnTypeChange(types.QValueKind(fd.Type), qkind, c.config.Flavor) {
```
dtunikov, line 1166: there is one weird case I would like to highlight
so, as tidb parser can't handle some maria specific types like UUID/INET4/INET6
we can't process DDL statements properly for new columns
but the thing is that later it will come as a `TABLE_MAP_EVENT` with type `MYSQL_TYPE_STRING` and bytes charset :sweat_smile: 
and.. we would simply add this new UUID column with `bytes` type, which doesn't make much sense

ilidemi: We do what we can 🤷‍♂️ 
Need to document all of this in https://clickhouse.com/docs/integrations/clickpipes/mysql/datatypes though, and table resync as mitigation

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -214,6 +216,39 @@ func processTime(str string) (time.Duration, error) {
...
 
+// padTrimmedMariaDBBinary restores a fixed-width binary value that MariaDB
+// trimmed when binlog-packing it. MariaDB stores BINARY(N) and the native
+// uuid/inet4/inet6 types (all arriving as MYSQL_TYPE_STRING) with trailing
+// 0x00 bytes stripped, so a value such as uuid 123e…4000 loses its final byte
+// and an all-zero value (uuid 0000…0000, inet 0.0.0.0, inet6 ::) arrives with
+// zero length. Right-padding with zeros to the column's declared width
+// reconstructs the original value. width comes from binaryColumnLength, which
+// also distinguishes inet4 (4) from inet6 (16) when the bytes alone cannot.
+func padTrimmedMariaDBBinary(data []byte, width int) []byte {
```
ilidemi: This logic is the same as in 
```
	case string:
		switch qkind {
		case types.QValueKindBytes:
```
branch, doesn't seem to be specific to Maria. Merge into a shared helper?

flow/connectors/mysql/cdc_test.go
ilidemi: Great selection. Last thing, let's assert the complete FieldDescription here, with type modifier for decimals and nullability. Not sure if there's a simple way to do nullability all up - if it's possible to blanket-append NOT NULL at the end, the test execution could do it, if it's not, a few types that exercise the other branch should be fine.

ilidemi (approved):
Awesome job

PR #4516, opened 2026-06-30
DBI-829: MariaDB as iteration dimension

pfcoperez:
alternative draft to https://github.com/PeerDB-io/peerdb/pull/4504
TODO: Figure out the best way to run MySQL tests with maria flavor

PR #4517, opened 2026-06-30
Separate MySQL and MariaDB tests

ilidemi:
Changes:
* Incorporate #4516 making Maria a separate DB in the matrix
* Spawn out the maria-roots of e2e suites
* Introduce TestIntegraion.../mysql and .../mariadb naming pattern for integration tests, filter from tilt with a matcher. MySQL-only are TestMySQLOnlyIntegration... to avoid running them from maria targets as parent tests are first ran to discover subtests. Unit tests are out of scope for tilt as they don't need any infra

Also GitHub seems to not handle indentation diffs very well, most of the line changes are just subtest indentation


dtunikov (approved):
looks good to me
i would maybe add smth to `claude.md` to make it more likely to catch cases when someone adds a test for mysql and forgets about mariadb (and vice versa), it mostly affects integration tests (not e2e suite):
```
func TestIntegrationGetTableSchemaCaseSensitiveIdentifiers(t *testing.T) {
	for _, tc := range []struct {
		name string
	}{
		{name: "mysql"},
		{name: "mariadb"},
	} {
		t.Run(tc.name, func(t *testing.T) {
```

flow/connectors/mysql/gtid_test.go
```diff
@@ -10,43 +10,52 @@ import (
...
-	gtidEnabled := connector.config.ReplicationMechanism == protos.MySqlReplicationMechanism_MYSQL_GTID
-
-	if !gtidEnabled {
-		_, err := connector.GetMasterGTIDSet(ctx)
-		require.ErrorContains(t, err, "gtid mode is not enabled")
-		return
+func TestIntegrationMySQLGetMasterGTIDSet(t *testing.T) {
+	for _, tc := range []struct {
+		name string
+	}{
```
pfcoperez, line 16: Can't this struct:

```go
{
		{name: "mysql"},
		{name: "mariadb"},
	}
```

Be factored out as a common constant or the table test structure factored out as a test helper? 

This way we don't only stop repeating ourselves but also prevent accidentally missing one of the flavors.

ilidemi: It can but we then would lose the IDE codelenses as the VS Code extensions are not that smart to match through the layers of indirection

pfcoperez (approved):
Proceeding to force merge and update the branch protection

ilidemi:
> to make it more likely to catch cases when someone adds a test for mysql and forgets about mariadb (and vice versa)

Addressed in https://github.com/PeerDB-io/peerdb/pull/4518. When this one is merged, I thiiink that one should switch to `main` as base and kick off the CI.

PR #4518, opened 2026-06-30
Enforce the naming convention for MySQL integration tests

ilidemi:
Follow-up to #4517. Add an AST test enforcing:
- Tests that run against both local MySQL and MariaDB must be named TestIntegration... and must declare inline subtests named "mysql" and "mariadb" using a []struct{ name string } table.
- Tests that intentionally run only against local MySQL must be named TestMySQLOnlyIntegration...
- Tests that intentionally run only against local MariaDB must be named TestMariaOnlyIntegration...
- Other Test... functions in this package are treated as unit tests and must not reach local MySQL/MariaDB test configuration helpers.

Add tests for the test as well, so it's clear from fixtures what's expected and not expected.


jgao54 (approved):
Another idea for enforcing "unit tests should not connect to db" and "mysql-only test should strictly talk to mysql"

```
func setupMyConnector(t t*testing.T, ...) {
    ...
    connector, err := NewMySqlConnector(ctx, config)
    base := connector.dialContext
    connector.dialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
        if !strings.Contains(t.Name(), "Integration") {
            t.Errorf("non-integration test dialed %s %s", network, addr) 
            return nil, fmt.Errorf("blocked: %s is not an integration test", t.Name())
        }
        if strings.StartsWith(t.Name(), "TestMySQLIntegration") {
            // check that it dials mysql
        }
        if strings.StartsWith(t.Name(), "TestMariaIntegration") {
            // check that it dials maria
        }
        return base(ctx, network, addr)
    }
    ...
```
may require some refactoring to set up the connector to go thru a common constructor; and the test becomes runtime checks instead of static syntax based; but main benefit is simplicity and not needing to maintain the db-related helpers and envs.

Just an idea, your call.  Long term ideally we separate out integration tests from unit tests at the directory level so we can block networking at the OS level when running unit tests.

ilidemi:
Didn't know about t.Name(), this could make things a lot simpler. Good find!

ilidemi:
We'll do the route of merging mysql and maria into one matrix leg and not using subtests to discriminate

PR #4520, opened 2026-06-30
fix: Add tests for non-pk table cdc works in append-only mode with Merge Tree

masterashu:
* fix: Add tests for non-pk table cdc works in append-only mode with Merge Tree
* Add validation to prevent null sharding key when setting up no-Primary Key table

### DIFF
--- a/flow/e2e/clickhouse_mysql_test.go
+++ b/flow/e2e/clickhouse_mysql_test.go
@@ -2031,3 +2031,76 @@ func (s ClickHouseSuite) Test_MySQL_String_Partition_Key_Arbitrary_FullTable() {
 	env.Cancel(s.t.Context())
 	RequireEnvCanceled(s.t, env)
 }
+
+func (s ClickHouseSuite) Test_MySQL_NoPrimaryKey_CDC() {
+	if _, ok := s.source.(*MySqlSource); !ok {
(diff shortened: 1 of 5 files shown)


flow/e2e/clickhouse_mysql_test.go
```diff
@@ -2031,3 +2031,81 @@ func (s ClickHouseSuite) Test_MySQL_String_Partition_Key_Arbitrary_FullTable() {
...
 	RequireEnvCanceled(s.t, env)
 }
+
+func (s ClickHouseSuite) Test_MySQL_NoPrimaryKey_CDC() {
+	if _, ok := s.source.(*MySqlSource); !ok {
+		s.t.Skip("only applies to mysql")
+	}
+
+	if s.cluster {
+		// TODO: Check how this behaviour works in cluster mode.
```
ilidemi: Does this fail on cluster or just because of the local testing? We usually test the cluster behavior in CI

ilidemi (approved):

flow/e2e/clickhouse_mysql_test.go
masterashu: This failed for the cluster on GitHub CI too, so now with clickhouse cluster working locally, will test it out.

ilidemi (approved):

PR #3616, opened 2025-10-17
mysql: no parameters

serprex:
work around https://github.com/go-mysql-org/go-mysql/issues/1056

PR #4522, opened 2026-06-30
support arbitrary string parallel snapshotting

jgao54:
Add parallel snapshotting support for arbitrary string using an adaptive bisection strategy. 

The partitions are determined by starting with a single partition using min/max fetched values; and then continuously bisecting the largest partition into 2 smaller partitions, until we reached the target number of partitions or until they are no longer splittable. 

- To determine largest partition, we leverage mysql's query planner's estimated row count for a given range.
- To determine the midpoint, we encode the first 8 characters (after shared prefix) as a base-95 integer (one digit per character over printable ASCII); bytes outside of the range get rounded to the nearest edge. 
- To ensure correctness in coverage, we always fetch the next real key after the interpolated midpoint to ensure server-side collation is used.

There are known edge cases that can lead to skewed partitions (see unit tests), but never compromises on correctness. We sacrifice some balancing optimization for a more simple, deterministic partitioning algorithm. 

Fixes: DBI-885

### DIFF
--- a/flow/connectors/mysql/mysql.go
+++ b/flow/connectors/mysql/mysql.go
@@ -281,6 +281,12 @@ func (c *MySqlConnector) setSessionSettings() error {
 	return nil
 }
 
+func escapeWithNoBackslashEscapes(s string) string {
+	// mysql.Escape must NOT be used because MySQL connector session has sql_mode set
+	// to NO_BACKSLASH_ESCAPES (see setSessionSettings). Only quotes needs to be escaped.
+	return strings.ReplaceAll(s, "'", "''")
+}
+
 // withRetries return an iterable over connections,
 // consumer should break out of loop on success or error,
 // to retry for mysql.ErrBadConn
--- a/flow/connectors/mysql/qrep.go
+++ b/flow/connectors/mysql/qrep.go
@@ -167,9 +167,7 @@ func (c *MySqlConnector) GetQRepPartitions(
 			stringPartitions, err := buildAdaptiveStringPartitions(
 				ctx, c, c.logger, parsedWatermarkTable, config.WatermarkColumn, start, end, numPartitions)
 			if err != nil {
-				c.logger.Warn("failed to build adaptive string partitions, falling back to full table partition",
-					slog.Any("error", err))
-				return utils.FullTablePartition(), nil
+				return nil, fmt.Errorf("failed to build adaptive string partitions: %w", err)
 			}
 			partitionHelper.AddPartitions(stringPartitions)
 		}
--- a/flow/connectors/mysql/qrep.go
+++ b/flow/connectors/mysql/qrep.go
@@ -370,8 +370,8 @@ func (c *MySqlConnector) PullQRepRecords(
 			rangeStart = "'" + x.TimestampRange.Start.AsTime().Format("2006-01-02 15:04:05.999999") + "'"
 			rangeEnd = "'" + x.TimestampRange.End.AsTime().Format("2006-01-02 15:04:05.999999") + "'"
 		case *protos.PartitionRange_StringRange:
-			rangeStart = "'" + mysql.Escape(x.StringRange.Start) + "'"
-			rangeEnd = "'" + mysql.Escape(x.StringRange.End) + "'"
+			rangeStart = "'" + escapeWithNoBackslashEscapes(x.StringRange.Start) + "'"
+			rangeEnd = "'" + escapeWithNoBackslashEscapes(x.StringRange.End) + "'"
 			if config.Query != "" {
 				// custom query is only possible for standalone QRepFlowWorkflow;
 				// this is a legacy feature and string partitioning is not supported
--- a/flow/connectors/mysql/qrep.go
+++ b/flow/connectors/mysql/qrep.go
@@ -164,8 +164,14 @@ func (c *MySqlConnector) GetQRepPartitions(
 			}
 			partitionHelper.AddPartitions(uuidPartitions)
 		} else {
-			c.logger.Info("string watermark column is not uuid, falling back to full table partition")
-			return utils.FullTablePartition(), nil
+			stringPartitions, err := buildAdaptiveStringPartitions(
+				ctx, c, c.logger, parsedWatermarkTable, config.WatermarkColumn, start, end, numPartitions)
+			if err != nil {
+				c.logger.Warn("failed to build adaptive string partitions, falling back to full table partition",
+					slog.Any("error", err))
+				return utils.FullTablePartition(), nil
+			}
+			partitionHelper.AddPartitions(stringPartitions)
 		}
 	} else if err := partitionHelper.AddPartitionsWithRange(val1.Value(), val2.Value(), numPartitions); err != nil {
 		return nil, fmt.Errorf("failed to add partitions: %w", err)
--- a/flow/connectors/mysql/qrep_partition.go
+++ b/flow/connectors/mysql/qrep_partition.go
@@ -9,7 +9,9 @@ import (
 	"math/big"
 	"regexp"
 	"strings"
+	"unicode/utf8"
 
+	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/google/uuid"
 	"go.temporal.io/sdk/log"
 
@@ -119,8 +121,13 @@ func stringMidpoint(s1 string, s2 string) string {
 	for i < len(s1) && i < len(s2) && s1[i] == s2[i] {
 		i++
 	}
-	// This can break multibyte characters in the middle but the queries still
-	// succeed and the whole range is covered, even if with a skew
+	// Back off so the prefix doesn't end in the middle of a multibyte character.
+	// The midpoint is sent to MySQL as utf8mb4 literal (the session charset),
+	// When the column charset differs from utf8mb4, the server transcodes the
+	// literal for comparison and fails on invalid UTF-8.
+	for i > 0 && i < len(s1) && !utf8.RuneStart(s1[i]) {
+		i--
+	}
 	sharedPrefix := s1[:i]
 	s1, s2 = s1[i:], s2[i:]
 	mid := (stringToBase95Integer(s1) + stringToBase95Integer(s2)) / 2
@@ -300,6 +307,12 @@ func (c *MySqlConnector) fetchPrevRealKey(
 func (c *MySqlConnector) fetchRealKey(ctx context.Context, query string) (string, bool, error) {
 	rs, err := c.Execute(ctx, query)
 	if err != nil {
+		// The interpolated midpoint may be unrepresentable in the column charset,
+		// in which case the server cannot transcode the string literal for comparison
+		// and fails with ERROR 3854 (ER_CANNOT_CONVERT_STRING). Treat it as "no key found".
+		if mErr, ok := errors.AsType[*mysql.MyError](err); ok && mErr.Code == 3854 {
+			return "", false, nil
+		}
 		return "", false, err
 	}
 	defer rs.Close()
--- a/flow/connectors/mysql/qrep_partition.go
+++ b/flow/connectors/mysql/qrep_partition.go
@@ -3,6 +3,7 @@ package connmysql
 import (
 	"container/heap"
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"math/big"
@@ -68,7 +69,7 @@ func buildUuidStringPartitions(
 		return nil, fmt.Errorf("failed to convert max uuid to bigint: %w", err)
 	}
 	if minInt.Cmp(maxInt) > 0 {
-		return nil, fmt.Errorf("min uuid (%s) greater than max uuid (%s)", minVal, maxVal)
+		return nil, errors.New("min uuid greater than max uuid")
 	}
 
 	var partitions []*protos.QRepPartition
@@ -119,6 +120,8 @@ func stringMidpoint(s1 string, s2 string) string {
 	for i < len(s1) && i < len(s2) && s1[i] == s2[i] {
 		i++
 	}
+	// This can break multibyte characters in the middle but the queries still
+	// succeed and the whole range is covered, even if with a skew
 	sharedPrefix := s1[:i]
 	s1, s2 = s1[i:], s2[i:]
 	mid := (stringToBase95Integer(s1) + stringToBase95Integer(s2)) / 2
@@ -209,7 +212,7 @@ func buildAdaptiveStringPartitions(
 
 	totalRows, err := prober.estimateRowsInRange(ctx, tableName, quotedCol, minVal, maxVal)
 	if err != nil {
-		return nil, fmt.Errorf("failed to estimate rows for [%s, %s]: %w", minVal, maxVal, err)
+		return nil, fmt.Errorf("failed to estimate rows: %w", err)
 	}
 	h := &stringPartitionHeap{{start: minVal, end: maxVal, rows: totalRows}}
 	heap.Init(h)
@@ -226,7 +229,7 @@ func buildAdaptiveStringPartitions(
 		mid := stringMidpoint(p.start, p.end)
 		k, found, err := prober.fetchNextRealKey(ctx, tableName, quotedCol, mid, p.start, p.end)
 		if err != nil {
-			return nil, fmt.Errorf("failed to fetch next real key for [%s, %s]: %w", p.start, p.end, err)
+			return nil, fmt.Errorf("failed to fetch next real key: %w", err)
 		}
 		if !found {
 			// The interpolated midpoint can overshoot every key in the range when
@@ -234,7 +237,7 @@ func buildAdaptiveStringPartitions(
 			// backwards before declaring the partition unsplittable.
 			k, found, err = prober.fetchPrevRealKey(ctx, tableName, quotedCol, mid, p.start, p.end)
 			if err != nil {
-				return nil, fmt.Errorf("failed to fetch prev real key for [%s, %s]: %w", p.start, p.end, err)
+				return nil, fmt.Errorf("failed to fetch prev real key: %w", err)
 			}
 		}
 		if !found {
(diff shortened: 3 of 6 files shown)


flow/connectors/mysql/qrep_partition.go
```diff
@@ -100,3 +106,220 @@ func bigIntToUUID(n *big.Int, casing hexCasing) (string, error) {
...
+}
+
+// estimateRowsInRange asks the query planner for the estimated number of rows in
+// [start, end). The estimate is best-effort: a NULL or zero estimate is returned
+// as 0 rather than an error, so the caller still treats the range as a real partition.
+func (c *MySqlConnector) estimateRowsInRange(
+	ctx context.Context, tableName string, quotedCol string, start string, end string,
+) (uint64, error) {
+	query := fmt.Sprintf(
+		"EXPLAIN FORMAT=TRADITIONAL SELECT 1 FROM %[1]s WHERE %[2]s >= '%[3]s' AND %[2]s < '%[4]s'",
```
jgao54, line 339: without FORMAT=TRADITIONAL, on newer version of mysql it will return:
```
mysql> explain select 1 from str_test where id > 'a' and id <= 'z';
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| EXPLAIN                                                                                                                                                                                |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| -> Filter: ((str_test.id > 'a') and (str_test.id <= 'z'))  (cost=18.9 rows=93)
    -> Covering index range scan on str_test using PRIMARY over ('a' < id <= 'z')  (cost=18.9 rows=93)
 |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

what we want:
```
mysql> explain format=traditional select 1 from str_test where id > 'a' and id <= 'z';
+----+-------------+----------+------------+-------+---------------+---------+---------+------+------+----------+--------------------------+
| id | select_type | table    | partitions | type  | possible_keys | key     | key_len | ref  | rows | filtered | Extra                    |
+----+-------------+----------+------------+-------+---------------+---------+---------+------+------+----------+--------------------------+
|  1 | SIMPLE      | str_test | NULL       | range | PRIMARY       | PRIMARY | 202     | NULL |   93 |   100.00 | Using where; Using index |
+----+-------------+----------+------------+-------+---------------+---------+---------+------+------+----------+--------------------------+
```

jgao54:
The last commit https://github.com/PeerDB-io/peerdb/pull/4522/changes/7e97b79ab00267bdd580568297e94f3319445887 also fixes the flaky test scenario TestBuildAdaptiveStringPartitions_MixedCase_UUID reported above, which happens occasionally if the generated UUIDs have a midpoint that does not return a valid `fetchNextRealKey`; but would return a valid `fetchPrevRealKey`.

flow/connectors/mysql/qrep.go
```diff
@@ -164,8 +164,14 @@ func (c *MySqlConnector) GetQRepPartitions(
 			}
 			partitionHelper.AddPartitions(uuidPartitions)
 		} else {
-			c.logger.Info("string watermark column is not uuid, falling back to full table partition")
-			return utils.FullTablePartition(), nil
+			stringPartitions, err := buildAdaptiveStringPartitions(
+				ctx, c, c.logger, parsedWatermarkTable, config.WatermarkColumn, start, end, numPartitions)
+			if err != nil {
+				c.logger.Warn("failed to build adaptive string partitions, falling back to full table partition",
```
ilidemi: We should still be deterministic in face of network errors.
Haven't we had this conversation before
https://github.com/PeerDB-io/peerdb/pull/4097#discussion_r3076969981
https://github.com/PeerDB-io/peerdb/pull/3972#discussion_r2843789152

ilidemi: This masks MariaDB rejecting `EXPLAIN FORMAT=TRADITIONAL`

flow/connectors/mysql/qrep_partition.go
```diff
@@ -100,3 +106,234 @@ func bigIntToUUID(n *big.Int, casing hexCasing) (string, error) {
...
+	minVal string,
+	maxVal string,
+	numPartitions int64,
+) ([]*protos.QRepPartition, error) {
+	tableName := table.MySQL()
+	quotedCol := common.QuoteMySQLIdentifier(watermarkColumn)
+
+	totalRows, err := prober.estimateRowsInRange(ctx, tableName, quotedCol, minVal, maxVal)
+	if err != nil {
+		return nil, fmt.Errorf("failed to estimate rows for [%s, %s]: %w", minVal, maxVal, err)
```
ilidemi: Logging raw strings here and below. Debuggability is nice but not sure what the solution is

flow/connectors/mysql/qrep_partition.go
```diff
@@ -100,3 +106,234 @@ func bigIntToUUID(n *big.Int, casing hexCasing) (string, error) {
...
+		heap.Push(h, stringPartitionEntry{start: k, end: p.end, rows: rightRows})
+	}
+
+	for h.Len() > 0 {
+		outputs = append(outputs, heap.Pop(h).(stringPartitionEntry))
+	}
+
+	partitions := make([]*protos.QRepPartition, 0, len(outputs))
+	for _, p := range outputs {
+		partitions = append(partitions, utils.CreateStringPartition(p.start, p.end, p.end == maxVal))
```
ilidemi, line 272: **🟠 Major — NUL byte in a boundary key → deterministic stuck snapshot**

MySQL VARCHAR permits `0x00`; `mysql.Escape` maps it to a valid literal so probes succeed and the NUL key lands in a StringRange boundary. Postgres then rejects it: the catalog insert of `partition_start`/`partition_end` (text columns, `monitoring.go:393`) fails with `invalid byte sequence for encoding "UTF8": 0x00` (reproduced live against the dev catalog). `GetQRepPartitions` fails, Temporal recomputes identical partitions, fails identically — stuck forever. **Bounded by defaults:** with `PEERDB_OFFLOAD_PARTITION_RANGES=true` (default) and `InitialCopyOnly` snapshots, ranges are offloaded to encrypted bytea and this insert is skipped. Trigger requires offload disabled *or* a continuous mirror, plus a NUL in the min/max key.

flow/connectors/mysql/qrep_partition_test.go
```diff
@@ -165,3 +174,266 @@ func TestBuildUuidStringPartitionsInvalid(t *testing.T) {
...
+	minVal, maxVal := planner.keys[0], planner.keys[len(planner.keys)-1]
+	table := &common.QualifiedTable{Namespace: "db", Table: "t"}
+	partitions, err := buildAdaptiveStringPartitions(
+		t.Context(), planner, log.NewStructuredLogger(slog.Default()),
+		table, "id", minVal, maxVal, numPartitions)
+	require.NoError(t, err)
+	return partitions
+}
+
+func verifyFullCoverage(t *testing.T, partitions []*protos.QRepPartition, keys []string, less func(a string, b string) bool) {
```
ilidemi, line 364: Verify [i].end == [i+1].start as well?

flow/connectors/mysql/qrep_partition.go
```diff
@@ -100,3 +106,234 @@ func bigIntToUUID(n *big.Int, casing hexCasing) (string, error) {
...
+	base95Radix = base95Max - base95Min + 1
+	base95Width = 8 // 95^8 to fit in an uint64
+)
+
+func stringMidpoint(s1 string, s2 string) string {
+	i := 0
+	for i < len(s1) && i < len(s2) && s1[i] == s2[i] {
+		i++
+	}
+	sharedPrefix := s1[:i]
```
ilidemi, line 131:
```suggestion
	// This can break multibyte characters in the middle but the queries still succeed and the whole range is covered, even if with a skew
	sharedPrefix := s1[:i]
```

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -2213,13 +2214,28 @@ func (s ClickHouseSuite) Test_MySQL_String_Partition_Key_Arbitrary_FullTable() {
...
 		fmt.Sprintf("CREATE TABLE %s (id VARCHAR(50) PRIMARY KEY, val INT NOT NULL)", srcFullName)))
-	for i := 1; i <= numRows; i++ {
-		require.NoError(s.t, s.source.Exec(s.t.Context(),
-			fmt.Sprintf("INSERT INTO %s (id, val) VALUES ('key_%04d', %d)", srcFullName, i, i)))
+	seen := make(map[string]struct{}, numRows)
+	values := make([]string, 0, numRows)
+	for len(seen) < numRows {
+		keyBytes := make([]byte, 10)
+		for i := range keyBytes {
+			keyBytes[i] = byte('a' + rand.IntN(26)) //nolint:gosec // test data
```
ilidemi, line 2554: I think this needs a fuzz integration test (quicker than e2e but real mysql/mariadb) and leave it overnight/weekend. LLMs can help with throwing in charsets/collations/index types (idk if relevant for PK)/padding and setting up the data generation.

ilidemi (commented):
A bunch of Fable finds

> ultracode an adversarial code review of the changes on this branch, especially the string wrangling logic. report the findings here

flow/connectors/mysql/qrep_partition.go
```diff
@@ -100,3 +106,234 @@ func bigIntToUUID(n *big.Int, casing hexCasing) (string, error) {
...
+// fetchNextRealKey returns the smallest real value of the watermark column that
+// is at or past the interpolated midpoint and strictly inside (start, end). The
+// bounds are enforced by MySQL using its column's collation. Return false when
+// no such key exists.
+func (c *MySqlConnector) fetchNextRealKey(
+	ctx context.Context, tableName string, quotedCol string, midpoint string, start string, end string,
+) (string, bool, error) {
+	query := fmt.Sprintf(
+		"SELECT %[1]s FROM %[2]s WHERE %[1]s >= '%[3]s' AND %[1]s > '%[4]s' AND %[1]s < '%[5]s' ORDER BY %[1]s LIMIT 1",
+		quotedCol, tableName, mysql.Escape(midpoint), mysql.Escape(start), mysql.Escape(end))
```
ilidemi: **🔴 Critical — Watermark values interpolated with the wrong escaping for the session's sql_mode**

`flow/connectors/mysql/qrep_partition.go:281` (and `:294`, `:325`, plus the pull query at `flow/connectors/mysql/qrep.go:379`)

Every new probe query and the StringRange pull query build SQL literals with `mysql.Escape(...)` wrapped in single quotes. But the connector unconditionally runs `SET sql_mode = 'ANSI,NO_BACKSLASH_ESCAPES'` (`mysql.go:237`). `mysql.Escape` does **backslash** escaping (`'` → `\'`), which under `NO_BACKSLASH_ESCAPES` is a no-op — the backslash is a literal character. This was masked at merge-base because non-UUID strings fell back to full-table and emitted no per-value SQL; UUID boundaries are hex-only. This branch routes arbitrary row data through these literals for the first time. Three concrete failure modes, **reproduced live on MySQL 8.0**:

- **Silent data loss (this is the critical core).** A max watermark value like `Z:\tmp` (Windows/UNC path as key) is stored unescaped, so partitioning *succeeds* — no fallback. The last partition's inclusive bound becomes `BETWEEN ... AND 'Z:\\tmp'`, which under `NO_BACKSLASH_ESCAPES` is the literal 7-char string `Z:\\tmp`. The real `Z:\tmp` sorts *above* it, so the max row (and neighbors) are **silently never replicated**. Verified: `SELECT ... BETWEEN 'Mmm' AND 'Z:\\tmp'` returned only `Mmm`, dropping the real max. Symmetric at the min bound with `\n`/`\t`/`\0`. Interior boundaries stay consistent (adjacent partitions share the identical escaped key), so the leak is concentrated at the two endpoints — but it's a successful query with no error and no fallback.
- **SQL injection.** A source row value like `x' OR '1'='1` escapes to `'x\' OR '1'='1'`; the backslash is literal, the string closes early, and the tail executes. Watermark values are attacker-influenceable table data. Reachable but harder — quote-bearing values usually syntax-error the probe phase first (see below).
- **Quote-free data breaks too.** `stringMidpoint` synthesizes bytes in `[0x20,0x7E]` as base95 digits — including `0x27 (')` and `0x5C (\)`. Verified: `stringMidpoint("&&&", "(((")` = `"'''"`, invented from quote-free input. The probe then throws `ERROR 1064` and the build silently degrades to a single full-table partition.

**Fix:** use parameterized `Execute(query, args...)` (already supported at `mysql.go:303`) or `''` quote-doubling instead of `mysql.Escape`. This fixes all three at once.

ilidemi (commented):
huh forgot to post

flow/connectors/mysql/qrep.go
jgao54: surprised that tests that expect N partitions did not catch this, will take a look what happened here.

jgao54: MariaDB seem to support it just fine:
```
MariaDB [test]> explain FORMAT=TRADITIONAL SELECT 1 FROM a where id1 > 0 and id1 < 1;
+------+-------------+-------+-------+---------------+---------+---------+------+------+-------------+
| id   | select_type | table | type  | possible_keys | key     | key_len | ref  | rows | Extra       |
+------+-------------+-------+-------+---------------+---------+---------+------+------+-------------+
|    1 | SIMPLE      | a     | range | PRIMARY       | PRIMARY | 4       | NULL | 1    | Using where |
+------+-------------+-------+-------+---------------+---------+---------+------+------+-------------+
```

ilidemi: My bad on Maria, gotta keep my AIs in check

flow/connectors/mysql/qrep_partition.go
jgao54: updated to to use string literal; this won't get triggered in the general case, but we still want to fix it for `PEERDB_OFFLOAD_PARTITION_RANGES=false`.

flow/connectors/mysql/qrep_partition.go
jgao54: This is a good catch. Looks like https://github.com/PeerDB-io/peerdb/pull/3616 was the original PR that switched parameterized queries to mysql.Escape because parameterized query hit a bug on mysql 8.0.26; but for string partitioning we are querying table content data instead of table metadata, the risk of data loss here seems non-trivial since any `\` in the min/max boundary values can potentially lead to data loss (interior boundaries containing `\` can lead to worse skew, but no data loss).

jgao54: Verified that the newly introduced e2e test `Test_MySQL_String_Partition_Key_Special_Chars_Parallel_Snapshot` [failed](https://github.com/PeerDB-io/peerdb/actions/runs/29630427324/job/88042805246?pr=4522) with 50 vs 49 rows as expected, and passed with fix introduced.

flow/e2e/clickhouse_mysql_test.go
jgao54: Have a [separate PR](https://github.com/PeerDB-io/peerdb/pull/4608) for the fuzz test. Addressed its findings [here](https://github.com/PeerDB-io/peerdb/pull/4522/changes/590974f634be292b8eb9ede99e1a0413d5323643) in this PR

flow/connectors/mysql/mysql.go
```diff
@@ -281,6 +281,12 @@ func (c *MySqlConnector) setSessionSettings() error {
 	return nil
 }
 
+func escapeWithNoBackslashEscapes(s string) string {
+	// mysql.Escape must NOT be used because MySQL connector session has sql_mode set
+	// to NO_BACKSLASH_ESCAPES (see setSessionSettings). Only quotes needs to be escaped.
```
ilidemi, line 286: Claude likes adding these types of comments but this really should be a task and a sweep (def too big for this PR)

flow/connectors/mysql/qrep_partition.go
```diff
@@ -300,6 +307,12 @@ func (c *MySqlConnector) fetchPrevRealKey(
 func (c *MySqlConnector) fetchRealKey(ctx context.Context, query string) (string, bool, error) {
 	rs, err := c.Execute(ctx, query)
 	if err != nil {
+		// The interpolated midpoint may be unrepresentable in the column charset,
+		// in which case the server cannot transcode the string literal for comparison
+		// and fails with ERROR 3854 (ER_CANNOT_CONVERT_STRING). Treat it as "no key found".
+		if mErr, ok := errors.AsType[*mysql.MyError](err); ok && mErr.Code == 3854 {
```
ilidemi, line 313: Awesome fallback, just don't split the partition if we're in a complete encoding rabbithole

flow/connectors/mysql/mysql.go
ilidemi, line 257: Not introduced in this PR but related: if this fails, this gives us more cases where we have to deal with invalid utf-8, so safer to just error out here. Statement itself is supported everywhere and the cases it's being hit are all network errors/driver bugs https://c.house/NEzPf5hrTg

ilidemi (approved):
Great round of improvements!

flow/connectors/mysql/mysql.go
jgao54: good point, will address this as a follow-up

flow/connectors/mysql/qrep_partition.go
jgao54: This is still an issue today that I didn't address in this PR yet. It doesn't impact correctness when that happens (this == consecutively invalid partition, not just transcode error), but we are unnecessarily creating a bunch of single-row partitions with only 1 bulk partition when that happen. I'll take this as a follow-up to not expand this PR further. Should be a relatively small fix where we check min/max to see if they are splittable upfront, and if they are both over, or both under the acceptable ascii, then fallback to single partition.

jgao54: actually nvm, the transcoding error rabbithole is not a real issue because with found == false, it would end partitioning logic quickly. (vs. the issue with all PK being outside of the ascii bound)

PR #4525, opened 2026-07-01
[DBI-798] MySQL Connector: Allow passing `server_id` as a peer configuration parameter

pfcoperez:
This PR allows setting `server_id` on MySQL/MariaDB peers. 

When it is present, CDC flows will use this value as the [replication `server_id`](https://dev.mysql.com/doc/mysql-replication-excerpt/5.7/en/replication-options.html#:~:text=server_id%20is%20set%20to%200,replica%20in%20the%20replication%20topology.)  in the declared MySQL replica.

This is an optional field in the configuration and offers a deterministic alternative to the default  "random id" select upon CDC flow initialization in the current implementation and in the new implementation when the value is missing.

Note that if a peer has this option, it can not be re-used in different mirrors. This PR also includes changes to prevent the creation of such mirrors through validation.

image

Commits are organized so they can be reviewed independently:

- bb949c0698a132e13d4b0416f190e4ce84c70fba Core functionality, adding optional parameter to the gRPC specification of PeerDB API and propagating it down to CDC initialization.
- 729225ce2e947a1ec792d1e27264862bea1ef7bb e2e tests for this functionality.
- 132c87e017ccafa82245640c6d6fbedefc5b947f  Add a validation step in the create/validate API endpoints that prevents re-using the same MySQL peer in two mirrors if it has `server_id` set.
- 35f565f5fda9b09e8db9bb6acee25ed7c4317d23 API e2e tests for this validation.
- 09507285701e219afa22991bba8e9f2798c14614 PeerDB UI changes to include this field in MySQL peer configuration

Part of: https://linear.app/clickhouse/issue/DBI-798

### DIFF
--- a/flow/cmd/validate_mirror.go
+++ b/flow/cmd/validate_mirror.go
@@ -8,6 +8,8 @@ import (
 	"regexp"
 	"slices"
 
+	"google.golang.org/protobuf/proto"
+
 	"github.com/PeerDB-io/peerdb/flow/connectors"
 	"github.com/PeerDB-io/peerdb/flow/generated/proto_conversions"
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
@@ -102,6 +104,10 @@ func (h *FlowRequestHandler) validateCDCMirrorImpl(
 		}
 	}
 
+	if apiErr := h.checkSourcePeerReuse(ctx, connectionConfigs); apiErr != nil {
+		return nil, apiErr
+	}
+
 	srcConn, srcClose, err := connectors.GetByNameAs[connectors.MirrorSourceValidationConnector](
 		ctx, connectionConfigs.Env, h.pool, connectionConfigs.SourceName,
 	)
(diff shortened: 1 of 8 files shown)


flow/cmd/validate_mirror.go
```diff
@@ -158,6 +164,73 @@ func (h *FlowRequestHandler) validateCDCMirrorImpl(
 	return &protos.ValidateCDCMirrorResponse{}, nil
 }
 
+// checkSourcePeerReuse rejects a CDC mirror whose MySQL source peer pins a fixed server_id while
+// that peer already backs another streaming CDC mirror. A fixed server_id can only be used by one
```
ilidemi, line 168: If the server_id is per cdc connection, is there a reason not to put the setting on the mirror?

pfcoperez: I thought about this possibility but didn't pursue it because `server_id` is a MySQL only concept. I kept it in the connector specific setting keeping the interface clean. It feels wrong to me that we can send over a `server_id` for a PG flow description. 

Let's assume this is a per mirror setting. In this situation API users have it easy to create mirrors sharing the same source and the same server_id and these are flows are going to fail after validation (unless it is explicitly implemented).
With the current approach, users using the same source server by just re-using the same peer will get the check at validation time.
In both cases, a cross mirror validation is required.

It's true that `replication_slot_name`  is PG specific while still living in `FlowConnectionConfigs`. But it's also true that `replication_mechanism`, which is CDC exclusive, (and I think correctly) lives  within  `MySqlConfig` peer protos. If I moved `server_id` to the mirror configuration I'd still add a validation check on it not being set on other sources than MySQL/MariaDB and another checking other mirrors usage of the same server_id.

Finally, while the guard I added is aimed to detect issues at validation time when using standalone PeerDB, ClickPipes managed use of this field will not trigger a clash by design.

pfcoperez: Actually, this conversation made me realize that we need to add a method at `flow/pkg/mysql` factoring out the validation of a pre-existing replica with the same service_id.

ilidemi: I'm not sure we'll be able to get through to PeerDB at validation time. If we want prevalidations for this (I think they're optional), control plane would probably be the best place (as disgusting as it may sound)

flow/connectors/mysql/cdc.go
```diff
@@ -330,9 +330,18 @@ func (c *MySqlConnector) startSyncer(ctx context.Context, env map[string]string)
...
 	}
-	//nolint:gosec
+
+	var serverId uint32
+	if c.config.ServerId != nil {
+		serverId = *c.config.ServerId
+	} else {
+		// If the configuration doesn't specify a server_id value, fallback to
+		// legacy behavior of generating a random server_id.
+		serverId = rand.Uint32() //nolint:gosec // G404: server_id does not require cryptographic randomness
```
ilidemi: Adding the rand(1000, max) would be a one-liner here but up to you

flow/connectors/mysql/cdc_test.go
```diff
@@ -136,6 +137,70 @@ func TestIntegrationANSIQuotesDDLParsedFromBinlog(t *testing.T) {
 	}
 }
 
+func sourceHasRegisteredReplica(ctx context.Context, c *MySqlConnector, serverID uint32) (bool, error) {
```
ilidemi: As it's single-use, maybe a func var in the test would be better?

ui/app/peers/create/[peerType]/helpers/my.ts
```diff
@@ -98,6 +98,29 @@ export const mysqlSetting: PeerSetting[] = [
...
   },
+  {
+    label: 'Server ID',
+    field: 'serverId',
+    type: 'number',
+    optional: true,
+    stateHandler: (value, setter) => {
+      const strValue = value as string;
+      if (!strValue) {
+        // Leave unset so a random server_id is generated on each CDC run (legacy behavior).
```
Comment (ilidemi) [inline ui/app/peers/create/[peerType]/helpers/my.ts:109]: It's more of a default behavior rather than legacy

ilidemi (approved):

flow/connectors/mysql/cdc.go
pfcoperez: Indeed, done in 8244b45ed56ede29b8d63d82810581661cb5ca1d

ui/app/peers/create/[peerType]/helpers/my.ts
Comment (pfcoperez) [reply ui/app/peers/create/[peerType]/helpers/my.ts:109]: That's true! Fixed in 8244b45ed56ede29b8d63d82810581661cb5ca1d

flow/connectors/mysql/cdc_test.go
pfcoperez: Actually, in 31289c9f5da5f266f05bcb0e8f40a00ac54eade8, I moved this logic to flow/pkg to be re-used for validation in ClickHouse services.

PR #4530, opened 2026-07-02
maria: require resync on partial row events

dtunikov:
We don't support partial row events. MariaDB 12.3+ emits those when payload size is bigger than `binlog_row_event_fragment_threshold` (default 1Gb).
This PR handles this event type and requires a resync.
In addition, it introduces `UnsupportedBinlogEventCounter(eventType)` metric to keep track of skipped events.


flow/connectors/mysql/cdc.go
```diff
@@ -566,12 +574,24 @@ func (c *MySqlConnector) PullRecords(
...
 				incident, message := parseIncidentEvent(ev.Data)
 				c.logger.Error("[mysql] received binlog incident event, resync required",
 					slog.Uint64("incident", uint64(incident)), slog.String("message", message))
 				return exceptions.NewMySQLBinlogIncidentError(incident, message)
+			case mariadbPartialRowDataEvent:
+				// MariaDB 12.3+ fragments an oversized rows event across multiple events; we don't
+				// reassemble fragments yet, so the row change is unrecoverable and a resync is required.
+				c.logger.Error("[mysql] received MariaDB partial row data event, resync required",
+					slog.Uint64("eventType", uint64(mariadbPartialRowDataEvent)))
+				return exceptions.NewMySQLUnsupportedPartialRowEventError(byte(mariadbPartialRowDataEvent))
```
dtunikov: I think it makes sense to require a resync here, right? (until we're able to process such events)
Otherwise we gonna silently skip data.

flow/connectors/mysql/cdc.go
```diff
@@ -566,12 +574,24 @@ func (c *MySqlConnector) PullRecords(
...
+			default:
+				c.logger.Warn("unknown generic event", slog.Any("type", event.Header.EventType))
+				otelManager.Metrics.UnsupportedBinlogEventCounter.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(
```
dtunikov: gonna add some chart to our mysql dashboard to have some visibility on those

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -1724,6 +1682,81 @@ func (s ClickHouseSuite) Test_MySQL_BinlogIncident() {
...
+	}
+	// We spin up our own MariaDB 12.3 below regardless of the suite source, so only run this once -
+	// under the MySQL suite variant - to avoid launching a second throwaway container needlessly.
+	if mySource.Config.Flavor == protos.MySqlFlavor_MYSQL_MARIA {
+		s.t.Skip("partial row event test spins up its own MariaDB; skip duplicate run under MariaDB suite")
+	}
+
+	// binlog_row_event_fragment_threshold is pinned to its 1024-byte minimum so a modest row
+	// overflows a single rows event and MariaDB splits it into PARTIAL_ROW_DATA_EVENT fragments.
+	src, suffix := SetupMySQLTestContainerSource(s.t, "mdbpart", MySQLTestContainerConfig{
```
dtunikov, line 2053: we could also put binlog-row-event-fragment-threshold on our main CI maria instance
but I'm afraid this might lead to unexpected behavior in other tests
afaik the testcontainer you put for `BinlogIncident` event seems to work fine, so I would use the same approach here

flow/alerting/classifier.go
```diff
@@ -1154,6 +1154,13 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
+	if _, ok := errors.AsType[*exceptions.MySQLUnsupportedPartialRowEventError](err); ok {
+		return ErrorNotifyBinlogInvalid, ErrorInfo{
```
dtunikov: it's not really "binlog invalid" case, but it's the closest fitting error
we could add a distinct error similar to `ErrorNotifyBinlogEventExceededMaxAllowedPacket` and point to FAQ page with an explanation that 'partial row events are not supported'

jgao54: i think it makes sense to use a distinct error with an FAQ link yeah!

flow/connectors/mysql/cdc.go
jgao54: erroring and asking for a resync makes sense.  the `binlog_row_event_fragment_threshold` is a server side config right?  if the customer bump this value and then resync, could that potentially allow them to workaround this issue for subsequent oversized events (if they know the upper bound is within a reasonable size)

flow/shared/exceptions/mysql.go
```diff
@@ -153,3 +153,17 @@ func (e *MySQLBinlogIncidentError) Error() string {
...
+}
+
+func NewMySQLUnsupportedPartialRowEventError(eventType byte) *MySQLUnsupportedPartialRowEventError {
+	return &MySQLUnsupportedPartialRowEventError{EventType: eventType}
+}
+
+func (e *MySQLUnsupportedPartialRowEventError) Error() string {
+	return fmt.Sprintf(
+		"unsupported MariaDB PARTIAL_ROW_DATA_EVENT (type %d): fragmented oversized row events cannot be processed; a resync is required",
+		e.EventType)
```
jgao54, line 191: nit: would be good to also supply schema/table name

flow/connectors/mysql/cdc.go
jgao54: ah nvm, i see that 1GB is the [limit](https://mariadb.com/docs/server/server-management/server-monitoring-logs/binary-log/binary-log-formats#splitting-large-row-format-replication-events)

flow/shared/exceptions/mysql.go
jgao54: I see that generic event doesn't provide this out-of-the-box and implementing our own parser may be out-of-scope, although technically possible according to Claude:
```
Pull the table ID out of the first fragment. The first fragment's content embeds the original rows event verbatim: its own 19-byte common header, then the rows-event post-header, which begins with a 6-byte little-endian table ID (same layout RowsEvent.Decode reads at row_event.go:1021). So inside ev.Data:

// ev.Data layout for PARTIAL_ROW_DATA_EVENT (172):
// [0:4] seq_no, [4:8] total_fragments, [8] flags,
// [9:17] orig size (only if flags&0x01, i.e. first fragment), then content.
pos := 9
if ev.Data[8]&0x01 != 0 {
    pos += 8
}
content := ev.Data[pos:]
// content = original rows event: 19-byte header, then post-header
tableID := mysql.FixedLengthInt(content[19 : 19+6])
```

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -1724,6 +1682,81 @@ func (s ClickHouseSuite) Test_MySQL_BinlogIncident() {
...
+	if s.cluster {
+		s.t.Skip("source-side partial row event coverage does not need to run against ClickHouse cluster")
+	}
+	mySource, ok := s.source.(*MySqlSource)
+	if !ok {
+		s.t.Skip("only applies to mysql")
+	}
+	// We spin up our own MariaDB 12.3 below regardless of the suite source, so only run this once -
+	// under the MySQL suite variant - to avoid launching a second throwaway container needlessly.
+	if mySource.Config.Flavor == protos.MySqlFlavor_MYSQL_MARIA {
```
jgao54: nit: may be a dumb question, but not sure if i follow: if we are spinning up our own MariaDB only once, why not run it under MARIA and skip when it's MySQL (instead of running it under MySQL and skip it for Maria)

jgao54 (approved):
a few nit comment but lgtm overall

flow/connectors/mysql/cdc.go
```diff
@@ -45,6 +46,13 @@ const (
 	binlogStalenessMultiplier    = 3
 )
 
+// mariadbPartialRowDataEvent is MariaDB 12.3+ PARTIAL_ROW_DATA_EVENT (Log_event_type = 172).
+// It fragments an oversized rows event across several binlog events, which a consumer must buffer
+// and reassemble before decoding. go-mysql lib has no constant or parser for it, so it
+// surfaces as a GenericEvent; we don't reassemble fragments yet, so we fail loudly rather than
+// silently dropping the row change.
```
ilidemi: The comment is very redundant, the one you added near the actual logic sums things up sufficiently

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -1724,6 +1682,81 @@ func (s ClickHouseSuite) Test_MySQL_BinlogIncident() {
...
+		s.t.Skip("partial row event test spins up its own MariaDB; skip duplicate run under MariaDB suite")
+	}
+
+	// binlog_row_event_fragment_threshold is pinned to its 1024-byte minimum so a modest row
+	// overflows a single rows event and MariaDB splits it into PARTIAL_ROW_DATA_EVENT fragments.
+	src, suffix := SetupMySQLTestContainerSource(s.t, "mdbpart", MySQLTestContainerConfig{
+		Image:                "mariadb:12.3",
+		Flavor:               protos.MySqlFlavor_MYSQL_MARIA,
+		ReplicationMechanism: protos.MySqlReplicationMechanism_MYSQL_GTID,
+		ExtraServerFlags: []string{
```
ilidemi, line 2057: The other test passes a bunch of params to lower the resource usage as this will add to all other containers on the machine. Should do it here as well, in the helper works if they're the same between MySQL and Maria.

Although the binlog incident one has introduced flakes and I haven't looked into those yet. CIDB would have data

flow/shared/exceptions/mysql.go
ilidemi: We would also start failing on tables outside of the pipe now without parsing the table name. Could either handle all at once (parsing and skipping if not included) or see if an escalation comes and handle both then

flow/connectors/mysql/cdc.go
```diff
@@ -566,12 +574,24 @@ func (c *MySqlConnector) PullRecords(
...
+			case mariadbPartialRowDataEvent:
+				// MariaDB 12.3+ fragments an oversized rows event across multiple events; we don't
+				// reassemble fragments yet, so the row change is unrecoverable and a resync is required.
+				c.logger.Error("[mysql] received MariaDB partial row data event, resync required",
+					slog.Uint64("eventType", uint64(mariadbPartialRowDataEvent)))
+				return exceptions.NewMySQLUnsupportedPartialRowEventError(byte(mariadbPartialRowDataEvent))
+			default:
+				c.logger.Warn("unknown generic event", slog.Any("type", event.Header.EventType))
+				otelManager.Metrics.UnsupportedBinlogEventCounter.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(
+					attribute.String(otel_metrics.BinlogEventTypeKey, strconv.Itoa(int(event.Header.EventType))),
```
ilidemi: Does this need to be an attribute string + itoa? If you're planning to include non-generic events here from the parent switch-case or other branches, these ones would be better as `Generic_%d`.

flow/e2e/clickhouse_mysql_test.go
ilidemi: Oh I see some of them moved and some stayed for some reason

flow/connectors/mysql/cdc_test.go
```diff
@@ -438,3 +439,35 @@ func TestIntegrationGetTableSchemaPrimaryKeyVariants(t *testing.T) {
...
+	t.Parallel()
+
+	// Real PARTIAL_ROW_DATA_EVENT payloads captured from MariaDB 12.3
+	decode := func(s string) []byte {
+		b, err := hex.DecodeString(s)
+		require.NoError(t, err)
+		return b
+	}
+	// first fragment: total=9, seq=1, flags=FL_ORIG_EVENT_SIZE, then origSize + embedded rows event
+	firstFragment := decode("09000000010000000126200000000000002cc0476a" +
```
dtunikov: these are real payloads captured during local testing, so I think we can keep this unit test in addition to e2e one (would be faster to run locally when we make changes to parsing logic)

flow/shared/exceptions/mysql.go
dtunikov: added logic to filter out partial raw failures related to the tables that aren't included in the pipe
the code is not very simple, but the e2e test covers it well
i can rollback if you think that it's too much complexity for such feature (as we actually don't know how often this event happens in production yet)

flow/e2e/clickhouse_mysql_test.go
dtunikov: moved all resource related flags to the helper

flow/connectors/mysql/cdc.go
```diff
@@ -540,6 +581,11 @@ func (c *MySqlConnector) PullRecords(
 
 	lastEventAt := time.Now()
 	var mysqlParser *parser.Parser
+	// table id -> "schema.table", maintained from TABLE_MAP_EVENTs
+	tableIdToName := make(map[uint64]string)
+	// When we ignore an out-of-pipe fragmented rows event, its continuation fragments carry no table
+	// id; skip this many remaining PARTIAL_ROW_DATA_EVENTs before resolving table ids again.
+	var partialRowSkipFragments uint32
```
dtunikov, line 622: this is the trickiest part
but it seems to work fine according to e2e test

flow/connectors/mysql/cdc.go
dtunikov: makes sense
added this:
https://github.com/PeerDB-io/peerdb/pull/4530/commits/71770d442647b273ab94bf4a4a62837c636d1456

flow/alerting/classifier.go
```diff
@@ -1154,6 +1157,13 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
+	if _, ok := errors.AsType[*exceptions.MySQLUnsupportedPartialRowEventError](err); ok {
+		return ErrorNotifyBinlogPartialRowEventUnsupported, ErrorInfo{
```
ilidemi, line 1182: Use AdditionalAttributes for table and schema, like in ErrorUnsupportedSchemaChange. This allows to template them in a user-facing notification in ClickPipes

flow/connectors/mysql/cdc.go
```diff
@@ -537,9 +578,22 @@ func (c *MySqlConnector) PullRecords(
 				time.Now().UTC().Sub(time.UnixMicro(int64(commitTs))).Microseconds())
 		}
 	}
+	recordUnsupportedEvent := func(ctx context.Context, event *replication.BinlogEvent, prefix string) {
+		otelManager.Metrics.UnsupportedBinlogEventCounter.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(
+			attribute.String(otel_metrics.BinlogEventTypeKey, fmt.Sprintf("%s_%d", prefix, int(event.Header.EventType))),
+		)))
+		if _, loaded := c.warnedUnsupportedEventTypes.LoadOrStore(event.Header.EventType, struct{}{}); !loaded {
+			c.logger.Warn("unsupported rows event", slog.Any("type", event.Header.EventType))
```
ilidemi: As we're now warning on all unhandled events, will need to add a whitelist of benign events. Could be a follow-up once we see logs, or sweep go-mysql code beforehand and add an empty case branch

ilidemi: Also is it "unsupported rows event" anymore

flow/otel_metrics/otel_manager.go
```diff
@@ -603,6 +605,14 @@ func (om *OtelManager) setupMetrics(ctx context.Context) error {
 		return err
 	}
 
+	if om.Metrics.UnsupportedBinlogEventCounter, err = om.GetOrInitInt64Counter(BuildMetricName(UnsupportedBinlogEventName),
+		metric.WithDescription(
+			"Counter of unsupported binlog generic events seen on the CDC path, with `eventType` label "+
+				"holding the numeric binlog event type"),
```
ilidemi: Needs to be updated

flow/connectors/mysql/cdc.go
```diff
@@ -566,12 +620,42 @@ func (c *MySqlConnector) PullRecords(
...
 					slog.Uint64("incident", uint64(incident)), slog.String("message", message))
 				return exceptions.NewMySQLBinlogIncidentError(incident, message)
+			case mariadbPartialRowDataEvent:
+				if partialRowSkipFragments > 0 {
+					// continuation fragments of an out-of-pipe group we already decided to ignore
+					partialRowSkipFragments--
+					break
+				}
+				tableID, totalFragments, ok := parsePartialRowEventTableID(ev.Data)
+				sourceTableName, known := tableIdToName[tableID]
```
ilidemi, line 659:
```suggestion
				// tableIdToName tracking relies on TABLE_MAP_EVENT and PARTIAL_ROW_DATA_EVENT coming within the same batch, so no checkpoints in between
				// Server sends events in groups with the usual sequence of begin - table map - rows - gtid/commit
				// So as of committing this, the unresolvable path is not expected to be hit 
				sourceTableName, known := tableIdToName[tableID]
```

flow/connectors/mysql/cdc_test.go
ilidemi: If we make this an integration test, it will still be much faster than running Temporal. The binlog incident one finishes in sub-second

flow/shared/exceptions/mysql.go
ilidemi: Trade-off seems ok to me, given the user-friendliness benefits. Thanks for doing the work

ilidemi (approved):

flow/connectors/mysql/cdc.go
dtunikov: added benign cases:
https://github.com/PeerDB-io/peerdb/pull/4530/commits/1546491e1e0c7effd14adb845903c7a1ea880cdf
wdyt?

flow/connectors/mysql/cdc_test.go
dtunikov: ok
i think it's enough to keep an e2e test as is and remove this unit test completely

ilidemi (approved):

PR #4534, opened 2026-07-03
clickhouse: fix date overflow for initial load

dtunikov:
clickhouse date types we currently use have min/max constraints (1900-2299)
this PR adds rounding logic for the dates outside this range


flow/model/qvalue/avro_converter.go
```diff
@@ -383,26 +392,44 @@ func (c *QValueAvroConverter) processGeneralTime(t time.Time, format string, avr
...
 	case protos.DBType_SNOWFLAKE:
 		// Snowflake has issues with avro timestamp types, returning as string form
 		// See: https://stackoverflow.com/questions/66104762/snowflake-date-column-have-incorrect-date-from-avro-file
 		s := t.Format(format)
 		return s, stringSize(s, so)
 	}
+	// Timestamps are micros since epoch; dates are days since epoch. Both encoded as Avro varint.
+	avroVal := t.UnixMicro()
+	if isDate {
+		avroVal = t.Unix() / 86400
```
dtunikov, line 405: had to move it here out of process[type] functions because we clamp `t` value for clickhouse

.github/workflows/flow.yml
```diff
@@ -352,9 +352,9 @@ jobs:
         id: ch-version
         run: |
           if [ "${{ matrix.db-version.ch }}" = "lts" ]; then
-            echo "ch_version=v25.8.11.66-lts" >> $GITHUB_OUTPUT
+            echo "ch_version=v25.8.28.1-lts" >> $GITHUB_OUTPUT
```
dtunikov, line 355: it was consistently failing on `v25.8.11.66-lts`: https://github.com/PeerDB-io/peerdb/actions/runs/28783586496/job/85344784176
```
    clickhouse_mysql_test.go:1822: 
        	Error Trace:	/home/runner/work/peerdb/peerdb/flow/e2e/clickhouse_mysql_test.go:1822
        	            				/home/runner/work/peerdb/peerdb/flow/e2e/clickhouse_mysql_test.go:1828
        	            				/opt/hostedtoolcache/go/1.26.4/x64/src/runtime/asm_amd64.s:1771
        	Error:      	Not equal: 
        	            	expected: "1900-01-01"
        	            	actual  : ""
```

and what is weird is that clickhouse holds correct value
but clickhouse-go driver seems to interpret it incorrectly...
v25.8.28.1 is the latest patch of the oldest supported release (clickhouse LTS support seems to be 1 year).
so, I think it's ok if we test on the patched version, right?
i can also investigate further why `v25.8.11.66` wasn't working properly, but it seems out-of-scope as it works fine on patched version.

jgao54 (approved):

ilidemi (approved):

.github/workflows/flow.yml
ilidemi: Seems niche enough, yeah

PR #4535, opened 2026-07-03
fix(mysql): align JSON representation between snapshot and CDC

dtunikov:
## Before -> After:
```
inserted: {"name": "a", "id": 1, "val": 1.0, "pi": 3.14}
SNAPSHOT  SELECT js: {"id": 1, "pi": 3.14, "val": 1.0, "name": "a"}
RenderJSONAsMySQLText OFF (old): {"id":1,"name":"a","pi":3.14,"val":1}
RenderJSONAsMySQLText ON  (fix): {"id":1,"pi":3.14,"val":1.0,"name":"a"}
SNAPSHOT compacted (fix): {"id":1,"pi":3.14,"val":1.0,"name":"a"}
```

## The three divergences (snapshot vs. old CDC, flag OFF)
Comparing {"id": 1, "pi": 3.14, "val": 1.0, "name": "a"} against {"id":1,"name":"a","pi":3.14,"val":1}:
- Key order — snapshot is MySQL's JSONB order (length-then-bytes: id(2), pi(2), val(3), name(4)); old CDC went through map[string]any → json.Marshal sorted lexicographically (id, name, pi, val). Different order.
- 1.0 → 1 — the val double lost its fractional part in the old CDC path (bare float64 → Go marshals whole numbers without .0), while snapshot keeps 1.0.
- Whitespace — snapshot has spaces after : and ,; CDC is compact.

## Notes
- On mysql 5.7 `SNAPSHOT  SELECT` returns `1` instead of `1.0`, not much we can do about that. 
- why not using json.Marshal? It introduces silent data corruption for big ints.
```
source (MySQL text) : {"id": 1, "sf": 1234567890123456789, "val": 1.0, "amt": 9007199254740993}
json.Marshal path   : {"amt":9007199254740992,"id":1,"sf":1234567890123456800,"val":1}
json.Compact path   : {"id":1,"sf":1234567890123456789,"val":1.0,"amt":9007199254740993}   ← current fix
```


flow/connectors/mysql/cdc.go
```diff
@@ -343,20 +343,21 @@ func (c *MySqlConnector) startSyncer(ctx context.Context, env map[string]string)
...
+		Host:                  config.Host,
+		Port:                  uint16(config.Port),
+		User:                  config.User,
+		Password:              config.Password,
+		Logger:                internal.SlogLoggerFromCtx(ctx),
+		Dialer:                c.Dialer(),
+		DisableRetrySync:      true,
+		UseDecimal:            true,
+		ParseTime:             true,
+		RenderJSONAsMySQLText: true,
```
dtunikov, line 395: this one was added (really weird git diff because of go fmt)
we can also put it behind mirror version gate to ensure that the old mirrors continue working the same way (in case someone relies on default parsing logic)

flow/connectors/mysql/qvalue_convert.go
```diff
@@ -331,7 +341,8 @@ func QValueFromMysqlFieldValue(qkind types.QValueKind, mytype byte, fv mysql.Fie
 		case types.QValueKindBytes:
 			return types.QValueBytes{Val: slices.Clone(v)}, nil
 		case types.QValueKindJSON:
-			return types.QValueJSON{Val: string(v)}, nil
+			// keep snapshot and cdc json representation consistent
+			return types.QValueJSON{Val: compactMySQLJSON(v)}, nil
```
dtunikov, line 346: same here, could gate behind internal mirror version to avoid inconsistency

flow/connectors/mysql/cdc.go
dtunikov: link to documentation for this flag if you want to read it:
https://github.com/go-mysql-org/go-mysql/blob/a57884f25cf0008867559df9012a6498bb17df70/replication/binlogsyncer.go#L111
the main reason we need it is to ensure that DOUBLEs are parsed properly

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -236,6 +236,142 @@ func (s ClickHouseSuite) Test_MySQL_Blobs() {
...
+		require.Equal(s.t, snapshotGot, cdcGot, "snapshot/cdc JSON representation differs for %q", variants[i])
+	}
+
+	env.Cancel(s.t.Context())
+	RequireEnvCanceled(s.t, env)
+}
+
+// Test_MySQL_JSON_SnapshotCDCConsistency_NativeJSON is the native-JSON companion to
+// Test_MySQL_JSON_SnapshotCDCConsistency.
+func (s ClickHouseSuite) Test_MySQL_JSON_SnapshotCDCConsistency_NativeJSON() {
```
dtunikov, line 371: Without PEERDB_CLICKHOUSE_ENABLE_JSON: true (i.e. JSON stored as String), snapshot and CDC can still disagree textually for doubles in magnitude ranges where the two paths format differently. The snapshot path uses MySQL's my_gcvt, which stays in plain decimal (e.g. 1000000.5), while the CDC path uses go-mysql, which switches to scientific notation (1.0000005e+06). The float64 value is identical — only the text differs — so it's harmless for native-JSON destinations (ClickHouse normalizes on parse) but visible for String-typed columns.

flow/connectors/mysql/cdc.go
jgao54: looks like a reasonably safe change to introduce it without putting it behind a flag, given this was an inconsistency between initial snapshot and CDC.

flow/e2e/clickhouse_mysql_test.go
jgao54: today we technically don't support `PEERDB_CLICKHOUSE_ENABLE_JSON` for pg/mysql because ClickHouse JSON does not support non-object JSON types (e.g. number, string); but still good to get test coverage here still.

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -236,6 +236,142 @@ func (s ClickHouseSuite) Test_MySQL_Blobs() {
...
+	srcFullName := s.attachSchemaSuffix(srcTableName)
+	dstTableName := "test_json_repr_dst"
+
+	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(`
+		CREATE TABLE IF NOT EXISTS %s (
+			id serial PRIMARY KEY,
+			js json NOT NULL
+		)`, srcFullName)))
+
+	variants := []string{
```
jgao54: nit: would be good to add a test case for `1.0` to test out type-faithfulness mentioned in the PR description

jgao54 (approved):
one nit, otherwise lgtm

flow/e2e/clickhouse_mysql_test.go
dtunikov: ok.. tried it out

it doesn't work on mysql 5.7 inside non native json test:
https://github.com/PeerDB-io/peerdb/actions/runs/28851748697/job/85568516983?pr=4535
mysql 5.7 collapses `1.0` into `1` on the snapshot path
on the cdc path we get `1.0` as expected with the new flag set to true

sadly, not much we can do about it
mysql 5.7 behavior:
```
INSERT INTO jtest.t (js) VALUES ('{\"n\": [3, 1, 2], \"obj\": {\"y\": 1.0, \"x\": 2}}');
SELECT js FROM jtest.t;
SELECT JSON_TYPE(JSON_EXTRACT(js,'\$.obj.y')) FROM jtest.t;
----
{"n": [3, 1, 2], "obj": {"x": 2, "y": 1}}
DOUBLE
```

the only thing we could do is to extract value type using JSON_TYPE+JSON_EXTRACT, but I think it's too expensive (also will quite complex code which doesn't seem to worth it)

dtunikov: we could also keep `RenderJSONAsMySQLText=OFF` for mysql 5.7
BUT it means that everything else will stay inconsistent on cdc path
i would rather keep it `ON` for all versions and maybe describe this snapshot `1.0` double inconsistency in docs if it's worth it

ilidemi:
Didn't get to this yet, agree the MySQL JSON nuances need a more thorough look

ilidemi (approved):
Cool that a recent change makes this easier.
CH supports extracting larger ints from JSON strings as well https://fiddle.clickhouse.com/1db6559b-7fe3-4640-b9e5-29411baf4c57
Ultra long decimals aren't supported and would need a custom decoder/encoder but we'll wait for a feature request, especially that it doesn't break the pipe like PG did

PR #4537, opened 2026-07-06
fix(mysql): classify event checksum error

dtunikov:
verify event checksum and throw binlon invalid error if invalid

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -17,6 +17,7 @@ import (
 	chproto "github.com/ClickHouse/ch-go/proto"
 	"github.com/ClickHouse/clickhouse-go/v2"
 	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/go-mysql-org/go-mysql/replication"
 	"github.com/jackc/pgerrcode"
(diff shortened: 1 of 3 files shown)


flow/alerting/classifier.go
```diff
@@ -1129,6 +1130,13 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
+	if strings.Contains(err.Error(), replication.ErrChecksumMismatch.Error()) {
```
dtunikov, line 1133: kinda hard to emulate this one on e2e level
but this if statement should handle an error go-mysql returns (it wraps it with %v so we have to assert with string.Contains)

jgao54 (approved):

flow/alerting/classifier.go
ilidemi: Can test with the debug image and
```
SET GLOBAL source_verify_checksum = 0;
SET GLOBAL debug = '+d,corrupt_read_log_event';
```
But yeah, no need for programmatic tests as we can trust go-mysql here

ilidemi (approved):

PR #4538, opened 2026-07-06
fix(mysql): fail fast on partial JSON binlog events (PARTIAL_JSON)

dtunikov:
## Problem

`binlog_row_value_options` is a dynamic, session-settable MySQL variable, so it can be set to `PARTIAL_JSON` at runtime **after** a mirror passes creation-time validation. When enabled, MySQL emits `PARTIAL_UPDATE_ROWS_EVENT` (event type 39) carrying JSON diffs instead of full column values.

go-mysql parses these into a `*replication.RowsEvent`, so the event reaches PeerDB's `case *replication.RowsEvent:` normally. But the inner event-type switch had **no case for `PARTIAL_UPDATE_ROWS_EVENT` and no `default`** — so the entire update (all columns, not just the JSON one) was **silently dropped** after its bytes were already counted as consumed. CDC advanced past it → silent data loss.

## Fix

- Add a `PARTIAL_UPDATE_ROWS_EVENT` case that fails fast with a new classified error (`MySQLUnsupportedBinlogRowValueOptionsError` → `NOTIFY_BINLOG_PARTIAL_JSON_UNSUPPORTED`), instructing the user to disable `binlog_row_value_options` and resync.

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -145,6 +145,9 @@ var (
 	ErrorNotifyBinlogEventExceededMaxAllowedPacket = ErrorClass{
 		Class: "NOTIFY_BINLOG_EVENT_EXCEEDED_MAX_ALLOWED_PACKET", action: NotifyUser,
 	}
+	ErrorNotifyBinlogPartialJsonUnsupported = ErrorClass{
+		Class: "NOTIFY_BINLOG_PARTIAL_JSON_UNSUPPORTED", action: NotifyUser,
+	}
 	ErrorNotifyBinlogRowMetadataInvalid = ErrorClass{
 		Class: "NOTIFY_BINLOG_ROW_METADATA_INVALID", action: NotifyUser,
 	}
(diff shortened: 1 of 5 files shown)


dtunikov:
Companion PRs (final links):
- Docs FAQ: https://github.com/ClickHouse/clickhouse-docs/pull/6494
- Platform notification mapping: https://github.com/ClickHouse/clickpipes-platform/pull/6647

jgao54 (approved):

flow/alerting/classifier.go
```diff
@@ -1119,6 +1122,13 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
+	if _, ok := errors.AsType[*exceptions.MySQLUnsupportedBinlogRowValueOptionsError](err); ok {
+		return ErrorNotifyBinlogPartialJsonUnsupported, ErrorInfo{
```
ilidemi, line 1126: If we're doing custom notification, can include AdditionalAttributes here as well

PR #4548, opened 2026-07-07
Renovate simulation script

pfcoperez:
This PR adds a script to run Renovate locally in order to validate `renovate.json5` configuration.

:memo: In the future, it can be added as a CI build validation workflow executed only upon changes to `renovate.json5` file.


scripts/renovate-simulation.sh
ilidemi, line 1: Let's have it in scripts/ - top level is definitely crowded and this one doesn't need to be called often

ilidemi (approved):

PR #4550, opened 2026-07-08
address partition overflow bug

jgao54:
This PR addresses the overflow bugs by moving the `computeNumericRanges` method into a shared generic function that applies for both int64 and uint64. Int64 can overflow if min value is negative. Also, both int64 and uint64 can overflow with `r.IntRange.Start+size*(i+1)`/`r.UintRange.Start+size*(i+1)` on the last partition. 

Another bug fix is when span is smaller than partition count, previous logic would blindly honors `numPartitions` and generate invalid partitions (where start exceeds end). This PR fixes this so it will stop as soon as max value is reached. 

Fixes: DBI-822

### DIFF
--- a/flow/connectors/utils/partition_test.go
+++ b/flow/connectors/utils/partition_test.go
@@ -1,0 +1,138 @@
+package utils
+
+import (
+	"log/slog"
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.temporal.io/sdk/log"
+
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+)
+
+func intRangeOf(t *testing.T, partition *protos.QRepPartition) *protos.IntPartitionRange {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


flow/connectors/utils/partition_test.go
```diff
@@ -0,0 +1,138 @@
...
+	})
+
+	t.Run("span smaller than partition count yields fewer partitions", func(t *testing.T) {
+		helper := NewPartitionHelper(logger)
+		require.NoError(t, helper.AddPartitionsWithRange(int64(0), int64(5), 10))
+
+		partitions := helper.GetPartitions()
+		require.Len(t, partitions, 5)
+		require.Equal(t, int64(5), intRangeOf(t, partitions[len(partitions)-1]).End)
+	})
```
jgao54, line 137: these tests were all failing before the fix

ilidemi: 😱

ilidemi (approved):

PR #4552, opened 2026-07-08
fix(tests): update peer config to avoid using stale config from previous local run

dtunikov:


flow/e2e/congen.go
```diff
@@ -41,7 +41,7 @@ func CreatePeer(t *testing.T, peer *protos.Peer) {
 	t.Helper()
 	pool, err := catalogTestAccessPool()
 	require.NoError(t, err)
-	res, err := utils.CreatePeerNoValidate(t.Context(), shared.CatalogPool{Pool: pool}, peer, false)
+	res, err := utils.CreatePeerNoValidate(t.Context(), shared.CatalogPool{Pool: pool}, peer, true)
```
dtunikov, line 44: otherwise local debug gets confusing
if you run with mysql-gtid and later with mysql-pos
mysql-pos would use the old config and try to reach port 3306 even if mysql_port env is set to 3307..
got this problem a few times when was working on mysql connector recently

ilidemi (approved):
Awesome

PR #4553, opened 2026-07-08
fix(mysql): pull INVISIBLE/GIPK columns during snapshot

dtunikov:
## Problem

MySQL/MariaDB QRep snapshot took a `SELECT *` fast path in `buildSelectedColumns`. Wildcard expansion **omits INVISIBLE columns** — including the generated invisible primary key (`my_row_id`) from `sql_generate_invisible_primary_key`. But the destination table (created from `information_schema.columns`) and CDC (binlog row images) both **include** those columns.

Result: snapshot rows keep destination defaults forever while CDC populates real values — silent divergence. Worst case is GIPK, where the invisible `my_row_id` is the table's only PK, so snapshot rows arrive with no key, breaking dedup/ordering.

## Fix

Always build an explicit column list from the fetched table schema (drop the `SELECT *` fast path). The schema already contains invisible/GIPK columns, and naming them explicitly bypasses wildcard invisibility, so the snapshot pulls them — matching CDC and the destination DDL.

## Tests

- Unit: `buildSelectedColumns` never returns `*`; invisible/GIPK columns are listed explicitly.
- E2E (`clickhouse_mysql_test.go`): `Test_MySQL_Invisible_Column_Consistency` and `Test_MySQL_GIPK_Consistency` compare snapshot vs CDC values (version-gated; GIPK is MySQL-only).

### DIFF
--- a/flow/e2e/clickhouse_mysql_test.go
+++ b/flow/e2e/clickhouse_mysql_test.go
@@ -2460,9 +2460,6 @@ func (s ClickHouseSuite) Test_MySQL_GIPK_Consistency() {
 	srcFullName := s.attachSchemaSuffix(srcTableName)
 	dstTableName := "test_my_gipk_dst"
 
-	// GIPK is controlled by a session variable; the connector holds a single persistent connection,
-	// so the setting sticks for the subsequent CREATE. Reset it afterwards to avoid leaking into
-	// other tables in the suite.
 	require.NoError(s.t, s.source.Exec(s.t.Context(), `SET SESSION sql_generate_invisible_primary_key = ON`))
 	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(`
 		CREATE TABLE IF NOT EXISTS %s (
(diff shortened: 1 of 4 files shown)


flow/e2e/clickhouse_mysql_test.go
```diff
@@ -2378,3 +2378,132 @@ func (s ClickHouseSuite) Test_MySQL_NoPrimaryKey_CDC() {
...
+		s.t.Skip("server version does not support generated invisible primary key")
+	}
+
+	srcTableName := "test_my_gipk"
+	srcFullName := s.attachSchemaSuffix(srcTableName)
+	dstTableName := "test_my_gipk_dst"
+
+	// GIPK is controlled by a session variable; the connector holds a single persistent connection,
+	// so the setting sticks for the subsequent CREATE. Reset it afterwards to avoid leaking into
+	// other tables in the suite.
```
ilidemi: Misleading comment, connections are per test. Ok to reset or not reset

ilidemi (approved):
The task will need e2e checking on how this shows up in ClickPipes UI and whether it passes validations

ilidemi:
Oh CI doesn't pass, something's up

dtunikov:
tested e2e with other clickpipes components, works fine

PR #4556, opened 2026-07-09
feat(api): add endpoint for pg-pg source and destination row count check

Amogh-Bharadwaj:
## Why
This is useful for validating mirror correctness after initial load and zero-lag CDC for PG - PG mirrors.

## What
- Adds new proto definitions for a route
- New handler in flow-api for mirror row count check, scoped to PG to PG mirrors.
- Adds api test for this new route

## What the endpoint does
User can pick a selection of tables to validate counts for.
Counts for source and destination obtained via two goroutines.
- For tables greater than 10GB -- gets an approximate row count as counts can take a long time
- For tables lesser than 10GB -- gets exact count
Count times capped to 30 seconds for each table. Response structure denotes what the counts are and whether they're approximate

- [x] Functionally tested

### DIFF
--- a/flow/cmd/row_counts.go
+++ b/flow/cmd/row_counts.go
@@ -25,6 +25,11 @@ const (
 	countTimeoutSeconds = 30
 )
 
+// GetMirrorRowCounts returns per-table source and destination row counts for a
+// Postgres-to-Postgres CDC mirror, so the caller can compare the two sides and spot drift.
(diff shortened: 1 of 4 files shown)


flow/cmd/row_counts.go
ilidemi, line 1: Codeowners please

ilidemi (approved):

PR #4559, opened 2026-07-09
maria: classify compressed columns errors + add validation to `create mirror`

dtunikov:
mariadb has `COMPRESSED` column feature
go-mysql can't decode such columns
this PR adds classified error when such column is added mid cdc stream
it also validates the source and errors out if there are compressed columns


dtunikov:
I went with catching COMPRESSED column error as a distinct error with it's own classifier type.
Another option could be to catch `unsupport type % in binlog and don't know how to handle` error that go-mysql returns when it encounters events it can't decode.
I prefer to be more explicit and keep error classification narrow to known issue.

flow/connectors/mysql/cdc.go
```diff
@@ -732,13 +733,17 @@ func (c *MySqlConnector) PullRecords(
 				}
 			}
 		case *replication.TableMapEvent:
-			tableIdToName[ev.TableID] = string(ev.Schema) + "." + string(ev.Table)
+			schemaTable := string(ev.Schema) + "." + string(ev.Table)
```
pfcoperez: nit/q: Any reason for this change?

dtunikov: not really, it's just left from the previous iteration on this branch (i tested a few approaches)
i'll rollback it

flow/connectors/mysql/cdc.go
```diff
@@ -1018,6 +1023,51 @@ func (c *MySqlConnector) PullRecords(
...
+func decodeRowsEvent(ev *replication.RowsEvent, data []byte) error {
+	pos, err := ev.DecodeHeader(data)
+	if err != nil {
+		return err
+	}
+
+	err = ev.DecodeData(pos, data)
+	if err != nil {
+		// avoid checking for every row event
+		if checkTableMapForCompressedColumns(ev.Table) != nil {
```
pfcoperez, line 1124: The rationale here is to skip upon detection as errors are bound to happen for these events, right?

flow/connectors/mysql/cdc.go
```diff
@@ -739,6 +740,11 @@ func (c *MySqlConnector) PullRecords(
 			exclusion := req.TableNameMapping[sourceTableName].Exclude
 			schema := req.TableNameSchemaMapping[destinationTableName]
 			if schema != nil {
+				if isMariaDb {
+					if err := checkTableMapForCompressedColumns(ev.Table); err != nil {
```
dtunikov, line 823: checkTableMapForCompressedColumns iterates over ev.ColumnType []byte to identify compressed columns
shoudn't be an issue to keep it on this RowsEvent path
though, we could add an additional guard like ev.Rows == nil before running the check
if we want to minimize unnecesary cpu consumption here

flow/connectors/mysql/cdc.go
dtunikov: yeah
if a column is compressed, DecodeData would return an error
but we don't want to fail with generic `unsupport type % in binlog and don't know how to handle` error
instead, we want to continue without an error and fail with a proper error later when we handle RowsEvent

the main reason to have such weird workflow is to avoid failing on tables that don't belong to the pipe

flow/pkg/mysql/validation.go
```diff
@@ -299,6 +299,42 @@ func IsVitess(conn *client.Conn) (bool, error) {
 	return true, nil // is a Vitess server
 }
 
+func IsCompressedColumnType(columnType string) bool {
+	return strings.Contains(strings.ToUpper(columnType), "COMPRESSED")
```
pfcoperez: Is it possible that a column type containing the word "compressed" matches this? e.g:

```sql
ENUM('flat', 'compressed')
```

dtunikov: yep, good point, fixed. 
using regex now to catch it in maria executable comment format only.
tested and column_type is always returned as executable comment from info schema table.

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -122,6 +122,151 @@ func (s ClickHouseSuite) Test_MySQL_MyISAM_NonTransactional() {
...
+	if !ok {
+		s.t.Skip("only applies to mysql")
+	}
+	if mySource.Config.Flavor != protos.MySqlFlavor_MYSQL_MARIA {
+		s.t.Skip("column compression is a MariaDB-only feature")
+	}
+
+	// A compressed column produces a valid MariaDB binlog event that go-mysql cannot decode.
+	// Because every CDC reader consumes the server-wide binlog before table filtering, the event
+	// can terminate unrelated mirrors using the shared CI server. Use a dedicated throwaway
```
ilidemi: As the events from unrelated tables are expected to be filtered out, it shouldn't impact other tests in the normal case (or will fail a lot of them loudly if there's a regression). I think it's ok to use the shared instance and not increase resource consumption

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -122,6 +122,151 @@ func (s ClickHouseSuite) Test_MySQL_MyISAM_NonTransactional() {
...
+	// Seeing the snapshot row at the destination does not guarantee that the CDC activity has
+	// started. Prove that CDC is consuming this source before introducing the unsupported type.
+	require.NoError(s.t, src.Exec(s.t.Context(), fmt.Sprintf(
+		`INSERT INTO %s (id, name) VALUES (2, 'cdc-ready')`, srcFullName)))
+	EnvWaitForCount(env, s, "waiting for CDC to start", dstTableName, "id,name", 2)
+
+	// Add a COMPRESSED column mid-stream, then write a row so a row event carrying the undecodable
+	// wire type reaches the pipe and trips the TABLE_MAP detection.
+	require.NoError(s.t, src.Exec(s.t.Context(), fmt.Sprintf(
+		`ALTER TABLE %s ADD COLUMN val TEXT COMPRESSED`, srcFullName)))
```
ilidemi: Would be cool to add an explicit test where the column is added to an unrelated table and the main one still does a batch without erroring

flow/connectors/mysql/type_conversion.go
```diff
@@ -5,11 +5,15 @@ import (
...
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	mysql_validation "github.com/PeerDB-io/peerdb/flow/pkg/mysql"
 	"github.com/PeerDB-io/peerdb/flow/shared"
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
 
 func QkindFromMysqlColumnType(ct string, binlogRowMetadataSupported bool, version uint32) (types.QValueKind, error) {
+	if mysql_validation.IsCompressedColumnType(ct) {
+		return types.QValueKindInvalid, fmt.Errorf("MariaDB COMPRESSED columns are not supported: %s", ct)
+	}
```
ilidemi, line 16: Am I getting it right that this one is exercised in setup (if snuck in after validation) and in DDL column add? If so, the e2e tests can validate the setup/snapshot path (mirror with validation disabled), the DDL add path (already done) and the binlog event parsing path (like in #4557) separately to make sure we have full coverage

ilidemi (approved):

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -206,6 +206,203 @@ func (s ClickHouseSuite) Test_MySQL_MyISAM_NonTransactional() {
...
+	catalogPool, err := internal.GetCatalogConnectionPoolFromEnv(s.t.Context())
+	require.NoError(s.t, err)
+
+	EnvWaitForCount(env, s, "waiting on initial snapshot", dstTableName, "id,name", 1)
+	// Seeing the snapshot row at the destination does not guarantee that the CDC activity has
+	// started. Prove that CDC is consuming this source before introducing the unsupported type.
+	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(
+		`INSERT INTO %s (id, name) VALUES (2, 'cdc-ready')`, srcFullName,
+	)))
+	EnvWaitForCount(env, s, "waiting for CDC to start", dstTableName, "id,name", 2)
```
ilidemi: We capture the initial binlog offset at snapshot start (not end) so if the add column + insert happens before CDC starts, the events will still get processed. Safe to remove

flow/pkg/mysql/validation.go
```diff
@@ -299,6 +300,48 @@ func IsVitess(conn *client.Conn) (bool, error) {
...
+// rather than a bare "COMPRESSED" substring, which would false-positive on values
+// like enum('compressed').
+var compressedColumnTypeRe = regexp.MustCompile(`(?i)/\*M!\d+\s+COMPRESSED\s*\*/`)
+
+func IsCompressedColumnType(columnType string) bool {
+	return compressedColumnTypeRe.MatchString(columnType)
+}
+
+func CheckCompressedColumns(conn *client.Conn, schema string, table string) error {
+	rs, err := conn.Execute(fmt.Sprintf(
```
ilidemi, line 329: Just realized that this would double the # of queries during validation (1->2 per table) - if there's a 1000 tables and the DB is across the continent, that can get slow. Would it work if we accepted a list of all tables and did ExecuteMultiple or a manual bulk Execute query?

ilidemi: Validation we have right now in ClickPipes could only work with ExecuteMultiple, but we've seen issues like https://github.com/go-mysql-org/go-mysql/issues/1056#issuecomment-4057351009 so would need to test with an old patch version before switching to that. So simple Execute is safer

ilidemi (approved):

flow/e2e/clickhouse_mysql_test.go
dtunikov: so, this one is a bit weird
i was struggling a lot to make this test work without waiting for this initial cdc row
if we remove these lines and run alter table add compressed right after mirror creation:
```
	env := ExecutePeerflow(s.t, tc, flowConnConfig)
	SetupCDCFlowStatusQuery(s.t, env, flowConnConfig)
	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(
		`ALTER TABLE %s ADD COLUMN val TEXT COMPRESSED`, notInPipeFullName,
	)))
	require.NoError(s.t, s.source.Exec(s.t.Context(), fmt.Sprintf(
		`INSERT INTO %s (id, name, val) VALUES (3, 'cdc', 'compressed value')`, notInPipeFullName,
	)))
```
the test will fail with:
```
message=MySQL execute error: Header &replication.EventHeader***Timestamp:0x6a511011, EventType:0x17, ServerID:0x1, EventSize:0x3e, LogPos:0x0, Flags:0x0***, Data "\xd8\x00\x00\x00\x00\x00\x01\x00\x03\a\xf8\x03\x00\x00\x00\x03\x00cdc\x11\x00\x00compressed value", Err: unsupport type 140 in binlog and don't know how to handle
```
(example: https://github.com/PeerDB-io/peerdb/actions/runs/29102567925/job/86394894861)

i don't see a stacktrace, but I think it might be happening at `syncer.StartSync` or so...

dtunikov: It looks like a very edge case to me, but I can try to confirm that assumption and maybe handle this error in other places in the code as well.

flow/pkg/mysql/validation.go
dtunikov: like this should be fine, right?
```
	rs, err := conn.Execute(fmt.Sprintf(
		"SELECT table_schema, table_name, column_name, column_type FROM information_schema.columns "+
			"WHERE (table_schema, table_name) IN (%s)",
		strings.Join(predicates, ", ")))
```

flow/e2e/clickhouse_mysql_test.go
ilidemi: AI tells me that this failure was before introducing RowsEventDecodeFunc and there weren't any after. Care to test again?

flow/pkg/mysql/validation.go
ilidemi: I think so

PR #4562, opened 2026-07-10
fix(ui): fix schema prefixing

### LINKED ISSUE #4561: UI: Target Schema override no longer applies destination prefix to all tables (ClickHouse) — regression from v0.36.2
### Describe the bug

When creating a CDC mirror to a **ClickHouse** destination, the **Target Schema** override (the `SchemaSettings` field in each schema box) no longer applies the destination-table prefix to the tables. Every destination table is stuck with the source-schema prefix (e.g. `public_<table>`), regardless of what the override is set to.

This is a regression. In older versions (verified on `v0.36.2`), editing the Target Schema override re-applied the new prefix to **all** tables in the schema at once. This is essential when a mirror has hundreds of tables — the alternative is editing every row's destination name by hand.

### Steps to reproduce

1. Create a new CDC mirror, Postgres source → ClickHouse destination.
2. Expand a source schema (e.g. `public`) so its tables load. Destination names default to `public_<table>`.
3. Change the **Target Schema** override to a custom value (e.g. `myprefix`).
4. Observe the destination table names.

### Expected behavior

All destination table names update to use the new prefix — `myprefix_<table>` — as they did in `v0.36.2`.

### Actual behavior

Nothing happens. The destination names stay `public_<table>`. There is no way to bulk-apply a prefix from the UI; each of the (potentially hundreds of) rows would have to be edited manually.

### Root cause (source analysis)

The destination-name logic itself is unchanged between `v0.36.2` and `v0.36.32` — `getDefaultDestinationTable` still builds `` `${schemaName}_${tableName}` `` from the passed `targetSchemaName`:
- https://github.com/PeerDB-io/peerdb/blob/v0.36.32/ui/app/mirrors/create/handlers.ts#L393-L401

The regression is in `ui/app/mirrors/create/cdc/schemabox.tsx`, in **when the fetch is triggered**:

- In `v0.36.2`, a `useEffect` re-ran the fetch (which recomputes destination names via the override) whenever `fetchTablesForSchema` changed — and `fetchTablesForSchema` depends on `defaultTargetSchema`. So changing the override re-applied the prefix to all tables:
  - https://github.com/PeerDB-io/peerdb/blob/v0.36.2/ui/app/mirrors/create/cdc/schemabox.tsx#L272-L276

- In `v0.36.32`, that `useEffect` was removed. Fetching now happens only once per schema, guarded by a new `fetchedSchemas` set in `handleSchemaClick`:
  - https://github.com/PeerDB-io/peerdb/blob/v0.36.32/ui/app/mirrors/create/cdc/schemabox.tsx#L257-L268

- Compounding this, the `SchemaSettings` (Target Schema override) input only renders **after** a schema is expanded, i.e. after that single fetch has already run with the default (`public`):
  - https://github.com/PeerDB-io/peerdb/blob/v0.36.32/ui/app/mirrors/create/cdc/schemabox.tsx#L292-L316

Result: there is no longer any code path that re-applies the Target Schema override to already-fetched tables, so the override is effectively non-functional for setting destination prefixes.

### Suggested fix

Re-fetch (or recompute destination names for) a schema's already-fetched tables when `defaultTargetSchema` changes — e.g. restore an effect keyed on `defaultTargetSchema`, or update existing rows' `destination` in place when the override changes, so the prefix applies to all tables without a manual per-row edit.

### Environment

- PeerDB version: `v0.36.32` (regressed) vs `v0.36.2` (working)
- Destination: ClickHouse
- Deployment: self-hosted (docker-compose)

---

cc @serprex @Amogh-Bharadwaj @iskakaushik — flagging as this blocks setting destination prefixes on large-table ClickHouse mirrors from the UI.

Amogh-Bharadwaj:
Fixes https://github.com/PeerDB-io/peerdb/issues/4561

### DIFF
--- a/ui/app/mirrors/create/cdc/schemabox.tsx
+++ b/ui/app/mirrors/create/cdc/schemabox.tsx
@@ -22,15 +22,17 @@ import {
   Dispatch,
   SetStateAction,
   useCallback,
-  useEffect,
   useMemo,
-  useRef,
   useState,
   useTransition,
 } from 'react';
 import ReactSelect from 'react-select';
 import { useTheme as useStyledTheme } from 'styled-components';
-import { fetchColumns, fetchTables, getDefaultDestinationTable } from '../handlers';
+import {
+  fetchColumns,
+  fetchTables,
+  getDefaultDestinationTable,
+} from '../handlers';
 import ColumnBox from './columnbox';
 import CustomColumnType from './customColumnType';
 import SchemaSettings from './schemasettings';
@@ -77,27 +79,25 @@ export default function SchemaBox({
   const [tableQuery, setTableQuery] = useState<string>('');
   const [defaultTargetSchema, setDefaultTargetSchema] =
     useState<string>(schema);
-  const prevTargetSchema = useRef<string>(schema);
 
-  useEffect(() => {
-    if (prevTargetSchema.current !== defaultTargetSchema && peerType) {
-      prevTargetSchema.current = defaultTargetSchema;
-      setRows((oldRows) =>
-        oldRows.map((row) => {
-          if (row.schema !== schema || row.editingDisabled) return row;
-          const tableName = row.source.split('.').slice(1).join('.');
-          return {
-            ...row,
-            destination: getDefaultDestinationTable(
-              peerType,
-              defaultTargetSchema,
-              tableName
-            ),
-          };
-        })
-      );
-    }
-  }, [defaultTargetSchema, peerType, schema, setRows]);
+  const applyTargetSchemaOverride = (newSchema: string) => {
+    setDefaultTargetSchema(newSchema);
+    if (peerType === undefined) return;
+    setRows((oldRows) =>
+      oldRows.map((row) =>
+        row.schema !== schema || row.editingDisabled
+          ? row
+          : {
+              ...row,
+              destination: getDefaultDestinationTable(
+                peerType,
+                newSchema,
+                row.source.slice(schema.length + 1)
+              ),
+            }
+      )
+    );
+  };
   const searchedTables = useMemo(() => {
     const tableQueryLower = tableQuery.toLowerCase();
     return rows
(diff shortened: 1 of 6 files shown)


ui/app/mirrors/create/cdc/schemabox.tsx
```diff
@@ -75,6 +77,27 @@ export default function SchemaBox({
   const [tableQuery, setTableQuery] = useState<string>('');
   const [defaultTargetSchema, setDefaultTargetSchema] =
     useState<string>(schema);
+  const prevTargetSchema = useRef<string>(schema);
+
+  useEffect(() => {
+    if (prevTargetSchema.current !== defaultTargetSchema && peerType) {
```
serprex:
```suggestion
    if (prevTargetSchema.current !== defaultTargetSchema && peerType !== undefined) {
```
avoids being false when peerType 0 (`DBType.BIGQUERY`)

ui/app/mirrors/create/cdc/schemabox.tsx
```diff
@@ -75,6 +77,27 @@ export default function SchemaBox({
   const [tableQuery, setTableQuery] = useState<string>('');
   const [defaultTargetSchema, setDefaultTargetSchema] =
     useState<string>(schema);
+  const prevTargetSchema = useRef<string>(schema);
```
serprex: instead we could wrap setDefaultTargetSchema:
```ts
const applyTargetSchemaOverride = (newSchema: string) => {
  setDefaultTargetSchema(newSchema);
  if (peerType === undefined) return;
  setRows((oldRows) =>
    oldRows.map((row) =>
      row.schema !== schema || row.editingDisabled
        ? row
        : {
            ...row,
            destination: getDefaultDestinationTable(
              peerType, newSchema, row.source.slice(schema.length + 1)
            ),
          }
    )
  );
};
```
no need for `useEffect` & `useRef`

serprex (approved):

PR #4564, opened 2026-07-10
DBI-798: Safer check on mirror validation with MySQL sources setting server_id

pfcoperez:
It adds a check on the utilization of the same `server_id` in the source MySQL server.
This offer an extra layer of protection against collisions with other replication actors.

### DIFF
--- a/flow/cmd/validate_mirror.go
+++ b/flow/cmd/validate_mirror.go
@@ -16,7 +16,6 @@ import (
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
 	"github.com/PeerDB-io/peerdb/flow/internal"
 	"github.com/PeerDB-io/peerdb/flow/pkg/common"
-	mysql_validation "github.com/PeerDB-io/peerdb/flow/pkg/mysql"
 	"github.com/PeerDB-io/peerdb/flow/shared"
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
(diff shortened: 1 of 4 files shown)


jgao54 (approved):

flow/cmd/validate_mirror.go
```diff
@@ -186,6 +188,25 @@ func (h *FlowRequestHandler) checkSourcePeerReuse(
 		return nil
 	}
 
+	// beyond other PeerDB mirrors (checked below), the pinned server_id must not be already registered
+	// as a replica on the source DB itself.
+	mysqlConn, mysqlClose, err := connectors.GetAs[*connmysql.MySqlConnector](ctx, cfg.Env, peer)
```
jgao54: nit: can avoid an extra connect/disconnect by moving `mysql_validation.HasReplicaWithServerId` inside `ValidateMirrorSource`  and pass in an already instantiated `conn`, although i understand that having this here put server_id validation check together. Your call.

jgao54:
(pending integrations test fixes)

flow/cmd/validate_mirror.go
pfcoperez: I followed your suggestion at 19f432586414050b5bf3d8033610a6132bbb1c31, it not only saves the extra connection cycle but also cleans-up several boundary leaks.

PR #4567, opened 2026-07-10
classify error as retryable when pipe closed in deadlineCappableConn

jgao54:
Classify a transient error that can happen with DeadlineCapableConn. In https://github.com/PeerDB-io/peerdb/pull/4380 we introduced an in-memory transport between connector and ssh using net.Pipe() to allow it to respect deadlines. This introduced a scenario where if the ssh tunnel is processing read/write when the pipe is closed, we get an `io.ErrClosedPipe` that ended up unhandled. The error is wrapped in an `SSHTunnelClosedError` to be precise, to avoid blindly handle all `io.ErrClosedPipe` that could hide future unrelated bugs. 

Fixes: DBI-896

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -442,6 +442,13 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		return ErrorOther, errInfo
 	}
 
+	if _, ok := errors.AsType[*exceptions.SSHTunnelClosedError](err); ok {
+		return ErrorRetryRecoverable, ErrorInfo{
+			Source: ErrorSourceSSH,
+			Code:   "TUNNEL_CONNECTION_CLOSED",
+		}
+	}
+
 	// An SSH-tunneled connection that goes bad (e.g. keepalive failure) is wrapped explicitly.
 	if _, ok := errors.AsType[*exceptions.SSHTunnelConnectionError](err); ok {
 		return ErrorNotifyConnectivity, ErrorInfo{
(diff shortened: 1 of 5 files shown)


flow/alerting/classifier.go
```diff
@@ -428,6 +428,13 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 		}
 	}
 
+	if _, ok := errors.AsType[*exceptions.SSHTunnelClosedError](err); ok {
+		return ErrorRetryRecoverable, ErrorInfo{
```
pfcoperez, line 446: This is a comment just for my own understanding.

Do these errors need to be notified to telemetry? 

From the PR description and the original PR it seems that they only happen in a certain situation where the pipe is going to be terminated anyway.

No action is required from us or the customer. It seems to me that we would only pay attention to them if we were inspecting logs manually, right?

pfcoperez (approved):
LGTM

flow/alerting/classifier.go
ilidemi: It could be argued to be overly cautious, but if the error is only to expected to occur once, `recoverable` allows us to still notice if it persists in case there's a bug in some of our logic. The "pipe closed" refers to the io.Pipe used for SSH, not the ClickPipe, and repeatedly hitting this error on read could mean the ClickPipe becomes blocked.

ilidemi (approved):

flow/alerting/classifier.go
jgao54: echo @ilidemi  that this leans overly cautious. Because this net.Pipe() solution is somewhat of an unconventional (and creative) workaround, this provides a peace of mind knowing we'd get notified if error persists (vs. unconditionally ignore it in the Read/Write methods).

PR #4569, opened 2026-07-11
fix mongo transient tls error

jgao54:
Mongo Atlas automatically manages TLS cert rotation [ref](https://www.mongodb.com/docs/atlas/setup-cluster-security/#preconfigured-security-features).  Classify this error as retryable. 

Fixes: DBI-874

### DIFF
--- a/flow/alerting/classifier.go
+++ b/flow/alerting/classifier.go
@@ -45,6 +45,7 @@ const (
 	MongoShutdownInProgress              = "(ShutdownInProgress) The server is in quiesce mode and will shut down"
 	MongoInterruptedDueToReplStateChange = "(InterruptedDueToReplStateChange) operation was interrupted"
 	MongoIncompleteReadOfMessageHeader   = "incomplete read of message header"
+	MongoTLSInvalidServerCertSignature   = "tls: invalid signature by the server certificate"
 
 	// mysqlGeometryLinearRingNotClosedError is the specific WKB parse failure raised by the
 	// go-geos library when a LinearRing's points do not close. Used to give a more specific code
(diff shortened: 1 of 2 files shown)


flow/alerting/classifier.go
```diff
@@ -817,6 +818,12 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 			return ErrorNotifyConnectivity, mongoErrorInfo
 		}
 
+		// TLS handshake failures during connection checkout is typically retryable. We've observed
+		// Atlas briefly serving a mismatched cert/key pair during rolling cert rotation that auto-recovers
+		if strings.Contains(err.Error(), MongoTLSInvalidServerCertSignature) {
```
pfcoperez, line 823: When matching errors by their message string we accept the risk of these messages changing in the most subtle ways, one extra space, or semicolon... 

Do you think it would be worth of matching by parts of the message that we  bet to be fairly constant? Something like:


```suggestion
   func mongoTLSInvalidServerCertSignature(err error) bool {
   signatureParts := []string{
	  "tls",
      "invalid signature",
	  "server certificate",
    }

	msg := err.Error()

	for _, chunk := range signatureParts {
		if !strings.Contains(msg, chunk) {
			return false
		}
	}

	return true
}
        
		if mongoTLSInvalidServerCertSignature(err.Error)
		 {
```

jgao54: I'm inclined to keep it as is, the string is not coming from mongo source  but is coming from go standard lib ([ref](https://github.com/search?q=repo%3Agolang%2Fgo%20%22tls%3A%20invalid%20signature%20by%20the%20server%20certificate%22&type=code)), which tend to be more stable. And in the case where the string changes, the blast radius is small (e.g. we get notified). So inclined to use the full string matching pattern here for consistency and searchability.

ilidemi (approved):

PR #4573, opened 2026-07-13
feat: add source/destination/e2e commit lag metrics

dtunikov:
This PR adds new source/destination/e2e commit lag metrics.
- source - lag between commit ts and ts of receiving an event on peerdb side. Recorded from connectors' code directly.
- destination - lag between receiving an event on peerdb side and writing to ch. Recorded from normalize batches shared activity code.
- e2e - roughly the sum of the two above. Recorded from normalize batches shared activity code.

Migration for cdc_batches table was added, it adds two new fields - first_row_received_at and first_row_commit_time. These fields allow us to report destination and e2e lag metrics.


flow/connectors/postgres/cdc.go
```diff
@@ -730,7 +755,7 @@ func PullCdcRecords[Items model.Items](
 
 		if err != nil && pgconn.Timeout(err) {
 			// On timeout, always send a ping before moving on to read more messages
-			if err := sendStandbyAfterReplLock("receive-timeout"); err != nil {
+			if err := sendStandbyAfterReplLock("receive-timeout", true); err != nil {
```
dtunikov: we should request immediate reply when we reached a timeout

flow/connectors/mongo/cdc.go
```diff
@@ -179,6 +203,10 @@ func (c *MongoConnector) PullRecords(
 
 	var resumeToken bson.Raw
 	var err error
+	mongoClockOffset, err := c.getMongoClockOffset(ctx)
+	if err != nil {
+		c.logger.Warn("failed to calculate MongoDB clock offset", slog.Any("error", err))
```
ilidemi, line 261: nit: skipping all errors could make the metric jump around on occasional network issues. I'd treat network errors as errors and "not supported" type of errors as warn

Also a flag, this adds a round-trip per batch, which could become noticeable in a case of far-away Mongo and short sync interval. I think it's ok to keep as is until it proves to be a problem. Maybe add a warn if this call takes >10% of the idle timeout budget, so it's immediately diagnosable?

flow/connectors/postgres/cdc.go
```diff
@@ -758,6 +783,10 @@ func PullCdcRecords[Items model.Items](
 				if err != nil {
 					return fmt.Errorf("ParsePrimaryKeepaliveMessage failed: %w", err)
 				}
+				// A keepalive means there is currently no source change event to lag behind.
+				// Reset the gauge so a previously backlogged, now-idle pipe does not retain
+				// the old source-lag value.
+				p.otelManager.Metrics.CommitLagGauge.Record(ctx, 0)
```
ilidemi: PG does reset the timeout countdown on data events, but when it's decoding a large tx or reading a large row from a throttled volume, it can still send these in between. When the source is backed up like this, this metric would be important in surfacing it. Debezium doesn't seem to be doing resets for PG at all, so maybe that's defensible, and I also wasn't able to find a solid one.

flow/connectors/mongo/cdc.go
ilidemi: Another option, store as a connector field and refresh every hour (still within Pull, so no worrying about locking)

flow/connectors/mysql/cdc.go
```diff
@@ -587,10 +619,34 @@ func (c *MySqlConnector) PullRecords(
...
+		case *replication.GtidTaggedLogEvent:
+			if ev.ImmediateCommitTimestamp > 0 {
+				timestampMicros = int64(ev.ImmediateCommitTimestamp)
+			}
+		}
+
+		if timestampMicros > 0 {
 			otelManager.Metrics.CommitLagGauge.Record(ctx,
-				time.Now().UTC().Sub(time.UnixMicro(int64(commitTs))).Microseconds())
+				time.Now().UTC().Add(time.Duration(mysqlClockOffset)*time.Microsecond).
```
ilidemi: My AI is flagging that mysqlClockOffset is already a parsed time and multiplying by Microsecond corrupts it

flow/otel_metrics/otel_manager.go
```diff
@@ -415,8 +415,8 @@ func (om *OtelManager) setupMetrics(ctx context.Context) error {
 
 	if om.Metrics.CommitLagGauge, err = om.GetOrInitInt64Gauge(BuildMetricName(CommitLagGaugeName),
 		metric.WithUnit("us"),
-		metric.WithDescription("Lag in microseconds from when a change event was committed on the source"+
-			" to when PeerDB processes it; subject to clock skew"),
+		metric.WithDescription("Lag in microseconds from a source event timestamp (transaction commit timestamp when available)"+
+			" to when PeerDB processes it; reset to zero on source heartbeats; subject to clock skew"),
```
ilidemi:
```suggestion
			" to when PeerDB processes it; reset to zero on source heartbeats; best effort is made to mitigate clock skew"),
```

flow/connectors/postgres/cdc.go
```diff
@@ -487,14 +505,21 @@ func PullCdcRecords[Items model.Items](
 	replLock *sync.Mutex,
 ) error {
 	logger := internal.LoggerFromCtx(ctx)
+	postgresClockOffset, err := p.getPostgresClockOffset(ctx)
```
ilidemi, line 531: PG has current server clock in the events, so we may not need to track this at all https://www.postgresql.org/docs/current/protocol-replication.html#PROTOCOL-REPLICATION-XLOGDATA

dtunikov: hm
not sure I understand why it would help us to get rid of `getPostgresClockOffset` logic.
pg doc says that ServerTime is "The server's system clock at the time of transmission, as microseconds since midnight on 2000-01-01.".
so, yeah, it gets stamped right before an event is sent to us, but it's all happening on pg side
i think we still want to capture the diff between the exact moment we got the message on our end (time.Now()), and the timestamp this change was committed on pg side.
`msg.ServerTime - msg.CommitTime` would be skew immune, but it doesn't take into account how long it took to transfer this message to us over a network. (idk what else can contribute to additional delay)
from my pov `time.Now().Add(postgresClockOffset).Sub(msg.CommitTime)` is a better option for our purpose

flow/connectors/postgres/cdc.go
dtunikov: yeah, but just removing this reset would lead to alerts in these cases: "previously backlogged, now-idle pipe".
wdyt about this approach?
```
	if totalRecords == 0 && int64(clientXLogPos) > req.ConsumedOffset.Load() {
		err := p.updateConsumedOffset(ctx, logger, req.FlowJobName, req.ConsumedOffset, clientXLogPos)
		if err != nil {
			return err
		}
		lastEmptyBatchPkmSentTime = time.Now()
		// Caught up with no records for this flow: clear any stale source lag.
		p.otelManager.Metrics.CommitLagGauge.Record(ctx, 0)
	}
```
`totalRecords == 0 && clientXLogPos > ConsumedOffset` means precisely "this whole batch produced no records for our publication, yet our position advanced" — i.e. our pipe is idle-but-progressing.
and another defensive reset on the keep alive path:
```
if pkm.ServerWALEnd <= clientXLogPos {
  p.otelManager.Metrics.CommitLagGauge.Record(ctx, 0)
}
```
this handles a case when we're fully caught up with ServerWALEnd. This complements previous check for the cases when ALL tables are idle.

flow/connectors/mongo/cdc.go
dtunikov: added logic to cache clock offset to every source connector
decided to keep it inside connector's code
another option would be to put it somewhere inside SyncFlow loop if we want to make this caching part "generic", but I think it's clearer to keep it inside PullRecords itself

nexus/peer-cursor/src/util.rs
```diff
@@ -48,7 +48,7 @@ fn encode_value(value: &Value, builder: &mut DataRowEncoder) -> PgWireResult<()>
         Value::Enum(_) | Value::Hstore(_) => Err(PgWireError::ApiError(
             format!(
                 "cannot write value {:?} in postgres protocol: unimplemented",
-                &value
+                value
```
dtunikov, line 51: clippy failed on this one, i guess should be fixed on the main branch

flow/connectors/postgres/cdc.go
ilidemi: That makes sense, I agree

flow/connectors/mongo/mongo.go
```diff
@@ -60,15 +60,17 @@ type createChangeStreamFunc func(
...
+	logger                    log.Logger
+	metadataStore             metadataStore
+	config                    *protos.MongoConfig
+	client                    *mongo.Client
+	ssh                       *utils.SSHTunnel
+	createChangeStream        createChangeStreamFunc
+	excludedOps               []operationType
+	totalBytesRead            atomic.Int64
+	deltaBytesRead            atomic.Int64
+	mongoClockOffset          time.Duration
```
ilidemi: No need for `mongo` on clockOffset and clockOffsetUpdatedAt

flow/connectors/postgres/cdc.go
ilidemi: Will need to think a bit more, brain fried atm

ilidemi (commented):
Looking great. The only piece left is thinking through Postgres carefully

.github/workflows/dev-docker.yml
```diff
@@ -2,7 +2,7 @@ name: Dev Docker Images
 
 on:
   push:
-    branches: [main]
+    branches: [main, feat/commit-lag-metric-improvements]
```
ilidemi: Don't forget to clean up please
Also a way to push a branch image is to prefix it with `customer-` (search for `customer-` in the upgrades doc)

flow/connectors/postgres/cdc.go
ilidemi:
```
if pkm.ServerWALEnd <= clientXLogPos {
  p.otelManager.Metrics.CommitLagGauge.Record(ctx, 0)
}
```
May be a bit confusing but ServerWALEnd is not WAL head but the current decoder position. It doesn't mean we've caught up
https://github.com/postgres/postgres/blob/8ac700acffc7b17d88414be47b8dff44fb1ea681/src/backend/replication/walsender.c#L3607-L3609

```
	if totalRecords == 0 && int64(clientXLogPos) > req.ConsumedOffset.Load() {
		err := p.updateConsumedOffset(ctx, logger, req.FlowJobName, req.ConsumedOffset, clientXLogPos)
		if err != nil {
			return err
		}
		lastEmptyBatchPkmSentTime = time.Now()
		// Caught up with no records for this flow: clear any stale source lag.
		p.otelManager.Metrics.CommitLagGauge.Record(ctx, 0)
	}
```
totalRecords == 0 can happen when PG is terribly terribly slow decoding a large transaction event (low on logical_decoding_work_mem, overloaded CPU, out of EBS iops budget and the event is 1GB). It is possible that we receive a primary keepalive advancing the XLogPos with some events outside of the publication before that, or receive a random row outside of our tables, the large tx data would still be pending and possibly stale in this case.

For this PR let's remove the zeroing out PG part + requestReply to unblock the rest of the work. If we find some creative and working solution later on, it can be PR'd separately.

ilidemi (approved):

PR #4578, opened 2026-07-14
Renovate check and simulation CI workflow

pfcoperez:
Wrap scripts/renovate-simulation.sh syntax check and update simulator script in a GH workflow triggered by PRs changing Renovate configuration.

It:

- Runs  a syntax check
- Runs a simulation of the updates the proposed configuration would generate, posting the results as a PR comment and uploading them and full simulation log as workflow artifacts.

### DIFF
--- a/.github/workflows/check-renovate-config.yml
+++ b/.github/workflows/check-renovate-config.yml
@@ -111,6 +111,12 @@ jobs:
               '',
               `Output of \`scripts/renovate-simulation.sh\` for ${context.payload.pull_request.head.sha} ([workflow run](${runUrl})).`,
               '',
+              '#### Proposed updates',
+              '',
+              'Dependency updates in the resulting PRs:',
+              '',
+              proposalsTable,
+              '',
               '#### `final-updated-packages.txt`',
               '',
               'Packages with updates in the resulting PRs (considering disabled/enabled rules):',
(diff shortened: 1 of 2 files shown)


dtunikov (approved):

dtunikov:
lgtm

just one more general question as I'm trying to learn how things work a bit more
i checked a bit how exactly this whole renovate workflow is structured
seems like we don't run renovate ourselves and rely on: https://www.mend.io/renovate/
is this correct?
if so, why can't we run renovate using gh workflows?
it doesn't block this PR, I just want to know the reasoning behind those decisions.

pfcoperez:
> just one more general question as I'm trying to learn how things work a bit more i checked a bit how exactly this whole renovate workflow is structured seems like we don't run renovate ourselves and rely on: https://www.mend.io/renovate/ is this correct? if so, why can't we run renovate using gh workflows? it doesn't block this PR, I just want to know the reasoning behind those decisions.

Renovate is an external service that does basically four things (oversimplification here):

On periodic schedule (CRON mode):

- Compare the source (dependency,version) list  with the upstream package repository -> diff
- Run a set of rules on the diff -> Proposed updates.
- Create PRs based on proposed updates.

Reactively (depending on the infra):

- Manage its own PR upon changes in the repository: React to merged updates, changes in the pre-conditions, etc.

Being this an open source project, we benefit from Mend provided infrastructure to run this service. We can use https://github.com/renovatebot/github-action to replace Mend CRON infra and move into self-hosted but not for this type of verification.

I am sharing with you this infra internal docs.

PR #4547, opened 2026-07-07
Pin go-mysql digest as well

ilidemi:
go-mysql was pinned as we're using a pseudoversion from main and don't want to follow the rest of main. Turns out just version is not enough for renovate (#4546), adding digest as well

PR #4579, opened 2026-07-14
Remove checks on pinned versions

pfcoperez:
Our dependencies updates come:
- Mostly from Renovate through automatic updates.
- When we want to remove a pinned version, through a manual PR removing the pin (less frequently with two remaining).

We aim to not having pinned versions and if we do instruct Renovate to stop updating them.
We also have some precedents of pinned packages that are just declared in Renovate: https://github.com/PeerDB-io/peerdb/pull/4547

After this PR, will trust Renovate and go.mod as unique sources of truth for pins with tests as barriers.


.claude/skills/update-dependencies/SKILL.md
ilidemi, line 1: Can remove the entire skill I think, renovate has been doing a good job

ilidemi (approved):

.claude/skills/update-dependencies/SKILL.md
pfcoperez: I was very close to delete it completely, 100% on removing it.

pfcoperez: Done in 92b1400aa47d19c9de376b16a8025f9ac09374aa
:tada: Determinism!

PR #4582, opened 2026-07-14
Classify "conn closed" errors as notify-user connectivity

dtunikov:
`conn closed` errors (dropped source connection) are currently classified as `ERROR_RETRY_RECOVERABLE` → `notify_telemetry`, which pages on-call. These are recoverable connectivity issues that auto-resolve on retry, with a root cause outside our control (source DB restart / network blip), so they shouldn't page us.

Reclassify to `ErrorNotifyConnectivity` (`notify_user`): still surfaced to the user (with threshold gating), consistent with other connectivity errors (`net.OpError`, `pgconn.ConnectError`, DNS, TLS), and no longer pages on-call.

DBI-917


flow/alerting/classifier.go
```diff
@@ -353,7 +353,7 @@ func GetErrorClass(ctx context.Context, err error) (ErrorClass, ErrorInfo) {
 	// Reference:
 	// https://github.dev/jackc/pgx/blob/master/pgconn/pgconn.go#L733-L740
 	if strings.Contains(err.Error(), "conn closed") {
-		return ErrorRetryRecoverable, ErrorInfo{
+		return ErrorNotifyConnectivity, ErrorInfo{
```
jgao54, line 354: "conn closed" is a bit too generic that could lead to false positives, since we know it's coming from postgres, it might be worth wrapping it around a custom error class in pg connector code.

dtunikov: handle with errors.Is instead
this should work fine according to pgx source code

image

PR #4583, opened 2026-07-14
qrep: derive pull columns from source instead of catalog schema

dtunikov:
## Problem

QRep pulls with excluded columns or child-table range partitions fail with:

```
[qrep] failed to pull records: failed to load table schema: no rows in result set
```

The pull path ([qrep_source.go](flow/connectors/postgres/qrep_source.go)) built the `SELECT` column list by loading the table schema from `table_schema_mapping`, keyed by `(ParentMirrorName, DestinationTableIdentifier)`. When no row exists for that key, `QueryRow().Scan()` returns `pgx.ErrNoRows` ("no rows in result set") and the pull hard-fails.

This has been observed on a CDC mirror. The in-repo setup paths normally populate that row under a matching key, so the exact reason the row was absent for this mirror isn't pinned down (candidates: version skew / older mirror, or attaching to a pre-existing mirror). Regardless, the pull shouldn't depend on the catalog row being present when it can derive the columns directly from the source.

## Fix

Derive the column list **directly from the source** via `GetSelectedColumns` (which reads `pg_attribute`) — the same primitive `getTableSchemaForTable` already uses to build the identical list, so column selection and ordering are unchanged.

This removes the `table_schema_mapping` dependency from the pull path entirely. `catalogPool` is no longer needed by `corePullQRepRecords`.

### DIFF
--- a/flow/connectors/postgres/postgres_source.go
+++ b/flow/connectors/postgres/postgres_source.go
@@ -410,7 +410,7 @@ func (c *PostgresConnector) GetSelectedColumns(
 	}
 	excludedColumnsSQL := ""
 	if len(excludedColumns) > 0 {
-		excludedColumnsSQL = "AND a.attname NOT IN (" + strings.Join(quotedExcludedColumns, ",") + ")"
+		excludedColumnsSQL = " AND a.attname NOT IN (" + strings.Join(quotedExcludedColumns, ",") + ")"
 	}
 
 	getColumnsSQL := `
@@ -421,7 +421,7 @@ func (c *PostgresConnector) GetSelectedColumns(
 		WHERE n.nspname = $1
 		AND c.relname = $2
 		AND a.attnum > 0
-		AND NOT a.attisdropped ` + excludedColumnsSQL
+		AND NOT a.attisdropped` + excludedColumnsSQL + ` ORDER BY a.attnum`
 
 	rows, err := c.conn.Query(ctx, getColumnsSQL, sourceTable.Namespace, sourceTable.Table)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


flow/connectors/postgres/postgres_source.go
```diff
@@ -421,7 +421,7 @@ func (c *PostgresConnector) GetSelectedColumns(
 		WHERE n.nspname = $1
 		AND c.relname = $2
 		AND a.attnum > 0
-		AND NOT a.attisdropped ` + excludedColumnsSQL
+		AND NOT a.attisdropped` + excludedColumnsSQL + ` ORDER BY a.attnum`
```
dtunikov, line 424: optional, suggested by copilot to keep columns order consistent across calls

ilidemi (commented):
Would prefer to first understand what led to the error before making a change just in case. What if there's a consistency issue

dtunikov:
@ilidemi 
> Would prefer to first understand what led to the error before making a change just in case. What if there's a consistency issue.

I explained it in internal discussion, posting here as well:
this is what has happened most probably:

- Resync cancels the active snapshot/QRep children.
- DropFlowWorkflow [deletes](https://github.com/PeerDB-io/peerdb/blob/a1f6b8dd6d2dc591e26ec5ff2f6f211739a19a3d/flow/activities/flowable.go#L1976) all table_schema_mapping rows, including during resync.
- An already-running QRep pull gets far enough to perform this new schema [lookup](https://github.com/PeerDB-io/peerdb/blob/a1f6b8dd6d2dc591e26ec5ff2f6f211739a19a3d/flow/connectors/postgres/qrep_source.go#L365) before its cancellation is observed.
- It sees the deleted row and emits this exact error.

ilidemi (approved):
Ah right, I missed that thread. Conceptually having a single source of truth is nicer, and waiting on full Temporal cancellation before starting resync would be a cleaner fix (if possible), but this is also bringing us closer to not needing an overarching snapshot tx and still supporting schema changes so ok enough.

PR #4585, opened 2026-07-15
support filtering out deletion operation types for mongo

jgao54:
Allow filtering out operations by type.  This is already possible today with Postgres publications. Here we extend this feature to MongoDB by filtering out operations in the pipeline definition. This allows data to be filtered out on the server-side. 

Adding a UI component in Edit Mirror page that supports updating the setting for a mirror (vs. the `settings` tab, that would update the setting for *all* pipes). If both are set, pipe-level setting will take precedence. Here we introduce a source-type-specific allowlist so the settings override shows up for applicable pipes (in this case, only Mongo pipe will display this option). Note that the `CDCFlowConfigUpdate` field in `FlowStateChangeRequest` already supports updatedEnv today, so this PR just surfaces it to the UI. 

Screenshot 2026-07-14 at 22 08 04@2x

Test: added uinit/e2e tests; also verified locally e2e that pipeline filtering works as expected and can be modified dynamically on pause/resume. 

Fixes: DBI-922


flow/connectors/mongo/cdc.go
```diff
@@ -569,7 +595,27 @@ func (c *MongoConnector) FinishExport(any) error {
 	return nil
 }
 
-func (c *MongoConnector) SetupReplConn(context.Context, map[string]string) error {
+func (c *MongoConnector) SetupReplConn(ctx context.Context, env map[string]string) error {
+	// Unlike Postgres, MongoDB doesn't need a dedicated replication connection:
+	// change streams are cursors served by the connector's pooled client.
+	// Since SetupReplConn is called once per SyncFlow activity, resolving
+	// dynamic config here avoids per-batch catalog reads.
```
jgao54, line 603: Also evaluated populating this in the mongo connector itself similar to Postgres, but this is a more logical place for it despite the `SetupReplConn` naming is a bit misleading. Not all mongo connectors instantiated need to worry about operation type exclusion, in fact only cdc does. So populating `excludedOps` here feels like the more natural placement. Although I can be persuaded otherwise. 

I think we have a few other cdc-related settings that are being fetched every batch today (when we expect them to be fetched every SyncFlow and only update on pause/resume), so they behave more like `APPLY_MODE_IMMEDIATE` even though they are labeled `APPLY_MODE_AFTER_RESUME`, resulting in a lot of db calls that could be avoided. So one of the follow-up would be to identify these dynconf and move them here as well.

masterashu (approved):

dtunikov (approved):

flow/connectors/mongo/cdc.go
```diff
@@ -569,7 +595,27 @@ func (c *MongoConnector) FinishExport(any) error {
...
+	excludedOps, err := internal.PeerDBMongoDBExcludedOperationTypes(ctx, env)
+	if err != nil {
+		return fmt.Errorf("failed to get excluded operation types: %w", err)
+	}
+	validOps := make([]operationType, 0, len(excludedOps))
+	for _, op := range excludedOps {
+		if parsed, ok := parseOperationType(op); ok {
+			validOps = append(validOps, parsed)
+		} else {
+			c.logger.Warn("ignoring invalid operation type", slog.String("operationType", op))
```
pfcoperez: This loop is to validate the exclusion filter elements but the current message might mislead to think the operation itself is being ignored at CDC processing:

```suggestion
			c.logger.Warn("ignoring invalid operation type in exclusion list", slog.String("operationType", op))
```

ui/app/mirrors/[mirrorId]/edit/EnvOverrides.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { TextField } from '@/lib/TextField';
+import { useEffect, useMemo, useState } from 'react';
+import { fieldStyle } from '../styles/edit.styles';
+
+// Settings that are safe to override per mirror via CDCFlowConfigUpdate.UpdatedEnv,
+// gated on the mirror's source type. The override is merged into the mirror's Env,
+// which dynLookup checks before the service-level dynamic_settings catalog value.
+const pipeLevelSettings: { name: string; sourceTypes: DBType[] }[] = [
+  {
+    name: 'PEERDB_MONGODB_EXCLUDED_OPERATION_TYPES',
```
Comment (morsapaes) [inline ui/app/mirrors/[mirrorId]/edit/EnvOverrides.tsx]: Can we expose the setting with a more user-friendly name in the UI vs. the raw setting name? In a similar direction, "Setting Overrides" bleeds implementation details; if we expose it as a configurable setting, it should be implied it's a safe override. Maybe "Replication Settings" would be more appropriate and future-proof, with a helper text like "Changes apply to this mirror only and take precedence over global defaults"?

flow/connectors/mongo/cdc.go
```diff
@@ -569,7 +595,27 @@ func (c *MongoConnector) FinishExport(any) error {
...
+func (c *MongoConnector) SetupReplConn(ctx context.Context, env map[string]string) error {
+	// Unlike Postgres, MongoDB doesn't need a dedicated replication connection:
+	// change streams are cursors served by the connector's pooled client.
+	// Since SetupReplConn is called once per SyncFlow activity, resolving
+	// dynamic config here avoids per-batch catalog reads.
+	excludedOps, err := internal.PeerDBMongoDBExcludedOperationTypes(ctx, env)
+	if err != nil {
+		return fmt.Errorf("failed to get excluded operation types: %w", err)
+	}
+	validOps := make([]operationType, 0, len(excludedOps))
```
pfcoperez: nit Q: Is there a reason for not doing just `c.excludedOps := make([]operationType, 0, len(...`

flow/internal/dynamicconf.go
```diff
@@ -509,6 +509,15 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_AFTER_RESUME,
 		TargetForSetting: protos.DynconfTarget_ALL,
 	},
+	{
+		Name: "PEERDB_MONGODB_EXCLUDED_OPERATION_TYPES",
```
morsapaes, line 513: Since the set of options is bound and known, should this be a multi-select field instead to reduce the risk of typos? Not sure if we have building blocks for this in the UI components, so a comma-separated list is fine to start; just wanted to flag.

flow/connectors/mongo/cdc.go
```diff
@@ -528,6 +547,13 @@ func createPipeline(tableNameMapping map[string]model.NameAndExclude) (mongo.Pip
 		}}})
 	}
 
+	// filter out excluded operation types
+	if len(excludedOps) > 0 {
+		pipeline = append(pipeline, bson.D{{Key: "$match", Value: bson.D{
+			{Key: "operationType", Value: bson.D{{Key: "$nin", Value: excludedOps}}},
```
pfcoperez, line 554: If the user input has duplicates (e.g: `insert, insert, insert,insert`), will this filter just ignore them or can that lead to issues?

If not issues, which I imagine to be the case, can be performance affected in any significant way?

flow/connectors/mongo/cdc.go
pfcoperez: Depending on https://github.com/PeerDB-io/peerdb/pull/4585/changes#r3585820029, maybe we could change `validOps` to be a set to remove duplicates.

pfcoperez (approved):
LGTM, conditioned on the log message change and on the confirmation that dups are inconsequential.

flow/connectors/mongo/cdc.go
jgao54: no reason, will update :)

ui/app/mirrors/[mirrorId]/edit/EnvOverrides.tsx
Comment (jgao54) [reply ui/app/mirrors/[mirrorId]/edit/EnvOverrides.tsx]: ah yeah, that's a good idea for the pipe-level setting override.

flow/internal/dynamicconf.go
jgao54: For global settings tab, this isn't super trivial to change; but for the pipe-level configs we can build something more user-friendly and translate it to env var on our end

ui/app/mirrors/[mirrorId]/edit/EnvOverrides.tsx
Comment (jgao54) [reply ui/app/mirrors/[mirrorId]/edit/EnvOverrides.tsx]: one downside is that this will make it less obvious it's referring to the same global setting here, which we already expose as raw envs in the Settings page, and it will also apply to other global envs that we want to enable on the pipe-level in the future. 

Screenshot 2026-07-15 at 10 24 09@2x

flow/connectors/mongo/cdc.go
jgao54: good idea to dedupe

ui/app/mirrors/[mirrorId]/edit/EnvOverrides.tsx
Comment (ilidemi) [reply ui/app/mirrors/[mirrorId]/edit/EnvOverrides.tsx]: We already have a system for overriding settings on the mirror level, some of our OSS customers are using that for the multi-tenancy setup. The UI is even clunkier (text box accepting raw JSON) but there is a precedent and if we're introducing support in editing, those probably should be unified (at some UX level).

Image

https://github.com/PeerDB-io/peerdb/blob/58babbd99f52a93342ffb9571e5b27fc100dcaf4/ui/app/mirrors/create/helpers/cdc.ts#L227-L238

jgao54:
Removed the UI component for now, will address as a follow-up. In the mean time the env can be updated via flow-api:
```
// pause the pipe
curl -X POST localhost:8113/v1/mirrors/state_change \
  -H "Content-Type: application/json" \
  -d '{"flow_job_name": "mongo_filter_test", "requested_flow_state": "STATUS_PAUSED"}'

// edit env + resume the pipe
curl -X POST localhost:8113/v1/mirrors/state_change \
  -H "Content-Type: application/json" \
  -d '{
    "flow_job_name": "mongo_filter_test",
    "requested_flow_state": "STATUS_RUNNING",
    "flow_config_update": {
      "cdc_flow_config_update": {
        "updated_env": {"PEERDB_MONGODB_EXCLUDED_OPERATION_TYPES": "delete"}
      }
    }
  }'

```

PR #4590, opened 2026-07-16
Include rust builder in renovate, fix script on mac

ilidemi:
* Rust builder image has a custom naming scheme Renovate doesn't recognize. Not updating the image blocks rust upgrades (#4572)
* sed on mac doesn't support inline replacement so use a temp file for portability
* gitignore one more script output

### DIFF
--- a/scripts/renovate-simulation.sh
+++ b/scripts/renovate-simulation.sh
@@ -20,12 +20,12 @@ fi
 npx --yes --package renovate -- renovate-config-validator renovate.json5 || exit 1
 
 ## (2) Temporarily change the renovate.json5 to point to the local extension source
-## Use perl for portable in-place editing (GNU `sed -i` and BSD/macOS `sed -i` are incompatible).
-perl -i -pe 's/local>PeerDB-io/github>PeerDB-io/' renovate.json5
+## Edit via a temp file for portability (GNU `sed -i` and BSD/macOS `sed -i` are incompatible).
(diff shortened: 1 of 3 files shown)


pfcoperez (approved):

scripts/renovate-simulation.sh
```diff
@@ -20,11 +20,12 @@ fi
...
 
 ## (2) Temporarily change the renovate.json5 to point to the local extension source
-sed -i 's/local>PeerDB-io/github>PeerDB-io/' renovate.json5
+## Edit via a temp file for portability (GNU `sed -i` and BSD/macOS `sed -i` are incompatible).
+sed 's/local>PeerDB-io/github>PeerDB-io/' renovate.json5 > renovate.json5.tmp && mv renovate.json5.tmp renovate.json5
 
 function cleanup {
   echo "Restoring renovate.json5 to point to the GitHub extension source"
-  sed -i 's/github>PeerDB-io/local>PeerDB-io/' renovate.json5
+  sed 's/github>PeerDB-io/local>PeerDB-io/' renovate.json5 > renovate.json5.tmp && mv renovate.json5.tmp renovate.json5
```
pfcoperez: According to this: https://stackoverflow.com/a/7573438

It can be simplified. I don't have a Mac to try it out though.

pfcoperez:
```suggestion
  sed -i.tmp 's/github>PeerDB-io/local>PeerDB-io/' renovate.json5
```

ilidemi: Yeah this is shorter. Temp file still needs to be removed though with `.tmp`

PR #4593, opened 2026-07-16
fix: respect clickhouse max_query_size

dtunikov:


flow/connectors/clickhouse/object_sync.go
```diff
@@ -35,6 +36,13 @@ func (b *objectBatch) urls() []string {
 
 const defaultBatchFlushInterval = 5 * time.Second
 
+// maxBatchURLBytes bounds the cumulative length of the URLs packed into a single
+// batch. The URLs are inlined into the INSERT query via brace expansion
+// (url('{url1,url2,...}', ...)), so a batch with too many URLs overflows
+// ClickHouse's max_query_size (256 KiB by default). This budget stays well under
+// that limit, leaving headroom for the headers, column list and query prefix.
+const maxBatchURLBytes int64 = 128 * 1024
```
dtunikov: hm, I think we can query it from ch settings and don't rely on default value here
will adopt it a bit later

flow/connectors/clickhouse/object_sync.go
```diff
@@ -35,6 +37,66 @@ func (b *objectBatch) urls() []string {
...
+//   - table_function_remote_max_addresses: the brace list is expanded like
+//     remote() shard addresses and rejected past this count.
+type urlBatchLimits struct {
+	maxURLBytes int64
+	maxURLCount int
+}
+
+// fetchURLBatchLimits reads the effective max_query_size and
+// table_function_remote_max_addresses from ClickHouse directly.
+func (c *ClickHouseConnector) fetchURLBatchLimits(ctx context.Context) urlBatchLimits {
```
ilidemi: Add an integration test for this one? Possible to override settings within a connection and not affect other tests

flow/connectors/clickhouse/object_sync.go
```diff
@@ -35,6 +37,66 @@ func (b *objectBatch) urls() []string {
...
+	maxQuerySize := defaultMaxQuerySize
+	remoteMaxAddresses := defaultRemoteMaxAddresses
+
+	rows, err := c.query(ctx,
+		"SELECT name, value FROM system.settings WHERE name IN ('max_query_size', 'table_function_remote_max_addresses')")
+	if err != nil {
+		c.logger.Warn("failed to read ClickHouse batch-limit settings, using defaults", slog.Any("error", err))
```
ilidemi: Would prefer to stay deterministic in face of network hiccups. I guess if the defaults are not good enough, the worst that's going to happen is either failing and successfully retrying, or too small batching, but the error we'd receive would be the same as today and we'd get paged. Let's just return the error

ilidemi (approved):

PR #4597, opened 2026-07-19
fix(clickhouse): preserve legacy soft-delete column casing

### LINKED ISSUE #4596: [Bug]: Normalization Error: failed to normalize records: error while inserting into target clickhouse table <tablle>: code: 16, message: No such column _PEERDB_IS_DELETED in table <table>
### Description

I upgraded from `v0.27.1` to `v0.37.1` and all my postgres -> clickhouse mirrors broke with the following error:

```
Normalization Error: failed to normalize records: error while inserting into target clickhouse table simulation_jobs: code: 16, message: No such column _PEERDB_IS_DELETED in table tdb.simulation_jobs (b6ee3e4b-4fd2-449f-9172-82a7fcfb818c)
```

In my original mirror I never changed the (old) default columns so my destination tables use `_peerdb_is_deleted` and `_peerdb_synced_at` (lowercase).

**UPDATE:** I was able to fix it by creating a new mirror as follows. My original query below had a typo that caused it to fail:
```sql
CREATE MIRROR <mirror_name>
FROM <postgres peer>
TO <clickhouse peer>
WITH TABLE MAPPING
(
  -- mappings
)
WITH (
  sync_interval = 60,
  do_initial_copy = false,
  publication_name = 'peerdb_publication',
  snapshot_max_parallel_workers = 2,
  soft_delete = true,
  soft_delete_col_name = '_peerdb_is_deleted',
  synced_at_col_name = '_peerdb_synced_at '
);
```

I suspect this was caused by #4365 which came out in release `v0.36.25` because rolling back to `v0.36.24` fixed the issue.

### Fix
1. Pause the mirror
2. Follow the steps to [upgrade](https://docs.peerdb.io/mirror/upgrade).
3. Recreate the mirror using the query I gave before using the same mirror name.

### Steps to reproduce

1. Use v0.27.1 and create a mirror from postgres to clickhouse
2. Update to v0.37.1
3. Check logs

### PeerDB version

ghcr.io/peerdb-io/flow-api:stable-v0.37.1

### Source connector

Postgres

### Destination connector

ClickHouse

floze-the-genius:
### Summary

- resolve the configured soft-delete column against the physical ClickHouse destination schema before normalization
- prefer exact matches, then case-insensitive matches, so both legacy lowercase tables and newer uppercase/custom tables keep working
- apply the same resolution during destination validation

### Root cause

Older PeerDB releases stored the UI default `_PEERDB_IS_DELETED` in mirror config, but ClickHouse table creation and normalization hardcoded the physical column as `_peerdb_is_deleted`. After #4365 began honoring the stored config, upgraded legacy mirrors generated inserts for the uppercase identifier even though their existing tables only had the lowercase column.

### Tests

- `go test ./connectors/clickhouse -skip '^TestCreateRawTableHasTTL$' -count=1`
- `go vet ./connectors/clickhouse`
- `go test -race ./connectors/clickhouse -run 'Test(ResolveSoftDeleteColumnName|BuildQuery_UsesDestinationSoftDeleteColumnCase|ProcessTableComparison_WithResolvedSoftDeleteColumnCase)$' -count=1`

Fixes #4596

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -28,6 +28,7 @@ import (
 
 const (
 	defaultIsDeletedColName = "_peerdb_is_deleted"
+	legacyIsDeletedColName  = "_PEERDB_IS_DELETED"
 	isDeletedColType        = "UInt8"
 	versionColName          = "_peerdb_version"
 	versionColType          = "UInt64"
@@ -464,6 +465,13 @@ func (c *ClickHouseConnector) NormalizeRecords(
 		return model.NormalizeResponse{}, err
 	}
 
+	destinationColumns, err := peerdb_clickhouse.GetTableColumnsMapping(
+		ctx, c.logger, c.database, destinationTableNames,
+	)
+	if err != nil {
+		return model.NormalizeResponse{}, fmt.Errorf("failed to get destination table columns: %w", err)
+	}
+
 	enablePrimaryUpdate, err := internal.PeerDBEnableClickHousePrimaryUpdate(ctx, req.Env)
 	if err != nil {
 		return model.NormalizeResponse{}, err
@@ -577,6 +585,10 @@ func (c *ClickHouseConnector) NormalizeRecords(
 				continue
 			}
 
+			softDeleteColName := resolveSoftDeleteColumnName(
+				req.SoftDeleteColName,
+				destinationColumns[tbl],
+			)
 			queryGenerator := NewNormalizeQueryGenerator(
 				tbl,
 				req.TableNameSchemaMapping,
@@ -589,7 +601,7 @@ func (c *ClickHouseConnector) NormalizeRecords(
 				rawTbl,
 				c.chVersion,
 				c.Config.Cluster != "",
-				req.SoftDeleteColName,
+				softDeleteColName,
 				req.Version,
 				req.Flags,
 			)
@@ -633,6 +645,35 @@ func (c *ClickHouseConnector) NormalizeRecords(
 		StartBatchID: lastNormBatchID + 1,
 		EndBatchID:   endBatchID,
 	}, nil
+}
+
+func resolveSoftDeleteColumnName(
+	configuredName string,
+	destinationColumns []peerdb_clickhouse.ClickHouseColumn,
+) string {
+	if configuredName == "" {
+		configuredName = defaultIsDeletedColName
+	}
+
+	for _, column := range destinationColumns {
+		if column.Name == configuredName {
+			return column.Name
+		}
+	}
+	if configuredName != defaultIsDeletedColName && configuredName != legacyIsDeletedColName {
+		return configuredName
+	}
+
+	compatibleName := defaultIsDeletedColName
+	if configuredName == defaultIsDeletedColName {
+		compatibleName = legacyIsDeletedColName
+	}
+	for _, column := range destinationColumns {
+		if column.Name == compatibleName {
+			return column.Name
+		}
+	}
+	return configuredName
 }
 
 func (c *ClickHouseConnector) getDistinctTableNamesInBatch(
--- a/flow/connectors/clickhouse/normalize_test.go
+++ b/flow/connectors/clickhouse/normalize_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	peerdb_clickhouse "github.com/PeerDB-io/peerdb/flow/pkg/clickhouse"
 	"github.com/PeerDB-io/peerdb/flow/shared"
 	"github.com/PeerDB-io/peerdb/flow/shared/types"
 )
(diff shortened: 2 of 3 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -631,6 +642,27 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
+		configuredName = defaultIsDeletedColName
+	}
+
+	for _, column := range destinationColumns {
+		if column.Name == configuredName {
+			return column.Name
+		}
+	}
+	for _, column := range destinationColumns {
+		if strings.EqualFold(column.Name, configuredName) {
```
pfcoperez: This successfully bridges the gap between the old UI default `_PEERDB_IS_DELETED` and default normalization value `_peerdb_is_deleted`  but it opens the space of matched columns to any case variant too. You reflect one example in the tests below:  `_PeErDb_Is_DeLeTeD` 

This space is also expanded now to custom columns names that have nothing to do with the defaults.

I am not sure we want these side effects. Depending on the use case, case might be relevant, specially in custom soft delete columns.

For example, if configuredName is `my_deleted`  but the table has `MY_deleted" column we should rather fail validation.

I'd limit resolution flexibility to the known two assumed to be equivalent values: `_PEERDB_IS_DELETED` and `_peerdb_is_deleted`.

pfcoperez (commented):
Thanks for your contribution! I've dropped a comment about the implementation detail.

floze-the-genius:
Narrowed the compatibility behavior in `ac5fe774` as requested:

- exact destination-column matches still win;
- fallback resolution is now limited to the two known equivalent defaults, `_PEERDB_IS_DELETED` and `_peerdb_is_deleted`;
- mixed-case variants and custom column names remain strict, so `my_deleted` will not resolve to `MY_deleted`.

Added positive tests for both legacy directions and negative tests for mixed/default and custom-case variants. Targeted ClickHouse tests and `go vet ./connectors/clickhouse` pass.

pfcoperez:
@fable review this PR

flow/connectors/clickhouse/normalize_test.go
```diff
@@ -325,6 +326,85 @@ func TestBuildQuery_WithPrimaryUpdate(t *testing.T) {
...
+	columns := []peerdb_clickhouse.ClickHouseColumn{
+		{Name: "_peerdb_is_deleted"},
+		{Name: "_PEERDB_IS_DELETED"},
+	}
+
+	require.Equal(t, "_PEERDB_IS_DELETED", resolveSoftDeleteColumnName("_PEERDB_IS_DELETED", columns))
+	require.Equal(t, "_peerdb_is_deleted", resolveSoftDeleteColumnName("_PEERDB_IS_DELETED", columns[:1]))
+	require.Equal(t, "_PEERDB_IS_DELETED", resolveSoftDeleteColumnName("_peerdb_is_deleted", columns[1:]))
+	require.Equal(t, "_peerdb_is_deleted", resolveSoftDeleteColumnName("", columns[:1]))
+	require.Equal(t, "_PeErDb_Is_DeLeTeD", resolveSoftDeleteColumnName("_PeErDb_Is_DeLeTeD", columns))
```
pfcoperez, line 339: 👍

pfcoperez (commented):
> 1. For a custom soft-delete column name (anything other than `""`/`_peerdb_is_deleted`/`_PEERDB_IS_DELETED`), `resolveSoftDeleteColumnName` is always the identity, yet every `NormalizeRecords` cycle now pays an extra `system.columns` round-trip and gains a new hard-failure point. Gating the fetch on the configured name being one of the two defaults (or empty) would keep the fix free for custom-named mirrors.

This is a good point @floze-the-genius  and it seems a simple enough change.

Besides that, the PR looks good to me, maybe @ilidemi or @jgao54 want to take a look too.

flow/connectors/clickhouse/normalize.go
```diff
@@ -464,6 +465,13 @@ func (c *ClickHouseConnector) NormalizeRecords(
 		return model.NormalizeResponse{}, err
 	}
 
+	destinationColumns, err := peerdb_clickhouse.GetTableColumnsMapping(
```
jgao54, line 468: this introduces a clickhouse query per table table on every normalize iteration, and the destination column name does not change from iteration to iteration. so this would add latency to all pipes. 

the history around upper case _peerdb_is_deleted is a bit unfortunate. initially, the `SoftDeleteColName`  field was originally never used by clickhouse, it was only for snowflake/bigquery/postgres destinations.  then https://github.com/PeerDB-io/peerdb/pull/2008 started to use `SoftDeleteColName` for ClickHouse to support the delete-on-merge feature, which accidentally made it possible for CH to also have upper case _peerdb_is_deleted. while this bug was fixed immediately in https://github.com/PeerDB-io/peerdb/pull/3574, it was only fixed in the UI path so it's still technically possible in the API path (and still possible for legacy existing pipes that have the config set to upper case)

my preference is actually to revert change introduced in https://github.com/PeerDB-io/peerdb/pull/4365 and also tighten the validation on the server-side to prevent upper case from being introduced. 

@ilidemi  wdyt?

ilidemi: Agreed. I don't have as much context about this feature, so if you think tightening validation is going to remove degrees of freedom in a good way, that sounds easier.

Does this mean that the existing setups would be recommended to rename the CH column, or would we keep some kind of handling to accommodate for both?

PR #4600, opened 2026-07-20
bug: CDC Date clamping consistent with initial load limits

pfcoperez:
In https://github.com/PeerDB-io/peerdb/pull/4534/ we added logic to proactively clamp dates to CH Date32 limits at the time.

However, the scope of that change didn't include CDC. CDC normalization was therefore constrained by ClickHouse's `parseDateTime64BestEffortOrNull` limits.

These limits changed with with https://github.com/clickHouse/clickhouse/pull/107907 leaving initial loads and cdc supported date ranges inconsistent with each other.

This PR aligns CDC to use the same limits as initial load.

From now this is a constant to `[1900-01-01,  2299-12-31]` but  might want to follow up with a piece of functionality probing destination CH valid ranges to make the most of the target CH version capabilities.

### DIFF
--- a/flow/connectors/clickhouse/normalize_query.go
+++ b/flow/connectors/clickhouse/normalize_query.go
@@ -86,6 +86,26 @@ func clampDates(dateExpr string) string {
 	return fmt.Sprintf("if(isNull(%s), NULL, %s)", dateExpr, upperAndLowerBounded)
 }
 
+// clampTimestamps bounds a DateTime64(6) SQL expression to PeerDB's supported
+// range, matching the Go-side clamp applied during initial load
+// (processGeneralTime): out-of-range values move to the boundary date with
+// time-of-day preserved. On ClickHouse < 26.7 parseDateTime64BestEffort*
+// already clamps, making this a no-op; on >= 26.7 the parse passes
+// out-of-range values through.
+func clampTimestamps(timestampExpr string) string {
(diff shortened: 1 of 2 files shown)


flow/connectors/clickhouse/normalize_query.go
```diff
@@ -74,6 +74,38 @@ func NewNormalizeQueryGenerator(
...
+// clampTimestamps bounds a DateTime64(6) SQL expression to PeerDB's supported
+// range, matching the Go-side clamp applied during initial load
+// (processGeneralTime): out-of-range values move to the boundary date with
+// time-of-day preserved. On ClickHouse < 26.7 parseDateTime64BestEffort*
+// already clamps, making this a no-op; on >= 26.7 the parse passes
+// out-of-range values through.
+func clampTimestamps(timestampExpr string) string {
+	minBound := fmt.Sprintf("toDateTime64('%d-01-01 00:00:00',6,'UTC')", qvalue.ClickHouseMinYear)
+	maxBound := fmt.Sprintf("toDateTime64('%d-12-31 23:59:59.999999',6,'UTC')", qvalue.ClickHouseMaxYear)
+	maxDayStart := fmt.Sprintf("toDateTime64('%d-12-31 00:00:00',6,'UTC')", qvalue.ClickHouseMaxYear)
```
serprex: these can be globals, no need to Sprintf every call

pfcoperez: Great performance point!  Addressed.

ilidemi (approved):

PR #4605, opened 2026-07-21
CI: Replace flow service orchestration with Tilt

pfcoperez:
Unifies local dev env with CI tests env by leveraging the existing CI on-demand Tilt execution infrastructure:

- https://github.com/PeerDB-io/peerdb/pull/4474
- https://github.com/PeerDB-io/peerdb/pull/4477

To orchestrate the test services required for e2e testing. By doing so, we use the same environment embryogenesis at CI and local dev envs. This come with several advantages, including:

- Removes the need of maintaining two different development environments.
- Opens the way to on demand resource allocation in CI runs.

:memo:  Commits  such as d4c2616d31e4fc327d80b3ffbae8f5e43529180b and ad2db0c7633d9c86bb0bd00b66e6d8debe9c151b migrate Kafka and ES connector tests to Tilt CI. We might aswell just remove these tests and services given the deprecation notice -> This is exercised in commit 41e87c8700c74222e8581a06a0106e9ae4e95360 
:warning: So this PR removes CI Kafka/Elasticsearch infrastructure  and its associated tests

:memo: Commits can be reviewed independently.


pfcoperez:
## QA Validation

### Run with ES and kafka infra (feature parity with dedicated CI env)

https://github.com/PeerDB-io/peerdb/actions/runs/29845595421/job/88685294892?pr=4605
Duration: 23'

### Removing ES/Kafka tests and Infra

https://github.com/PeerDB-io/peerdb/actions/runs/29848110355?pr=4605
Duration: 22'

.github/workflows/tilt-flow.yml
```diff
@@ -280,30 +391,68 @@ jobs:
...
           echo "All requested ancillary services are up and running."
 
       - name: Install gotestsum
         run: go install gotest.tools/gotestsum@latest
 
+      # ----------------------------------------------------------------------
+      # Credentials for the cloud-destination suites (BigQuery/PubSub via the
+      # GCP service account, Snowflake, GCS, EventHubs).
+      # ----------------------------------------------------------------------
+      - name: setup gcp service account
```
pfcoperez, line 445: Given that we have decided to remove Kafka and ES test infra, I think we can remove these too

pfcoperez: This will be a follow-up, I'd like to be able to revert that removal partially in case of  need.

.github/workflows/flow.yml
```diff
@@ -18,757 +18,68 @@ on:
...
-      - name: Ingest tests results for analysis
-        if: success() || failure()
-        uses: ./.github/actions/ingest-test-results
-        with:
-          combination-id: "pg${{ matrix.db-version.pg }}-my${{ matrix.db-version.mysql }}-ma${{ matrix.db-version.mariadb }}-mo${{ matrix.db-version.mongo }}-ch${{ matrix.db-version.ch }}"
-          o11y-api-key-id: ${{ secrets.CI_O11Y_TARGET_API_KEY_ID }}
-          o11y-api-key-secret: ${{ secrets.CI_O11Y_TARGET_API_KEY_SECRET }}
-          o11y-query-endpoint: ${{ secrets.CI_O11Y_TARGET_QUERY_ENDPOINT }}
-
-      - name: Check all present test functions in the source branch were exercised in CI
```
pfcoperez, line 728: Notice how this didn't get migrated. That's because I wanted to keep the inputs surface on flow-tilt clean. There will be a follow-up adding a version of this step that reads from global execution list.

ilidemi (approved):
Great work and ty for the commits

ilidemi:
Asked friendo Fable to take a look:
* BQ tests run thrice because of the old PG-hooked check not working anymore
* CH TLS and IAM tests are silently skipped
* OTel tests aren't executed anymore (previously were on one configuration)
* Keeps complaining about ancillary env not autorefreshing - maybe we can add a version field to it, autoregen when it's bumped and have a ci job nagging to bump it?
* The workflow-level env block exports POSTGRES_IMAGE/MYSQL_GTID_IMAGE/etc. as set-but-empty strings when the corresponding inputs are omitted, and in docker compose interpolation the (empty) process environment takes precedence over the ancillary.env env-file that generate-test-environment.sh fills with the intended defaults, so the documented 'Empty values fall back to the script's defaults' path never works.
* Concurrency group cancels queued main runs. GitHub cancels the older pending run when a new one queues (cancel-in-progress: false only protects running runs), so of 3 quick merges to main, the middle commit gets no CI results. The standard fix is to make the group key unique for pushes, e.g. include github.event_name == 'push' && github.sha || '' in the group expression — pushes then never share a group, while PR runs still coalesce.
* Every test_env KEY=VALUE override is echoed verbatim to the public Actions log (`echo "test env override: $override"`); test_env is a generic pass-through input, so any credential a caller routes through it is printed in plaintext
* Replacing the digest-pinned 'mariadb:lts-ubi9@sha256:55a81b2d...' with the interpolated ${MARIADB_IMAGE} (resolved to tag-only 'mariadb:lts-ubi9' from the flow.yml matrix) drops the digest pin and, being a variable now, removes the image from Renovate's digest-pinning management entirely.
* Noticed outside of this PR: add `|| true` to ingest-test-results upload log processing, as no grep match will lead to step silently not continuing: `WRITTEN_ROWS=$(grep "$SUMMARY_HEADER" upload.log | sed "s/$SUMMARY_HEADER: //" | jq -r '.written_rows' || true)`

Fable clearly is better CI reviewer than I am

pfcoperez:
> Fable clearly is better CI reviewer than I am

Maybe it just haves a lower activation threshold and the exhaustiveness only machines can offer :stuck_out_tongue_winking_eye: 

Funny thing is that I used Opus to self review and it didn't catch this stuff either legit or not.

Some of the comments make more sense than others. Commenting and addressing below.

pfcoperez:
> * BQ tests run thrice because of the old PG-hooked check not working anymore

Prior to these changes, `CI_PG_VERSION`  was being set and we have:

https://github.com/PeerDB-io/peerdb/blob/3a343b8ebcd09431980a4fc014f29edf4d9d6742/flow/e2e/bigquery_test.go#L21-L23

At  BigQuery tests. This has the code smell resulting from:

1. Having different CI environment than local.
2. Pigging on version combinations to  test components outside the combinations.

`Only running in PG17 to reduce flakiness from high concurrency!` So the day that version goes away from the matrix so silently go the tests! :man_facepalming: 

I feel tempted to "fix" this by just matching on version 17 withing `POSTGRES_IMAGE` but TBH, I'd leave it "running thrice". If we really observed the so called flakiness, we can do this but I don't really see the point because it aimplies that many runs of the same test might also lead to flakiness.

Left filter as it is and commented about this situation at 4f4d8a65fa6ed8aab1563bab0828f0a0559b3a61

pfcoperez:
> * Keeps complaining about ancillary env not autorefreshing - maybe we can add a version field to it, autoregen when it's bumped and have a ci job nagging to bump it?

Opus complained about this but this is completely irrelevant as environments are ephemeral. `ancillary.env` files are generated on each run.

This is only relevant for local development Tilt environments and the fact is not autorefreshed is intentional as this file might be edited by developers. Tilt generates it when it's missing. If we want to be resilient to re-using workers in the future we just need to add a `rm ancillary.env` to the flow-tilt.yml file.

pfcoperez:
> * Noticed outside of this PR: add `|| true` to ingest-test-results upload log processing, as no grep match will lead to step silently not continuing: `WRITTEN_ROWS=$(grep "$SUMMARY_HEADER" upload.log | sed "s/$SUMMARY_HEADER: //" | jq -r '.written_rows' || true)`

This means that it will not show the error right below:

`echo "No results uploaded. Query client logs:"`

And it is a good catch, this is something limited to a very specific action, so I am fixing this in a separate PR for history hygiene: https://github.com/PeerDB-io/peerdb/pull/4622

pfcoperez:
> * The workflow-level env block exports POSTGRES_IMAGE/MYSQL_GTID_IMAGE/etc. as set-but-empty strings when the corresponding inputs are omitted, and in docker compose interpolation the (empty) process environment takes precedence over the ancillary.env env-file that generate-test-environment.sh fills with the intended defaults, so the documented 'Empty values fall back to the script's defaults' path never works.

This only applies to CI. It is not broken because when these inputs are omitted the associated services are also not listed to be started and therefore don't start. In ant case, protected against it in: cbd87a4bea4ec9a142f5193f248f474e4a724643

This is an interesting case because generate-test-environment.sh takes into consideration empty strings and its output is fed to docker-compose as --enf-file through Tilt. It is Docker compose where the conflict happens as it gives precedence to the empty environment variables over the values it obtains from the file.

pfcoperez:
> * OTel tests aren't executed anymore (previously were on one configuration)

There are no OTEL tests. Not having `ENABLE_OTEL_METRICS` means that flow-worker is not ingesting OTEL metrics.
In the previous flow `PEERDB_OTEL_METRICS_PANIC_ON_EXPORT_FAILURE` was set to `true` leads to using an exporter implementation that panics upon failure:

```go
func (p *panicOnFailureExporter) Export(ctx context.Context, metrics *metricdata.ResourceMetrics) error {
	if err := p.Exporter.Export(ctx, metrics); err != nil {
		panic(fmt.Sprintf("[panicOnFailureExporter] failed to export metrics: %v", err))
	}
	return nil
}
```

So this basically boils down to running the e2e tests in an environment were OTel is present and the services are sending metrics to it, panicking upon failure.

I've added OTEL support for CI in 28fe5733561af6b21755965db57679850eda94bd. Note that, by default, the local env does not start this service to avoid clashes with environments integrating with ClickPipes envs.

pfcoperez:
> * Every test_env KEY=VALUE override is echoed verbatim to the public Actions log (`echo "test env override: $override"`); test_env is a generic pass-through input, so any credential a caller routes through it is printed in plaintext

This is a bit of sloppy AI, the overrides are declared publicly in flow.yml : 

```yaml
      test_env: |
        CI_MYSQL_VERSION=${{ matrix.db-version.mysql }}
        CI_MYSQL_PORT=${{ matrix.db-version.mysql == 'mysql-pos' && '3307' || '3306' }}
        CI_MARIADB_VERSION=${{ matrix.db-version.mariadb }}
```

pfcoperez:
> * Replacing the digest-pinned 'mariadb:lts-ubi9@sha256:55a81b2d...' with the interpolated ${MARIADB_IMAGE} (resolved to tag-only 'mariadb:lts-ubi9' from the flow.yml matrix) drops the digest pin and, being a variable now, removes the image from Renovate's digest-pinning management entirely.

This is AI slop because:

A) That digest was only used in local env defaults and our intention is to align local and CI versions. This is actually correcting a bug.
B) The claim that Renovated updated this digest is completely hallucinated.
C) If we want to pass digests we should just add them to the test matrix configuration.

pfcoperez:
> CH TLS and IAM tests are silently skipped

Restored in in:

- CH TLS support: 91c1b9a0ec2fd9ebeb1fa68e7952377e333e2d82 making TLS tests execution unconditional in 034360b9ee26375081e015ea0dfa6d7abb9f0471
- RDS IAM support: 8e3efa4abf9767de788e7930eaf411ada0d6d21d f585523a43b22fc73081f6c9acdad8339415ff91

pfcoperez:
@fable review this PR considering the comment history

serprex (approved):

pfcoperez:
Branch protection rules updated:

image

pfcoperez:
Must be followed-up by: https://linear.app/clickhouse/issue/DBI-967

PR #4606, opened 2026-07-21
CI: Monitor CI stages times with actions-timeline action

pfcoperez:
Introduces https://github.com/marketplace/actions/actions-timeline action to our CI build thus allowing for analysis of CI stages performance.

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -142,6 +142,9 @@ jobs:
           DOCKER_MODS: linuxserver/mods:openssh-server-ssh-tunnel
 
     steps:
+      - uses: Kesin11/actions-timeline@v3.1.1
+        with:
+          expand-composite-actions: true
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
         with:
           ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.ref }}


pfcoperez:
## Claude Fable driven secret leak analysis

```
Verdict: no telemetry, no secret-leak paths — the action is clean, and it's a good fit for the CI performance comparison you're running. Analysis of Kesin11/actions-timeline at HEAD (e2445da, tags up to v3.1.1):

What it actually does

The main phase is an explicit no-op (dist/main.js — 460 bytes, dummy). Everything happens in the post phase, at job end: it fetches the current run's jobs/steps timings from the GitHub Actions API, renders a Mermaid Gantt chart, and writes it to $GITHUB_STEP_SUMMARY. With expand-composite-actions: true it additionally fetches workflow/action YAML files via repos.getContent — still the same API.

Telemetry: none

- The source (~1.4k lines of TypeScript) references exactly one service: the GitHub API. The base URL honors GITHUB_API_URL (and a host option → /api/v3 for GHES), so traffic stays inside your GitHub domain, including on GHES.
- I swept the full 1.4 MB shipped bundle (dist/post.js) for every host string: only GitHub endpoints (api.github.com, uploads.github.com — octokit constants) plus inert spec/doc URLs inside bundled polyfill comments (fetch.spec.whatwg.org etc.). Zero hits for telemetry keywords (sentry, segment, posthog, analytics, beacon, track-event, …).

Secrets: no leak path

- The github-token input (default github.token) is passed solely to Octokit auth against that base URL — never logged, never sent anywhere else. Documented minimum permission is actions: read (plus contents: read for the composite-expansion option).
- What it writes out is job/step names and timestamps into the run summary — exactly the data anyone with run-read access already sees on the run page. No new exposure surface.

Bundle scrutiny (the part that actually runs)

dist/ is a committed Deno→Node (dnt) build. The child_process/net/tls/dns references inside it are generic dnt Deno shims and octokit/undici transport polyfills — the action's own source never spawns a process (verified by grep). No eval/obfuscation patterns. The build is reproducible (deno task bundle → commit), so the dist-matches-source property is verifiable.
```

pfcoperez:
## QA Validation

Check charts at _Summary_ page: https://github.com/PeerDB-io/peerdb/actions/runs/29851047636?pr=4606

jgao54 (approved):

PR #4607, opened 2026-07-21
feat(bigquery): support GCP workload identity

jkaflik:
## Problem

BigQuery peers currently require long-lived service-account keys. ClickPipes needs a keyless mode where a single-tenant PeerDB deployment uses its projected Kubernetes token to impersonate its deployment-scoped tenant GSA.

## Solution

- recognize `service_account_workload_identity` through the existing BigQuery `auth_type` field
- build external-account credentials from trusted deployment configuration and fixed Google STS/IAM endpoints
- preserve ambient ADC for internal platform access and preserve the existing service-account-key path
- refresh projected tokens and impersonated credentials without restarting workers
- reject missing or unknown authentication modes with actionable errors

Target service account and cluster context remain deployment configuration; peers only select the authentication mode.


.env.example
```diff
@@ -55,6 +55,15 @@ PEERDB_CATALOG_DATABASE=postgres
 
 PEERDB_CLICKHOUSE_AWS_S3_BUCKET_NAME=peerdb
 
+# BigQuery workload identity deployment settings. The projected token must use
```
jkaflik:
```suggestion
# GCP workload identity deployment settings. The projected token must use
```

flow/connectors/bigquery/auth.go
```diff
@@ -0,0 +1,269 @@
...
+) (*bigQueryCredentialConfig, error) {
+	authType := config.GetAuthType()
+	if authType != BigQueryAuthTypeServiceAccount && authType != BigQueryAuthTypeServiceAccountWorkloadIdentity {
+		return nil, fmt.Errorf(
+			"unsupported BigQuery auth_type %q: expected %q or %q",
+			authType,
+			BigQueryAuthTypeServiceAccount,
+			BigQueryAuthTypeServiceAccountWorkloadIdentity,
+		)
+	}
```
jkaflik, line 101: BigQuery auth type is a string proto type.
It does not default into `BigQueryAuthTypeServiceAccount` value, meaning this check might be not backward compatible.

I think it is backward compatible since we unmarshal service account JSON that contains a `type = "service_account"`

jkaflik (commented):
Review follow-ups addressed.

.env.example
jkaflik: Applied in d57cba6e.

flow/connectors/bigquery/auth.go
jkaflik: Confirmed. Legacy peer creation paths populate `auth_type = "service_account"`: the OSS UI maps the service-account JSON `type`, and ClickPipes assigns the decrypted service account type explicitly. Also, before this PR an empty type already failed `GcpServiceAccount.Validate`, where every field is required, so the strict check does not reject a previously valid legacy configuration. The tests retain coverage for the explicit legacy value and for rejecting empty or unknown values.

jkaflik:
Published the current PR head (`d57cba6e`) to [`customer-clp-962`](https://github.com/PeerDB-io/peerdb/tree/customer-clp-962) to produce testable GHCR images.

[Customer Docker workflow](https://github.com/PeerDB-io/peerdb/actions/runs/30425755038) completed successfully. Images are available with the tags `latest-clp-962` and `clp-962-d57cba6e`, for example `ghcr.io/peerdb-io/flow-worker:clp-962-d57cba6e`.

flow/connectors/bigquery/auth_test.go
```diff
@@ -0,0 +1,415 @@
...
+	)
+}
+
+func TestResolveWorkloadIdentityDeploymentConfig(t *testing.T) {
+	environment := map[string]string{
+		workloadIdentityServiceAccountEnv: "tenant@tenant-project.iam.gserviceaccount.com",
+		workloadIdentityTokenFileEnv:      "/token",
+	}
+	config, err := resolveWorkloadIdentityDeploymentConfig(t.Context(), workloadIdentityConfigSource{
+		lookupEnv: func(name string) (string, bool) {
```
ilidemi: Looks a bit strange to thread a stdlib function as a callback. Can the same be achieved with t.Setenv()?

flow/connectors/bigquery/auth.go
```diff
@@ -0,0 +1,269 @@
...
+	tokenFile            string
+	projectID            string
+	clusterLocation      string
+	clusterName          string
+}
+
+type workloadIdentityConfigSource struct {
+	lookupEnv              func(string) (string, bool)
+	projectID              func(context.Context) (string, error)
+	instanceAttributeValue func(context.Context, string) (string, error)
```
ilidemi: Have you explored GCE_METADATA_HOST? Callbacks are a bit hard on the brain

RomeroGaliza:
I understand this is a change on a deprecated connector, but it would be immensely helpful to have workload identity supported, relying exclusively on JSON keys is an operational hassle. Thank you @jkaflik!

coderanger:
https://github.com/PeerDB-io/peerdb/issues/4104 is a related issue, and just to confirm you looked at it, the underlying crate does support all of this natively.

morsapaes:
@RomeroGaliza, just to clarify: BQ is only deprecated as a target, not as a source. We're evolving the source connector in the coming months to support query-based CDC.

masterashu (approved):

PR #4609, opened 2026-07-21
fix: handle drop column on clickhouse and remove from schema in clickpipes

masterashu:
### Problem
When customer removes a column from table in Clickhouse, Normalize fails due to SQL query error mentioning unknown column.

In order to deal with this, we had to change the source schema for the table to remove that column.

This PR aims to make this easier by identifying if the column is deleted and retrying and updating the schema.

We check the errors in Normalization step and match if its occurring from a missing column, if it is we refresh the schema and retry normalization and also update the schema.

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -757,6 +757,37 @@ func (a *FlowableActivity) startNormalize(
 		}
 
 		a.recordDeliveryLag(ctx, config.FlowJobName, res.StartBatchID, res.EndBatchID)
+		// If any destination columns were auto-removed during normalize (user dropped them), persist
+		// the updated schemas to the catalog and warn the user.
+		if len(res.UpdatedSchemaMapping) > 0 {
+			tableNames := make([]string, 0, len(res.UpdatedSchemaMapping))
+			for name := range res.UpdatedSchemaMapping {
+				tableNames = append(tableNames, name)
+			}
+			if updateErr := internal.ReadModifyWriteTableSchemasToCatalog(
+				ctx, a.CatalogPool, logger, config.FlowJobName, tableNames,
+				func(schemas map[string]*protos.TableSchema) (map[string]*protos.TableSchema, error) {
+					result := make(map[string]*protos.TableSchema, len(schemas))
+					for name, schema := range schemas {
+						if updated, ok := res.UpdatedSchemaMapping[name]; ok {
+							result[name] = updated
+						} else {
+							result[name] = schema
+						}
+					}
+					return result, nil
+				},
+			); updateErr != nil {
+				logger.Error("failed to persist auto-refreshed table schemas to catalog",
+					slog.Any("error", updateErr))
+			}
+			for name, schema := range res.UpdatedSchemaMapping {
+				tableNameSchemaMapping[name] = schema
+				a.Alerter.LogFlowWarning(ctx, config.FlowJobName,
+					fmt.Errorf("destination column(s) were dropped from table %s and automatically removed "+
+						"from the mirror schema; future syncs will omit those columns", name))
+			}
+		}
 
 		logger.Info("normalized batches",
 			slog.Int64("startBatchID", res.StartBatchID), slog.Int64("endBatchID", res.EndBatchID), slog.Int64("syncBatchID", batchID))
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"slices"
 	"sync/atomic"
 	"time"
 
@@ -757,36 +758,8 @@ func (a *FlowableActivity) startNormalize(
 		}
 
 		a.recordDeliveryLag(ctx, config.FlowJobName, res.StartBatchID, res.EndBatchID)
-		// If any destination columns were auto-removed during normalize (user dropped them), persist
-		// the updated schemas to the catalog and warn the user.
-		if len(res.UpdatedSchemaMapping) > 0 {
-			tableNames := make([]string, 0, len(res.UpdatedSchemaMapping))
-			for name := range res.UpdatedSchemaMapping {
-				tableNames = append(tableNames, name)
-			}
-			if updateErr := internal.ReadModifyWriteTableSchemasToCatalog(
-				ctx, a.CatalogPool, logger, config.FlowJobName, tableNames,
-				func(schemas map[string]*protos.TableSchema) (map[string]*protos.TableSchema, error) {
-					result := make(map[string]*protos.TableSchema, len(schemas))
-					for name, schema := range schemas {
-						if updated, ok := res.UpdatedSchemaMapping[name]; ok {
-							result[name] = updated
-						} else {
-							result[name] = schema
-						}
-					}
-					return result, nil
-				},
-			); updateErr != nil {
-				logger.Error("failed to persist auto-refreshed table schemas to catalog",
-					slog.Any("error", updateErr))
-			}
-			for name, schema := range res.UpdatedSchemaMapping {
-				tableNameSchemaMapping[name] = schema
-				a.Alerter.LogFlowWarning(ctx, config.FlowJobName,
-					fmt.Errorf("destination column(s) were dropped from table %s and automatically removed "+
-						"from the mirror schema; future syncs will omit those columns", name))
-			}
+		if len(res.RemovedColumnsMapping) > 0 {
+			a.applyRemovedColumnsFromNormalize(ctx, logger, config.FlowJobName, res.RemovedColumnsMapping, tableNameSchemaMapping)
 		}
 
 		logger.Info("normalized batches",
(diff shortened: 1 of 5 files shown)


jgao54:
@masterashu  could you add a description for the expected behavior for this PR?  Thanks!

flow/connectors/clickhouse/normalize.go
```diff
@@ -625,10 +671,98 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
+	colMapping, err := peerdb_clickhouse.GetTableColumnsMapping(ctx, c.logger, c.database, []string{gen.TableName})
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to query destination table columns for schema refresh: %w", err)
+	}
+	actualColSet := make(map[string]struct{}, len(colMapping[gen.TableName]))
+	for _, col := range colMapping[gen.TableName] {
+		actualColSet[col.Name] = struct{}{}
+	}
+
+	currentSchema := gen.tableNameSchemaMapping[gen.TableName]
```
jgao54: The `gen.tableNameSchemaMapping` does not represent the freshest schema stored in the catalog, but instead represent the schema that was fetched from catalog earlier at the start of the `startNormalize`. This introduces a race condition between Sync and Normalize flow: 
1. normalize fetch schemas from catalog
2. sync applies an `add column` DDL and updated the catalog 
3. normalize take this schemas from 1, removes the deleted column, and saves it back to catalog

The catalog is now stuck with a stale schema with a missing new column

Since we only care about column removal in the error handling, instead of passing back the updated schemas, pass back the removed column instead for the NormalizeResponse. Then in the readModifyLoop you can focus on just removing the deleted column from the fresh schema fetched from catalog, instead of replacing the entire schema today.

flow/connectors/clickhouse/normalize.go
```diff
@@ -625,10 +671,98 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
+		resp.UpdatedSchemaMapping = updatedSchemas
+	}
+	return resp, nil
+}
+
+// refreshSchemaAndRebuildQuery queries the actual columns present in the destination ClickHouse table,
+// removes any columns from the schema that no longer exist there, and returns a rebuilt normalize query
+// together with the filtered schema. This is used to recover from ErrNoSuchColumnInTable when the user
+// has manually dropped a column from the destination after it was dropped at the source.
+func (c *ClickHouseConnector) refreshSchemaAndRebuildQuery(
```
jgao54, line 738: the e2e tests are great; let's also add unit tests for `refreshSchemaAndRebuildQuery`

flow/connectors/clickhouse/normalize.go
```diff
@@ -524,6 +531,44 @@ func (c *ClickHouseConnector) NormalizeRecords(
 					slog.Int("parallelWorker", i))
 
 				if err := c.execWithConnection(errCtx, chConn, q.query); err != nil {
+					// If a destination column is missing (user dropped it), try to auto-recover
+					// by refreshing the schema and retrying the INSERT without that column.
+					var chEx *clickhouse.Exception
+					if errors.As(err, &chEx) && ch_proto.Error(chEx.Code) == ch_proto.ErrNoSuchColumnInTable {
```
jgao54: nit:  `errors.AsType[*clickhouse.Exception](err)`

flow/connectors/clickhouse/normalize.go
```diff
@@ -625,10 +671,98 @@ func (c *ClickHouseConnector) NormalizeRecords(
...
+
+	if len(removedCols) == 0 {
+		return "", nil, fmt.Errorf(
+			"no user columns were absent from destination table %s despite ErrNoSuchColumnInTable; "+
+				"the missing column may be a PeerDB internal column", gen.TableName)
+	}
+
+	c.logger.Warn("[clickhouse] auto-removing destination-dropped columns from schema",
+		slog.String("table", gen.TableName),
+		slog.Any("removedColumns", removedCols))
```
jgao54: iiuc, since we don't check against _source db_ schema,  if the user drops the column from ClickHouse today, but does not drop it from source, we will gracefully recover in this scenario as well and continue to replicate without the dropped column; and this column essentially becomes an excluded column, unless the table gets resynced which it would reappear again.

I think this is more of a product decision on whether we should fail loudly here, vs. we should continue.  My only concern here with continue is if customer accidentally dropped a column, then previously it would fail the pipe, they would have manually added it back and the pipe would recover. vs with this change manually adding back the column would do nothing, and would require a resync. 

If this is intended behavior we should have an user-facing log for it

cc @morsapaes  on thoughts here.

flow/activities/flowable_core.go
```diff
@@ -757,6 +757,37 @@ func (a *FlowableActivity) startNormalize(
...
+			); updateErr != nil {
+				logger.Error("failed to persist auto-refreshed table schemas to catalog",
+					slog.Any("error", updateErr))
+			}
+			for name, schema := range res.UpdatedSchemaMapping {
+				tableNameSchemaMapping[name] = schema
+				a.Alerter.LogFlowWarning(ctx, config.FlowJobName,
+					fmt.Errorf("destination column(s) were dropped from table %s and automatically removed "+
+						"from the mirror schema; future syncs will omit those columns", name))
+			}
```
jgao54: If `ReadModifyWriteTableSchemasToCatalog` fails above, this would still add a user-facing warning that destination columns were dropped. should only surface log if update succeeds

flow/connectors/clickhouse/normalize.go
```diff
@@ -524,6 +531,44 @@ func (c *ClickHouseConnector) NormalizeRecords(
 					slog.Int("parallelWorker", i))
 
 				if err := c.execWithConnection(errCtx, chConn, q.query); err != nil {
```
jgao54: nit: can factor out exec into a helper to improve readability, currently the error handling with continue is a bit hard to follow; and there are some duplicated loggings introduced that wouldn't be needed. e.g.
```
func (c *ClickHouseConnector) execNormalizeQuery(
      ctx context.Context,
      logger log.Logger,
      conn clickhouse.Conn,
      q queryInfo,
) (*protos.TableSchema, error) { 
      err := c.execWithConnection(ctx, conn, q.query)
      if err == nil {
              return nil, nil
      }

      chEx, ok := errors.AsType[*clickhouse.Exception](err)
      if !ok || chEx == nil || ch_proto.Error(chEx.Code) != ch_proto.ErrNoSuchColumnInTable {
              return nil, err
      }

      logger.Warn("[clickhouse] missing column in destination, attempting schema refresh and retry",
              slog.Any("error", err))
      retryQuery, refreshedSchema, refreshErr := c.refreshSchemaAndRebuildQuery(ctx, q.generator)
      if refreshErr != nil {
              logger.Error("[clickhouse] failed to refresh schema after missing column error",
                      slog.Any("error", refreshErr))
              return nil, err
      }
      if retryErr := c.execWithConnection(ctx, conn, retryQuery); retryErr != nil {
              logger.Error("[clickhouse] retry after schema refresh failed",
                      slog.String("query", retryQuery), slog.Any("error", retryErr))
              return nil, err
      }
      logger.Warn("[clickhouse] successfully retried normalize after dropping missing destination columns")
      return refreshedSchema, nil
}
```

flow/activities/flowable_core.go
```diff
@@ -757,6 +757,37 @@ func (a *FlowableActivity) startNormalize(
 		}
 
 		a.recordDeliveryLag(ctx, config.FlowJobName, res.StartBatchID, res.EndBatchID)
+		// If any destination columns were auto-removed during normalize (user dropped them), persist
```
jgao54: nit: better implemented as a helper as this method is getting quite long

flow/connectors/clickhouse/normalize.go
jgao54: Just realized that if user dropped the column and then add it back, the historical data would be lost anyways. so maybe continuing the pipe is the more graceful option here (and user will just have to resync if it was a mistake)

flow/activities/flowable_core.go
```diff
@@ -767,6 +771,66 @@ func (a *FlowableActivity) startNormalize(
...
+			result := make(map[string]*protos.TableSchema, len(schemas))
+			for name, schema := range schemas {
+				result[name] = removeColumnsFromSchema(schema, removedColumnsMapping[name])
+			}
+			return result, nil
+		},
+	)
+	if updateErr != nil {
+		logger.Error("failed to persist auto-removed destination columns to catalog",
+			slog.Any("error", updateErr))
```
jgao54, line 798: this should fail loudly and get retried. while it's also reasonable to just let the next iteration pick up the missing column if there is a transient error, if there's any internal errors we wouldn't be notified and it would always fail to persist to catalog.

jgao54: also means can remove the `updateErr == nil` check below

flow/activities/flowable_core.go
```diff
@@ -767,6 +771,66 @@ func (a *FlowableActivity) startNormalize(
...
+
+	for name, removed := range removedColumnsMapping {
+		if schema, ok := tableNameSchemaMapping[name]; ok {
+			tableNameSchemaMapping[name] = removeColumnsFromSchema(schema, removed)
+		}
+		if updateErr == nil {
+			a.Alerter.LogFlowWarning(ctx, flowJobName,
+				fmt.Errorf("destination column(s) were dropped from table %s and automatically removed "+
+					"from the mirror schema; future syncs will omit those columns", name))
+		}
```
jgao54, line 809: currently the warning message doesn't specify columns that are dropped, let's make the log more informative about that, since we now have this information

jgao54:
Codex identified a race condition worth mentioning:

Today the `NormalizeRecords` method will not only insert the rows into the destination table in CH, it will also update the offset to catalog on the _table_  level (see `SetLastNormalizedBatchIDForTable` that updates `metadata_last_sync_state` table). So if a batch fails but one or more tables already succeeded normalize in the table, on retry those succeeded table will be skipped in the batch, so far this is an intended optimization.

The edge case that can cause correctness issue is:
  1. A user manually drops a column in CH.
  2. Normalize recovers for that table, as intended by this PR.
  3. Another table in the same normalize batch fails and stuck in retry loop.
  4. Correction to the catalog never happened because of 3.
  5. Postgres drops and re-adds that same column during this time. 

Given catalog never removed the column, pg's ADD COLUMN ddl got ignored since no schema delta was detected, so ClickHouse never re-added that column. In this case, once the normalize error finally got unblocked, subsequent normalize will silently perform the "schema correction" introduced in this PR, and then permanently skip replicating the column that was added back.

The odds of this happening is pretty small, but not zero.  Here's a potential best-effort workaround: https://github.com/PeerDB-io/peerdb/pull/4678. It's still not perfect because if normalize fails _and_ catalog correction fails, catalog correction doesn't get retried. A more bulletproof solution would be to update schema _and_ advance offset together when there are column removals, but I am not sure if this additional  code complexity is worthwhile given the unlikely scenario.

I am also okay with not trying to handle this edge case at all, if we feel that this is getting too much into the weeds. If user delete a column in CH accidentally and added it back, is probably warrants a resync anyways.  if user delete a column in PG and added it back immediately, it's also very unlikely they also manually removed the column from ClickHouse in the same batch.

jgao54:
discussed separately in slack, would be safer to add a source-check to make sure columns are also removed from source table before updating the catalog; or keep erroring otherwise. 

This also allow us to keep the invariant of replicated columns = all columns - excluded columns, while allowing user to manually drop columns in clickhouse that have already been deleted from source, if they desire.

PR #4612, opened 2026-07-22
fix: estimate row counts for postgres partitioned/inherited tables

itsbilal:
Previously, Postgres tables that other tables inherit from, or that have manually-added partitions, did not go through the estimated row count logic when determining the number of partitions for snapshotting. Instead, we ran a precise COUNT(*) which was an expensive operation.

This change updates the estimation logic to walk through the tree of partitions and children tables and sum up estimated rows across all of them, instead of falling through to the COUNT(*) query.


itsbilal:
Closing and resubmitting to avoid the external-contributor flow

PR #4613, opened 2026-07-22
fix: estimate row counts for postgres partitioned/inherited tables

itsbilal:
Previously, Postgres tables that other tables inherit from, or that have manually-added partitions, did not go through the estimated row count logic when determining the number of partitions for snapshotting. Instead, we ran a precise COUNT(*) which was an expensive operation.

This change updates the estimation logic to walk through the tree of partitions and children tables and sum up estimated rows across all of them, instead of falling through to the COUNT(*) query.

### DIFF
--- a/flow/connectors/postgres/qrep_partition.go
+++ b/flow/connectors/postgres/qrep_partition.go
@@ -130,17 +130,18 @@ func CTIDBlockPartitioningFunc(ctx context.Context, pp PartitionParams) ([]*prot
 	}
 	switch {
 	case tc.relkind == "p":
-		blocksPerTable, err := getPartitionedTables(ctx, pp.tx, tc.oid)
+		blockStatsPerTable, err := getPartitionedTables(ctx, pp.tx, tc.oid)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get child partitioned tables: %w", err)
 		}
-		return ctidPartitionsForChildTables(pp, blocksPerTable)
+		return ctidPartitionsForChildTables(pp, blockStatsPerTable)
 	case tc.relhassubclass:
-		blocksPerTable, err := getInheritedTables(ctx, pp.tx, tc.oid, tc.qualifiedName)
+		visited := make(map[uint32]struct{})
+		blockStatsPerTable, err := getInheritedTables(ctx, pp.tx, tc.oid, tc.qualifiedName, visited)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get child inherited tables: %w", err)
 		}
-		return ctidPartitionsForChildTables(pp, blocksPerTable)
+		return ctidPartitionsForChildTables(pp, blockStatsPerTable)
 	default:
 		return ctidPartitionsForTable(ctx, pp)
 	}
@@ -227,11 +228,11 @@ func ctidPartitionsForTable(ctx context.Context, pp PartitionParams) ([]*protos.
 //   - partition 3: [T3:0-9]
 func ctidPartitionsForChildTables(
 	pp PartitionParams,
-	blocksPerTable map[string]int64,
+	blockStatsPerTable map[string]tableBlockStats,
 ) ([]*protos.QRepPartition, error) {
 	var totalBlocks int64
-	for _, blocks := range blocksPerTable {
-		totalBlocks += blocks
+	for _, blockStats := range blockStatsPerTable {
+		totalBlocks += blockStats.blockCount
 	}
 	if totalBlocks == 0 {
 		pp.logger.Warn("zero non-empty child tables found")
@@ -240,10 +241,10 @@ func ctidPartitionsForChildTables(
 
 	pp.logger.Info("child tables detected",
 		slog.String("table", pp.watermarkTable),
-		slog.Int("numChildTables", len(blocksPerTable)),
+		slog.Int("numChildTables", len(blockStatsPerTable)),
 		slog.Int64("totalBlocks", totalBlocks))
 
-	sortedTables := slices.Sorted(maps.Keys(blocksPerTable))
+	sortedTables := slices.Sorted(maps.Keys(blockStatsPerTable))
 	blocksPerPartition := max(int64(1), shared.DivCeil(totalBlocks, pp.numPartitions))
 
 	var partitions []*protos.QRepPartition
@@ -255,7 +256,7 @@ func ctidPartitionsForChildTables(
 
 		for remaining > 0 && childIdx < len(sortedTables) {
 			tableName := sortedTables[childIdx]
-			blocks := blocksPerTable[tableName]
+			blocks := blockStatsPerTable[tableName].blockCount
 			remainingBlocks := blocks - blockOffset
 			consume := min(remaining, remainingBlocks)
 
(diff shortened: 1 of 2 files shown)


flow/connectors/postgres/qrep_partition.go
```diff
@@ -470,35 +487,79 @@ func ComputeNumPartitions(ctx context.Context, pp PartitionParams, numRowsPerPar
...
-		} else if v.Valid {
-			totalRows = v.Int64
+		switch {
+		case tc.relkind == "p": // Partitioned table
+			blockStatsPerTable, err := getPartitionedTables(ctx, pp.tx, tc.oid)
+			if err != nil {
+				return 0, fmt.Errorf("failed to get child partitioned tables: %w", err)
+			}
+			for _, blockStats := range blockStatsPerTable {
+				totalRows += int64(blockStats.blockDensity * float64(blockStats.blockCount))
```
ilidemi: If reltuples = -1 for a partition/inherited table, this could add a negative term, right? Then only if we're net negative, we fall back to full count. How about coalescing negatives to zero, and also triggering the fallback if <80% came out non-negative?

flow/connectors/postgres/qrep_partition.go
```diff
@@ -431,7 +451,10 @@ func getInheritedTables(ctx context.Context, tx pgx.Tx, parentOID uint32, parent
 
 	for _, child := range children {
```
ilidemi, line 460: Codex flagged that the existing logic can get exponential if there's multiple inheritance (with dedup at the end). Could do a seen oid set if you're in the mood for graph traversal

flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -779,6 +779,160 @@ func TestCTIDPartitioningOnEmptyInheritedTable(t *testing.T) {
...
+	for _, tbl := range tables {
+		if _, err := conn.Exec(t.Context(), "ANALYZE "+tbl); err != nil {
+			t.Fatalf("Failed to ANALYZE %s: %v", tbl, err)
+		}
+	}
+}
+
+// TestComputeNumPartitionsForPartitionedTable verifies that ComputeNumPartitions sums the
+// estimated row counts of the leaf partitions (the partitioned parent itself stores no rows).
+func TestComputeNumPartitionsForPartitionedTable(t *testing.T) {
```
ilidemi: For the test set, let's replicate the cases used for partitioning above: + multilevel tables and empty children.
Also testing multiple inheritance and that we avoided count(*) (if that's possible to verify) would be good

flow/connectors/postgres/qrep_partition.go
jgao54: Another edge case we have encountered in the past was bigint out of range on a single table due to bad stats ( https://github.com/PeerDB-io/peerdb/pull/4393 fixed it by using numeric). With the numeric fix, if it exceeds MaxInt64, it would always fall back to exact count.  

With this change, depending on the number of time overflow happens and wraps around, could in theory end up with a value that's close to MaxInt64. Which will pass `totalRows <= 0`, but later on can fail in weird ways in `AdjustNumPartitions` with `DivCeil(totalRows, desiredRowsPerPartition)` overflowing and generating a negative adjustedRowsPerPartition. This is technically already possible today, but the code-level overflow can make it more likely.

itsbilal: @jgao54 good point about the overflow cases! I've addressed the code-side rollup overflow now, let me know if it looks good or if you also want to address the case where the SQL query returns an out-of-int64 bounds value.

flow/connectors/postgres/qrep_partition.go
```diff
@@ -470,35 +497,100 @@ func ComputeNumPartitions(ctx context.Context, pp PartitionParams, numRowsPerPar
...
 	var totalRows int64
 	var whereClause string
 	var queryArgs []any
+	rollupSubclassCounts := func(blockStatsPerTable map[string]tableBlockStats) int64 {
+		var invalidCounts int64
+		var runningSum int64
+		for _, blockStats := range blockStatsPerTable {
+			estimatedCount := int64(blockStats.blockDensity * float64(blockStats.blockCount))
+			if estimatedCount <= 0 {
+				invalidCounts++
```
jgao54, line 543: nit: I think we'd want to `continue` after incrementing invalid count

jgao54: ah nvm. i see that `runningSum += max(0, estimatedCount)` does the same thing below, that works too

flow/connectors/postgres/qrep_partition.go
```diff
@@ -470,35 +497,100 @@ func ComputeNumPartitions(ctx context.Context, pp PartitionParams, numRowsPerPar
...
+	const preciseCountTemplate = "SELECT COUNT(*) FROM %s %s"
+
 	var totalRows int64
 	var whereClause string
 	var queryArgs []any
+	rollupSubclassCounts := func(blockStatsPerTable map[string]tableBlockStats) int64 {
+		var invalidCounts int64
+		var runningSum int64
+		for _, blockStats := range blockStatsPerTable {
+			estimatedCount := int64(blockStats.blockDensity * float64(blockStats.blockCount))
```
jgao54, line 541: nit: Claude says int64(<value>) overflow behavior is different per architecture. on amd64 this is math.MinInt64, but on arm64 this is math.MaxInt64. I think this is unlikely to trip us because if we get max.MaxInt64 this will likely overflow in the next iteration.  but if you want the invalidCounts here to be precise, I'd also check for estimatedCount >= math.MaxInt64.

flow/connectors/postgres/qrep_partition_test.go
```diff
@@ -779,6 +779,380 @@ func TestCTIDPartitioningOnEmptyInheritedTable(t *testing.T) {
...
+			require.NoError(t, err)
+		}
+	}
+	analyzeTables(t, conn, allTables)
+
+	// The estimate should cover every table in the hierarchy.
+	estimated, tables := sumEstimatedRows(t, conn, grandparent)
+	require.Len(t, tables, len(allTables))
+	require.InDelta(t, totalRows, estimated, float64(totalRows)*0.1,
+		"estimated rows %d should be within 10%% of actual %d", estimated, totalRows)
```
jgao54, line 1122: ooc, could this introduce flakes? or is it guaranteed that EXPLAIN will keep it within 10%

jgao54 (approved):
a few nits, otherwise lgtm!

flow/connectors/postgres/qrep_partition.go
jgao54: > where the SQL query returns an out-of-int64 bounds value

Ah, so today this is handled when it's out-of-int64 bound [here](https://github.com/PeerDB-io/peerdb/blob/5ff0ef476d106f6e12713f7c68533da398253fc1/flow/connectors/postgres/qrep_partition.go#L496-L499), the unlikely edge case would be if it's just slightly under math.MaxInt64, and _then_ overflow in `AdjustNumPartitions`.  So I think it's okay to leave it out-of-scope for this PR

flow/connectors/postgres/qrep_partition.go
ilidemi: This would be super defensive I think, as PG supports 8KB per page * 2^32 pages, so someone would have to fudge the PG stats to break the math. But OTOH adding defenses is only a handful of LOC

ilidemi (approved):

flow/connectors/postgres/qrep_partition.go
itsbilal: Done - although the linter thinks no `int64` can exceed `math.MaxInt64` so that conditional should always be false, but defensively it's there now in case the value of MaxInt64 changes depending on arch.

flow/connectors/postgres/qrep_partition_test.go
itsbilal: The analyze at the start should keep it very close to the actual - and I've stressed these tests for a while and seen no failures so we should be good.

PR #4616, opened 2026-07-23
ttl_only_drop_parts=1

serprex:
It came up that putting this on raw table reduces system load for 90 day ttl


ilidemi (approved):
CH 19, nice to not need backcompat

PR #4617, opened 2026-07-24
fix string partitioning edge case to avoid rabbit-holing into single-row partitions

jgao54:
Follow up to https://github.com/PeerDB-io/peerdb/pull/4522

Main thing to clean up is the edge case where if both start and end boundary string's first differing byte (the most significant byte for determining the midpoint) are below or above printable ascii, continuing with midpoint interpolation in the extreme case can result in generating a lot of single-row partitions, until we are left with one large final partition. Terminate early when this happens to avoid introducing unnecessary latency. It may still be a full table snapshot, but we don't want to generate a lot of meaningless partitions.

### DIFF
--- a/flow/connectors/mysql/qrep_partition.go
+++ b/flow/connectors/mysql/qrep_partition.go
@@ -136,7 +136,7 @@ func stringMidpoint(s1 string, s2 string) (string, bool) {
 	s1, s2 = s1[i:], s2[i:]
 
 	// When the first differing byte in both strings are outside printable ASCII
-	// (both below or both above), treat it as unsplittable so we don't rabbit-holing
+	// (both below or both above), treat it as unsplittable so we aren't rabbit-holing
 	// into generating a lot of single-row partitions.
 	var b1, b2 byte
 	if s1 != "" {
--- a/flow/connectors/mysql/qrep_partition.go
+++ b/flow/connectors/mysql/qrep_partition.go
@@ -116,7 +116,11 @@ const (
 	base95Width = 8 // 95^8 to fit in an uint64
 )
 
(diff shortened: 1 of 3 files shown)


flow/connectors/mysql/qrep_partition_test.go
```diff
@@ -177,46 +177,48 @@ func TestBuildUuidStringPartitionsInvalid(t *testing.T) {
...
-		t.Run(tc.name, func(t *testing.T) {
-			mid := stringMidpoint(tc.s1, tc.s2)
-			assert.Equal(t, tc.expected, mid)
-			assert.LessOrEqual(t, tc.s1, mid)
-			assert.LessOrEqual(t, mid, tc.s2)
-		})
-	}
-}
-
-func TestStringMidpointRemainsValidUTF8(t *testing.T) {
```
jgao54, line 203: wrapped this up into test above; since we are now checking ascii boundaries, most of these cases would no longer be hit, as they would be deemed unsplittable before then.

flow/connectors/mysql/qrep_partition.go
```diff
@@ -130,8 +134,23 @@ func stringMidpoint(s1 string, s2 string) string {
 	}
 	sharedPrefix := s1[:i]
 	s1, s2 = s1[i:], s2[i:]
+
+	// When the first differing byte in both strings are outside printable ASCII
+	// (both below or both above), treat it as unsplittable so we don't rabbit-holing
```
ilidemi:
```suggestion
	// (both below or both above), treat it as unsplittable so we aren't rabbit-holing
```

ilidemi (approved):

PR #4619, opened 2026-07-25
fix bug in row count estimation

jgao54:
`pgtype.Numeric.Int64Value()` does not automatically convert floating values to int64, and will return an error for any non-integer values.  Reproduced error locally and verified that this is fixed.

Fixes: DBI-956

### DIFF
--- a/flow/connectors/postgres/qrep_partition.go
+++ b/flow/connectors/postgres/qrep_partition.go
@@ -492,7 +492,7 @@ func estimatedNumRowsForTable(ctx context.Context, pp PartitionParams) (int64, e
 			-- row estimation formula
 			CASE
 				WHEN c.reltuples >= 0 AND c.relpages > 0 AND NOT c.relhassubclass
-				    THEN c.reltuples::numeric / c.relpages * (pg_relation_size(c.oid) / current_setting('block_size')::integer)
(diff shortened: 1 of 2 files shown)


ilidemi (approved):

flow/connectors/postgres/qrep_partition.go
```diff
@@ -492,7 +492,7 @@ func estimatedNumRowsForTable(ctx context.Context, pp PartitionParams) (int64, e
 			-- row estimation formula
 			CASE
 				WHEN c.reltuples >= 0 AND c.relpages > 0 AND NOT c.relhassubclass
-				    THEN c.reltuples::numeric / c.relpages * (pg_relation_size(c.oid) / current_setting('block_size')::integer)
+				    THEN round(c.reltuples::numeric / c.relpages * (pg_relation_size(c.oid) / current_setting('block_size')::integer))
```
ilidemi, line 495: Are the #4613 additions affected as well?

jgao54: nope that code is fine. it doesn't use the same logic (the blockDensity * blockCount generates a float, and then casted to int64 via code), so doesn't get impacted by this bug.

PR #4622, opened 2026-07-27
CI/Upload test results action: Fail with error when the upload log doesn't contain the summary header

pfcoperez:
This addressed the improbable corner case of missing results reports being silently ignored resulting in not failed builds were test results ingestion haven't actually happened.


pfcoperez:
## QA Validation

Ingestion is still working for run `30254798351`: https://github.com/PeerDB-io/peerdb/actions/runs/30254798351?pr=4622

```sql
SELECT MIN(timestamp), MAX(timestamp) FROM ci_peerdb_test_runs WHERE workflow_run_id = '30254798351';
```
 
```
2026-07-27 09:41:57	2026-07-27 09:42:04
```

ilidemi (approved):

PR #4623, opened 2026-07-27
Retry catalog and Temporal connections at startup

dtunikov:
## Why

`WorkerSetup`, `SnapshotWorkerMain` and `APIMain` each dial the catalog and Temporal exactly once. Any error propagates to `main.go`, which panics — so a single slow dial takes the pod down.

We hit this during releases, most often on `maintenance-worker` (a Helm pre/post-upgrade hook, so it cold-starts twice per release while everything else is churning):

```
panic: unable to create Temporal client: failed reaching server: context deadline exceeded
panic: unable to create catalog connection pool: unable to establish connection with catalog: failed to connect to ...
```

Neither dependency recovers on its own: Temporal's `GetSystemInfo` deadline is a single-shot 5s, and pgxpool defaults `ConnectTimeout` to 2 minutes.

## What

- `connectWithRetry` in `flow/cmd/startup.go`: exponential backoff (1s → 10s) over a 3 minute budget, each attempt bounded at 30s so a hanging dial can't eat the budget. Logs a warning per failed attempt so a genuinely misconfigured secret still surfaces.
- Applied to the catalog and Temporal dials in worker, snapshot-worker and api.
- `client.Dial` → `client.DialContext`. `Dial` ignores the caller's context and uses `context.Background()`, so without this the per-attempt timeout and SIGTERM handling had no effect on the Temporal dial.

Unit tests cover immediate success, success after a failure, abort on cancellation, and the per-attempt deadline.

### DIFF
--- a/flow/cmd/api.go
+++ b/flow/cmd/api.go
@@ -204,7 +204,9 @@ func APIMain(ctx context.Context, args *APIServerParams) error {
 		Meter: metricsProvider.Meter("temporal-sdk-go"),
 	})
 
-	tc, err := setupTemporalClient(ctx, clientOptions)
+	tc, err := connectWithRetry(ctx, "temporal", func(ctx context.Context) (client.Client, error) {
+		return setupTemporalClient(ctx, clientOptions)
+	})
 	if err != nil {
 		return fmt.Errorf("unable to create Temporal client: %w", err)
 	}
@@ -240,7 +242,7 @@ func APIMain(ctx context.Context, args *APIServerParams) error {
 		)),
 	)
 
-	catalogPool, err := internal.GetCatalogConnectionPoolFromEnv(ctx)
+	catalogPool, err := connectWithRetry(ctx, "catalog", internal.GetCatalogConnectionPoolFromEnv)
 	if err != nil {
 		return fmt.Errorf("unable to get catalog connection pool: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


ilidemi (approved):
Does MaintenanceMain need to be updated as well?

dtunikov:
Added a second commit extending the same retry to the maintenance CLI (`flow/cmd/maintenance.go`): the Temporal dial in `MaintenanceMain`, plus the catalog pool in `WriteMaintenanceOutputToCatalog` and `ReadLastMaintenanceOutput`.

The maintenance Jobs have `backoffLimit: 1000`, so these were already eventually-retried — but at process granularity, with the kubelet backing off exponentially from 10s while Helm blocks on the hook.

`WriteMaintenanceOutputToCatalog` is the case that actually needed fixing. It runs *after* the start-maintenance workflow has completed, and `getMaintenanceWorkflowOptions` uses `ALLOW_DUPLICATE` reuse with `USE_EXISTING` conflict resolution. `USE_EXISTING` only attaches to a *running* workflow, so once the run has closed a restarted pod starts a fresh execution — meaning a transient catalog blip on a one-row insert re-runs the whole `StartMaintenanceWorkflow`.

Not touched: `GetVersion` / `ListMirrors` in `skipStartMaintenanceIfNeeded` are unretried gRPC calls to flow-api that fail fast with `Unavailable` if it's mid-rollout. Different dependency, wants a gRPC retry policy or `WaitForReady` rather than this helper.

PR #4629, opened 2026-07-27
fix(clickhouse): check for S3 grant before validating CH-bound pipe

itsbilal:
Previously, if custom roles were selected instead of all access when creating a pipe, we did not validate that the ClickHouse user can read from the staging S3 bucket in the final step of ingestion. This would result in pipes getting stuck in provisioning without an obvious user-facing error.

This change updates the ValidateClickHousePeer function to also run a CHECK GRANT that's as version-agnostic as possible, to confirm if the user can actually read from S3. If the CHECK GRANT returns a 0, we throw a validation failure.

### DIFF
--- a/flow/pkg/clickhouse/validation.go
+++ b/flow/pkg/clickhouse/validation.go
@@ -137,12 +137,49 @@ func ValidateClickHouseHost(ctx context.Context, chHost string, allowedDomainStr
 		chHost, strings.Join(allowedDomains, ","))
 }
 
+func validateBucketGrant(ctx context.Context, logger log.Logger, conn clickhouse.Conn, storageType string) error {
+	// First check under the new syntax, where the object to check is storageType.
+	// Eg. CHECK GRANT READ on S3. This also passes for users holding the legacy
+	// GRANT S3 ON *.* style grant, so a definitive failure here is final.
+	var grantExists bool
+	if err := QueryRow(ctx, logger, conn, "CHECK GRANT READ ON "+storageType).Scan(&grantExists); err != nil {
+		// Do not return an error on syntax error; this could mean we're on a
+		// CH version that does not support this syntax.
+		var chException *clickhouse.Exception
+		if !errors.As(err, &chException) || chproto.Error(chException.Code) != chproto.ErrSyntaxError {
+			return fmt.Errorf("failed to validate %s read grant: %w", storageType, err)
+		}
+		// NB: syntax error falls through to the next check.
+	} else if !grantExists {
+		return fmt.Errorf("failed to validate %s read grant: user lacks READ on %s (fix with GRANT READ ON %s)",
+			storageType, storageType, storageType)
+	} else {
+		// grantExists and no error.
+		return nil
+	}
+	// Now check under the old syntax, where the object to check is *.*.
+	// Eg. CHECK GRANT S3 on *.*.
+	if err := QueryRow(ctx, logger, conn, fmt.Sprintf("CHECK GRANT %s ON *.*", storageType)).Scan(&grantExists); err != nil {
+		// Similarly, do not error on syntax errors; instead, just log that the check failed.
+		var chException *clickhouse.Exception
+		if !errors.As(err, &chException) || chproto.Error(chException.Code) != chproto.ErrSyntaxError {
+			return fmt.Errorf("failed to validate %s read grant: %w", storageType, err)
+		}
+		logger.Warn("[clickhouse] CHECK GRANT not supported by this ClickHouse version, skipping grant validation")
+	} else if !grantExists {
+		return fmt.Errorf("failed to validate %s read grant: user lacks READ on %s (fix with GRANT %s on *.*)",
+			storageType, storageType, storageType)
+	}
+	return nil
+}
+
 func ValidateClickHousePeer(
 	ctx context.Context,
 	logger log.Logger,
 	allowedDomains string,
 	serviceHost string,
 	conn clickhouse.Conn,
+	stagingAccessMethod string,
 ) error {
 	// Hostname validation
 	if err := ValidateClickHouseHost(ctx, serviceHost, allowedDomains); err != nil {
@@ -211,6 +248,11 @@ func ValidateClickHousePeer(
 	// drop the table
 	if err := Exec(ctx, logger, conn, "DROP TABLE IF EXISTS "+validateDummyTableNameRenamed); err != nil {
 		return fmt.Errorf("failed to drop validation table %s: %w", validateDummyTableNameRenamed, err)
+	}
+
+	// Validate that ClickHouse has access permissions to the staging access bucket.
+	if err := validateBucketGrant(ctx, logger, conn, stagingAccessMethod); err != nil {
+		return err
 	}
 
 	return nil
--- a/flow/pkg/clickhouse/validation_test.go
+++ b/flow/pkg/clickhouse/validation_test.go
@@ -267,3 +267,108 @@ func TestBuildPartitionByValidationQuery(t *testing.T) {
 		require.Equal(t, "SELECT (id % 2) FROM (SELECT CAST(NULL, 'Dynamic') AS `id`) LIMIT 0", query)
 	})
 }
+
+func TestCheckBucketGrantInValidatePeer(t *testing.T) {
+	ctx := t.Context()
+	addr := fmt.Sprintf("%s:%d", testutil.ClickHouseTestHost(), testutil.ClickHouseTestPort())
+	adminConn, err := clickhouse.Open(&clickhouse.Options{Addr: []string{addr}})
+	require.NoError(t, err)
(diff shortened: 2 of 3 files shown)


itsbilal:
Marking as draft; I'm trying to refactor this code in such a way that it's easily reusable by the GCS side without making the imports too annoying. Should be done very soon.

itsbilal:
Test failure aside, this is ready for a review.

flow/pkg/objectstore/validation.go
```diff
@@ -10,27 +11,78 @@ import (
...
 // supply a constructed *s3.Client and the bucket+prefix to test.
 func NewS3StagingValidator(client *s3.Client, bucket, prefix string) StagingValidator {
-	return func(ctx context.Context) error {
+	return func(ctx context.Context, log log.Logger, checker ClickhouseGrantChecker) error {
 		key := strings.TrimPrefix(prefix+"/"+stagingCheckObjectPrefix+uuid.NewString(), "/")
 		body := strings.NewReader(time.Now().Format(time.RFC3339))
 
+		if err := validateBucketGrant(ctx, log, checker, "S3"); err != nil {
+			return err
+		}
```
jgao54: I don't think we want to thread the clickhouse permission check here thru the objectstore validators. In Discovery today we do not use the s3/gcs validator, this was intentional because s3 is considered internal infra, whereas discovery validation is intended for validating user setup. So adding it here would have it silently skipped today. 

Instead, can call the check directly inside `ValidateClickHousePeer`:
```
if stagingBucketIsS3 {
        if err := validateS3ReadGrant(ctx, logger, conn); err != nil {
            return err
        }
}
```

One more NIT is when i read:
```
	if err := validateBucketGrant(ctx, log, checker, "GCS"); err != nil {
		return err
	}
``` 
i assumed that we were trying to check  `READ ON GCS` until i saw:
```
if checker == nil {
	return nil
}
```
I find this a bit implicit and it's better to just hard-code it to S3 given that's all we care about right now.

flow/pkg/objectstore/validation.go
```diff
@@ -10,27 +11,78 @@ import (
...
+	// staging store (e.g. GCS staging is read via url() signed URLs).
+	if checker == nil {
+		return nil
+	}
+	// First check under the new syntax, where the object to check is storageType.
+	// Eg. CHECK GRANT READ on S3. This also passes for users holding the legacy
+	// GRANT S3 ON *.* style grant, so a definitive failure here is final.
+	result, err := checker.CheckGrant("READ", storageType)
+	if err != nil {
+		return fmt.Errorf("failed to validate S3 read grant: %w", err)
```
ilidemi: If storageType is passed as an opaque string, the errors should be formatted with it too

flow/pkg/objectstore/validation.go
ilidemi: Agreed with moving to ValidateClickHousePeer and the duplication between checker and the object. For GCS, we actually need to validate READ ON URL instead, it's a separate grant that's not guaranteed

flow/pkg/clickhouse/validation.go
```diff
@@ -138,12 +138,34 @@ func ValidateClickHouseHost(ctx context.Context, chHost string, allowedDomainStr
 		chHost, strings.Join(allowedDomains, ","))
 }
 
+type grantChecker struct {
```
ilidemi: Was trying to understand the need for more abstractions here and in the tests, and realized it's because we want to test _something_ but the package doesn't have the integration test infra hooked up to do proper CH calls and keep the actual validation change to 1-2 simple functions. We only sorted out the test infra in the recent couple of months, so the only reason it's not simple is historical. Added scaffolding in #4639, let's make the tests set up an actual user with/without grants and see how the validation responds.

flow/pkg/objectstore/validation.go
itsbilal: If we do need a validation in the GCS case (CHECK GRANT READ on URL) should we keep the broad structure the way it is right now, remove the `if checker == nil { return nil } `, and just have the gcs path also call `validateBucketGrant` correctly?

I think that'd be cleanest, but also happy to move the logic to `ValidateClickHousePeer` if y'all think that's better - only reason I didn't want to encode it there is because it feels off-place to have an S3/GCS switch there when we already delegate S3/GCS specific logic to the appropriate `StagingStore`.

itsbilal: I've made this change - let me know how you feel about this approach. Also happy to move this logic out of `StagingStore` entirely and move it in `ValidateClickHousePeer` if that's still preferred.

jgao54: >  move the logic to ValidateClickHousePeer

To clarify, today in Discovery we use `NoStagingValidator`, not `S3StagingValidator`/`GCSStagingValidator`, so `validateBucketGrant` would never be called here; which is why moving it to `ValidateClickHousePeer` might make more sense.

jgao54: >  it feels off-place to have an S3/GCS switch there when we already delegate S3/GCS specific logic to the appropriate StagingStore

if there's a better way to introduce just the s3/gcp perm check logic into the staging store and only only trigger that path, that works too. (I don't feel too strong either way, because currently the validator check write perm using the object storage client; vs. the read perm check you introduced is purely a clickhouse permission check; so i can see it belong to either)

ilidemi: There are two kinds of staging store validations: can PeerDB access the store and does CH user have permissions to do read queries. The first one is only relevant in OSS, as people would do ground-up S3 setups or self-host MinIO, and in our managed environment S3 is set up uniformly and the service doing the validation can't even access it. Only the second one is relevant there.

Reading through the code again, I realized that the whole StagingValidator infra was a result of a misunderstanding of how it all works e2e, and it doesn't need to be in pkg. This must've been confusing to read and try to align with, submitted a cleanup in #4643. ValidateClickHousePeer would lose the last arg, and you'd be able to cleanly add the "S3"/"URL" (or a boolean/enum) instead.

Hopefully now it's clearer how we don't really need extra abstractions besides doing a CH query and validating it with various kinds of users in real CH.

itsbilal: @jgao54 Good catch, I had forgotten that that method is called directly from discovery and not through StagingStore 🤦 

I've simplified it to just the check in `ValidateClickHousePeer` now after @ilidemi 's cleanup in #4643. Added the extra arg there as suggested. The query is run right from that package so this change is now down to just 2 files now.

jgao54 (approved):

ilidemi (commented):
Can you add an integration test please? Like https://github.com/PeerDB-io/peerdb/blob/0863a763f26f303607ae385e4a232ba555d9cdcc/flow/pkg/clickhouse/validation_test.go#L28 but with creating users with various grants and seeing what it returns

itsbilal:
@ilidemi added the test! this is ready for another look

flow/pkg/clickhouse/validation_test.go
```diff
@@ -267,3 +267,74 @@ func TestBuildPartitionByValidationQuery(t *testing.T) {
...
+	database := "pkgch_" + strings.ToLower(common.RandomString(8))
+	require.NoError(t, adminConn.Exec(ctx, "CREATE DATABASE "+QuoteIdentifier(database)))
+	t.Cleanup(func() {
+		cleanupCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+		require.NoError(t, adminConn.Exec(cleanupCtx, "DROP DATABASE IF EXISTS "+QuoteIdentifier(database)))
+	})
+
+	// Create a test user to test grants with.
+	require.NoError(t, adminConn.Exec(ctx, "CREATE USER testuser IDENTIFIED BY 'testpassword';"))
```
ilidemi: Should the username also have a suffix in case the test crashes locally midway, or another test goes for the same name?

flow/pkg/clickhouse/validation_test.go
```diff
@@ -267,3 +267,74 @@ func TestBuildPartitionByValidationQuery(t *testing.T) {
...
+			"clickhouse.cloud",
+			"something.clickhouse.cloud",
+			conn,
+			storageAccessType,
+		)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), fmt.Sprintf("failed to validate %s read grant", storageAccessType))
+
+		// Grant the appropriate privilege, then verify there's no error.
+		require.NoError(t, adminConn.Exec(ctx, fmt.Sprintf("GRANT %s ON *.* TO testuser;", storageAccessType)))
```
ilidemi: Nit: have cases for both `S3` and `READ ON S3`?

ilidemi (approved):

flow/pkg/clickhouse/validation_test.go
itsbilal: Added

PR #4632, opened 2026-07-28
feat(mysql): carry MySQL column DEFAULT through to ClickHouse on ADD COLUMN

dtunikov:
`ALTER TABLE ... ADD COLUMN c INT DEFAULT 5` is DDL, not DML: MySQL makes existing rows read back as `5` without emitting a single row event for them. CDC therefore has nothing to replicate, and rows untouched since the ALTER show ClickHouse's type default (`0`/`''`/`NULL`) indefinitely — until a full table resync.

Emitting the `DEFAULT` on the destination `ADD COLUMN` fixes exactly that set: ClickHouse evaluates a column default at read time for parts written before the column existed, which is by construction the rows whose last change predates the ALTER. Rows changed after it carry real binlog values and win on version dedup, so the two mechanisms don't overlap.

## Changes

- `FieldDescription.default_expr` (`optional string`) carries a rendered SQL literal; unset means no default, or one we declined to translate.
- MySQL DDL parsing renders **constant** defaults only, using standard SQL quote doubling so the literal stays dialect neutral. Signed literals arrive as a unary-minus wrapper and decimals as `MyDecimal`, both handled.
- ClickHouse appends ` DEFAULT <literal>` to `ADD COLUMN`, for shard and distributed tables.

Declined (unset, no DDL change): `NULL` (nullable columns already read back as `NULL`), `CURRENT_TIMESTAMP`/`NOW()`/`UUID()` and other expressions, bit/hex literals, and strings needing escapes beyond quote doubling. Expression defaults are deliberately not translated — read-time evaluation would give a different value per query, and MySQL gives each pre-existing row a distinct value for things like `UUID()`.

If ClickHouse rejects a literal we translated, the `ADD COLUMN` is retried without it and a warning is logged. Losing a default only forfeits the historical backfill; stalling schema replay would stop the table.

## Scope

MySQL source, ClickHouse destination, CDC DDL path only. Not covered: snapshot/resync `CREATE TABLE` (so a resync drops the default from the DDL — data stays correct since it re-copies real values), the gh-ost path (binlog row metadata carries no defaults), Postgres sources, other destinations.

## Tests

`TestAlterTableAddColumnDefault` covers 27 rendering cases including the declined ones. Two e2e tests: `Test_MySQL_AlterTableAddColumnDefault` inserts a row before the ALTER so a plain source↔destination comparison only passes if the default carried across (29 type/default combinations), and `Test_MySQL_AlterTableAddColumnDefaultUntranslated` asserts declined and rejected defaults don't stall replication.

### DIFF
--- a/flow/connectors/clickhouse/cdc.go
+++ b/flow/connectors/clickhouse/cdc.go
@@ -274,9 +274,19 @@ func (c *ClickHouseConnector) ReplayTableSchemaDeltas(
 				return fmt.Errorf("failed to convert column type %s to ClickHouse type: %w", addedColumn.Type, err)
 			}
 
+			defaultExpr := addedColumn.DefaultExpr
+			if defaultExpr != nil && (qvKind == types.QValueKindTime || qvKind == types.QValueKindTimeTZ) &&
+				!slices.Contains(flags, shared.Flag_ClickHouseTime64Enabled) {
+				// on the legacy path time lands in DateTime64 as an offset from the epoch,
+				// which the source's 'HH:MM:SS' literal does not describe
+				c.logger.Warn("[schema delta replay] omitting source default for time column without Time64 support",
+					slog.String("column", addedColumn.Name), slog.String("default", *defaultExpr))
+				defaultExpr = nil
+			}
+
 			columnDef := clickHouseColType
-			if addedColumn.DefaultExpr != nil {
-				columnDef += " DEFAULT " + *addedColumn.DefaultExpr
+			if defaultExpr != nil {
+				columnDef += " DEFAULT " + *defaultExpr
 			}
 
 			addColumn := func(tableName, def string) error {
@@ -289,11 +299,11 @@ func (c *ClickHouseConnector) ReplayTableSchemaDeltas(
 			// retry without default if ch rejected the expression
 			addColumnWithDefaultFallback := func(tableName string) error {
 				err := addColumn(tableName, columnDef)
-				if err == nil || addedColumn.DefaultExpr == nil {
+				if err == nil || defaultExpr == nil {
 					return err
 				}
 				c.logger.Warn("[schema delta replay] retrying added column without its source default",
-					slog.String("column", addedColumn.Name), slog.String("default", *addedColumn.DefaultExpr),
+					slog.String("column", addedColumn.Name), slog.String("default", *defaultExpr),
 					slog.String("destination table name", tableName), slog.Any("error", err))
 				return addColumn(tableName, clickHouseColType)
 			}
--- a/flow/connectors/clickhouse/cdc.go
+++ b/flow/connectors/clickhouse/cdc.go
@@ -274,26 +274,42 @@ func (c *ClickHouseConnector) ReplayTableSchemaDeltas(
 				return fmt.Errorf("failed to convert column type %s to ClickHouse type: %w", addedColumn.Type, err)
 			}
 
-			if shardTableName != "" {
-				if err := c.execWithLogging(ctx,
+			columnDef := clickHouseColType
+			if addedColumn.DefaultExpr != nil {
+				columnDef += " DEFAULT " + *addedColumn.DefaultExpr
+			}
+
+			addColumn := func(tableName, def string) error {
+				return c.execWithLogging(ctx,
 					fmt.Sprintf("ALTER TABLE %s%s ADD COLUMN IF NOT EXISTS %s %s",
-						peerdb_clickhouse.QuoteIdentifier(shardTableName), onCluster,
-						peerdb_clickhouse.QuoteIdentifier(addedColumn.Name), clickHouseColType),
-				); err != nil {
+						peerdb_clickhouse.QuoteIdentifier(tableName), onCluster,
+						peerdb_clickhouse.QuoteIdentifier(addedColumn.Name), def))
+			}
+
+			// retry without default if ch rejected the expression
+			addColumnWithDefaultFallback := func(tableName string) error {
+				err := addColumn(tableName, columnDef)
+				if err == nil || addedColumn.DefaultExpr == nil {
+					return err
+				}
+				c.logger.Warn("[schema delta replay] retrying added column without its source default",
+					slog.String("column", addedColumn.Name), slog.String("default", *addedColumn.DefaultExpr),
+					slog.String("destination table name", tableName), slog.Any("error", err))
+				return addColumn(tableName, clickHouseColType)
+			}
+
+			if shardTableName != "" {
+				if err := addColumnWithDefaultFallback(shardTableName); err != nil {
 					return fmt.Errorf("failed to add column %s for table shards %s: %w", addedColumn.Name, schemaDelta.DstTableName, err)
 				}
 			}
 
-			if err := c.execWithLogging(ctx,
-				fmt.Sprintf("ALTER TABLE %s%s ADD COLUMN IF NOT EXISTS %s %s",
-					peerdb_clickhouse.QuoteIdentifier(schemaDelta.DstTableName), onCluster,
-					peerdb_clickhouse.QuoteIdentifier(addedColumn.Name), clickHouseColType),
-			); err != nil {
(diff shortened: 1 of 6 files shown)


flow/connectors/clickhouse/cdc.go
```diff
@@ -274,26 +274,42 @@ func (c *ClickHouseConnector) ReplayTableSchemaDeltas(
...
-						peerdb_clickhouse.QuoteIdentifier(shardTableName), onCluster,
-						peerdb_clickhouse.QuoteIdentifier(addedColumn.Name), clickHouseColType),
-				); err != nil {
+						peerdb_clickhouse.QuoteIdentifier(tableName), onCluster,
+						peerdb_clickhouse.QuoteIdentifier(addedColumn.Name), def))
+			}
+
+			// retry without default if ch rejected the expression
+			addColumnWithDefaultFallback := func(tableName string) error {
+				err := addColumn(tableName, columnDef)
```
dtunikov, line 301: try to add with full col def (including default)
then retry with simple clickHouseColType if it's failed

if we want to, we can also add a metric + alert when this happens
or just fail with an error

dtunikov:
### Which defaults are carried through

| MySQL column | Example `DEFAULT` | Emitted on ClickHouse |
| --- | --- | --- |
| `TINYINT`…`BIGINT`, signed and unsigned | `-128`, `18446744073709551615` | `-128`, `18446744073709551615` |
| `TINYINT(1)` / `BOOL` | `TRUE` | `1` |
| `YEAR` | `2021` | `2021` |
| `FLOAT`, `DOUBLE` | `-3.14`, `1e5` | `-3.14`, `100000` |
| `DECIMAL(p,s)` | `1.50` | `1.50` (scale as written) |
| `DATE`, `DATETIME(n)`, `TIME` | `'2020-01-02 03:04:05.678'` | `'2020-01-02 03:04:05.678'` |
| `CHAR`, `VARCHAR`, any charset | `'it''s'` | `'it''s'` |
| `ENUM`, `SET` | `'b'`, `'x,z'` | `'b'`, `'x,z'` — needs binlog row metadata, see below |
| `BINARY`, `VARBINARY` | `'abcd'` | `'abcd'` |

Literals are rendered with SQL quote doubling rather than backslash escapes, so the value stays valid in any destination dialect.

### Which are declined

The column is still added, just without a `DEFAULT` — identical to the behaviour before this PR.

| Case | Example | Why |
| --- | --- | --- |
| Expression defaults | `CURRENT_TIMESTAMP`, `NOW()`, `(UUID())` | These reach us as function calls. ClickHouse evaluates a column default when reading parts written before the column existed, so passing the expression through would make pre-`ALTER` rows return a different value on every query instead of the one constant MySQL froze at `ALTER` time. |
| `DEFAULT NULL` | `INT DEFAULT NULL` | Nothing to fill; the column already reads back `NULL`. |
| Bit literals | `BIT(8) DEFAULT b'101'` | Arrives as a `BinaryLiteral` with no dialect-neutral rendering. |
| Strings needing escapes | `'a\tb'`, `'a\\b'` | Declined rather than rendered ambiguously, since only quote doubling is applied. |
| `ENUM`/`SET` on MySQL < 8.0.1 | `ENUM('a','b') DEFAULT 'b'` | Without binlog row metadata the member name can't be resolved from the binlog, so the destination column is `UInt16`/`UInt64` holding the ordinal or bitmask. The label MySQL gives as the default doesn't describe a value that column can hold. |

`TEXT`, `BLOB` and `JSON` are absent because MySQL rejects a plain `DEFAULT` on them outright.

If a literal is rendered but ClickHouse rejects it anyway, `ADD COLUMN` is retried without the `DEFAULT` so schema replay for the table can't stall.

flow/connectors/mysql/cdc.go
```diff
@@ -1175,10 +1225,19 @@ func fieldDescriptionFromMysqlColumn(
...
 	nullable := true
+	var defaultExpr *string
 	for _, option := range col.Options {
-		if option.Tp == ast.ColumnOptionNotNull {
+		switch option.Tp {
+		case ast.ColumnOptionNotNull:
 			nullable = false
-			break
+		case ast.ColumnOptionDefaultValue:
+			// Servers without binlog row metadata carry enums and sets as their ordinal or
```
dtunikov, line 1241: that is unfortunate
but not much we can do here
on the qrep path we handled it using enum as unsigned conversion
but here can't avoid getting a plain string

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -1496,6 +1496,176 @@ func (s ClickHouseSuite) Test_MySQL_AlterTableAddColumnTypes() {
 	RequireEnvCanceled(s.t, env)
 }
 
+type mysqlAddColumnDefaultCase struct{ name, def, val string }
+
+func mysqlAddColumnDefaultTestCases(rowMetadata bool) []mysqlAddColumnDefaultCase {
```
ilidemi: This is single use, can be inlined

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -1496,6 +1496,176 @@ func (s ClickHouseSuite) Test_MySQL_AlterTableAddColumnTypes() {
 	RequireEnvCanceled(s.t, env)
 }
 
+type mysqlAddColumnDefaultCase struct{ name, def, val string }
```
ilidemi: Test defs are trivial and can be duplicated inline

flow/e2e/clickhouse_mysql_test.go
```diff
@@ -1496,6 +1496,176 @@ func (s ClickHouseSuite) Test_MySQL_AlterTableAddColumnTypes() {
...
+}
+
+// Test_MySQL_AlterTableAddColumnDefault covers rows that existed before an ADD COLUMN with a
+// DEFAULT.
+func (s ClickHouseSuite) Test_MySQL_AlterTableAddColumnDefault() {
+	mysource, ok := s.source.(*MySqlSource)
+	if !ok {
+		s.t.Skip("only applies to mysql")
+	}
+	cmp, err := mysource.CompareServerVersion(s.t.Context(), mysql_validation.MySQLMinVersionForBinlogRowMetadata)
```
ilidemi:
```suggestion
	supportsBinlogRowMetadata, err := mysource.CompareServerVersion(s.t.Context(), mysql_validation.MySQLMinVersionForBinlogRowMetadata)
```
nit, as the result is used many lines down

flow/connectors/clickhouse/cdc.go
ilidemi: Metric and a non-paging alert sounds like a good idea. Could also be a task to check these logs in a month, but alert is a more reliable way to have it happen and we want our defaults to be good

flow/connectors/mysql/cdc.go
```diff
@@ -1162,6 +1165,53 @@ func checkTableMapForCompressedColumns(ev *replication.TableMapEvent) error {
...
+		literal = strconv.FormatFloat(v, 'g', -1, 64)
+	case *tidbtypes.MyDecimal:
+		literal = v.String()
+	case string:
+		if negate || strings.ContainsAny(v, "\\\t\n\r\x00") {
+			return "", false
+		}
+		return "'" + strings.ReplaceAll(v, "'", "''") + "'", true
+	default:
+		// DEFAULT NULL is a nil datum, bit literals are types.BinaryLiteral
```
ilidemi: Nit: could make exhaustive and warn/alert if something changes

ilidemi (approved):
Not a blocker for this PR, but is there a plan for ensuring we don't skip over DDL?

@fable Please look through the logic of mysql cdc.go and determine all the cases where a DDL with a column add might not get propagated as a schema delta. Ignore the ones where we fail to parse the query and emit a metric, but past that everything that belong to our tables should result in a delta applied on the destination

ilidemi:
Fable is ignoring us but there's another option https://claude.ai/code/session_01Sf6nJJgyfu4T8z1KxSmNar

PR #4635, opened 2026-07-28
feat(postgres): handle postgres DEFAULT in schema changes

dtunikov:
Stacked on #4632, which added `FieldDescription.default_expr` and the ClickHouse-side replay. This fills the field for the Postgres source so rows that predate an `ADD COLUMN ... DEFAULT` read back the same value on both sides instead of ClickHouse's type zero.

## How

Relation messages carry neither nullability nor the value Postgres returns for rows that physically predate an added column. Both now come from one `pg_catalog.pg_attribute` lookup over the added columns, replacing the nullable-only query and its replica-identity heuristic:

- `attnotnull` is authoritative for column nullability
- `atthasmissing` / `attmissingval` contain the value Postgres substitutes when the column is absent from an older tuple

Volatile or generated defaults cause Postgres to rewrite physical rows and do not have one `attmissingval` that can be represented by a destination column default. Those values remain untranslated. Defaults are only translated for the Q type system; PG-type-system destinations splice the source type in verbatim.


flow/connectors/postgres/cdc.go
```diff
@@ -1396,36 +1420,6 @@ func processRelationMessage[Items model.Items](
 				schemaDelta.SrcTableName))
 		}
 	}
-	if len(potentiallyNullableAddedColumns) > 0 {
```
dtunikov, line 1399: this part is replaced with the code above (Nullable:       !catalogInfo.notNull)

flow/connectors/postgres/cdc.go
```diff
@@ -1437,6 +1431,123 @@ func processRelationMessage[Items model.Items](
...
+	// PostgreSQL uses this value when the column is physically absent from a row that predates
+	// ADD COLUMN. nil means either there is no missing value or that value is SQL NULL.
+	missingValue *string
+	hasMissing   bool
+	notNull      bool
+}
+
+// fetchAddedColumnCatalogInfo reads nullability and the value PostgreSQL uses for rows that predate
+// the named columns. pg_attrdef is deliberately not used: it describes the current default for new
+// rows, which can differ from the value stored in pg_attribute.attmissingval for existing rows.
```
ilidemi, line 1444: Add/change a comment to mention that attmissingval is the original one, it wasn't clear from the name or looking at the query.

flow/connectors/postgres/cdc.go
```diff
@@ -1437,6 +1431,123 @@ func processRelationMessage[Items model.Items](
...
+		return quoteDefaultLiteral(normalizePostgresMissingTimestamp(value))
+
+	case types.QValueKindString, types.QValueKindEnum, types.QValueKindQChar, types.QValueKindUUID,
+		types.QValueKindJSON, types.QValueKindJSONB, types.QValueKindHStore, types.QValueKindINET,
+		types.QValueKindCIDR, types.QValueKindMacaddr, types.QValueKindDate:
+		return quoteDefaultLiteral(value)
+
+	default:
+		// bytea, arrays, interval, time and the geo types all have a text form the destination reads
+		// differently, if at all
```
ilidemi, line 1540: Is the plan to document this near type support?

flow/connectors/postgres/cdc.go
```diff
@@ -1437,6 +1431,123 @@ func processRelationMessage[Items model.Items](
...
+
+func normalizePostgresMissingTimestamp(value string) string {
+	if len(value) > len("2006-01-02") && value[len("2006-01-02")] == 'T' {
+		return value[:len("2006-01-02")] + " " + value[len("2006-01-02")+1:]
+	}
+	return value
+}
+
+// quoteDefaultLiteral quotes value for splicing into DDL, declining anything whose escaping is not
+// portable across dialects.
```
ilidemi: Curious, what are dialects here? It seems the cost of handling those is having to document that support depends on the value, and it would be great to not go there

ilidemi (commented):
Looks good overall, only a few comments

flow/connectors/postgres/cdc.go
dtunikov: oh, yeah, that was about escaping some characters differently on pg and ch sides
but i looked into this a bit more and implemented proper escaping that works for both backslashes and '

flow/connectors/postgres/cdc.go
dtunikov: yeah, gonna open docs PR as well

ilidemi (approved):

flow/connectors/postgres/cdc.go
ilidemi: Nice

PR #4639, opened 2026-07-29
Use integration tests in pkg/clickhouse, fix CI for pkg

ilidemi:
We're in the integrations business, so we should be using real DBs instead of mocks as much as we can. When pkg was first added, we didn't have a fleshed out approach on integration testing and local infra, now there's tilt, .env and all the test helpers.


flow/pkg/testutil/env.go
```diff
@@ -79,3 +80,23 @@ func loadEnvOnce() {
 
 	slog.WarnContext(ctx, "LoadEnv: no .env file found above flow directory")
 }
+
+func ClickHouseTestHost() string {
```
pfcoperez, line 84: Not a blocker but now we are reading different connection strings from envvars at yet another package. I think, in the future, we might want to give some thought on how to group together the collection of test relevant env vars.

pfcoperez (approved):
e2e testing for flow/pkg, nice! 

I left a generic comment related to the spread of test env vars listing across the source code. Not blocking.

pfcoperez:
@fable review this PR

flow/pkg/testutil/env.go
ilidemi: Agreed. Could be a split between internal and pkg, or dump everything into pkg. Idk if a new package outside pkg could work, given that pkg is imported externally and Go doesn't have test-only dependencies

pfcoperez:
> `flow/pkg` is a nested Go module (own `go.mod`), and the only Go test invocation in CI is `gotestsum ... ./...` with `working-directory: ./flow` (`.github/workflows/tilt-flow.yml:565`) — `go test ./...` does not descend into nested modules. I confirmed against this PR's passing "Flow build and test" job log: `flow/connectors/clickhouse` tests ran, but no `flow/pkg/...` test package appears anywhere in the output, and `TestCheckIfTablesEmptyAndEngine` is absent.

@ilidemi This is a good point!

ilidemi:
Eagle eye Fable, fixing

flow/pkg/common/tls.go
```diff
@@ -8,28 +8,23 @@ import (
 	"net"
 )
 
-// modified from https://github.com/golang/go/blob/master/src/crypto/tls/example_test.go
+// Modified from https://github.com/golang/go/blob/master/src/crypto/tls/example_test.go.
 // https://github.com/PeerDB-io/peerdb/pull/2805
-func verifyPeerCertificateWithoutHostname(rootCAs *x509.CertPool) func(certificates [][]byte, _ [][]*x509.Certificate) error {
-	return func(certificates [][]byte, _ [][]*x509.Certificate) error {
+func verifyConnectionWithoutHostname(rootCAs *x509.CertPool) func(tls.ConnectionState) error {
```
ilidemi, line 13: Not sure why that changed, looking

ilidemi: gosec G123 — TLS resumption may bypass VerifyPeerCertificate when VerifyConnection is unset
VerifyPeerCertificate is not called on TCP resumption. We're not using resumption so should be no effect, so just making the linter calm down.
https://github.com/golang/go/issues/77377

flow/pkg/common/tls.go
```diff
@@ -8,28 +8,23 @@ import (
...
+	return func(state tls.ConnectionState) error {
+		if len(state.PeerCertificates) == 0 {
+			return errors.New("tls: server provided no certificates")
+		}
```
ilidemi, line 17: This would be a panic before, an error seems more appropriate

.github/workflows/tilt-flow.yml
pfcoperez, line 635: We need to ingest the rest results for the flow/pkg run , they are on a different file.

I think a clean way of doing this is just repeating an invocation to the action, as this step passing the test output file path for the `test-results-file` action input parameter: 

https://github.com/PeerDB-io/peerdb/blob/bb14273e8490c79e7a861f90a3f7c94c590a4826/.github/actions/ingest-test-results/action.yml#L14-L17

pfcoperez (changes requested):

.github/workflows/tilt-flow.yml
ilidemi: Right, I confused it with the codecov upload that we removed. Fixed in [07edf5b](https://github.com/PeerDB-io/peerdb/pull/4639/commits/07edf5b400e7f8b9c0dceca4fe2490dfcadf0a1a)

ilidemi:
Claude picked up that the test job is working already ^
Lint job confirmation: https://github.com/PeerDB-io/peerdb/actions/runs/30432492473/job/90512674461

Amogh-Bharadwaj (approved):

PR #4643, opened 2026-07-30
Refactor staging store validation

ilidemi:
Consumers of pkg don't need to actually access the staging store, so the whole machinery of passing closures can be replaced with simple method calls. On the next pkg upgrade the consumer will just need to drop the dummy validator arg.


flow/connectors/clickhouse/staging_validate_test.go
```diff
@@ -9,26 +9,26 @@ import (
...
 	"github.com/stretchr/testify/require"
 	"google.golang.org/api/option"
+
+	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
 )
 
 // S3 Tests:
 
-func newFakeS3Client(t *testing.T, server *httptest.Server) *s3.Client {
-	t.Helper()
```
ilidemi, line 21: Was unused as there are no assertions in the func

jgao54 (approved):
makes sense to me!

PR #4644, opened 2026-07-30
feat: add temporal cli in flow-worker image

masterashu:


ilidemi:
Any context on why? For manual operations we have CLI access and worker itself has the full client

PR #4662, opened 2026-08-05
Add native CockroachDB connector support

viragtripathi:
Adds CockroachDB as a source peer. Continues #3861, moved in repo so CI runs without the external contributor gate.

Split into five stacked commits, each builds on its own:
1. Peer connector: pgx based, TLS, SSH tunnel, validation, version gating, variant detection
2. Schema introspection and query replication with AS OF SYSTEM TIME partitioned reads
3. CDC via sinkless changefeeds: resolved timestamp checkpoints, cursor resume with backoff, snapshot handoff at a captured cluster logical timestamp, PTS protection during initial load using the same mechanism as MOLT, delete key fallback, replay dedup, SQLSTATE based error classification, gc.ttlseconds validation at mirror creation, resolved lag and record metrics
4. UI for peer creation including a changefeeds toggle
5. Unit tests, e2e suite, CockroachDB in the CI matrix, Tilt and dev tooling

Tested against CockroachDB v25.4.13 (e2e suite, plus a full CDC mirror to ClickHouse on a five node cluster with node kills mid stream, destination matched source on counts and checksums) and a CockroachDB Cloud v26.2.1 cluster over verify-full TLS.

Known limitations: schema changes beyond added columns require a resync, sinkless changefeeds only for now (webhook and Kafka sink transports are planned follow ups), cursor expiry beyond gc.ttlseconds requires a resync.


viragtripathi:
Note for reviewers on shared code: this PR's types test is the first to compare a Nullable(UUID) column, which exposed a missing case in the shared e2e ClickHouse helper; added the six line handler following the existing pattern (flow/e2e/clickhouse.go). The otel_manager.go diff looks large but is two new metric registrations plus gofmt realignment of the struct. Everything else outside the connector package is standard connector registration wiring. Happy to split the helper change into its own PR if preferred.

viragtripathi:
Restacked as GitHub stacked PRs for easier review per reviewer request: #4667 #4668 #4669 #4670 #4671 (stack merges bottom up, top merges everything). Identical commits to the green run here.

PR #4667, opened 2026-08-05
crdb stack/1 peer connector

viragtripathi:

### DIFF
--- a/flow/connectors/cockroachdb/client.go
+++ b/flow/connectors/cockroachdb/client.go
@@ -1,0 +1,111 @@
+package conncockroachdb
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net/url"
+
+	"github.com/jackc/pgx/v5"
+
+	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/pkg/common"
+	"github.com/PeerDB-io/peerdb/flow/shared"
+)
+
+func ParseConfig(connectionString string, config *protos.CockroachDBConfig) (*pgx.ConnConfig, error) {
+	connConfig, err := pgx.ParseConfig(connectionString)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse connection string: %w", err)
+	}
+	// the password is set here instead of the connection string so it never
+	// appears in errors or logs
+	connConfig.Password = config.Password
+	// Pin the wire protocol version so a pgx dependency update cannot switch
+	// it silently: CockroachDB only negotiates protocol 3.2 from v26.1 on,
+	// while this connector supports much older versions. Matches the pin in
+	// the postgres connector.
+	connConfig.Config.MaxProtocolVersion = "3.0"
+	if !config.DisableTls {
+		var clientCert *common.ClientCertificate
+		if clientTls := config.GetClientTls(); clientTls != nil {
+			clientCert, err = common.NewClientCertificate(clientTls.GetCertificate(), clientTls.GetPrivateKey())
+			if err != nil {
+				return nil, err
+			}
+		}
+		tlsConfig, err := common.CreateTlsConfig(
+			tls.VersionTLS12, config.RootCa, connConfig.Host, config.TlsHost, config.SkipCertVerification, clientCert)
+		if err != nil {
+			return nil, err
+		}
+		connConfig.TLSConfig = tlsConfig
+	}
+	return connConfig, nil
+}
+
+func NewCockroachDBConnFromConfig(
+	ctx context.Context,
+	connConfig *pgx.ConnConfig,
+	tunnel *utils.SSHTunnel,
+) (*crdbConn, error) {
+	if tunnel.IsActive() {
+		connConfig.DialFunc = tunnel.DialContext
+		// DNS lookup seems to happen before connection is established which can be an issue if given host
+		// can only be resolved on the SSH host https://github.com/jackc/pgx/issues/1724
+		connConfig.LookupFunc = func(ctx context.Context, host string) ([]string, error) {
+			return []string{host}, nil
+		}
+	}
+	conn, err := pgx.ConnectConfig(ctx, connConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to CockroachDB: %w", toCrdbError(err))
+	}
+	return &crdbConn{conn}, nil
+}
+
+func GetCRDBConnectionString(config *protos.CockroachDBConfig, flowName string) string {
+	applicationName := "peerdb"
+	if flowName != "" {
+		applicationName = "peerdb_" + flowName
+	}
+
+	// the password is intentionally left out of the URL (ParseConfig sets it on
+	// the parsed config) so it never appears in errors or logs
+	u := &url.URL{
+		Scheme: "postgres",
+		User:   url.User(config.User),
+		Host:   shared.JoinHostPort(config.Host, config.Port),
+		Path:   "/" + config.Database,
+	}
+	q := u.Query()
+	q.Set("application_name", applicationName)
+	q.Set("client_encoding", "UTF8")
+	if config.DisableTls {
+		q.Set("sslmode", "disable")
+	} else {
+		q.Set("sslmode", "require")
+	}
+	u.RawQuery = q.Encode()
+	return u.String()
+}
+
+func (c *CockroachDBConnector) GetMajorVersion(ctx context.Context) (int, error) {
+	var version string
+	if err := c.conn.QueryRow(ctx, "SELECT version()").Scan(&version); err != nil {
+		return 0, fmt.Errorf("failed to get CockroachDB version: %w", err)
+	}
+	return parseCrdbMajorVersion(version)
+}
+
+// parses version strings like "CockroachDB CCL v25.4.13 ..." or "CockroachDB v23.1.0 ..."
+func parseCrdbMajorVersion(version string) (int, error) {
+	var major, minor, patch int
+	if _, err := fmt.Sscanf(version, "CockroachDB CCL v%d.%d.%d", &major, &minor, &patch); err != nil {
+		if _, err := fmt.Sscanf(version, "CockroachDB v%d.%d.%d", &major, &minor, &patch); err != nil {
+			return 0, fmt.Errorf("failed to parse CockroachDB version %q: %w", version, err)
+		}
+	}
+	return major, nil
+}
--- a/flow/connectors/cockroachdb/cockroachdb.go
+++ b/flow/connectors/cockroachdb/cockroachdb.go
@@ -1,0 +1,141 @@
+package conncockroachdb
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"time"
+
+	"github.com/jackc/pgerrcode"
+	"github.com/jackc/pgx/v5/pgconn"
+	"go.temporal.io/sdk/log"
+
+	metadataStore "github.com/PeerDB-io/peerdb/flow/connectors/external_metadata"
+	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/internal"
+	"github.com/PeerDB-io/peerdb/flow/shared"
+	"github.com/PeerDB-io/peerdb/flow/shared/exceptions"
+)
+
+type CockroachDBConnector struct {
+	*metadataStore.PostgresMetadata
+	logger      log.Logger
+	ssh         *utils.SSHTunnel
+	conn        *crdbConn
+	config      *protos.CockroachDBConfig
+	connStr     string
+	crdbVersion string
+}
+
+func NewCockroachDBConnector(ctx context.Context, env map[string]string, config *protos.CockroachDBConfig) (*CockroachDBConnector, error) {
+	logger := internal.LoggerFromCtx(ctx)
+	flowNameInApplicationName, err := internal.PeerDBApplicationNamePerMirrorName(ctx, nil)
+	if err != nil {
+		logger.Error("Failed to get flow name from application name", slog.Any("error", err))
+	}
+	var flowName string
+	if flowNameInApplicationName {
+		flowName, _ = ctx.Value(shared.FlowNameKey).(string)
+	}
+	connectionString := GetCRDBConnectionString(config, flowName)
+	connConfig, err := ParseConfig(connectionString, config)
+	if err != nil {
+		return nil, err
+	}
+
+	connConfig.Config.RuntimeParams["timezone"] = "UTC"
+	connConfig.Config.RuntimeParams["idle_in_transaction_session_timeout"] = "0"
+	connConfig.Config.RuntimeParams["statement_timeout"] = "0"
+	connConfig.Config.RuntimeParams["DateStyle"] = "ISO, DMY"
+
+	pgMetadata, err := metadataStore.NewPostgresMetadata(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	tunnel, err := utils.NewSSHTunnel(ctx, config.SshConfig)
+	if err != nil {
+		logger.Error("failed to create ssh tunnel", slog.Any("error", err))
+		return nil, fmt.Errorf("failed to create ssh tunnel: %w", err)
+	}
+
+	conn, err := NewCockroachDBConnFromConfig(ctx, connConfig, tunnel)
+	if err != nil {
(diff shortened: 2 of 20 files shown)


viragtripathi:
Bottom of the CockroachDB connector stack (5 PRs: #4667 peer connector, #4668 schema introspection and query replication, #4669 CDC via changefeeds, #4670 UI, #4671 tests and CI). Same commits that were green as a single PR in #4662, restacked for review. Full context, testing evidence and reviewer notes are on #4662.

README.md
```diff
@@ -93,13 +93,14 @@ You can use Postgres’ ecosystem to manage your ETL —
 
 ## Status
 
-We have expanded our connector ecosystem to support multiple source connectors beyond Postgres, including MySQL and MongoDB. You can check the status of connectors [here](https://docs.peerdb.io/sql/commands/supported-connectors).
+We have expanded our connector ecosystem to support multiple source connectors beyond Postgres, including MySQL, MongoDB, and CockroachDB. You can check the status of connectors [here](https://docs.peerdb.io/sql/commands/supported-connectors).
```
jgao54: nit: could we delay the doc update to when it's ready in the UI

protos/peers.proto
```diff
@@ -139,6 +139,25 @@ message PostgresConfig {
...
+  // defaults to _peerdb_internal
+  optional string metadata_schema = 7;
+  optional SSHConfig ssh_config = 8;
+  optional string root_ca = 9 [(peerdb_redacted) = true];
+  bool require_tls = 10;
+  // CockroachDB-specific settings
+  optional uint32 max_retries = 11;
+  optional uint32 retry_base_delay_ms = 12;
+  optional string cluster_name = 13;
+  bool use_changefeeds = 14;
```
jgao54: it's not obvious to me how these crdb-specific settings are used by this connector, since i don't see them being used when setting up the connector.  could you provide more context here?

flow/connectors/cockroachdb/validate.go
```diff
@@ -0,0 +1,55 @@
...
+	if err == nil && settingValueIsTrue(serverless) {
+		return protos.DatabaseVariant_COCKROACHDB_SERVERLESS, nil
+	}
+
+	// managed CockroachDB Cloud clusters always have cluster.organization set
+	if clusterOrg != "" {
+		return protos.DatabaseVariant_COCKROACHDB_CLOUD, nil
+	}
+
+	return protos.DatabaseVariant_VARIANT_UNKNOWN, nil
```
jgao54, line 98: COCKROACHDB_DEDICATED was added but not used anywhere, can it be removed?

flow/connectors/cockroachdb/validate.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func (c *CockroachDBConnector) GetDatabaseVariant(ctx context.Context) (protos.DatabaseVariant, error) {
+	// SHOW CLUSTER SETTING works without allow_unsafe_internals, unlike
+	// crdb_internal.cluster_settings
+	var clusterOrg string
+	err := c.conn.QueryRow(ctx, "SHOW CLUSTER SETTING cluster.organization").Scan(&clusterOrg)
+	if err != nil {
+		return protos.DatabaseVariant_VARIANT_UNKNOWN, nil
+	}
```
jgao54: what is the expected behavior for `SHOW CLUSTER SETTTING cluster.organization` on a non-cloud flavored cockroachdb? if it's expected to fail, is it possible to differentiate between that vs. network error? 

same question with `SHOW CLUSTER SETTING server.serverless.enabled` below.

flow/connectors/cockroachdb/cockroachdb.go
```diff
@@ -0,0 +1,141 @@
...
+	return c.conn.Ping(ctx)
+}
+
+func (c *CockroachDBConnector) GetVersion(ctx context.Context) (string, error) {
+	if c.crdbVersion != "" {
+		return c.crdbVersion, nil
+	}
+	var version string
+	err := c.conn.QueryRow(ctx, "SELECT version()").Scan(&version)
+	if err != nil {
```
jgao54: nit:` if err := ...; err != nil { ... }` for codebase convention

flow/connectors/cockroachdb/cockroachdb.go
```diff
@@ -0,0 +1,141 @@
...
+	if pgErr, ok := errors.AsType[*pgconn.PgError](err); ok {
+		switch pgErr.Code {
+		case pgerrcode.InvalidAuthorizationSpecification,
+			pgerrcode.InvalidPassword,
+			pgerrcode.InsufficientPrivilege:
+			return exceptions.NewAuthError(err)
+		}
+	}
+	return err
+}
```
jgao54, line 107: this is me getting ahead a bit, but realizing that this will be the first postgres-compatible connector that shares the same exception code as Postgres, but has its own connector implementation; one new issues this introduces is that all crdb errors will get classified into postgres errors in `classifier.go` during error classification, because it uses the same client.

Instead of having a`*pgx.Conn`, it may be a good idea to introduce a light connection wrapper and a dedicated CockrachDBError to wrap all the error messages coming this source connector. e.g.: 
```
type crdbConn struct {
      conn *pgx.Conn
}

func (c crdbConn) Ping(ctx context.Context) error {
      return toCrdbError(t.conn.Ping(ctx))
}

func (c crdbConn) Exec(ctx context.Context, sql string, args ...any) (pgconn.CommandTag, error) {
      res, err := t.conn.Exec(ctx, sql, args...)
      return res, toCrdbError(err)
}

// wrap more methods here that are used by the connector
```
So the struct becomes:
```
type CockroachDBConnector struct {
    ...
	conn           *crdbConn
	...
}
```
this way the source type can be identified based on error class in error classification.

flow/connectors/cockroachdb/client.go
```diff
@@ -0,0 +1,93 @@
...
+	connConfig, err := pgx.ParseConfig(connectionString)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse connection string: %w", err)
+	}
+	// the password is set here instead of the connection string so it never
+	// appears in errors or logs
+	connConfig.Password = config.Password
+	if config.RequireTls || config.RootCa != nil {
+		tlsConfig, err := common.CreateTlsConfig(
+			tls.VersionTLS12, config.RootCa, connConfig.Host, config.TlsHost, false, nil)
```
jgao54: nit: can add support for `skip_cert_verification` (we support this for other connectors) by adding the field in the proto instead of hard-coding it to false here;  and `client_tls` as well  if mutualTls is supported

flow/connectors/cockroachdb/client.go
```diff
@@ -0,0 +1,93 @@
...
+	return conn, nil
+}
+
+func GetCRDBConnectionString(config *protos.CockroachDBConfig, flowName string) string {
+	applicationName := "peerdb"
+	if flowName != "" {
+		applicationName = "peerdb_" + flowName
+	}
+
+	connString := fmt.Sprintf(
```
jgao54: instead of generating the string by hand could we do something similar to GetPGConnectionString for consistency?

README.md
viragtripathi: Moved the README update out of this PR, it now lands together with the UI PR further up the stack.

protos/peers.proto
viragtripathi: Restructured this so each field is introduced where it's used: this PR's proto now only has the connection fields plus skip_cert_verification and client_tls, while use_changefeeds, max_retries and retry_base_delay_ms move to the CDC PR where the changefeed retry loop reads them. cluster_name turned out to be unused so I removed it entirely and reserved its field number.

flow/connectors/cockroachdb/validate.go
viragtripathi: Verified against a live v25.4.13: cluster.organization exists on every flavor and comes back as an empty string on self hosted, so an error there is never the expected negative, while server.serverless.enabled fails on non serverless clusters with an "unknown setting" error under SQLSTATE XXUUU. I now treat unknown setting as the expected miss and log a warning for anything else like network or privilege failures before returning UNKNOWN, with unit tests on the helper.

flow/connectors/cockroachdb/cockroachdb.go
viragtripathi: Fixed here and swept the rest of the connector. The remaining split assignments are ones where err feeds two branches.

flow/connectors/cockroachdb/client.go
viragtripathi: Added both to the proto and plumbed them through ParseConfig into common.CreateTlsConfig the same way postgres.go does, mutual TLS included. Also wired into the UI form, the nexus WITH option parsing, and covered with unit tests.

flow/connectors/cockroachdb/client.go
viragtripathi: Rebuilt it on url.URL with query Set like GetPGConnectionString. The one deliberate difference is that the password stays out of the URL and is set on the parsed config instead so it can't leak into errors or logs.

flow/connectors/cockroachdb/validate.go
viragtripathi: Removed and reserved enum number 12. GetDatabaseVariant only ever returns CLOUD, SERVERLESS or UNKNOWN, and cloud advanced clusters are just detected as CLOUD via cluster.organization.

flow/connectors/cockroachdb/cockroachdb.go
viragtripathi: Done, essentially as you sketched: crdbConn wraps *pgx.Conn (plus the returned Rows/Row so errors surfaced during iteration get tagged too) and toCrdbError tags everything with a new exceptions.CockroachDBError that unwraps cleanly. classifier.go keeps the Postgres SQLSTATE handling but reports source cockroachdb when the tag is present, the dedicated changefeed and history retention connections go through the same wrapper, and there are unit tests for toCrdbError and the classifier path.

protos/peers.proto
```diff
@@ -139,6 +139,22 @@ message PostgresConfig {
...
+  uint32 port = 2;
+  string user = 3;
+  string password = 4 [(peerdb_redacted) = true];
+  string database = 5;
+  string tls_host = 6;
+  // defaults to _peerdb_internal
+  optional string metadata_schema = 7;
+  optional SSHConfig ssh_config = 8;
+  optional string root_ca = 9 [(peerdb_redacted) = true];
+  bool require_tls = 10;
```
jgao54: i missed this earlier, `requrie_tls` was a legacy name that we've since deprecated and unified to `disable_tls` in https://github.com/PeerDB-io/peerdb/pull/4388, so let's make sure we use `disable_tls` here as well

flow/connectors/cockroachdb/validate.go
jgao54: since PR hasn't been merged, don't need to reserve

flow/connectors/cockroachdb/validate.go
jgao54: let's make sure the network/transient errors (errors that are not "unknown setting") are propagated as error so it an opportunity to retry instead of defaulting to unknown.

flow/connectors/cockroachdb/cockroachdb.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/PeerDB-io/peerdb/flow/shared"
+	"github.com/PeerDB-io/peerdb/flow/shared/exceptions"
+)
+
+type CockroachDBConnector struct {
+	*metadataStore.PostgresMetadata
+	logger         log.Logger
+	ssh            *utils.SSHTunnel
+	conn           *crdbConn
+	Config         *protos.CockroachDBConfig
```
jgao54: missed this earlier: can keep this private as it's not being used outside of this package

flow/connectors/cockroachdb/cockroachdb.go
```diff
@@ -0,0 +1,139 @@
...
+	connConfig.Config.RuntimeParams["idle_in_transaction_session_timeout"] = "0"
+	connConfig.Config.RuntimeParams["statement_timeout"] = "0"
+	connConfig.Config.RuntimeParams["DateStyle"] = "ISO, DMY"
+
+	pgMetadata, err := metadataStore.NewPostgresMetadata(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	tunnel, err := utils.NewSSHTunnel(ctx, config.SshConfig)
```
jgao54, line 58: missing a ssh tunnel keepalive goroutine in case ssh tunnel is used, see `newPostgresConnector` for reference (the fact it requires an explicit call is not great API design, something we've been meaning to improve, but for now let's add it below, before returning the connector)

flow/connectors/cockroachdb/client.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/jackc/pgx/v5"
+
+	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/pkg/common"
+	"github.com/PeerDB-io/peerdb/flow/shared"
+)
+
+func ParseConfig(connectionString string, config *protos.CockroachDBConfig) (*pgx.ConnConfig, error) {
```
jgao54, line 17: should also pin MaxProtocolVersion to 3.0

it looks like cockroachdb only supports 3.2 since [v26.1](https://www.cockroachlabs.com/docs/releases/v26.1#sql-language-changes-7) but connector supports >= v22; so if we bump pgx version in the future and the default protocol version becomes 3.2, it can break existing pipes that are on 22.x–25.x

flow/connectors/cockroachdb/cockroachdb.go
```diff
@@ -0,0 +1,139 @@
...
+			c.logger.Error("failed to close connection", slog.Any("error", err))
+			errs = append(errs, fmt.Errorf("failed to close connection: %w", err))
+		}
+	}
+	if err := c.ssh.Close(); err != nil {
+		c.logger.Error("failed to close SSH tunnel", slog.Any("error", err))
+		errs = append(errs, fmt.Errorf("failed to close SSH tunnel: %w", err))
+	}
+	if len(errs) > 0 {
+		return fmt.Errorf("errors closing CockroachDB connector: %v", errs)
```
jgao54: nit: use  `%w` and `errors.Join` so classifier's error.As still works

jgao54 (commented):
looks good overall, I picked up a few things with claude that I missed earlier and added comment.

flow/connectors/cockroachdb/validate.go
jgao54: also for some of the other connectors we use domain suffix to determine db variant, could also use `.cockroachlabs.cloud` here which doesn't require querying the db; up to you

protos/peers.proto
viragtripathi: Switched to disable_tls with TLS on by default, matching the newer connectors. Since nothing shipped, I also dropped the reserved field numbers and renumbered the message to be contiguous, same for the variant enum. UI, nexus and the provision script follow.

flow/connectors/cockroachdb/cockroachdb.go
viragtripathi: Made it unexported.

flow/connectors/cockroachdb/cockroachdb.go
viragtripathi: Added, same shape as newPostgresConnector: keepalive failure closes the connector's conn with a short timeout.

flow/connectors/cockroachdb/client.go
viragtripathi: Pinned to 3.0 with the same rationale comment as postgres, plus a test. Good catch on 3.2 being v26.1+ only.

flow/connectors/cockroachdb/cockroachdb.go
viragtripathi: Done, Close now returns errors.Join so errors.As sees the wrapped close errors.

flow/connectors/cockroachdb/validate.go
viragtripathi: Took this, it's cleaner. Hosts ending in .cockroachlabs.cloud now classify as Cloud up front without touching the database, and the serverless probe only runs to refine Cloud into serverless. Verified against the Cloud cluster. The query path stays as the fallback for anything not on that domain, like private DNS in front of a Cloud cluster.

protos/peers.proto
viragtripathi: Right, dropped the reserves and renumbered while the PR is still open.

jgao54 (approved):

viragtripathi:
> looks good overall, I picked up a few things with claude that I missed earlier and added comment.

Thank you for taking the time to go through all of this so carefully. Really appreciate it!

PR #4668, opened 2026-08-05
crdb stack/2 schema qrep

viragtripathi:

### DIFF
--- a/flow/connectors/cockroachdb/qrep.go
+++ b/flow/connectors/cockroachdb/qrep.go
@@ -1,0 +1,534 @@
+package conncockroachdb
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/jackc/pgx/v5/pgtype"
+	"go.temporal.io/sdk/temporal"
+
+	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/model"
+	"github.com/PeerDB-io/peerdb/flow/otel_metrics"
+	"github.com/PeerDB-io/peerdb/flow/pkg/common"
+	"github.com/PeerDB-io/peerdb/flow/shared"
+	"github.com/PeerDB-io/peerdb/flow/shared/exceptions"
+	"github.com/PeerDB-io/peerdb/flow/shared/types"
+)
+
+// gcThresholdErrorSubstring appears when CREATE CHANGEFEED is given a cursor
+// older than the tables' gc.ttlseconds window: the MVCC history needed to
+// resume has been garbage collected and retrying can never succeed.
+const gcThresholdErrorSubstring = "must be after replica GC threshold"
+
+func isCursorTooOldError(err error) bool {
+	return err != nil && strings.Contains(err.Error(), gcThresholdErrorSubstring)
+}
+
+// classifySnapshotReadError marks AS OF SYSTEM TIME reads whose timestamp has
+// fallen behind the replica GC threshold as non-retryable for Temporal: the
+// MVCC history at the snapshot timestamp is garbage collected, so retries can
+// never succeed and the mirror needs a resync. This mirrors the postgres
+// invalid-snapshot handling; shared.WrapError re-raises the ApplicationError
+// even when it gets wrapped along the way.
+func classifySnapshotReadError(err error) error {
+	if err != nil && isCursorTooOldError(err) {
+		return temporal.NewNonRetryableApplicationError(
+			"snapshot timestamp is older than the replica GC threshold, resync the mirror",
+			exceptions.ApplicationErrorTypeIrrecoverableInvalidSnapshot.String(), err)
+	}
+	return err
+}
+
+func (c *CockroachDBConnector) GetQRepPartitions(
+	ctx context.Context,
+	config *protos.QRepConfig,
+	last *protos.QRepPartition,
+) ([]*protos.QRepPartition, error) {
+	if config.WatermarkColumn == "" || config.NumPartitionsOverride == 1 {
+		// if no watermark column is specified, return a single partition
+		return utils.FullTablePartition(), nil
+	}
+
+	if config.NumPartitionsOverride == 0 && config.NumRowsPerPartition == 0 {
+		return nil, errors.New("num rows per partition must be greater than 0")
+	}
+
+	numPartitions := int64(config.NumPartitionsOverride)
+	numRowsPerPartition := int64(config.NumRowsPerPartition)
+
+	parsedWatermarkTable, err := common.ParseTableIdentifier(config.WatermarkTable)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse watermark table %s: %w", config.WatermarkTable, err)
+	}
+
+	systemTime, err := c.snapshotSystemTime(ctx, config)
+	if err != nil {
+		return nil, err
+	}
+
+	var lastRangeEnd any
+	resuming := last != nil && last.Range != nil
+	if resuming {
+		switch lastRange := last.Range.Range.(type) {
+		case *protos.PartitionRange_IntRange:
+			lastRangeEnd = lastRange.IntRange.End
+		case *protos.PartitionRange_UintRange:
+			lastRangeEnd = lastRange.UintRange.End
+		case *protos.PartitionRange_TimestampRange:
+			lastRangeEnd = lastRange.TimestampRange.End.AsTime()
+		default:
+			return nil, fmt.Errorf("unsupported range type %T in last partition after resuming QRep", lastRange)
+		}
+	}
+
+	// numPartitions was not overridden, derive it from the row count
+	if numPartitions == 0 {
+		var totalRows int64
+		if resuming {
+			// resuming mid load is only reachable through the standalone QRep
+			// flow, a deprecated feature, so it is kept in its own branch to be
+			// easy to remove; the rows past lastRangeEnd cannot be estimated
+			// from table statistics, count them exactly instead. Both this
+			// count and the min/max below read AS OF the same system time, so
+			// they stay consistent without sharing a scan.
+			countQuery := fmt.Sprintf("SELECT COUNT(*) FROM %s AS OF SYSTEM TIME '%s' WHERE %s > $1",
+				parsedWatermarkTable.String(), systemTime, common.QuoteIdentifier(config.WatermarkColumn))
+			if err := c.conn.QueryRow(ctx, countQuery, lastRangeEnd).Scan(&totalRows); err != nil {
+				return nil, classifySnapshotReadError(fmt.Errorf("failed to query for remaining rows: %w", err))
+			}
+		} else {
+			// starting from the beginning: use the statistics based estimate,
+			// fall back to a precise count when it is unusable
+			estimate, err := c.tableRowEstimate(ctx, parsedWatermarkTable)
+			if err != nil || estimate <= 0 {
+				// a NULL estimate scans as 0 without an error, so this branch covers
+				// both a failed estimate and a genuinely empty (or unanalyzed) table
+				c.logger.Warn("[cockroachdb] failed to estimate row count or table is empty, falling back to precise count",
+					slog.Any("error", err))
+				countQuery := fmt.Sprintf("SELECT COUNT(*) FROM %s AS OF SYSTEM TIME '%s'", parsedWatermarkTable.String(), systemTime)
+				if err := c.conn.QueryRow(ctx, countQuery).Scan(&estimate); err != nil {
+					return nil, classifySnapshotReadError(fmt.Errorf("failed to query for total rows: %w", err))
+				}
+			}
+			totalRows = estimate
+		}
+
+		if totalRows == 0 {
+			c.logger.Warn("no records to replicate, only using 1 partition")
+			numPartitions = 1
+		} else {
+			adjustedPartitions := shared.AdjustNumPartitions(totalRows, numRowsPerPartition)
+			c.logger.Info("[cockroachdb] partition details",
+				slog.Int64("totalRows", totalRows),
+				slog.Int64("desiredNumRowsPerPartition", numRowsPerPartition),
+				slog.Int64("adjustedNumPartitions", adjustedPartitions.AdjustedNumPartitions),
+				slog.Int64("adjustedNumRowsPerPartition", adjustedPartitions.AdjustedNumRowsPerPartition))
+			numPartitions = adjustedPartitions.AdjustedNumPartitions
+		}
+	}
+
+	minmaxQuery := buildMinMaxQuery(parsedWatermarkTable, config.WatermarkColumn, systemTime, resuming)
+	var queryArgs []any
+	if resuming {
+		queryArgs = []any{lastRangeEnd}
+	}
+	c.logger.Info("querying min/max", slog.String("query", minmaxQuery))
+
+	var minVal, maxVal any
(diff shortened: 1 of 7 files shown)


flow/connectors/cockroachdb/qrep.go
```diff
@@ -0,0 +1,552 @@
...
+	"github.com/PeerDB-io/peerdb/flow/model"
+	"github.com/PeerDB-io/peerdb/flow/otel_metrics"
+	"github.com/PeerDB-io/peerdb/flow/pkg/common"
+	"github.com/PeerDB-io/peerdb/flow/shared"
+	"github.com/PeerDB-io/peerdb/flow/shared/exceptions"
+	"github.com/PeerDB-io/peerdb/flow/shared/types"
+)
+
+// crdbSystemTimeRegex matches a CockroachDB HLC timestamp, e.g. 1712345678901234567.0000000001
+var crdbSystemTimeRegex = regexp.MustCompile(`^\d+(\.\d+)?$`)
```
Jeremyyang920: Is this regex still needed now that the thrid PR is using the HLC parsing logic?

viragtripathi: It turned into the format gate inside parseHLC, which checks the shape before splitting into the typed struct, so there's one validation point shared by everything. In this PR it's used directly because the typed parser only lands in the CDC one, by the top of the stack qrep goes through parseHLC too and the direct use is gone. The definition lives here since both layers need it.

flow/connectors/cockroachdb/qrep.go
```diff
@@ -0,0 +1,552 @@
...
+		default:
+			return nil, fmt.Errorf("unsupported range type %T in last partition after resuming QRep", lastRange)
+		}
+	}
+
+	if numPartitions == 0 && lastRangeEnd == nil {
+		// starting replication from the beginning: estimate row count to calculate partitions
+		totalRows, err := c.tableRowEstimate(ctx, parsedWatermarkTable)
+		if err != nil {
+			c.logger.Warn("[cockroachdb] failed to estimate row count, falling back to precise count", slog.Any("error", err))
```
itsbilal: nb: if the estimated_row_count was null, we'd not get an error here so even though we fell to the precise count, this warning would not be logged. Let's reword it a little so it applies to the true zero case too (eg. `failed to estimate row count or table is empty, falling back to precise count`), and move it to the `if totalRows <= 0` case below.

flow/connectors/cockroachdb/qrep.go
```diff
@@ -0,0 +1,552 @@
...
+		c.logger.Info("[cockroachdb] partition details",
+			slog.Int64("totalRows", totalRows),
+			slog.Int64("desiredNumRowsPerPartition", numRowsPerPartition),
+			slog.Int64("adjustedNumPartitions", adjustedPartitions.AdjustedNumPartitions),
+			slog.Int64("adjustedNumRowsPerPartition", adjustedPartitions.AdjustedNumRowsPerPartition))
+		numPartitions = adjustedPartitions.AdjustedNumPartitions
+	}
+
+	switch minVal.(type) {
+	case nil, int16, int32, int64, time.Time:
```
itsbilal: The unsigned types should be supported too, right?

flow/connectors/cockroachdb/qrep.go
```diff
@@ -0,0 +1,552 @@
...
+
+	c.logger.Info("[cockroachdb] pulled records",
+		partitionIdLog,
+		slog.Int64("records", totalRecords),
+		slog.Int64("bytes", totalBytes),
+		slog.Int("channelLen", len(stream.Records)))
+	return totalRecords, totalBytes, nil
+}
+
+func (c *CockroachDBConnector) clusterLogicalTimestamp(ctx context.Context) (string, error) {
```
itsbilal: Let's use the HLC parser here too - so we don't return just a string

flow/connectors/cockroachdb/qrep.go
```diff
@@ -0,0 +1,552 @@
...
+		return "", fmt.Errorf("failed to get cluster logical timestamp: %w", err)
+	}
+	return systemTime, nil
+}
+
+// snapshotSystemTime returns the HLC timestamp QRep queries read at via AS OF SYSTEM TIME,
+// pinning reads in the recent past so long scans neither block nor get pushed by writers.
+// config.SnapshotName carries an exported system time when set, otherwise the current
+// cluster timestamp is captured.
+func (c *CockroachDBConnector) snapshotSystemTime(ctx context.Context, config *protos.QRepConfig) (string, error) {
```
itsbilal: Same here on the return type

flow/connectors/cockroachdb/qvalue_convert.go
```diff
@@ -0,0 +1,590 @@
...
+	"github.com/PeerDB-io/peerdb/flow/shared/datatypes"
+	"github.com/PeerDB-io/peerdb/flow/shared/types"
+)
+
+// crdbTypeToQValueKind maps CockroachDB information_schema column types to QValueKinds.
+// data_type is information_schema.columns.data_type; udt_name disambiguates
+// arrays (data_type 'ARRAY') and user-defined types (data_type 'USER-DEFINED').
+func crdbTypeToQValueKind(dataType string, udtName string) types.QValueKind {
+	switch strings.ToLower(dataType) {
+	case "bigint", "int8", "oid":
```
itsbilal: `oid` should be QValueKindUInt32

flow/connectors/cockroachdb/qvalue_convert.go
```diff
@@ -0,0 +1,590 @@
...
+func crdbOIDToQValueKind(oid uint32) types.QValueKind {
+	switch oid {
+	case pgtype.BoolOID:
+		return types.QValueKindBoolean
+	case pgtype.Int2OID:
+		return types.QValueKindInt16
+	case pgtype.Int4OID:
+		return types.QValueKindInt32
+	case pgtype.Int8OID, pgtype.OIDOID:
+		return types.QValueKindInt64
```
itsbilal, line 91: Same here

itsbilal (approved):
Thanks for doing this! Looks mostly good to me, save for the comments below. I'd prefer to wait for a review from either @jgao54 or someone else on the team before being confident with merging this though.

flow/connectors/cockroachdb/qrep.go
viragtripathi: You are right, a NULL estimate scans as 0 without an error so the warning never fired. Moved it into the totalRows <= 0 branch and reworded it to "failed to estimate row count or table is empty, falling back to precise count".

flow/connectors/cockroachdb/qrep.go
viragtripathi: Yes, added them. supportsRangePartition and the min/max value check now accept the unsigned kinds the same way the mysql connector does. This goes together with the oid change from the other two comments, since oid columns are now uint32, and the changefeed JSON path also got a uint32 parse case.

flow/connectors/cockroachdb/qrep.go
viragtripathi: Changed, clusterLogicalTimestamp now parses the text and returns the typed crdbHLC. The crdbHLC type and parser moved down into this commit so it is available here, and the string form only appears where it is embedded into SQL.

flow/connectors/cockroachdb/qrep.go
viragtripathi: Same change here, snapshotSystemTime returns the typed crdbHLC now and callers render it only at the SQL boundary.

flow/connectors/cockroachdb/qvalue_convert.go
viragtripathi: Changed to QValueKindUInt32. The value conversion returns QValueUInt32 directly now and the old uint32 to int64 shim is removed. The same fix covers this and the other two oid comments.

flow/connectors/cockroachdb/qvalue_convert.go
viragtripathi: Done, pgtype.OIDOID maps to QValueKindUInt32 now. Same fix as the other oid comments. Verified against a cloud cluster that pg_class oids above the int32 range come through correctly.

viragtripathi:
> Thanks for doing this! Looks mostly good to me, save for the comments below. I'd prefer to wait for a review from either @jgao54 or someone else on the team before being confident with merging this though.

Thank you for taking the time to go through all of this so carefully. Appreciate your help!

flow/connectors/cockroachdb/schema.go
```diff
@@ -0,0 +1,247 @@
...
+				return nil, fmt.Errorf("failed to scan column info: %w", err)
+			}
+
+			qkind := crdbTypeToQValueKind(dataType, udtName)
+			colType := udtName
+			if system == protos.TypeSystem_Q {
+				colType = string(qkind)
+			}
+			// capture DECIMAL(p,s) so destinations pick a matching decimal type;
+			// -1 means unbounded (like Postgres atttypmod)
```
jgao54: nit: it looks like `DECIMAL(precision)` is also [supported](https://www.cockroachlabs.com/docs/stable/decimal#precision-and-scale) in CockroachDB; and currently the check for typeModifier is `numericPrecision != nil && numericScale != nil`. iiuc, currently the code would ignore precision if scale is not set (or does it always get set to 0 automatically server-side?)

flow/connectors/cockroachdb/qrep.go
```diff
@@ -0,0 +1,547 @@
...
+	return err != nil && strings.Contains(err.Error(), gcThresholdErrorSubstring)
+}
+
+// classifySnapshotReadError marks AS OF SYSTEM TIME reads whose timestamp has
+// fallen behind the replica GC threshold as non-retryable for Temporal: the
+// MVCC history at the snapshot timestamp is garbage collected, so retries can
+// never succeed and the mirror needs a resync. This mirrors the postgres
+// invalid-snapshot handling; shared.WrapError re-raises the ApplicationError
+// even when it gets wrapped along the way.
+func classifySnapshotReadError(err error) error {
```
jgao54, line 43: `ApplicationErrorTypeIrrecoverableInvalidSnapshot` is hard-coded to `Source:  ErrorSourcePostgres`, can do a sweep and update all `ErrorSourcePostgres` to `pgErrSource` in `GetErrorClass`

flow/connectors/cockroachdb/qrep.go
```diff
@@ -0,0 +1,547 @@
...
+		}
+
+		adjustedPartitions := shared.AdjustNumPartitions(totalRows, numRowsPerPartition)
+		c.logger.Info("[cockroachdb] partition details",
+			slog.Int64("totalRows", totalRows),
+			slog.Int64("desiredNumRowsPerPartition", numRowsPerPartition),
+			slog.Int64("adjustedNumPartitions", adjustedPartitions.AdjustedNumPartitions),
+			slog.Int64("adjustedNumRowsPerPartition", adjustedPartitions.AdjustedNumRowsPerPartition))
+		numPartitions = adjustedPartitions.AdjustedNumPartitions
+	}
```
jgao54: nit: this duplication of `AdjustNumPartitions` and the several conditional based on `numPartitions` and `lastRangeEnd` is a bit hard to follow,  a simpler way to organize this logic: 

```
var lastRangeEnd any
resuming := last != nil && last.Range != nil
// numPartitions not overriden, therefore needs to be computed
 if numPartitions == 0 {
      var totalRows int64
      if resuming {
          // handle the resuming case (this is actually a deprecated
          // feature so at some point we'll probably take this out; this
          // will make it easier to do so)
          switch lastRange := last.Range.Range.(type) { /*compute lastRangeEnd*/}
          totalRows = /* compute exact row count with filter, since it can't be estimated */                                                                 
      } else {
          // handle the general case
          totalRows = /* compute estimate row count, fallback to exact row count */                  
      }

      if totalRows == 0 {                                                            
          numPartitions = 1
      } else {                                                                       
          numPartitions = AdjustNumPartitions()
          ...                                                    
      }                                                                              
  }                                                                                  
 
  // compute min/max values, passing in `resuming`
  minMaxQuery := buildMinMaxQuery()
  ... 
```

flow/connectors/cockroachdb/qrep.go
```diff
@@ -0,0 +1,547 @@
...
+	// add null values partition to the end, if nulls aren't present it will be an empty partition
+	// that gets skipped during replication
+	if config.AddNullPartition {
+		partitionHelper.AddNullPartition()
+	}
+
+	return partitionHelper.GetPartitions(), nil
+}
+
+func (c *CockroachDBConnector) GetDefaultPartitionKeyForTables(
```
jgao54, line 175: Realized this function is currently not doing much because `input.TableSchemaMapping` is empty and is guarded behind [this conditional](https://github.com/PeerDB-io/peerdb/blob/d81e12525106dc1e3092dc7cd78cb5bac30a5e68/flow/activities/snapshot_activity.go#L201).

This was a footgun on our end; we'll separately merge https://github.com/PeerDB-io/peerdb/pull/4686 so no action needed here.

jgao54:
i see that the ci test infra is not added til https://github.com/PeerDB-io/peerdb/pull/4671 which makes it hard to validate changes in the PR merged so far, 

if possible, would be good to have the test infra set up (the minimum required to run e2e tests), and have the e2e tests be merged gradually with the PRs. 

not blocking for this PR, but can be done as a follow-up PR to add e2e test coverage for stack/1 and stack2/ so far!

flow/connectors/cockroachdb/schema.go
viragtripathi: Good question. Checked on v25.4 and v26.2: the server treats DECIMAL(p) as DECIMAL(p, 0) and information_schema reports numeric_scale as 0 rather than NULL, so precision was already being captured. I still changed the check to key on precision alone and default a missing scale to 0, so precision can never be silently dropped, and added a test covering DECIMAL(10), DECIMAL(10,2) and bare DECIMAL.

flow/connectors/cockroachdb/qrep.go
viragtripathi: Done, every Source: ErrorSourcePostgres inside GetErrorClass now uses pgErrSource, so classifications reached through a wrapped CockroachDB error are attributed to cockroachdb, including the invalid snapshot application error from this path. Added a classifier test for that case. The change sits in the CDC commit since that is where pgErrSource is introduced, so it shows up in the next PR of the stack.

flow/connectors/cockroachdb/qrep.go
viragtripathi: Reorganized along the lines you sketched. lastRangeEnd and a resuming flag are computed once up front, the numPartitions block keeps the deprecated resume path in its own branch with an exact filtered count so it is easy to remove later, the general path keeps the estimate with a precise fallback, and the min/max query is back to plain min/max with the duplicate AdjustNumPartitions block gone. The count and the min/max both read AS OF the same system time, so dropping the shared scan does not lose consistency.

flow/connectors/cockroachdb/qrep.go
viragtripathi: Thanks for tracking that down, and for #4686. Leaving this as is; once that merges the schema mapping will flow through and this code starts doing its job.

jgao54 (approved):

jgao54:
Hey @viragtripathi  just realizing i didn't see the changes in this PR mentioned in https://github.com/PeerDB-io/peerdb/pull/4668#discussion_r3765753075, looks like it wasn't checked in. Could you follow-up with another PR for this? thanks!

EDIT: looks like it got bundled with stack/3, all good 👍

viragtripathi:
Heads up on how the test coverage landed after 4668 merged. The plan was for the CI wiring and the introspection and qrep e2e tests to ride with 4668 itself. The merge happened before I pushed that restructure, so those tests now arrive one PR later than I told you they would. Sorry about that. 4669 is now the bottom of the stack and it carries the flow.yml matrix legs and compose plumbing, the e2e scaffolding, the introspection and qrep tests, and the CDC tests it was always going to have. So the CI legs turn on with 4669 and every remaining PR still lands with its tests. The UI and dev tooling PRs on top stay thin.

PR #4669, opened 2026-08-05
crdb stack/3 cdc changefeeds

viragtripathi:


flow/connectors/cockroachdb/changefeed.go
```diff
@@ -0,0 +1,849 @@
...
+	if !crdbSystemTimeRegex.MatchString(hlc) {
+		return 0, fmt.Errorf("invalid CockroachDB HLC timestamp %q", hlc)
+	}
+	wall, _, _ := strings.Cut(hlc, ".")
+	nanos, err := strconv.ParseInt(wall, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("invalid CockroachDB HLC timestamp %q: %w", hlc, err)
+	}
+	return nanos, nil
+}
```
Jeremyyang920: https://github.com/cockroachdb/cockroachdb-parser/blob/97d0c5a3e2190b6135858dac489886d2324e4ade/pkg/util/hlc/hlc.go#L639

You could possible incorporate usage of the actual CRDB parser to do much of this parsing and validation. It could also help for other things like constructing the create changefeed statement to use the AST.

flow/connectors/cockroachdb/cdc.go
```diff
@@ -0,0 +1,662 @@
...
+)
+
+// changefeedSlotName is a placeholder slot identifier: CockroachDB has no
+// replication slots, but a non-empty name makes the snapshot workflow carry
+// the captured system time into the QRep AS OF SYSTEM TIME reads.
+const changefeedSlotName = "changefeed"
+
+const resolvedLagWarnThreshold = time.Hour
+
+func (c *CockroachDBConnector) clusterLogicalTimestamp(ctx context.Context) (string, error) {
```
itsbilal: Can we just have this method do the parse and return an `hlc.Timestamp{}` (we can make a quick struct in this package that has `{WallTime, Logical uint64}` even if we only use the WallTime in it). That way any errors during parsing will be returned here too, and we don't need to have a separate `parseHLCWallNanos` method. String return types can be brittle.

flow/connectors/cockroachdb/cdc.go
```diff
@@ -0,0 +1,662 @@
...
+// failure point are delivered again within the same PullRecords batch.
+type changefeedDedupKey struct {
+	table   string
+	key     string
+	updated string
+}
+
+// maxChangefeedDedupEntries bounds the replay dedup set; overflowing it trades
+// possible duplicates for bounded memory.
+const maxChangefeedDedupEntries = 1 << 20
```
itsbilal: @jgao54 correct me if I'm wrong but we don't usually do dedups like this in the connector, instead as long as we have at-least-once delivery we let the sink handle it, correct?

flow/connectors/cockroachdb/cdc.go
```diff
@@ -0,0 +1,662 @@
...
+	}
+
+	err = rows.Err()
+	switch {
+	case ctx.Err() != nil:
+		return false, ctx.Err()
+	case queryCtx.Err() != nil:
+		// watchdog fired: with records the batch is complete on the last
+		// checkpoint, without records reconnect and keep waiting
+		return state.recordCount > 0, nil
```
itsbilal: Should we verify that we only do this when the error is actually a watchdog `context deadline exceeded` and not any error from the parent context? Should be a pretty easy error check

flow/connectors/cockroachdb/cdc.go
```diff
@@ -0,0 +1,662 @@
...
+	state.req.RecordStream.UpdateLatestCheckpointID(wallNanos)
+	// a reconnect replays at most back to this resolved timestamp, so older
+	// dedup entries can never be delivered again
+	clear(state.seenSinceResolved)
+
+	if lag := time.Since(time.Unix(0, wallNanos)); lag > resolvedLagWarnThreshold &&
+		time.Since(state.lastLagWarnAt) > time.Minute {
+		state.lastLagWarnAt = time.Now()
+		c.logger.Warn("[cockroachdb] changefeed resolved timestamp lags far behind;"+
+			" if it falls behind gc.ttlseconds the mirror cannot resume",
```
itsbilal: Let's clarify `gc.ttlseconds` is a CRDB setting

flow/connectors/cockroachdb/changefeed.go
itsbilal: +1, let's import the CRDB HLC parser and just use hlc.Timestamp wherever possible (I'm particularly wary of passing around string timestamps)

flow/connectors/cockroachdb/changefeed.go
```diff
@@ -0,0 +1,849 @@
...
+// between TIMESTAMP and TIMESTAMPTZ zoneless values can reach the zoned path,
+// where they parse as UTC.
+var changefeedTimestampTZLayouts = [...]string{
+	time.RFC3339Nano,
+	"2006-01-02T15:04:05.999999999-0700",
+	"2006-01-02T15:04:05.999999999-07",
+	"2006-01-02 15:04:05.999999999Z07:00",
+	"2006-01-02 15:04:05.999999999-0700",
+	"2006-01-02 15:04:05.999999999-07",
+	"2006-01-02T15:04:05.999999999",
```
itsbilal, line 756: We should have the `2006-01-02T15:04:05.999999999Z` option too for UTC

itsbilal (commented):
Thanks for doing this! Still going through but wanted to flush some comments including one about eliminating the in-memory dedup map

flow/connectors/cockroachdb/validate.go
```diff
@@ -24,6 +27,170 @@ func (c *CockroachDBConnector) ValidateCheck(ctx context.Context) error {
...
+	return c.checkRangefeedEnabled(ctx)
+}
+
+// snapshotGCTTLFloorSeconds is the hard lower bound on the effective
+// gc.ttlseconds of source tables for mirrors with an initial snapshot: the
+// snapshot reads AS OF SYSTEM TIME at one fixed timestamp, and those reads
+// fail once the timestamp falls behind the replica GC threshold. Snapshot
+// duration cannot be predicted reliably, but a window this small cannot
+// cover any realistic initial load.
+const snapshotGCTTLFloorSeconds = 600
```
itsbilal: 10 mins is too little I'd argue; I know we aren't thinking about large tables here, but we'd at least want this to be 1h+. The default is 25h and I recall gc.ttlseconds less than 4h being very rare at CRDB.

flow/connectors/cockroachdb/changefeed.go
```diff
@@ -0,0 +1,849 @@
...
+	return num, true
+}
+
+// decodeChangefeedBytes decodes the Postgres hex escape form ("\x0102")
+// CockroachDB uses for BYTES values in changefeed JSON.
+func decodeChangefeedBytes(s string) ([]byte, error) {
+	if hexPart, ok := strings.CutPrefix(s, `\x`); ok {
+		return hex.DecodeString(hexPart)
+	}
+	return []byte(s), nil
```
itsbilal: This will convert the ascii values of the individual characters themselves - will we ever want that?

flow/connectors/cockroachdb/changefeed.go
```diff
@@ -0,0 +1,849 @@
...
+	Cursor           string
+	ResolvedInterval time.Duration
+}
+
+func buildChangefeedStatement(tables []*common.QualifiedTable, opts changefeedOptions) (string, error) {
+	if len(tables) == 0 {
+		return "", errors.New("changefeed requires at least one table")
+	}
+	// cursor is embedded in the statement; the regex both validates the HLC
+	// format and guards against injection
```
itsbilal: Let's also use the parser here

flow/connectors/cockroachdb/cdc.go
viragtripathi: Done. There's now a crdbHLC struct with WallNanos and Logical, parsed once at the boundary, so malformed text fails right here. Cursors, resolved timestamps and dedup entries all carry the typed value; strings only exist at the SQL and catalog persistence edges.

flow/connectors/cockroachdb/cdc.go
viragtripathi: The wrinkle here is that a reconnect mid batch replays everything since the last resolved timestamp, and CockroachDB reconnects are routine (node drains, watchdog), not crash-only. For ClickHouse the sink converges anyway, but for Kafka and friends those replays would be visible duplicates on every reconnect. So the set only spans one batch and exists to keep the common case clean. Happy to drop it if you'd rather lean on at-least-once everywhere.

flow/connectors/cockroachdb/cdc.go
viragtripathi: Good catch. The watchdog now cancels with a dedicated cause and the session end check uses context.Cause, so parent cancellation propagates as an error and only a genuine watchdog fire completes the batch. Covered in TestClassifySessionEnd.

flow/connectors/cockroachdb/cdc.go
viragtripathi: Reworded to "the CockroachDB gc.ttlseconds zone setting" in the error and the lag warning.

flow/connectors/cockroachdb/changefeed.go
viragtripathi: Added, both the T and space forms with a trailing Z, plus a fixture. Shows up after an ALTER between TIMESTAMPTZ and TIMESTAMP.

flow/connectors/cockroachdb/validate.go
viragtripathi: Raised the floor to 3600. Agreed 10 minutes was meaningless given the 25h default; the warning path still nudges people to size it against expected load duration.

flow/connectors/cockroachdb/changefeed.go
viragtripathi: No, we never want that. Verified live that BYTES always arrives with the \x prefix (empty values included), so a missing prefix means the column type drifted under us. It's now a hard error instead of silent corruption, with a test.

flow/connectors/cockroachdb/changefeed.go
viragtripathi: This one dissolved: the cursor is a typed value now and its rendering is digits only, so the regex guard at statement build time went away entirely, parseHLC validates at the string boundary instead.

flow/connectors/cockroachdb/changefeed.go
viragtripathi: It pulls in a big chunk of crdb's dependency tree and all we need here is splitting the decimal into wall nanos and the logical counter, so I went with the small struct Bilal suggested. It round trips the full HLC and is fuzz tested. The changefeed statement itself is a fixed template, only quoted identifiers and the typed cursor go in, so there isn't much parsing surface left. Happy to revisit if statements ever get built from user input.

flow/connectors/cockroachdb/history_retention.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return nil, err
+	}
+	conn, err := NewCockroachDBConnFromConfig(ctx, connConfig, c.ssh)
+	if err != nil {
+		return nil, err
+	}
+	if _, err := conn.Exec(ctx, "SET allow_unsafe_internals = on"); err != nil {
+		c.logger.Debug("[cockroachdb] could not set allow_unsafe_internals", slog.Any("error", err))
+	}
```
Jeremyyang920, line 113: I don't recall if this is scoped to a session or a cluster setting, but just thinking if we want to do a set = off at the end of the snapshsot.

viragtripathi: It's session scoped, just verified on 26.2 set it in one session, a fresh session still gets the gate error and shows off, and the cluster wide override (sql.override.allow_unsafe_internals.enabled) stays untouched. On our side the set also only ever happens on a dedicated connection opened for the retention builtins and closed right after each call, so the elevation ends with the connection and the shared connection never sees it.

flow/connectors/cockroachdb/cdc.go
itsbilal: I'll let @jgao54 make the call here (and don't drop it without a confirmation from someone else), but I'd have a slight preference to drop it just to keep the connectors themselves lightweight. I recognize that duplicates are more expected with cockroachdb but we should still be logically equivalent

flow/connectors/cockroachdb/cdc.go
```diff
@@ -0,0 +1,749 @@
...
+	watchdog := time.AfterFunc(state.watchdogWindow(), func() { cancelQuery(errChangefeedWatchdog) })
+	defer watchdog.Stop()
+
+	rows, err := conn.Query(queryCtx, stmt)
+	if err != nil {
+		return false, fmt.Errorf("failed to create changefeed: %w", err)
+	}
+	defer rows.Close()
+
+	if c.historyProtectionChecked.CompareAndSwap(false, true) {
```
itsbilal, line 394: There's no way in which the `CockroachDBConnector` gets reused in such a way that we run a changefeed session again with a new cursor, right? If yes, we'd probably want to set this to `false` again, but I don't see any code going in that direction so I'm assuming it's a non-issue.

flow/connectors/cockroachdb/cdc.go
```diff
@@ -0,0 +1,749 @@
...
+	if state.recordCount == 1 {
+		state.req.RecordStream.SignalAsNotEmpty()
+		state.batchDeadline = time.Now().Add(state.req.IdleTimeout)
+	}
+	if state.recordCount >= state.req.MaxBatchSize && state.graceDeadline.IsZero() {
+		// full batch: allow a short grace period for a final resolved
+		// timestamp so the checkpoint covers the pulled records
+		state.graceDeadline = time.Now().Add(2 * state.resolvedInterval)
+	}
+	if state.recordCount%50000 == 0 {
```
itsbilal: Let's make this % 50000 a const that we define at the top of the file.

flow/connectors/cockroachdb/validate.go
```diff
@@ -23,6 +27,171 @@ func (c *CockroachDBConnector) ValidateCheck(ctx context.Context) error {
...
+			slog.String("minGcTtlConfiguredOn", minTTLTarget),
+			slog.Int64("totalRowEstimate", totalRowEstimate))
+	}
+	return nil
+}
+
+// effectiveGCTTLSeconds resolves the gc.ttlseconds that applies to a table.
+// SHOW ZONE CONFIGURATION FOR TABLE already returns the effective cascaded
+// zone config with its origin (table, database or RANGE default) in the
+// target column; the database and default-range lookups are fallbacks for
```
itsbilal: This should never happen, right, unless the table itself does not exist? We shouldn't have the quiet fallback logic and should always just do `... FOR TABLE`.

itsbilal (approved):
Thanks for working on this, this looks good to me! As with the other PR I'd want @jgao54 or someone else on the team to also review this before we merge, but from the CockroachDB side of things this looks good.

flow/connectors/cockroachdb/validate.go
itsbilal: Actually, I realized you can have partitions of a table and those can have a `gc.ttlseconds` that's different from the table-level default. So we'd probably need to rethink this function a bit. I'm open to either option here:

1. Block partitioned tables from being validated in the first place
2. List all partitions for a table, get the `gc.ttlseconds` for each, and return the min of those.

flow/connectors/cockroachdb/cdc.go
viragtripathi: Good question. A connector instance lives inside a single SyncFlow activity execution. flowable.go creates it once and calls PullRecords in a loop until the activity ends. Every changefeed session it runs resumes from the same flow's stored cursor lineage, and that only advances. A genuinely new cursor requires a new SetupReplication, which only happens on mirror create or resync, and by then the running flow and this connector instance are already torn down. The fresh instance starts with the flag false. PullFlowCleanup also releases the job on drop as the backstop. I added this as a comment on the field declaration so it lives next to the code.

flow/connectors/cockroachdb/cdc.go
viragtripathi: Extracted it into a named const. pullProgressLogInterval lives in qrep.go next to the other pull tuning and both the qrep pull loop and the changefeed streaming loop use it now. One progress line every 50k records keeps large pulls observable without flooding the logs.

viragtripathi:
While gating this stack against 25.4.14 and 26.2.5 we found a correctness bug in the snapshot history protection. SHOW JOBS middle truncates job descriptions at roughly 69 characters, and the extend and release paths looked the job up with an exact suffix match on that description. For any flow name long enough to push the description past the truncation point the match never hit, so the heartbeat and the release were silent no-ops and the retention job just sat there until the 4 hour orphan window expired. That meant a long initial load on a long named mirror could lose its GC protection mid load, and jobs lingering from earlier runs were also behind the cursor past GC test flake.

The fix moves the lookup to crdb_internal.jobs, which carries the full untruncated description. Verified live on v24.1.30, v25.4.14, v26.2.5 and a CockroachDB Cloud cluster on 26.2. The lookup runs on the dedicated session that already sets allow_unsafe_internals for the retention builtins on 26.1 and later, and it degrades to the old SHOW JOBS query if crdb_internal.jobs is not readable. The release path now uses that session as well instead of the shared connection. The e2e test drives a flow name over 100 characters and pins the premise, asserting SHOW JOBS cannot find the job while protect, extend and release all still work through crdb_internal.jobs.

Two smaller fixes rode along from edge case testing. A NULL element inside an array came through the qrep pull path as a nil placeholder string while the changefeed path produced an empty string, both now produce an empty string. And a changefeed hitting a missing CHANGEFEED grant now fails the pull immediately instead of burning the reconnect retry budget, a later grant still recovers through the normal activity retry.

protos/peers.proto
```diff
@@ -151,6 +151,10 @@ message CockroachDBConfig {
   bool disable_tls = 9;
   bool skip_cert_verification = 10;
   optional ClientTlsConfig client_tls = 11;
+  bool use_changefeeds = 12;
+  // changefeed reconnect tuning; defaults to 5 retries with a 500ms base delay
+  optional uint32 max_retries = 13;
+  optional uint32 retry_base_delay_ms = 14;
```
jgao54: This is probably not where you wanted to add these settings; if they are intended to be configurable by user, the recommended way to support this is to make them dynamic config changes (see `dynamicconf.go`).  the peer config is intended for connection-related config during specified setup. 

also not sure how important is it to make these configurable vs. just leaving it with reasonable defaults, if `PullRecord` fails it will retry indefinitely anyways on the Activity level, so may be able to get away with not making them configurable to keep the logic simple.

.github/workflows/flow.yml
```diff
@@ -64,6 +64,10 @@ jobs:
               lts: clickhouse/clickhouse-server:25.8.28.1
               stable: clickhouse/clickhouse-server:26.3.16.16
               latest: clickhouse/clickhouse-server:latest
+            cockroachdb:
+              lts: cockroachdb/cockroach:v25.4.14
+              stable: cockroachdb/cockroach:v26.2.5
+              latest: cockroachdb/cockroach:latest
```
jgao54: We use ClickHouse's `latest` image given it's the most heavily used destination connector; and is beneficial for identifying latest compatibility issues with ClickHouse, but at the expense of the maintenance overhead of having to address them as soon as CI detects an issue. We do not do this for source connectors. so woudl recommend using fixed version(s) here that are popular

jgao54: this means can also remove `COCKROACHDB_IMAGE` in `.env.example`

protos/peers.proto
```diff
@@ -151,6 +151,10 @@ message CockroachDBConfig {
   bool disable_tls = 9;
   bool skip_cert_verification = 10;
   optional ClientTlsConfig client_tls = 11;
+  bool use_changefeeds = 12;
```
jgao54: iiuc, the `use_changefeeds` isn't needed and can be derived based on if it's a snapshot-only pipe or a cdc pipe, since cdc requires changefeed.

flow/connectors/cockroachdb/cdc.go
jgao54: I agree with @itsbilal , would prefer on dropping for simplicity as well. Reason is peerdb does not provide exactly once guarantee so even if everything is perfectly deduped here on the source, downstream actions can still fail and retry and result in duplication.

flow/connectors/cockroachdb/validate.go
```diff
@@ -23,6 +27,171 @@ func (c *CockroachDBConnector) ValidateCheck(ctx context.Context) error {
...
+	var minTTLTable, minTTLTarget string
+	var totalRowEstimate int64
+	for _, table := range tables {
+		ttl, target, err := c.effectiveGCTTLSeconds(ctx, table)
+		if err != nil {
+			// insufficient privileges to read zone configurations shouldn't block
+			// mirror creation; the snapshot surfaces GC threshold errors otherwise
+			c.logger.Warn("[cockroachdb] unable to determine effective gc.ttlseconds",
+				slog.String("table", table.String()), slog.Any("error", err))
+			continue
```
jgao54: in general, i'd say we try to make validation catch as many things as we can upfront.  

if insufficient privilege is an issue, then granting privilege itself should be a requirement and be validated, unless the privilege needed is to powerful (e.g. require write or admin access)

flow/connectors/cockroachdb/validate.go
```diff
@@ -23,6 +27,171 @@ func (c *CockroachDBConnector) ValidateCheck(ctx context.Context) error {
...
+}
+
+// snapshotGCTTLFloorSeconds is the hard lower bound on the effective
+// gc.ttlseconds of source tables for mirrors with an initial snapshot: the
+// snapshot reads AS OF SYSTEM TIME at one fixed timestamp, and those reads
+// fail once the timestamp falls behind the replica GC threshold. Snapshot
+// duration cannot be predicted reliably, so anything below an hour cannot be
+// trusted to cover an initial load; CockroachDB's default is 25 hours and
+// production values below 4 hours are rare.
+const snapshotGCTTLFloorSeconds = 3600
```
jgao54: is there a reason why the floor is so low?  seems reasonable to match cockroachdb's default

jgao54: i think 24 hours might actually be the minimum that we want here for validation, if my understanding of `gc.ttlseconds` is correct that:
- snapshot that takes less than this TTL may fail (assuming they are on a version where history retention isn't supported)
- cdc that pause longer than this TTL may not be able to resume

flow/connectors/cockroachdb/validate.go
jgao54: +1 to the fallback logic being a bit odd here, unless that's how the config behaves where setting a more granular ttl overrides a coarse ttl, and the config is applied in a cascading way. then i can see this making sense.

in postgres we don't list/validate partitioned tables; but this because just validating the source table is sufficient, where as in this case it looks like ttl can be set on the partition level, and that overrides the parent table's ttl? if so, then validating the min seems a bit safer

flow/connectors/cockroachdb/history_retention.go
```diff
@@ -0,0 +1,249 @@
...
+	connConfig, err := ParseConfig(c.connStr, c.config)
+	if err != nil {
+		return nil, err
+	}
+	conn, err := NewCockroachDBConnFromConfig(ctx, connConfig, c.ssh)
+	if err != nil {
+		return nil, err
+	}
+	if _, err := conn.Exec(ctx, "SET allow_unsafe_internals = on"); err != nil {
+		c.logger.Debug("[cockroachdb] could not set allow_unsafe_internals", slog.Any("error", err))
```
jgao54: nit: this should be a warn instead of debug

flow/connectors/cockroachdb/history_retention.go
```diff
@@ -0,0 +1,249 @@
...
+			c.logger.Debug("[cockroachdb] failed to close history protection connection", slog.Any("error", err))
+		}
+	}()
+
+	var jobID int64
+	if err := conn.QueryRow(ctx, protectMVCCHistoryQuery(historyRetentionWindow),
+		systemTime, historyRetentionDescription(flowJobName)).Scan(&jobID); err != nil {
+		c.logger.Warn("[cockroachdb] could not protect the snapshot timestamp from garbage collection,"+
+			" the initial load relies on gc.ttlseconds alone; protection requires CockroachDB 24.1+,"+
+			" the REPLICATION privilege, and on 26.1+ allow_unsafe_internals = on"+
```
jgao54, line 134: should REPLICATION privilege be part of validation requirement ?

flow/connectors/cockroachdb/validate.go
```diff
@@ -23,6 +27,171 @@ func (c *CockroachDBConnector) ValidateCheck(ctx context.Context) error {
...
+		" once it falls behind the replica GC threshold; raise it to cover the expected initial load duration with"+
+		" ALTER TABLE %s CONFIGURE ZONE USING gc.ttlseconds = <seconds>",
+		table, ttlSeconds, target, snapshotGCTTLFloorSeconds, table)
+}
+
+// validateSnapshotGCTTL checks that the MVCC GC window of every source table
+// leaves the initial snapshot's AS OF SYSTEM TIME reads room to complete,
+// failing hard below snapshotGCTTLFloorSeconds and warning with sizing
+// guidance otherwise.
+func (c *CockroachDBConnector) validateSnapshotGCTTL(ctx context.Context, tables []*common.QualifiedTable) error {
```
jgao54: nit: should rename to `validateGCTTL` since this matters for CDC as well, not just for snapshot

(and should always be checked, not just `cfg.DoInitialSnapshot == true`

flow/connectors/cockroachdb/validate.go
```diff
@@ -23,6 +27,171 @@ func (c *CockroachDBConnector) ValidateCheck(ctx context.Context) error {
...
+	for _, table := range tables {
+		ttl, target, err := c.effectiveGCTTLSeconds(ctx, table)
+		if err != nil {
+			// insufficient privileges to read zone configurations shouldn't block
+			// mirror creation; the snapshot surfaces GC threshold errors otherwise
+			c.logger.Warn("[cockroachdb] unable to determine effective gc.ttlseconds",
+				slog.String("table", table.String()), slog.Any("error", err))
+			continue
+		}
+		if err := gcTTLFloorError(table.String(), target, ttl); err != nil {
```
jgao54: nit: this naming is a bit confusing (thought it was jsut fetching the ttl floor), maybe `validateGCTTLFloor`

flow/connectors/cockroachdb/changefeed.go
```diff
@@ -0,0 +1,1008 @@
...
+) (model.RecordItems, error) {
+	items := model.NewRecordItems(len(row))
+	for col, raw := range row {
+		if _, excluded := exclude[col]; excluded {
+			continue
+		}
+		field, ok := schema.fields[col]
+		if !ok {
+			continue
+		}
```
jgao54: when a column is not in the schema?  does this indicate a ADD COLUMN DML?  if so are we intentionally not supporting column additions for some reason?

flow/connectors/cockroachdb/cdc.go
jgao54: Was reading a bit a more about changefeed and how resolve works and realizing with the peerdb framework where `PullRecord` is called every sync interval and we are creating a new changefeed each time from last persisted checkpoint, it is almost guaranteed that even in the happy path there will be dupes emitted every sync interval, since the resolved timestamp doesn't necessary represent the timestamp of the most recently processed event. The current dedupe logic is not going to help with that since it only helps dedupe _within_ a batch. 

Given this behavior, having a fixed 1-second checkpoint (instead of currently computing a value between 1s-10s based on sync interval) is probably preferred to reduce duplicated events, assuming it doesn't introduce too much server-side burden. This would limit the duplicated messages per batch to up to 1-second worth of events.

Another idea would be to buffer the processed change events in-memory,  when a resolved timestamp is received, forward the buffered events _up to the resolved timestamp_, and then checkpoint the timestamp in-memory. This would avoid the happy-path duplicates between PullRecords. To avoid memory pressure, the buffer can be bounded so once full it gets flushed to the stream anyways. Just an idea for consideration.

flow/connectors/cockroachdb/cdc.go
```diff
@@ -0,0 +1,750 @@
...
+
+	updatedTs, err := parseHLC(envelope.Updated)
+	if err != nil {
+		return newTerminalChangefeedError(
+			fmt.Errorf("changefeed message for %s missing updated timestamp: %w", source, err))
+	}
+	commitWallNanos := updatedTs.WallNanos
+
+	nameAndExclude := state.req.TableNameMapping[source]
+	if unknown := schema.unknownColumns(envelope.After, envelope.Before); len(unknown) > 0 {
```
jgao54, line 543: i think the current implementation will deem excluded columns as unknown (because the the schema computed from `TableNameSchemaMapping` already has excluded removed) and try to add it to the schema,  so you'd want to make sure `unknownColumns` checks for that.

(i see that the data properly filters out the excluded columns, but we should make sure the schema doesn't include them either)

flow/connectors/cockroachdb/cdc.go
```diff
@@ -0,0 +1,750 @@
...
+// announce DDL) and emits a TableSchemaDelta for the added columns.
+func (c *CockroachDBConnector) emitSchemaDelta(
+	ctx context.Context,
+	state *changefeedPullState,
+	source string,
+	destination string,
+	schema *changefeedTableSchema,
+	unknown []string,
+) error {
+	freshSchemas, err := c.GetTableSchema(ctx, state.req.Env, state.req.InternalVersion, protos.TypeSystem_Q,
```
jgao54: iiuc, the reason we need to fetch schema from source is because the change events only have field names and values, but we don't know anything about the type of the added fields from the change event themselves, right?

is it possible to fetch table schema AT SYSTEM TIME (of the event timestamp)? because that would give you not the latest schema, but the correct schema at the time of the DDL, which would remove race conditions caused by fetching the latest schema when multiple add/remove DDLs are issued in the same batch.

protos/peers.proto
viragtripathi: Removed both fields from the peer config. The reconnect loop now uses fixed defaults, 5 retries with a 500ms base and a 30s cap, and once that budget is spent the error goes back to the activity which retries anyway. Didn't feel worth a dynamic setting, happy to wire it into dynamicconf later if someone actually needs to tune it.

protos/peers.proto
viragtripathi: You're right, it's derivable. Dropped the field entirely. Snapshot only mirrors skip the changefeed checks and everything else validates rangefeeds. The UI toggle and the nexus option went away with it.

flow/connectors/cockroachdb/validate.go
viragtripathi: Changed both spots to hard failures. I checked on 25.4 that a plain SELECT grant is enough to read the effective zone config and the partitions, so nothing extra to require there. The rangefeed check needs VIEWCLUSTERSETTING and the error now says that instead of logging a warning and moving on.

.github/workflows/flow.yml
viragtripathi: Makes sense. The matrix now pins v24.1.30, v25.4.14 and v26.2.5, which are the versions this was verified against, and latest is gone. COCKROACHDB_IMAGE is out of .env.example too, local runs get the image from the flow.yml matrix like the other databases.

flow/connectors/cockroachdb/cdc.go
viragtripathi: Dropped the dedup map entirely, replays just pass through now. And your reading is right, every new session starts from the last resolved timestamp so cross batch dupes were always there, the map only ever hid the in batch ones. I took the fixed 1 second resolved interval suggestion, it was scaled between 1s and 10s off the sync interval before, so the happy path replay is now bounded to about a second of events. I left the buffering idea out for now since it trades memory and latency for dupes the sinks already absorb, but I'm happy to try it as a follow up if you want. The tests treat delivery as at least once now.

flow/connectors/cockroachdb/validate.go
viragtripathi: Raised it to 86400 and it now runs for CDC mirrors too, not just snapshots. Two things I hit while testing. Single node clusters lower the default window to 4 hours, so the CI provisioning script now puts it back to 25. And Cloud Standard defaults to 4500 seconds, so on a stock cloud cluster validation fails until you raise the table zone configs. That does work there, I verified on a Standard cluster, and the error spells out the exact statement. If failing on cloud defaults feels too harsh I can soften that case to a warning, let me know.

flow/connectors/cockroachdb/validate.go
viragtripathi: Went with the min across partitions. The fallback queries are gone, the table level statement already returns the effective cascaded config with its origin. Partition configs override the table so the effective value is the min of the table and its partitions. When a partition is the culprit the error names it and the hint uses the partition form of the alter, since raising it on the table wouldn't help. Covered by a live test that sets a low ttl on one partition.

flow/connectors/cockroachdb/history_retention.go
viragtripathi: Changed to warn.

flow/connectors/cockroachdb/history_retention.go
viragtripathi: I'd rather not require it. The protection is best effort by design, it only exists on 24.1 and later, and the snapshot still works off gc.ttlseconds alone, which validation now enforces at 24 hours. REPLICATION also feels like a big grant to demand for a belt and suspenders feature. When protection can't be created we log a warning with the reason so it's visible.

flow/connectors/cockroachdb/validate.go
viragtripathi: Done, renamed to validateGCTTL and it runs for every mirror now. The sizing warning with the row estimate stays scoped to mirrors that do an initial load.

flow/connectors/cockroachdb/validate.go
viragtripathi: Renamed to validateGCTTLFloor.

flow/connectors/cockroachdb/cdc.go
viragtripathi: Good catch, that is exactly what would have happened. The exclude list is now seeded into the schema's ignored set when the cached schema is built, so excluded columns never register as unknown and never make it into a delta. Unit test added, and I ran a live check with an excluded column emitted alongside a genuinely added one, only the real one produced a delta.

flow/connectors/cockroachdb/cdc.go
viragtripathi: Right, the events only carry names and values. And yes it works, information_schema reads accept AS OF SYSTEM TIME, so the schema refresh now reads at the row's updated timestamp. That timestamp is always inside the GC window because the cursor has to be for the feed to run at all. There's a new e2e that adds a column, writes a row, drops the column again before the pull, and the row still comes out typed correctly with the delta emitted.

flow/connectors/cockroachdb/cdc.go
Jeremyyang920: > we are creating a new changefeed each time from last persisted checkpoint

Is this going to have performance impact on large clusters where we are constantly creating these? Joy's suggestions is almost similar to what Replicator does, but it directly writes the events to the staging table and pops them out when the resolved message comes in. It also acts as the dedupe directly. There was a new version that used Pebble to act as that store but that may be a bit overkill for this use case.

viragtripathi: You are right and there is some per session cost, mostly re-establishing rangefeeds over the table's ranges plus a catch up scan from the cursor, so it scales with the number of ranges and how much changed in the interval rather than table size. Sinkless also doesn't go through the jobs system so the restarts don't churn the job registry. With the default sync interval that's one setup per minute which held up fine in my testing, though I haven't measured it on a really large cluster. On the Replicator point, do you think the same staging table approach is worth doing here with sinkless, buffer the events and pop them at resolved so it handles the dupes too? If that's the right shape I can take a crack at it as a follow up.

flow/connectors/cockroachdb/validate.go
jgao54: i think having the error spells out the fix needed is sufficient, otherwise it's a worse experience to fail snapshot after 75 minutes (when history retention isn't supported)

jgao54 (approved):
great work, thanks!

viragtripathi:
> great work, thanks!

Thanks a lot for all your reviews and help!

PR #4676, opened 2026-08-07
fix(clickhouse): quorum writes on replicated clusters to stop normalize dropping rows

andreyzhelnin-st:
## Problem

On a multi-replica self hosted ReplicatedMergeTree cluster, normalize populates the raw table and then reads it back via `INSERT ... SELECT`. Those two statements can run on different connections/replicas (e.g. behind a load-balanced host), so the read may hit a replica that has not yet replicated the just-written raw parts.

`select_sequential_consistency` (already set on reads) only guarantees visibility of quorum-inserted blocks. With the default `insert_quorum=0`, it enforces nothing, so the normalize `SELECT` can return fewer/zero rows, mark the batch processed, advance the normalize pointer, and silently drop those rows permanently.

## Fix

- Write raw/normalize inserts with `insert_quorum=auto` and `insert_quorum_parallel=0` (required for sequential-consistency reads to be honored) on `Replicated` clusters. Gated behind a new dynamic setting `PEERDB_CLICKHOUSE_ENABLE_REPLICATED_QUORUM` (default `true`).
- Explicitly pin `select_sequential_consistency=1` alongside the quorum writes so the read side of the read-your-writes contract stays paired with them regardless of the global settings default.

## Test plan

- [ ] Verified against a multi-replica ReplicatedMergeTree cluster that normalize no longer drops rows under replica lag
- [ ] Existing ClickHouse connector tests pass


andreyzhelnin-st:
Could you please review this @jgao54 - it works good in our prod for a few days?

andreyzhelnin-st:
@serprex could you please have a look?

flow/internal/dynamicconf.go
```diff
@@ -305,6 +305,17 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_IMMEDIATE,
 		TargetForSetting: protos.DynconfTarget_CLICKHOUSE,
 	},
+	{
+		Name: "PEERDB_CLICKHOUSE_ENABLE_REPLICATED_QUORUM",
+		Description: "On Replicated ClickHouse clusters, write raw/normalize inserts with quorum " +
+			"(insert_quorum=auto, insert_quorum_parallel=0) so that select_sequential_consistency reads " +
+			"see them regardless of which replica serves the read. Prevents normalize silently dropping " +
+			"rows when the read hits a replica that has not yet replicated the just-written raw parts.",
+		DefaultValue:     "true",
```
jgao54: The default value should be false here to ensure backwards-compatibility (it's also the default setting in ClickHouse so would be good to keep it consistency)

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -227,6 +213,24 @@ func Connect(ctx context.Context, env map[string]string, config *protos.Clickhou
...
+		// immediately reads it back via INSERT ... SELECT. Those two statements may run on different
+		// connections/replicas (e.g. behind a load-balanced host), so the read can hit a replica that
+		// has not yet replicated the just-written parts. select_sequential_consistency (set above) only
+		// guarantees visibility of quorum-inserted blocks, so without quorum writes it enforces nothing
+		// and normalize silently drops rows it marks as processed. Writing with quorum closes the race:
+		// insert_quorum_parallel=0 is required for sequential-consistency reads to be honored.
+		if quorum, err := internal.PeerDBClickHouseEnableReplicatedQuorum(ctx, env); err != nil {
+			return nil, fmt.Errorf("failed to load replicated quorum config: %w", err)
+		} else if quorum {
+			settings["insert_quorum"] = "auto"
```
jgao54, line 240: "auto" was introduced in ClickHouse [22.9](https://presentations.clickhouse.com/2022-release-22.9/), so on older version I suspect it may cause connect to fail. Since this is behind a feature flag, should be fine to add a note at the end of the Description like "Supported by ClickHouse 22.9+"

jgao54:
two nits otherwise lgmt 👍

flow/internal/dynamicconf.go
andreyzhelnin-st:
```suggestion
		DefaultValue:     "false",
```

andreyzhelnin-st:
> two nits otherwise lgmt 👍

@jgao54 thanks for review, addressed.

jgao54 (approved):

andreyzhelnin-st:
@jgao54 how can I proceed with merge ?

PR #4677, opened 2026-08-07
Add test cases for detecting incompatible Temporal history

jgao54:
Add tests for detecting Temporal history replay incompatibility. We've been bitten by this in the past. And this will make it easier to refactor CDCFlowWorkflow, which is currently a 1000+ loc that is very difficult to reason about, especially the continue-as-new logic that we currently employ to avoid patching. 

Verified by removing critical code (that are not deleted by continue-as-new) has caused CI to fail as expected.

### DIFF
--- a/flow/workflows/cdc_flow_replay_fixtures_test.go
+++ b/flow/workflows/cdc_flow_replay_fixtures_test.go
@@ -30,10 +30,6 @@ import (
 // The replay test TestCDCFlowWorkflowReplay re-executes the workflow code
 // against these recorded histories and checks for the same command sequence.
 //
-// To generate fixture, start the catalog and run the following command:
-//
-//	PEERDB_GENERATE_REPLAY_FIXTURES=1 go test -count=1 -timeout 10m -run TestGenerateCDCFlowReplayFixtures ./workflows/
-//
 // Note that fake activity implementations are registered under production names
 // to generate history. This is sufficient because replay never re-executes
 // activities, it validates only the command sequence the workflow  code issues.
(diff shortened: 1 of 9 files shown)


jgao54:
[Verified](https://github.com/PeerDB-io/peerdb/actions/runs/31208857331/job/92966582307?pr=4677) that we got "replay diverged from recorded history" as expected.

flow/workflows/cdc_flow_replay_fixtures_test.go
```diff
@@ -0,0 +1,318 @@
...
+		// the fake SyncFlow returns promptly for this flow name, exercising the
+		// steady-state "sync finished -> ContinueAsNew" path
+		wfID, firstRunID := startRunningFlow(ctx, t, c, "running-sync-finish")
+		waitContinuedAsNew(ctx, t, c, wfID, firstRunID)
+		exportHistory(ctx, t, c, wfID, firstRunID, "running_sync_finish")
+		terminateWorkflow(ctx, t, c, wfID)
+	})
+}
+
+func setCatalogEnvDefaults(t *testing.T) {
```
ilidemi: In other tests we aligned on the pattern like [this one](https://github.com/tpanetti/peerdb/blob/74e64fa6ae108724614a9ec1f957a9f5cf7a3b62/flow/connectors/utils/autoenv.go#L9), would it work here? Idea is that locally env variables come from an `.env` file and in CI the yaml sets them, and if code needs to provide a default, the shared helper does that

flow/workflows/cdc_flow_replay_fixtures_test.go
```diff
@@ -0,0 +1,318 @@
...
+//	PEERDB_GENERATE_REPLAY_FIXTURES=1 go test -count=1 -timeout 10m -run TestGenerateCDCFlowReplayFixtures ./workflows/
+//
+// Note that fake activity implementations are registered under production names
+// to generate history. This is sufficient because replay never re-executes
+// activities, it validates only the command sequence the workflow  code issues.
+// The one exception is local activities, which is why the catalog must be running.
+//
+// IMPORTANT: fixtures must be generated from the workflow code that is currently
+// deployed, never from the branch under test. Otherwise, replay compares new code
+// against history that new code generates and trivially passes.
```
ilidemi: Good callout. Would it be possible to add the intended workflow on how to update fixtures properly? I.e. is it two PRs, is a deployment in between required, or the same PR with links in the description on how was it verified, etc.

ilidemi (approved):

flow/workflows/cdc_flow_replay_fixtures_test.go
jgao54: `TestCDCFlowWorkflowReplay` always runs in CI; so by default it will catch incompatibility issues.

for adding new tests, what i did was:
1) checkout out the code to the currently released version
2) starts the catalog and run the command described in this comment
3) it will update the replay json files (populate new files, and update existing ones, but the existing ones will be modified trivially that doesn't impact test correctness, like the `identity` field)

I'll add the steps here in the comments

flow/workflows/cdc_flow_replay_fixtures_test.go
jgao54: good catch

PR #4679, opened 2026-08-07
fix(clickhouse): validate staging read grants by access method

tpanetti:
## Summary

Moves ClickHouse staging-read grant selection into the `StagingStore` access-method contract instead of inferring it with connector-side concrete-type checks.

- S3 staging reports `S3`, matching direct ClickHouse S3 reads.
- GCS staging reports `URL`, matching the signed-URL path ClickHouse actually reads.
- Grant validation accepts only the supported `S3` and `URL` methods and fails closed before issuing SQL for any unsupported method.
- Existing new-style `CHECK GRANT READ ON <method>` and legacy `CHECK GRANT <method> ON *.*` compatibility behavior is preserved.
- Unit coverage verifies S3/GCS method selection, modern and legacy grants, missing grants, unsupported ClickHouse grant syntax, and unsupported staging methods.

This matters for migrated CDC pipes because the destination must validate the grant for the object-access mechanism it will actually use. In particular, a GCS-backed staging store is consumed through signed URLs and therefore requires `URL`, not a provider-name-derived grant.

## Scope

This is defensive PeerDB hardening, not a strict blocker for the CDC migration E2E wiring. The current S3 and GCS implementations already map to known methods; this change makes that mapping explicit at the staging-store boundary and ensures a future implementation cannot silently skip or construct an invalid grant check.

## Validation

- `gofmt -d` on every changed Go file: clean
- `git diff --check`: clean
- `go test ./clickhouse -run '^TestValidateStagingAccessGrant$'` from `flow/pkg`: passed
- Added `TestStagingStoreClickHouseAccessMethod` for the S3 and signed-URL GCS implementations

Local environment limitations:

- The full `flow/pkg/clickhouse` suite reaches existing integration tests that require ClickHouse on `localhost:9000`; no local ClickHouse instance was available.
- The connector package test binary could not link on this workstation because the installed GEOS library is `x86_64` while Go is building for `arm64`. Disabling CGO is not supported by the `go-geos` dependency.

CI should run the complete package suites in the repository's supported build environment.


itsbilal (approved):
LGTM

ilidemi (approved):
Nice refactor, pushing the access method into the StagingStore contract is the right call, and the fail-closed guard is a good addition.

One change before merging: I dropped validation_grant_test.go. The cases that matter for the migration — S3 and URL, modern and legacy grant syntax — are already covered against a real ClickHouse by TestCheckBucketGrantInValidatePeer. The two that aren't, the syntax-error fallback and the unsupported-method guard, can't fire in a managed environment: CH is ≥25.8 there, and Azure staging isn't wired up yet, and will be by the time we need to migrate onto it.

The other reason is mocking the server response. I'd rather not set that precedent for this area specifically — we won't know for sure what the server response looks like without running it, and generating tests from the code risks false confidence. If we're worried about regressions, local tilt and CI config can be extended to test against a real legacy instance.

PR #4688, opened 2026-08-13
support parallel_view_processing

jgao54:
`parallel_view_processing` allow for concurrent view processing, and can benefit from `max_threads` and `max_insert_threads`. 

The setting had been available since clickhouse 19.4 (2019) so is available on most ClickHouse versions. It is also behind a setting and not set by default. 

Fixes: DBI-1033


Jeremyyang920 (approved):
Thanks for adding!

PR #4692, opened 2026-08-14
 WIP feat(bigquery): add BigqueryCdcConfig mirror-config oneof

dtunikov:
Introduces a connector-specific mirror-config extension point on
FlowConnectionConfigs/FlowConnectionConfigsCore, mirroring how Peer
already does this for peer-level config. BigqueryCdcConfig (with a
cdc_mode of APPENDS or CHANGES) is the first variant.

Extends the codegen tool that keeps FlowConnectionConfigs and
FlowConnectionConfigsCore in sync: a oneof field can't be copied with
a plain `Core.X = api.X` since protoc-gen-go generates a distinct,
message-scoped interface type per parent message for the same oneof.
The generator now detects oneof fields (via their protoc-gen-go marker
methods) and emits a type switch per variant instead.

No behavior change yet - BigqueryCdcConfig is not read anywhere.

---


dtunikov:
Closing this in favour of splitting it in two, so the generic PeerDB change doesn't ride along with the BigQuery work:

- **#4700** — `refactor(codegen): convert FlowConnectionConfigs by field number, drop generator`. The codegen extension this PR originally added is gone entirely. `FlowConnectionConfigs` and `FlowConnectionConfigsCore` are wire-compatible by construction, so the converter now copies by field *number* via protoreflect, and a `oneof` needs no special-casing at all — no type switch, no generator. Parity is enforced by a descriptor-comparison test instead of at codegen time.
- **#4695** (`bq-cdc/3-validation`) — now carries the `BigqueryCdcConfig` proto oneof, the nexus `grpc.rs` fix, and the oneof round-trip test. Validation is the first branch that actually reads `cdc_mode`, so the proto change lands with its first consumer rather than as a standalone PR.

`bq-cdc/2-snapshot-timestamp` (#4693) is now based directly on #4700 and no longer depends on the oneof. The rest of the stack (#4695 → #4696 → #4698) was rebased with no conflicts; every branch was verified to build and vet clean, and branch 2 was checked to compile without the oneof present.

The `bq-cdc/1-proto-oneof` branch is being deleted.

PR #4695, opened 2026-08-14
WIP feat(bigquery): source validation

dtunikov:


dtunikov:
Heads up: this PR now also carries the `BigqueryCdcConfig` proto oneof, previously PR #4692 / `bq-cdc/1-proto-oneof`. That branch was dropped and its content split:

- The generic converter work moved to **#4700**, at the bottom of the stack.
- The proto change lands here, since validation is the first branch that actually reads `cdc_mode` — there's no reason for it to be a standalone PR.

Three commits were added below the existing validation work:

- `feat(bigquery): add BigqueryCdcConfig mirror-config oneof` — `protos/flow.proto` only (+18)
- `fix(nexus): set source_connector_config in FlowConnectionConfigs literal` — required for the Rust struct literal to keep compiling
- `test(proto_conversions): cover the source_connector_config oneof`

The codegen changes the original PR needed are gone: after #4700, `FlowConnectionConfigs` ↔ `FlowConnectionConfigsCore` conversion copies by field *number*, so a oneof is carried across with no converter work at all. The test commit only asserts the variant is rebuilt against the destination's own wrapper type in both directions, and that an unset oneof stays unset.

Rebased with no conflicts; `go build ./...` and `go vet ./...` are clean on this branch.

PR #4697, opened 2026-08-14
 WIP feat(bigquery): implement CHANGES-mode CDC pull

dtunikov:
CHANGES() reports an UPDATE as a delete+insert pair sharing the same PK
and _CHANGE_TIMESTAMP, so PullRecords needs to pair those back into one
UpdateRecord instead of emitting two records -- otherwise every update
on a CHANGES-mode mirror would be replicated as a delete followed by an
unrelated insert. Reads cdc_mode once per PullRecords call (it's a
per-mirror setting) to pick APPENDS or CHANGES per the "Decisions
locked in" plan.

---

### DIFF
--- a/flow/connectors/bigquery/bigquery.go
+++ b/flow/connectors/bigquery/bigquery.go
@@ -52,8 +52,13 @@ func NewBigQueryServiceAccount(bqConfig *protos.BigqueryConfig) (*utils.GcpServi
 }
 
 type BigQueryConnector struct {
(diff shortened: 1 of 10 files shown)


dtunikov:
Folding CHANGES-mode into #4696 (bq-cdc/4-appends-pull) — no need to split APPENDS and CHANGES pull modes into separate PRs. This branch's commit now lives there.

PR #4700, opened 2026-08-14
refactor(codegen): convert FlowConnectionConfigs by field number, drop generator

dtunikov:
`FlowConnectionConfigs` and `FlowConnectionConfigsCore` are wire-compatible by construction: every field they share has the same number and type. So converting between them doesn't need a field list at all — it's a copy by field number, which protoreflect does in a handful of lines:

```go
core := &protos.FlowConnectionConfigsCore{}
copyFieldsByNumber(api.ProtoReflect(), core.ProtoReflect())
```

This deletes `flow/cmd/codegen/flow_config_converter.go` — 209 lines that parsed the generated `flow.pb.go` with `go/ast` and templated out one `Core.X = api.X` line per field — and replaces it with a hand-written `flow/proto_conversions` package. `cmd/codegen` still exists for the gRPC wrapper.

Composite kinds that a per-field copy has to special-case — maps, repeated fields, and `oneof`, whose wrapper types protoc-gen-go scopes per parent message — need no special-casing when copying by number.

## Parity enforcement moves from codegen to a test

`TestFlowConnectionConfigsEquivalence` compares the two descriptors field by field: name, kind, cardinality, `optional` keyword, `oneof` membership, and message/enum type per shared number. That is strictly stronger than the check it replaces, which compared Go field *names* only and relied on the generated file failing to compile to catch type drift.

It also makes the property documented on `FlowConnectionConfigsCore` real. The proto says it's "safe to remove (mark reserved) fields from it", but the generator demanded exact name-set equality, so every removal needed another hardcoded skip (which is what the existing `TableMappings` / `TableMappingsVersion` special cases were). A removal is now declared in `removedFromCore` instead.

I verified the guard fires by regenerating protos twice with deliberate breakage:

- Core-only `table_mappings_version = 30` → `field table_mappings_version (30) is missing from FlowConnectionConfigs: add it there, or record it in coreOnly`
- Core's `max_batch_size` switched to `uint64` → `field max_batch_size (5) has a different type in Core`

## Tradeoff worth knowing

A field added to only one of the two messages used to break `go generate`; it now fails `go test ./proto_conversions/...` instead. That's later feedback, in exchange for a converter that needs no work per new proto feature.

## Notes

- The `ruleguard` rule banning `protos.FlowConnectionConfigs` outside `cmd/` and `e2e/` needed an exemption for the new package. The old converter was invisible to it only because `flow/generated` is lint-excluded; `proto_conversions` is by definition the boundary that handles both types.
- `flow/generated/proto_conversions/stub.go` is removed — the package is now real and committed, so `go mod tidy` resolves it without a placeholder.
- Verified: `go build ./...`, `go vet ./...`, `golangci-lint run` clean on touched packages, `go mod tidy` produces no diff.


ilidemi (approved):
Nice solution!

PR #4705, opened 2026-08-18
CDCFlowWorkFlow refactor

jgao54:
This PR looks like it's making a lot of change but it's a refactor that leaves the logic unchanged (use hide whitespace to reduce the diff). The main goal is to make CDCFlowWorkFlow more maintainable and readable, by:
- decomposing each "phase" (pipe creation, steady-state CDC, and handling pause signal) into its own helpers; leaving the core logic in `CDCFlowWorkFlow` simple. 
- make the helpers methods more explicit about when error vs CaN is returned (previously ContinueAsNewError was mixed with generic error). the new code will always return a `nextRun` in the helpers to indicate a CaN separate from error handling, and  inside `CDCFlowWorkFlow` the actual CaN will be called. 
- dedupe resync/terminate request processing logic

### DIFF
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -677,6 +677,7 @@ func startSetupAndSnapshot(
 		case protos.FlowStatus_STATUS_RESYNC:
 			// this is the only place where we can have a resync during a resync
 			// so we need to NOT sync the tableMappings to catalog to preserve original names
+			cfg.TableMappings = originalTableMappings
 			state.SyncFlowOptions.TableMappings = originalTableMappings
 			processResync(ctx, cfg, state, val, true)
 		}
--- a/flow/workflows/cdc_flow.go
+++ b/flow/workflows/cdc_flow.go
@@ -486,110 +486,92 @@ func addCdcPropertiesSignalListener(
 	})
 }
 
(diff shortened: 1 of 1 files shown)


flow/workflows/cdc_flow.go
```diff
@@ -626,219 +634,217 @@ func CDCFlowWorkflow(
...
+	flowSignalStateChangeChan.AddToSelector(setupSnapshotSelector, func(val *protos.FlowStateChangeRequest, _ bool) {
+		switch val.RequestedFlowState {
+		case protos.FlowStatus_STATUS_PAUSED:
+			logger.Warn("pause requested during setup, ignoring")
+		case protos.FlowStatus_STATUS_TERMINATING:
+			processTerminate(ctx, cfg, state, val)
+		case protos.FlowStatus_STATUS_RESYNC:
+			// this is the only place where we can have a resync during a resync
+			// so we need to NOT sync the tableMappings to catalog to preserve original names
+			state.SyncFlowOptions.TableMappings = originalTableMappings
```
ilidemi, line 681: nit: up to you, but could be a tiny bit less fragile to set cfg.TableMappings here as well

ilidemi (approved):
Awesome refactor, especially resync branches were hard to reason about

PR #4706, opened 2026-08-18
feat(bigquery): implement PullRecords for APPENDS/CHANGES modes

dtunikov:
This PR makes BigQuery a valid source connector.
Mainly it implements PullRecords method.
It supports two change capture modes:
- [APPENDS](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/time-series-functions#appends) - available for all BQ tables with time-travel enabled (7d window by default). Captures only INSERTs.
- [CHANGES](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/time-series-functions#changes) - available only when [enable_change_history option](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#table_option_list) is set for the table. Captures INSERTs/UPDATEs/DELETEs.
For now the mode can be configured per mirror, later we might move it to the table mapping config.


flow/activities/flowable_core.go
```diff
@@ -278,12 +281,23 @@ func pullAndSyncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDC
...
 			return nil, fmt.Errorf("failed to sync schema: %w", err)
 		}
 
-		return nil, a.applySchemaDeltas(ctx, config, recordBatchSync.SchemaDeltas)
+		if err := a.applySchemaDeltas(ctx, config, recordBatchPull.SchemaDeltas); err != nil {
+			return nil, err
+		}
+
+		lastCheckpoint := recordBatchPull.GetLastCheckpoint()
+		if lastCheckpoint != lastOffset {
```
dtunikov, line 294: for BigQuery it's important to advance the checkpoint even when no records were received
otherwise we'd query the same interval again and again, which will affect BQ costs

this change won't be needed if we decide to introduce a separate branch for `query-based` sources

flow/internal/dynamicconf.go
```diff
@@ -281,6 +281,24 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_AFTER_RESUME,
 		TargetForSetting: protos.DynconfTarget_BIGQUERY,
 	},
+	{
+		Name: "PEERDB_BIGQUERY_CDC_SAFETY_LAG_SECONDS",
```
dtunikov, line 285: microsoft fabric documents 10 mins delay [here](https://learn.microsoft.com/en-us/fabric/data-factory/cdc-copy-job-bigquery)
but I couldn't find anything about it in the official BQ docs + local testing didn't reveal any constant delay.

PR #4707, opened 2026-08-19
feat(bigquery): implement appends/changes event based cdc modes

dtunikov:
# What 
This PR makes BigQuery a valid source connector.
Mainly it implements PullRecords method and BigQuery CDC mirror validation. (check if ordering key is set for cdc mirrors, etc)
It supports two change capture modes:
- [APPENDS](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/time-series-functions#appends) - available for all BQ tables with time-travel enabled (7d window by default). Captures only INSERTs.
- [CHANGES](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/time-series-functions#changes) - available only when [enable_change_history option](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#table_option_list) is set for the table. Captures INSERTs/UPDATEs/DELETEs.

The new replication mode setting was added to BigQuery mirrors:
```
enum BigQueryReplicationMode {
  // Default: table-valued function (APPENDS()/CHANGES()) based CDC polling.
  BIGQUERY_REPLICATION_MODE_EVENTS = 0;
  // query based: standard "SELECT ... FROM table WHERE col > ? AND col <= ?"
  BIGQUERY_REPLICATION_MODE_QUERY = 1;
}
```
and the new BQ specific setting was added to the table mapping struct (only relevant for `BIGQUERY_REPLICATION_MODE_EVENTS` replication mode):
```
enum BigqueryCdcEventsFunction {
  BIGQUERY_CDC_EVENTS_FUNCTION_APPENDS = 0;
  BIGQUERY_CDC_EVENTS_FUNCTION_CHANGES = 1;
}
```

# Important files for review
- flow/connectors/bigquery/cdc.go - core BQ cdc logic
- flow/connectors/bigquery/source.go - changes to mirror validation
- flow/connectors/bigquery/qrep_object_pull.go - move checkpoint by 1ms forward to ensure that we don't have an overlap with the snapshot
- flow/e2e/bigquery_cdc_test.go - e2e test cases

# What is not included in this PR
- changes to the PeerDB UI to enable BigQuery CDC feature yet.
- error classification, need to explore it a bit more and don't want to overload this PR
- table isolation (if tableA facing errors - we should continue cdc for other pipe's tables) + parallel table processing

Resolves: DBI-1038


flow/activities/flowable_core.go
```diff
@@ -278,12 +281,23 @@ func pullAndSyncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDC
 
 		syncState.Store(new("updating schema"))
 		if err := dstConn.ReplayTableSchemaDeltas(
-			ctx, config.Env, flowName, options.TableMappings, recordBatchSync.SchemaDeltas, config.Flags,
+			ctx, config.Env, flowName, options.TableMappings, recordBatchPull.SchemaDeltas, config.Flags,
```
dtunikov: for BigQuery it's important to advance the checkpoint even when no records were received
otherwise we'd query the same interval again and again, which will affect BQ costs

this change won't be needed if we decide to introduce a separate branch for query-based sources

flow/internal/dynamicconf.go
```diff
@@ -281,6 +281,24 @@ var DynamicSettings = [...]*protos.DynamicSetting{
 		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_AFTER_RESUME,
 		TargetForSetting: protos.DynconfTarget_BIGQUERY,
 	},
+	{
+		Name: "PEERDB_BIGQUERY_CDC_SAFETY_LAG_SECONDS",
```
dtunikov, line 285: microsoft fabric documents 10 mins delay [here](https://learn.microsoft.com/en-us/fabric/data-factory/cdc-copy-job-bigquery)
but I couldn't find anything about it in the official BQ docs + local testing didn't reveal any constant delay.

flow/connectors/bigquery/qvalue_convert.go
```diff
@@ -1,9 +1,15 @@
 package connbigquery
 
 import (
+	"encoding/base64"
```
dtunikov, line 4: i'll PR more type related e2e tests separately

flow/connectors/bigquery/source.go
```diff
@@ -60,5 +60,91 @@ func (c *BigQueryConnector) ValidateMirrorSource(ctx context.Context, cfg *proto
 		return fmt.Errorf("failed to access staging bucket: %w", exceptions.NewBigQueryError(err))
 	}
 
+	// snapshot-only mirrors never touch CDC, so no need to validate mode-specific requirements below
+	if cfg.DoInitialSnapshot && cfg.InitialSnapshotOnly {
+		return nil
+	}
+
+	replicationMode := cfg.GetBigqueryCdcConfig().GetReplicationMode()
```
dtunikov, line 68: I might move this whole validation logic to pkg to make it available to other services

flow/activities/flowable_core.go
```diff
@@ -278,12 +281,23 @@ func pullAndSyncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDC
...
 		); err != nil {
 			return nil, fmt.Errorf("failed to sync schema: %w", err)
 		}
 
-		return nil, a.applySchemaDeltas(ctx, config, recordBatchSync.SchemaDeltas)
+		if err := a.applySchemaDeltas(ctx, config, recordBatchPull.SchemaDeltas); err != nil {
+			return nil, err
+		}
+
+		lastCheckpoint := recordBatchPull.GetLastCheckpoint()
```
dtunikov: for BQ it's important to not scan the same period twice as it causes additional costs on the customer side
so, even if we've got zero records, we should advance the offset

flow/activities/flowable_core.go
```diff
@@ -187,6 +187,9 @@ func pullAndSyncCore[TPull connectors.CDCPullConnectorCore, TSync connectors.CDC
 		return nil, fmt.Errorf("failed to get CDC channel buffer size: %w", err)
```
dtunikov, line 187: for BigQuery it's important to advance the checkpoint even when no records were received
otherwise we'd query the same interval again and again, which will affect BQ costs

this change won't be needed if we decide to introduce a separate branch for query-based sources

flow/connectors/bigquery/bigquery.go
```diff
@@ -52,8 +52,14 @@ func NewBigQueryServiceAccount(bqConfig *protos.BigqueryConfig) (*utils.GcpServi
 }
 
 type BigQueryConnector struct {
+	// lastPollAt is the wall-clock time PullRecords last actually queried BigQuery.
+	lastPollAt time.Time
+	// droppedExcludeColumns remembers, per source table, excluded columns that
+	// BigQuery has reported as no longer existing, so later polls stop asking BigQuery
+	// to EXCEPT them.
+	droppedExcludeColumns map[string]map[string]struct{}
```
dtunikov, line 60: needed since we explicitly provide the list of excluded column in BQ queries:
`SELECT a, b EXCEPT d FROM table ...`
ofc, we could also exclude on the client, but it means more network traffic + customer would have to pay for each scanned byte on BQ side.

dtunikov: later I will PR schema evolution separately to not overload this PR
there we could update the list of excluded columns in the catalog after the first occurrence of such error

flow/connectors/bigquery/bigquery.go
```diff
@@ -52,8 +52,14 @@ func NewBigQueryServiceAccount(bqConfig *protos.BigqueryConfig) (*utils.GcpServi
 }
 
 type BigQueryConnector struct {
+	// lastPollAt is the wall-clock time PullRecords last actually queried BigQuery.
+	lastPollAt time.Time
```
dtunikov, line 56: will disappear in the next [PR](https://github.com/PeerDB-io/peerdb/pull/4727/changes), so no need to review this lastPollAt logic thoroughly.

flow/connectors/bigquery/cdc.go
```diff
@@ -0,0 +1,546 @@
...
+			DestinationTableName: nameAndExclude.Name,
+		}); err != nil {
+			return 0, err
+		}
+	}
+}
+
+// recordItemsApproxBytes estimates the storage size of one converted row
+func recordItemsApproxBytes(items model.RecordItems) (int64, error) {
+	b, err := items.MarshalJSON()
```
dtunikov, line 286: this is the most expensive operation in the whole bq cdc pipeline in terms of CPU (i guess similar to mongo?)
another option could be to use `TotalBytesProcessed` from BQ job stats (every query has associated job in BQ). But `TotalBytesProcessed` is not always going to be equal to how many bytes we transferred via PeerDB. So, I'd `MarshalJSON` in the first iteration and see if it really is a bottleneck for customer on prod.

flow/activities/flowable_core.go
jgao54: we also periodically persist/advance on an idle stream for other connectors, but logic is implemented inside PullRecord since empty batch stays open until at least one record is received, so for consistency this could be moved to inside PullRecord.

flow/internal/dynamicconf.go
```diff
@@ -281,6 +281,24 @@ var DynamicSettings = [...]*protos.DynamicSetting{
...
+		TargetForSetting: protos.DynconfTarget_BIGQUERY,
+	},
+	{
+		Name: "PEERDB_BIGQUERY_CDC_MAX_QUERY_WINDOW_SECONDS",
+		Description: "BigQuery CDC only: caps how much time a single APPENDS()/CHANGES() poll can cover, " +
+			"bounding one BigQuery job's row-scan cost even if a mirror falls far behind",
+		DefaultValue:     "86400",
+		ValueType:        protos.DynconfValueType_INT,
+		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_IMMEDIATE,
+		TargetForSetting: protos.DynconfTarget_BIGQUERY,
```
jgao54: this way TargetForSetting is used is essentially "which _destination_ feature does this belong to", not source, so setting this to BQ here will filter it out. so should be `DynconfTarget_ALL` here (as well as other BQ settings above)

flow/connectors/bigquery/qvalue_convert.go
jgao54, line 124: nit: could be shortened:

```
var bqTypeKinds = map[bigquery.FieldType]struct{ scalar, array types.QValueKind }{
    bigquery.StringFieldType:  {types.QValueKindString, types.QValueKindArrayString},
    bigquery.BytesFieldType:   {types.QValueKindBytes, types.QValueKindArrayString},
    bigquery.IntegerFieldType: {types.QValueKindInt64, types.QValueKindArrayInt64},
    // ...
}

func BigQueryTypeToQValueKind(f *bigquery.FieldSchema) types.QValueKind {
    kinds, ok := bqTypeKinds[f.Type]
    if !ok {
        return types.QValueKindInvalid
    }
    if f.Repeated {
        return kinds.array
    }
    return kinds.scalar
}
```

flow/connectors/bigquery/cdc.go
```diff
@@ -0,0 +1,546 @@
...
+func (c *BigQueryConnector) PullRecords(
+	ctx context.Context,
+	catalogPool shared.CatalogPool,
+	otelManager *otel_metrics.OtelManager,
+	req *model.PullRecordsRequest[model.RecordItems],
+) error {
+	defer req.RecordStream.Close()
+	// Recorded regardless of how this call returns below (including the
+	// "nothing new to scan yet" early return) so pacing always advances.
+	defer func() { c.lastPollAt = time.Now() }()
```
jgao54, line 102: nit: iiuc, there's a subtle difference here between how sync interval is respected in BQ vs other connectors.

Given in BQ the records are polled at the end of the sync interval, and tables are iterated synchronously, the actual processing time grows with number of records, so the true sync interval is more like sync interval + process time (and process time is dynamic based on number of records). To match closer to the semantic of other connectors, could be better to record after `waitForNextPoll` at the start of processing, rather than on exit.

flow/connectors/bigquery/cdc.go
jgao54: for mongo (and other connectors) for bytes processed, the raw byte length is used so no serialization is needed. Cheaper way would be to a local hand-rolled estimator (similar to how we estimate inside avro writer [here](https://github.com/PeerDB-io/peerdb/blob/6b52fb6cd6db6226c2678e7a522799b317d9ca14/flow/model/qvalue/avro_converter.go#L207)). An even simpler way would be to measure the wired bytes, since Google BQ API is just HTTP request/response, can pass in an http client that computes the wired bytes to the BQ client, and then measure the wired bytes delta before/after every `pullTable`. e.g.

```
// create an http client
httpClient, _, _ := htransport.NewClient(ctx, option.WithAuthCredentials(creds))
client, _ := bigquery.NewClient(ctx, project, option.WithHTTPClient(httpClient))

// create the wrapper for Transport (`RoundTripper` interface)
transport := &meteredTransport{base: httpClient.Transport}
httpClient.Transport = transport

// implement `RoundTripper` and accumulate response bytes that can be used to measure the wired bytes delta
```

dtunikov: oh, yeah, that is a good idea!
opened a separate PR to not overload this one:
https://github.com/PeerDB-io/peerdb/pull/4724

flow/connectors/bigquery/cdc.go
dtunikov: lastPollAt logic will disappear in the next [PR](https://github.com/PeerDB-io/peerdb/pull/4727/changes), so no need to review this lastPollAt logic thoroughly.

flow/connectors/bigquery/cdc.go
```diff
@@ -0,0 +1,555 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to parse BigQuery CDC checkpoint %q: %w", req.LastOffset.Text, err)
+	}
+
+	now, err := c.currentBigQueryTimestamp(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get current BigQuery timestamp: %w", err)
+	}
+
+	safetyLag, err := internal.PeerDBBigQueryCDCSafetyLag(ctx, req.Env)
```
ilidemi: Do we expect the default to work for all customers and the config basically be a break-glass tool? Or could there be a need to make this a pipe/flow config somewhere in advanced settings

flow/connectors/bigquery/cdc.go
```diff
@@ -0,0 +1,555 @@
...
+		req.RecordStream.SignalAsEmpty()
+	}
+
+	trace.SpanFromContext(ctx).SetAttributes(
+		attribute.Int64(otel_metrics.RowsInBatchKey, int64(recordCount)),
+		attribute.Int64(otel_metrics.BytesPulledKey, bytesProcessed),
+	)
+	c.logger.Info("[bigquery] PullRecords polled window",
+		slog.Time("start", checkpoint), slog.Time("end", upper),
+		slog.Int("records", recordCount), slog.Int64("bytes", bytesProcessed))
```
ilidemi, line 206: Nit: please add channelLen (to have telemetry for bottlenecks) and log end-start as elapsedMinutes float. slog.Time logs nanoseconds so it's less intuitive to skim when reading logs

flow/connectors/bigquery/cdc.go
```diff
@@ -0,0 +1,555 @@
...
+			for i, field := range it.Schema {
+				qfields[i] = BigQueryFieldToQField(field)
+			}
+			changeCols = locateBigQueryChangeColumns(it.Schema)
+		}
+
+		// _CHANGE_TIMESTAMP is APPENDS()'s own commit-time signal for the row;
+		// used as this record's CommitTimeNano. Falls back to the poll window's
+		// end if, unexpectedly, the column isn't present.
+		commitTimeNano := end.UnixNano()
```
ilidemi: Should commit time default be conservative and default to start time? I know this specific value is not used for latency tracking and the changeTimestamp is expected to be present, but just to reduce confusion

flow/connectors/bigquery/bigquery.go
ilidemi: Keep in mind the columns can come back too, so client-side filtering would be needed as well.

Also, maybe more of a product note: excluding columns in a way that we don't even see the data was an often requested feature in other source types, so EXCEPT here works great, but seems if we advertise it we'd need to caveat that it only works as long as the column is already present at the first batch and doesn't disappear-reappear. If it becomes an important first-party scenario, maybe could have some option to always query the known columns and not support column adds. Couldn't think of a way to have SELECT * EXCEPT fully work here, as even if we always do the elimination loop, an excluded column could get re-added in between retries and we'd receive the bytes.

flow/connectors/bigquery/cdc.go
```diff
@@ -0,0 +1,555 @@
...
+			record = &model.InsertRecord[model.RecordItems]{
+				BaseRecord: baseRecord, Items: items,
+				SourceTableName: sourceTableIdentifier, DestinationTableName: nameAndExclude.Name,
+			}
+		case bigQueryChangeTypeUpdate:
+			record = &model.UpdateRecord[model.RecordItems]{
+				BaseRecord: baseRecord, NewItems: items,
+				SourceTableName: sourceTableIdentifier, DestinationTableName: nameAndExclude.Name,
+			}
+		default: // bigQueryChangeTypeDelete, not flagged for update: a genuine delete
```
ilidemi: Nit: exhaustive switch with default error?

flow/connectors/bigquery/qvalue_convert.go
```diff
@@ -88,66 +94,60 @@ func qValueKindToBigQueryType(columnDescription *protos.FieldDescription, nullab
...
+// numericRoundingScale fallback to default scale if precision and scale are not set.
+func numericRoundingScale(qfield types.QField) int16 {
+	if qfield.Precision == 0 && qfield.Scale == 0 {
+		switch bigquery.FieldType(qfield.OriginalType) {
+		case bigquery.NumericFieldType:
+			return bigquery.NumericScaleDigits
+		case bigquery.BigNumericFieldType:
+			return bigquery.BigNumericScaleDigits
+		default:
+			return datatypes.PeerDBBigQueryScale
```
ilidemi: Do we expect this branch to be hit? If it's not exercised today, better have it as an error we'd get notified about

flow/connectors/bigquery/qvalue_convert.go
```diff
@@ -88,66 +94,60 @@ func qValueKindToBigQueryType(columnDescription *protos.FieldDescription, nullab
...
+var bqTypeKinds = map[bigquery.FieldType]struct{ scalar, array types.QValueKind }{
+	bigquery.StringFieldType:     {types.QValueKindString, types.QValueKindArrayString},
+	bigquery.BytesFieldType:      {types.QValueKindBytes, types.QValueKindArrayString},
+	bigquery.IntegerFieldType:    {types.QValueKindInt64, types.QValueKindArrayInt64},
+	bigquery.FloatFieldType:      {types.QValueKindFloat64, types.QValueKindArrayFloat64},
+	bigquery.BooleanFieldType:    {types.QValueKindBoolean, types.QValueKindArrayBoolean},
+	bigquery.TimestampFieldType:  {types.QValueKindTimestamp, types.QValueKindArrayTimestamp},
+	bigquery.DateTimeFieldType:   {types.QValueKindTimestamp, types.QValueKindArrayTimestamp},
+	bigquery.DateFieldType:       {types.QValueKindDate, types.QValueKindArrayDate},
+	bigquery.TimeFieldType:       {types.QValueKindTime, types.QValueKindArrayString},
```
ilidemi: Seems to be a miss that we don't have an QValueKindArrayTime. Mind adding it for BQ and creating a ticket for other connectors?

flow/connectors/bigquery/qvalue_convert.go
```diff
@@ -88,66 +94,60 @@ func qValueKindToBigQueryType(columnDescription *protos.FieldDescription, nullab
...
+	bigquery.IntervalFieldType:   {types.QValueKindInterval, types.QValueKindArrayInterval},
+	bigquery.RangeFieldType:      {types.QValueKindString, types.QValueKindArrayString},
+}
+
 // BigQueryTypeToQValueKind converts a bigquery.FieldType to a QValueKind
 func BigQueryTypeToQValueKind(fieldSchema *bigquery.FieldSchema) types.QValueKind {
-	switch fieldSchema.Type {
-	case bigquery.StringFieldType:
+	if fieldSchema.Type == bigquery.RecordFieldType {
+		// Preserve field names and values as JSON text in STRING or ARRAY<STRING> values.
```
ilidemi: Did we clear it with Product that JSON strings are desired here? BQ records are typed, CH has Nested, so seems like our internal bottleneck

flow/connectors/bigquery/qvalue_convert.go
```diff
@@ -201,11 +201,435 @@ func qValueKindToBigQueryTypeString(columnDescription *protos.FieldDescription,
...
+	default:
+		return value, nil
+	}
+}
+
+func bigQueryRangeString(value *bigquery.RangeValue) string {
+	return fmt.Sprintf("[%s, %s)", bigQueryRangeBoundString(value.Start), bigQueryRangeBoundString(value.End))
+}
+
+func bigQueryRangeBoundString(value bigquery.Value) string {
```
ilidemi, line 618: Please add a subissue to [DBI-295](https://linear.app/clickhouse/issue/DBI-295/type-modeling-asks) or augment DBI-296 so we pretty it up at the same time

flow/connectors/bigquery/source.go
```diff
@@ -60,5 +60,91 @@ func (c *BigQueryConnector) ValidateMirrorSource(ctx context.Context, cfg *proto
...
+
+	for _, tableMapping := range cfg.TableMappings {
+		dstDatasetTable := dstDatasetTables[tableMapping.SourceTableIdentifier]
+		table := c.client.DatasetInProject(c.projectID, dstDatasetTable.dataset).Table(dstDatasetTable.table)
+		metadata, err := table.Metadata(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get metadata for table %s: %w", tableMapping.SourceTableIdentifier, err)
+		}
+		hasPK := tableHasPrimaryKey(metadata)
+		destinationHasOrderingKey := hasPK || tableHasOrderingKey(tableMapping)
```
ilidemi, line 101: This seems like it should be aligned with CH validations. Or you think it's better at the source here? https://github.com/PeerDB-io/peerdb/blob/ba4ea5838a405ad8866e620db0dbeaf25fbebcbb/flow/pkg/clickhouse/validation.go#L317

flow/internal/dynamicconf.go
ilidemi: Seems there was a 10 min restriction for CHANGES that got removed with GA: https://www.linkedin.com/feed/update/urn:li:activity:7488597379827445760/

For appends though, the outcome of underestimating this would be silently missing rows, right? Would it make sense to make it user-configurable somewhere in advanced settings? Also wonder if there's any telemetry we could insert to track this, even at a higher cost during private preview

Edit: reading further, seems CHANGES also doesn't document any guarantees wrt end_timestamp coverage, just that you can pass null and not get an error

flow/connectors/bigquery/cdc.go
ilidemi: Haven't looked at the further PRs yet, but would be cool to also report records/bytes/channelLen in ongoing manner like in other connectors. This helps investigate tickets about a single batch taking long and OOMs

jgao54 (approved):

jgao54:
ah, didn't see the new review from @ilidemi  coming in, just approving what I have reviewed so far. Please make sure to address the rest.

dtunikov:
@ilidemi regarding [this](https://github.com/PeerDB-io/peerdb/pull/4707#discussion_r3847361394)
Yep, I've left a comment in the next PR about the same thing:
https://github.com/PeerDB-io/peerdb/pull/4727/changes#r3862432408
I agree that it'd make sense to make safety_lag/max_query_window (ofc with some reasonable default value)
As I mentioned in the comment above, we could even have this logic available for all future query-based sources. I think this might pop up for other DWHs as well.

flow/connectors/bigquery/cdc.go
dtunikov: oh yeah, I'll make sure we have all necessary logs in the follow-up PR
this PullRecords function is removed anyways in the follow-up in favor of PullTableRecords for isolated tables CDC

dtunikov:
@ilidemi regarding `EXCEPT`.
I put a PR to select columns explicitly instead of relying on `*` with EXCEPT.
https://github.com/PeerDB-io/peerdb/pull/4740

flow/connectors/bigquery/qvalue_convert.go
dtunikov: nope, it shouldn't be hit
removed and added a fallback to error

flow/connectors/bigquery/qvalue_convert.go
dtunikov: https://github.com/PeerDB-io/peerdb/pull/4707/commits/2f93adcfdc49f5769458e0ae48797d30dcaaa797

+ created a ticket for other connectors

flow/connectors/bigquery/source.go
dtunikov: oh, yeah, makes sense
removed this branch from the source

flow/connectors/bigquery/qvalue_convert.go
ilidemi: Tuple would be the best typed one, you're right

dtunikov: changed to native JSON for Struct and ArrayString for repeated Struct (since we don't support Array(JSON) yet in peerdb)

ilidemi (approved):

PR #4708, opened 2026-08-19
SyncFlow graceful termination handling

jgao54:
PR can be reviewed commit-by-commit:

1. As part of introducing graceful termination during worker shutdown, we want to make sure that context cancellation are immediately processed and not delayed due to time.sleep pausing execution

2. Adding a Temporal replay test coverage to catch code changes that could lead to indeterminism error during graceful termination. We already have a test case for `running_sync_finish` which is the happy path; this `running_sync_error` is added specifically to address pipe that are `RUNNING` but SyncFlow is [erroring out](https://github.com/PeerDB-io/peerdb/blob/113ffc792a9457434f5792c811cf15d4b0206118/flow/workflows/cdc_flow.go#L865-L894).

3. adds a `WorkerStopTimeout` of 30s (so worker doesn't shutdown immediately on `SIGTERM`) and a goroutine in SyncFlow that watches worker stop channel for worker stopping and triggers context cancellation. With this change, worker with running pipes can shutdown gracefully; workflow that already processed the pause signal (today's case during upgrade) doesn't have running activities, so will shutdown immediately, effectively same behavior as not setting a `WorkerStopTimeout` today.

Fixes: DBI-1050

Testing: 
- test service upgrade without maintenance worker with this graceful termination code, observed logs for `worker stopping, draining sync flow` and `sync flow drained for worker shutdown`. 
- test service upgrade with start-/end-maintenance, observed `sync canceled` when StartMaintenance workflow paused the pipe, and the graceful termination path did not get triggered.

### DIFF
--- a/flow/activities/flowable_core.go
+++ b/flow/activities/flowable_core.go
@@ -841,21 +841,18 @@ func (a *FlowableActivity) normalizeLoop(
 			normalizingBatchID.Store(reqBatchID)
 			if err := a.startNormalize(ctx, config, reqBatchID, normalizeResponses); err != nil {
 				_ = a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
-				for {
-					// update req to latest normalize request & retry
-					select {
-					case <-syncDone:
-						logger.Info("[normalize-loop] syncDone closed before retry")
-						return
-					case <-ctx.Done():
-						logger.Info("[normalize-loop] context closed before retry")
-						return
-					default:
-						time.Sleep(retryInterval)
-						retryInterval = min(retryInterval*2, 5*time.Minute)
-						reqBatchID = normalizeRequests.Load()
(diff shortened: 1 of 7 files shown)


flow/activities/flowable_core.go
```diff
@@ -841,21 +841,18 @@ func (a *FlowableActivity) normalizeLoop(
 			normalizingBatchID.Store(reqBatchID)
 			if err := a.startNormalize(ctx, config, reqBatchID, normalizeResponses); err != nil {
 				_ = a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
-				for {
```
jgao54, line 844: the inner for-loop here is intentionally removed because it was never actually meaningful, none of the cases below would hit it.

ilidemi (approved):

PR #4709, opened 2026-08-19
remove dead code in cdcflowworkflow and reduce replay surface area

jgao54:
Removed:
- `state.SyncFlowOptions.NumberOfSyncs` is a deprecated feature from and can actually be fully removed from the proto
- migrate postgres table oid was a one-time migration when we introduced table addition/removal (and been a no-op since) so can be removed
- remove no longer used `SetQueryHandler`

All these changes in CDCFlowWorkflow are safe to remove today because they do not touch the pause-resume path (`handlePaused` already doesn't require flow metadata, and `SetQueryHandler` doesn't introduce a history event). The existing replay history tests have broader coverage and are failing because it enforces replay determinism in the other paths as well. 

Testing:
- ran smoke test in staging to verify that pipes are not impacted by this upgrade.

Fixes: DBI-1051

(edit: initially I also moved GetFlowMetadataContext logic inside SyncFlow in the startCDC path to reduce replay surface area for RUNNING pipes, but decided against this optimization for now. Given the flow metadata context impact log and metrics, having all the metadata fetched upfront in a single place makes the code easier to reason about, and remove the risk of drifts across code paths over time.)


jgao54:
Looking at [failed tests](https://github.com/PeerDB-io/peerdb/actions/runs/32318501249/job/96275780205?pr=4709) so far, on all the jobs we have the following test cases failed:
- running_paused
- running_sync_finished
- running_terminate

This is expected, also confirmed that `parked_resume` and `parked_config_update` didn't fail, since those are the ones that matter for upgrades when pipe is parked in paused state.

Safe to update replay fixtures to reflect the latest change (generated against current code).

issue Claude review found below is benign because of this.

flow/workflows/qrep_flow.go
```diff
@@ -440,12 +440,6 @@ func setWorkflowQueries(ctx workflow.Context, state *protos.QRepFlowState) error
 		return fmt.Errorf("failed to set `%s` query handler: %w", shared.QRepFlowStateQuery, err)
 	}
 
-	// Support a Query for the current status of the qrep flow.
-	_ = workflow.SetQueryHandler(ctx, "q-flow-status", func() (protos.FlowStatus, error) {
-		// no longer used, handler kept to avoid nondeterminism
```
jgao54, line 445: workflow.SetQueryHandler (or the act of issuing querying) is a purely worker-local registration — it does not add to event history.

ilidemi (commented):
Re OID migration, the caveat with those is when someone is upgrading from an old version, we'd still want it to run. Plus we could see more proto migrations, like the one with supporting a lot more table mappings by moving them out of Temporal contracts. The pattern for DB migrations is an accumulating list of sql's, and it would be ideal to have these there too, but so far it seems they need to be in Go code, and having a figured-out pattern to easily add new migrations is valuable IMO. Could even change the activity to "run all migrations" and keep the code itself tucked away somewhere.

jgao54:
agree it's worth figuring out a pattern for migrating from proto -> db ; but in this case i think the solution is a bit simpler. looking into lazily backfill this from source for table addition cancellation.

flow/activities/cancel_table_addition_activity.go
```diff
@@ -461,3 +471,69 @@ func (a *CancelTableAdditionActivity) getRunIDOfLatestRunningPeerFlow(ctx contex
...
+	ctx context.Context,
+	flowJobName string,
+	sourcePeerName string,
+	sourcePeerType protos.DBType,
+	tableMappings []*protos.TableMapping,
+	tableSchemas map[string]*protos.TableSchema,
+) error {
+	if sourcePeerType != protos.DBType_POSTGRES {
+		return nil
+	}
```
jgao54, line 485: this could have been moved to earlier, to make `GetTableOIDsFromCatalog` activity itself only applicable to PG, since we are logging `"Retrieved PostgreSQL table OIDs"` for non-PG cancellations as well; but left as is for this PR, given it's a bit out-of-scope

ilidemi (approved):
The solution is simple indeed

PR #4710, opened 2026-08-20
crdb stack/6 resolved buffering

viragtripathi:
Follow-up to #4669, implementing the buffering discussed on the dedupe thread there and on Slack.

Changefeed events now wait in an in-memory buffer keyed by commit timestamp and row key. When a resolved timestamp arrives, everything it covers is emitted in commit order and the checkpoint advances to that same value, so nothing handed to the stream ever lies past the persisted checkpoint. Batch boundaries stop producing duplicates: records still buffered when a batch ends are dropped and re-read by the next batch from the cursor. Session reconnect replays land on the same buffer keys and collapse.

The buffer is bounded in bytes computed from the raw changefeed message sizes, not record count, since single events can be large. Crossing the cap (128 MiB) or a resolved stream that stalls past the batch deadline force-emits the buffer past the checkpoint with a warning, which is exactly the pre-existing at-least-once behavior, so the degraded mode is the previously reviewed semantics. Delivery overall remains at least once: a crash between emit and checkpoint persist still replays.

Validation: new unit tests for the replay-collapse, partial-emit-at-resolved and force-emit ordering paths, plus a new e2e test (Test_CDC_Exactly_Once_Across_Batches) asserting 82 events arrive exactly once across two consecutive pulls. Full suite run against a 5-node v26.2.5 cluster behind haproxy, 17/17 passing, and unit suites green.

### DIFF
--- a/flow/connectors/cockroachdb/cdc.go
+++ b/flow/connectors/cockroachdb/cdc.go
@@ -99,11 +99,18 @@ func (c *CockroachDBConnector) PullFlowCleanup(ctx context.Context, jobName stri
 // changefeedPullState carries the per-batch state shared between changefeed
 // sessions (a session being one sinkless changefeed query on one connection).
 //
-// A session reconnect replays every message since the last resolved timestamp,
-// so rows delivered before the failure point can be emitted again within one
-// PullRecords batch. Those duplicates are deliberately passed through: PeerDB
-// guarantees at-least-once delivery and destinations already absorb replays,
-// so deduplicating here would only duplicate that responsibility.
+// Converted records are buffered keyed by commit timestamp and row key, and
+// only handed to the record stream once a resolved timestamp at or beyond
+// their commit timestamp arrives; the checkpoint then advances to that same
+// resolved timestamp. Nothing emitted therefore ever lies past the persisted
+// checkpoint, so neither batch boundaries nor session reconnects re-deliver
+// records on the happy path (a reconnect's replays land on the same buffer
+// keys and collapse). Records still buffered when the batch ends are dropped,
+// not emitted: the next batch re-reads them from the cursor, trading a little
+// re-fetch work for exactly-once emission across batches. Delivery overall
+// stays at least once: a crash between emitting and persisting the checkpoint
+// replays, and force-emits under memory pressure (see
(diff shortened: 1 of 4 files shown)


flow/connectors/cockroachdb/cdc.go
```diff
@@ -122,6 +132,35 @@ type changefeedPullState struct {
...
+	record model.Record[model.RecordItems]
+	ts     crdbHLC
+	seq    uint64
+	bytes  int64
+}
+
+// stashRecord buffers a converted record under its commit timestamp and row
+// key. A replayed message (same timestamp, table and key) overwrites its
+// earlier copy instead of queueing a duplicate.
+func (state *changefeedPullState) stashRecord(
```
jgao54, line 148: ah this is neat, since it also dedupes duplicated messages send from server-side retries

flow/connectors/cockroachdb/cdc.go
```diff
@@ -258,6 +298,14 @@ func (c *CockroachDBConnector) PullRecords(
 			attribute.Int64(otel_metrics.RowsInBatchKey, int64(state.recordCount)),
 			attribute.Int64(otel_metrics.BytesPulledKey, state.totalBytes.Load()),
 		)
+		if len(state.buffer) > 0 {
+			// no data loss: nothing buffered was emitted, so the persisted
+			// cursor still precedes these records and the next batch re-reads
+			// them from the changefeed
+			c.logger.Info("[cockroachdb] leaving records past the final resolved timestamp for the next batch",
```
jgao54: nit: this logline would get a bit noisy as it would be emitted every PullRecord on a busy pipe, can be removed or emit at debug level

jgao54 (approved):
one small nit, otherwise lgtm!

flow/connectors/cockroachdb/cdc.go
viragtripathi: Moved it to debug. And thanks for the review, and for all of them across the stack, the connector came out a lot better for how thorough you've been.

flow/connectors/cockroachdb/cdc.go
viragtripathi: Yeah that fell out of keying the buffer by timestamp and row key, reconnect and server side replays just land on the same slot.

PR #4712, opened 2026-08-20
feat(bigquery): isolate CDC table polling

dtunikov:
## Summary

- Store BigQuery CDC progress per source table in a versioned JSON checkpoint.
- Keep polling healthy tables when one table fails before emitting records. The failed table keeps its last confirmed cursor and reports one `LogFlowError` per failure episode.
- Add `RUNNING`, `PARTIAL`, and `COMPLETED` batch states. Partial batches show their completed-table count and lagging source tables in the mirror UI.
- Add an end-to-end test that drops one BigQuery source table and proves that its sibling continues CDC.
- Add an end-to-end test that removes a table from the mirror mid stream and check that other tables continue replication without issues + that the checkpoint doesn't contain the removed table anymore.

## UI/API changes

cdc batches handler now returns such result:
```
{
    "cdcBatches": [
        {
            "startLsn": "0",
            "endLsn": "0",
            "numRows": "1",
            "startTime": "2026-08-20T14:16:48.405186Z",
            "endTime": "2026-08-20T14:16:54.846275Z",
            "batchId": "1",
            "status": "CDC_BATCH_STATUS_PARTIAL",
            "tablesCompleted": 1,
            "tablesTotal": 2,
            "laggingTables": [
                "do_not_remove_peerdb_source_test_dataset.a_cdc_isolation_failed_bqch_8v9fg4zx"
            ]
        }
    ],
    "total": 1,
    "page": 1
}
```

The UI renders it as:
image

## Flow

```mermaid
flowchart TD
    A[Load the per-table checkpoint] --> B{Next mapped table}
    B --> C[Poll from the table's synced_through timestamp]
    C -->|Success| D[Advance this table and emit its records]
    C -->|Failure before records| E[Keep synced_through, record the attempted target, and call LogFlowError]
    D --> B
    E --> B
    B -->|All tables attempted| F[Send the mixed stream to the destination]
    F -->|Destination sync succeeds| G[Persist the new checkpoint JSON]
    F -->|Destination sync fails| H[Persist none of the new table progress]
    G --> I[Derive batch table progress]
    I --> J[Show Partial with the lagging-table list]
```

Each table owns its source cursor. Destination sync remains the shared confirmation boundary, so a destination failure cannot commit source progress ahead of durable records.

If every attempted table fails, `PullRecords` returns the combined error. A failure after a table has emitted records remains flow-wide because retrying only part of that table's window could lose or duplicate the emitted portion.

## Scope

This change isolates BigQuery source-table polling failures. Destination normalization and backpressure still use their existing global watermarks.

Resolves: DBI-1045


flow/connectors/bigquery/cdc.go
```diff
@@ -151,38 +151,82 @@ func (c *BigQueryConnector) PullRecords(
...
 		tableBytesProcessed, err := pullTable(
-			ctx, sourceTableIdentifier, req.TableNameMapping[sourceTableIdentifier], checkpoint, upper, addRecord,
+			ctx, sourceTableIdentifier, req.TableNameMapping[sourceTableIdentifier], start, upper, addRecord,
 		)
 		if err != nil {
-			return err
+			if ctx.Err() != nil {
+				return ctx.Err()
+			}
+			if recordCount > recordCountBeforeTable {
```
dtunikov, line 187: this is only necessary if we want to fail loudly when we encounter an error after writing partial result to the channel

flow/connectors/bigquery/cdc_checkpoint.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+
+	if sourceTables != nil {
+		checkpoint.retainAndInitializeTables(sourceTables)
+	}
+	return &checkpoint, nil
+}
+
+func (c *bigQueryCDCCheckpoint) retainAndInitializeTables(sourceTables []string) {
```
ilidemi: I don't fully understand the fallback logic here. Can it apply to tables during their normal operation, or only table additions/manual state edits? If it's basically init logic for brand new tables, there is probably a moment when we can write time.Now() and be explicit about it. If it's a fallback-fallback, there should be an error/warn log that it happened. There shouldn't be code that both looks like it can silently lose data and is hard to reason about

dtunikov: Yes, you're right, it's rather confusing. 
It was meant to handle table additions: when a new table is added we used the oldest SyncedThrough as its initial checkpoint. 
But I think that `now()` makes more sense for such cases. (we'd have a gap between the moment user added a table and a moment the pipe was resumed, but it sounds fine).
https://github.com/PeerDB-io/peerdb/pull/4712/commits/5d5195afc961a96c2f61a89dc3dc571c98de1fd0

dtunikov: closing this PR in favor of a separate `SyncFlow` branch for query-based cdc.

PR #4715, opened 2026-08-20
support legacy 0-byte geometry data

jgao54:
MySQL 5.7 (and MariaDB) can create zero-byte values for a NOT NULL GEOMETRY column when the column is omitted on insert:
```
mysql> CREATE TABLE `geo` (`id` int primary key, `g` geometry NOT NULL);
Query OK, 0 rows affected (0.02 sec)

mysql> insert into geo (id) values (1);
ERROR 1364 (HY000): Field 'g' doesn't have a default value

mysql> insert ignore into geo (id) values (1);
Query OK, 1 row affected, 1 warning (0.00 sec)

mysql> show warnings;
+---------+------+----------------------------------------+
| Level   | Code | Message                                |
+---------+------+----------------------------------------+
| Warning | 1364 | Field 'g' doesn't have a default value |
+---------+------+----------------------------------------+
1 row in set (0.00 sec)

mysql> select * from geo;
+----+---+
| id | g |
+----+---+
|  1 |   |
+----+---+

```

These values can persist after an upgrade to MySQL 8.x and cause snapshot failures when converting geometry.

Because geometry is propagated as a string in ClickHouse, we can be more lenient here and add support for this. Note that this doesn't fix other destination types where data is propagated as Geometry (e.g. MySQL-> PG, MySQL -> Snowflake), but it is already broken for those cases today so this is not introducing a regression. 

Testing: add unit test; locally test against a mysql 5.7 instance, repro-ed the error,  and after the fix was able to successfully snapshot and cdc into table with empty string in destination.

Fixes: DBI-1043

### DIFF
--- a/flow/connectors/mysql/qvalue_convert.go
+++ b/flow/connectors/mysql/qvalue_convert.go
@@ -152,6 +152,12 @@ func QRecordSchemaFromMysqlFields(tableSchema *protos.TableSchema, fields []*mys
 // MySQL's internal geometry format is 4-byte SRID (little-endian) followed by standard WKB.
 // SRID is stripped because destinations like ClickHouse don't support EWKT (SRID=N;WKT).
 func processGeometryData(data []byte) (types.QValueGeometry, error) {
+	if len(data) == 0 {
+		// MariaDB and MySQL 5.X store a zero-length value when a NOT NULL geometry column is filled
+		// by an implicit default (e.g. non-strict INSERT, INSERT IGNORE), and such rows survive
+		// upgrades even though they can no longer be created. Map them to an empty string.
+		return types.QValueGeometry{}, nil
+	}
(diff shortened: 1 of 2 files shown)


masterashu (approved):

ilidemi:
@jgao54 0-byte geometry can be reproduced on 5.7 and Maria in our CI, right? Should it be an e2e or integration test instead of crafting bytes in unit?

jgao54:
Yeah, I tested e2e locally, but could be an e2e test. i didn't assuming it's a rare edge case so was trading off between CI time vs. CI coverage.

ilidemi:
Didn't realize we could be treating CI as a limited resource. I would shoot for adding more compute or parallelism to CI rather than making coverage trade-offs

PR #4716, opened 2026-08-21
fix(pg-pg): classify pgdump extension missing

Amogh-Bharadwaj:
If the target Postgres does not have a custom extension installed which is installed on source, the user must switch to manual schema dump instead. 
This PR wires the notification for this case


jgao54 (approved):

jgao54:
just merge conflict, but lgtm

PR #4719, opened 2026-08-21
feat(bigquery): poll CDC source tables in parallel

dtunikov:
Stacked on #4712.

## What

BigQuery CDC polled one source table at a time. A slow or large table blocked every table queued behind it in the same batch, so a single heavy table set the pace for the whole mirror. Table polls now run concurrently.

## How

PulLRecords spins up a goroutine per table with bounded parallelism controlled via mirror settings (or dynamic config as a fallback).
Error reporting behavior stays the same (one failing table doesn't block the others). When all tables have failed -> an error is returned from PullRecords.

Resolves: DBI-1046

### DIFF
--- a/flow/connectors/bigquery/cdc.go
+++ b/flow/connectors/bigquery/cdc.go
@@ -9,11 +9,13 @@ import (
 	"regexp"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"cloud.google.com/go/bigquery"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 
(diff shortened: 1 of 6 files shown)


flow/connectors/bigquery/cdc.go
```diff
@@ -132,33 +208,43 @@ func (c *BigQueryConnector) PullRecords(
...
+
+	parallelism := int(cfg.GetBigqueryCdcConfig().GetQueryCdcTablesParallelism())
+	if parallelism <= 0 {
+		parallelism, err = internal.PeerDBBigQueryCDCTableParallelism(ctx, req.Env)
+		if err != nil {
+			return fmt.Errorf("failed to get BigQuery CDC table parallelism: %w", err)
+		}
+	}
+	if parallelism <= 0 {
+		parallelism = -1 // errgroup treats a negative limit as unlimited
```
dtunikov, line 227: only possible if we set PeerDBBigQueryCDCTableParallelism via peerdb settings to -1 or so
probably won't be needed ever
but still nice to have an option to override it for some large pipes

PR #4722, opened 2026-08-21
fix(mongo): pipeline document decode in CDC PullRecords

itsbilal:
Previously, we serially decoded the change event, then the full document that was changed, inside the PullRecords loop for a given batch. This would leave a lot of CPU underutilized as the bulk of work would happen on just one thread.

This change moves the full document decode part to a different goroutine, and the record publish to another goroutine, which should allow for greater CPU utilization by the decoding function.

Existing tests pass; currently looking into benchmarking this with a fast-changing table on Mongo.

### DIFF
--- a/flow/connectors/mongo/cdc.go
+++ b/flow/connectors/mongo/cdc.go
@@ -6,8 +6,10 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"runtime"
 	"slices"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -16,6 +18,7 @@ import (
 	"go.mongodb.org/mongo-driver/v2/mongo/options"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/PeerDB-io/peerdb/flow/alerting"
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
@@ -230,12 +233,281 @@ func decodeEvent(
 	return nil
 }
 
+// Constants used by PullRecords.
+//
+// Buffered channel size for channels to pass records to decode and send loops (see below).
+// This should ideally be larger than decodeWorkerBufSize.
+const workerBufferedChanSize = 10
+
+// Number of recordItems to pass in one batch to decode/send loops. Doing per-item channel sends
+// results in too much coordination and reduces effective concurrency in practice.
+const pullRecordsItemsBatchSize = 256
+
+// maxNumDecodeWorkers is the maximum number of goroutines that decodeLoop spins up to parallelize
+// decoding of record batches.
+const maxNumDecodeWorkers = 6
+
+// decodeWorkerBufSize is the size of the channel to each decode worker spun up by decodeLoop.
+const decodeWorkerBufSize = 4
+
+// Amount of time to wait for workers to drain before timing out.
+const workerDrainTimeout = 2 * time.Minute
+
+type decodeBatch struct {
+	// base64-encoded resume token
+	resumeToken string
+	items       []recordItems
+}
+
+type sendBatch struct {
+	// base64-encoded resume token
+	resumeToken string
+	// records to send to RecordStream.
+	records []model.Record[model.RecordItems]
+}
+
+// Two additional loops are created by PullRecords, each running in their own goroutines.
+// One is decodeLoop, which takes records from the main PullRecords goroutine through `records`
+// that have already had the operation type decoded, but not the full document. decodeLoop manages
+// decodeWorkers running in their own goroutines that handle the actual bson document parsing. After
+// a decodeWorker has parsed a batch of items, it passes the batch to sendLoop, which assembles records
+// from all decodeWorkers into req.RecordStream. req.RecordStream.AddRecord could block on the downstream
+// channel send, so separating `sendLoop` out this way from `decodeWorker` makes sense to speed up the
+// relatively CPU-heavy `decodeWorker`. `sendLoop` is also responsible for advancing the resume token
+// inside req.RecordStream, except in cases when `sendLoop` has been confirmed to be terminated,
+// in which case `PullRecords` itself can do the advancement.
+//
+// decodeLoop is expected to propagate channel closures through to sendLoop for graceful
+// draining to ensure no records are lost. Context cancellation is only used in the error
+// path, where losing inflight records is acceptable with the expectation that a future
+// restart from the last checkpoint will see those events again. Note that in case of
+// recreation of the upstream changestream, we could drain and restart these two worker loops
+// plus any decodeWorkers within the execution of one `PullRecords`.
+func (c *MongoConnector) sendLoop(
+	ctx context.Context,
+	records <-chan chan sendBatch,
+	errChan chan<- error,
+	req *model.PullRecordsRequest[model.RecordItems],
+	wg *sync.WaitGroup,
+) {
+	defer wg.Done()
+
+	for {
+		select {
+		case recordChan, ok := <-records:
+			if !ok {
+				return
+			}
+			select {
+			case sendBatch := <-recordChan:
+				for i := range sendBatch.records {
+					if err := req.RecordStream.AddRecord(ctx, sendBatch.records[i]); err != nil {
+						select {
+						case errChan <- err:
+						case <-ctx.Done():
+						}
+						return
+					}
+				}
+				if sendBatch.resumeToken != "" {
+					req.RecordStream.UpdateLatestCheckpointText(sendBatch.resumeToken)
+				}
+			case <-ctx.Done():
+				return
+			}
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+type recordItems struct {
+	maybeFullDocument    *bson.Raw
+	operationType        operationType
+	sourceTableName      string
+	destinationTableName string
+	documentKey          bson.Raw
+	commitTimeNanos      int64
+}
+
+// decodeWorker is spun up by decodeLoop in separate goroutines, up to
+// maxNumDecodeWorker in parallel. Batches of items to decode are passed in
+// through recv, and the output is sent through `send` to `sendLoop` directly.
+func (c *MongoConnector) decodeWorker(
+	ctx context.Context,
+	recv <-chan decodeBatch,
+	send chan<- sendBatch,
+	req *model.PullRecordsRequest[model.RecordItems],
+) error {
+	// Utils used by this routine.
+	converter := NewDirectBsonConverter()
+	fullDocumentColumnName := DefaultFullDocumentColumnName
+	if req.InternalVersion < shared.InternalVersion_MongoDBFullDocumentColumnToDoc {
+		fullDocumentColumnName = LegacyFullDocumentColumnName
+	}
+	parseItem := func(item recordItems) (model.Record[model.RecordItems], error) {
+		items := model.NewRecordItems(2)
+
+		if len(item.documentKey) > 0 {
+			rv := item.documentKey.Lookup(DefaultDocumentKeyColumnName)
+			if rv.IsZero() || rv.Type == bson.TypeNull {
+				return nil, exceptions.NewInvalidIdValueError(item.sourceTableName)
+			}
+			qValue, err := converter.QValueStringFromId(rv, req.InternalVersion)
+			if err != nil {
+				return nil, err
+			}
+			items.AddColumn(DefaultDocumentKeyColumnName, qValue)
+		} else {
+			return nil, fmt.Errorf("document key is nil")
+		}
+
+		if item.maybeFullDocument != nil && len(*item.maybeFullDocument) > 0 {
+			qValue, err := converter.QValueJSONFromDocument(*item.maybeFullDocument)
+			if err != nil {
+				return nil, fmt.Errorf("failed to convert document: %w", err)
+			}
+			items.AddColumn(fullDocumentColumnName, qValue)
+		} else {
+			// `fullDocument` field will not exist in the following scenarios:
+			// 1) operationType is 'delete'
+			// 2) document is deleted / collection is dropped in between update and lookup
+			// 3) update changes the values for at least one of the fields in that collection's
+			//    shard key (although sharding is not supported today)
+			items.AddColumn(fullDocumentColumnName, types.QValueJSON{Val: "{}"})
+		}
+		var record model.Record[model.RecordItems]
+		switch item.operationType {
+		case operationTypeInsert:
+			record = &model.InsertRecord[model.RecordItems]{
+				BaseRecord:           model.BaseRecord{CommitTimeNano: item.commitTimeNanos},
+				Items:                items,
+				SourceTableName:      item.sourceTableName,
+				DestinationTableName: item.destinationTableName,
+			}
+
+		case operationTypeUpdate, operationTypeReplace:
+			record = &model.UpdateRecord[model.RecordItems]{
+				BaseRecord:           model.BaseRecord{CommitTimeNano: item.commitTimeNanos},
+				NewItems:             items,
+				SourceTableName:      item.sourceTableName,
+				DestinationTableName: item.destinationTableName,
+			}
+		case operationTypeDelete:
+			record = &model.DeleteRecord[model.RecordItems]{
+				BaseRecord:           model.BaseRecord{CommitTimeNano: item.commitTimeNanos},
+				Items:                items,
+				SourceTableName:      item.sourceTableName,
+				DestinationTableName: item.destinationTableName,
+			}
+		}
+		return record, nil
+	}
+	for {
+		select {
+		case batch, ok := <-recv:
+			if !ok {
+				return nil
+			}
+			items := batch.items
+
+			modelRecords := make([]model.Record[model.RecordItems], len(items))
+			for i := range items {
+				var err error
+				if modelRecords[i], err = parseItem(items[i]); err != nil {
+					return err
+				}
+			}
+
+			select {
+			case send <- sendBatch{records: modelRecords, resumeToken: batch.resumeToken}:
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+}
+
+func (c *MongoConnector) decodeLoop(
+	parentCtx context.Context,
+	recv <-chan decodeBatch,
+	errChan chan<- error,
+	req *model.PullRecordsRequest[model.RecordItems],
+	wg *sync.WaitGroup,
+) {
+	defer wg.Done()
+
+	// NB: child workers are tracked by this errGroup, not by
+	// the parent WaitGroup (shared with PullRecords and sendLoop).
+	eg, ctx := errgroup.WithContext(parentCtx)
+	defer func() {
+		err := eg.Wait()
+		if err != nil {
+			select {
+			case errChan <- err:
+			case <-parentCtx.Done():
+			}
+		}
+	}()
+
+	// This function spins up half as many workers as the number of cores, capped
+	// at maxNumDecodeWorkers.
+	numWorkers := max(1, min(maxNumDecodeWorkers, runtime.GOMAXPROCS(0)/2))
+	workerChan := make([]struct {
+		req chan decodeBatch
+		res chan sendBatch
+	}, numWorkers)
+	for i := range workerChan {
+		workerChan[i].req = make(chan decodeBatch, decodeWorkerBufSize)
+		workerChan[i].res = make(chan sendBatch, decodeWorkerBufSize)
+		// Start worker.
+		eg.Go(func() error {
+			return c.decodeWorker(ctx, workerChan[i].req, workerChan[i].res, req)
+		})
+	}
+
+	// Start up sendLoop. Have a buffered channel in case decoding runs faster
+	// than the downstream addition of records to req.RecordStream.
+	sender := make(chan chan sendBatch, decodeWorkerBufSize*maxNumDecodeWorkers)
+	go c.sendLoop(parentCtx, sender, errChan, req, wg)
+	// Index of next worker to send a result to.
+	nextWorker := 0
+
+	for {
+		select {
+		case item, ok := <-recv:
+			if !ok {
+				// Draining. Close all worker channels.
+				for i := range workerChan {
+					close(workerChan[i].req)
+				}
+				close(sender)
+				return
+			}
+
+			select {
+			case workerChan[nextWorker].req <- item:
+				sender <- workerChan[nextWorker].res
+				nextWorker = (nextWorker + 1) % numWorkers
+			case <-ctx.Done():
+				return
+			}
+		case <-parentCtx.Done():
+			// Context cancellation usually means an error, not just cutting a batch (which would be
+			// a graceful drain via recv getting closed).
+			return
+		}
+	}
+}
+
 func (c *MongoConnector) PullRecords(
 	ctx context.Context,
 	catalogPool shared.CatalogPool,
 	otelManager *otel_metrics.OtelManager,
 	req *model.PullRecordsRequest[model.RecordItems],
-) error {
+) (retErr error) {
 	defer req.RecordStream.Close()
 
 	var alerter *alerting.Alerter
@@ -247,6 +519,7 @@ func (c *MongoConnector) PullRecords(
 	if req.InternalVersion < shared.InternalVersion_MongoDBFullDocumentColumnToDoc {
 		fullDocumentColumnName = LegacyFullDocumentColumnName
 	}
+	var wg sync.WaitGroup
 
 	c.logger.Info("[mongo] started PullRecords for mirror "+req.FlowJobName,
 		slog.Any("table_mapping", req.TableNameMapping),
@@ -324,7 +597,12 @@ func (c *MongoConnector) PullRecords(
 			attribute.Int64(otel_metrics.RowsInBatchKey, int64(recordCount)),
 			attribute.Int64(otel_metrics.BytesPulledKey, cumulativeBytesProcessed.Load()),
 		)
-		if changeStream != nil {
+		if changeStream != nil && retErr == nil {
+			// NB: We don't set the otel metrics ResumeTokenKey in the error case because
+			// it's possible for the changeStream to have been advanced past the last record
+			// send to RecordStream (and therefore the actual resume token set in RecordStream).
+			// It's safer to not report a resume token than to report an incorrect one that can
+			// skip records if used.
 			if rt := changeStream.ResumeToken(); rt != nil {
 				rtStr := base64.StdEncoding.EncodeToString(rt)
 				if len(rtStr) > 64 {
@@ -339,6 +617,15 @@ func (c *MongoConnector) PullRecords(
 			slog.Int("channelLen", req.RecordStream.ChannelLen()),
 			slog.Float64("elapsedMinutes", time.Since(pullStart).Minutes()))
 	}()
+	// Context inheritance tree: There's a parent ctx (always called ctx in this function).
+	// Off of that, we create two child contexts, one for workers (called workerCtx), that's passed
+	// to children workers (eg. decodeLoop). The other, timeoutCtx, is managed by us for timeouts.
+	// Note that when we hit a timeout, we don't want to cancel the workerCtx; we want the workers
+	// to gracefully drain.
+	//
+	// Also note that timeoutCtx is occasionally rewritten, such as when a record in a batch arrives
+	// or when we reset the changestream.
+	workerCtx, workerCtxCancel := context.WithCancel(ctx)
 	// before the first record arrives, we wait for up to an hour before resetting context timeout
 	// after the first record arrives, we switch to configured idleTimeout
 	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, time.Hour)
@@ -351,6 +638,8 @@ func (c *MongoConnector) PullRecords(
 
 	defer func() {
 		cancelTimeout()
+		workerCtxCancel()
+		wg.Wait()
 		reportBytesShutdown()
 		read := deltaBytesProcessed.Swap(0)
 		otelManager.Metrics.FetchedBytesCounter.Add(ctx, read)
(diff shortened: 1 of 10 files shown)


flow/connectors/mongo/cdc.go
```diff
@@ -227,19 +228,177 @@ func decodeEvent(
 	return nil
 }
 
-func (c *MongoConnector) PullRecords(
+// Two additional loops are created by PullRecords, each running in their own goroutines.
+// One is decodeLoop, which takes records from
```
pfcoperez: _One is decodeLoop, which takes records from ..._  has this code comment been cut?

itsbilal: ah yes, my bad - forgot to finish up this comment.

flow/connectors/mongo/cdc.go
```diff
@@ -227,19 +228,189 @@ func decodeEvent(
...
+	recv <-chan recordItems,
+	errChan chan<- error,
+	req *model.PullRecordsRequest[model.RecordItems],
+	wg *sync.WaitGroup,
+) {
+	defer wg.Done()
+
+	// Start up sendLoop. Have a buffered channel in case decoding runs faster
+	// than the downstream addition of records to req.RecordStream.
+	sender := make(chan model.Record[model.RecordItems], 10)
```
pfcoperez: I'd isolate the size magic number in a constant above.

flow/connectors/mongo/cdc.go
```diff
@@ -227,19 +228,191 @@ func decodeEvent(
...
+	ctx context.Context,
+	recv <-chan recordItems,
+	errChan chan<- error,
+	req *model.PullRecordsRequest[model.RecordItems],
+	wg *sync.WaitGroup,
+) {
+	defer wg.Done()
+
+	// Start up sendLoop. Have a buffered channel in case decoding runs faster
+	// than the downstream addition of records to req.RecordStream.
```
pfcoperez: I wonder if a a buffer of size 10 really makes a difference vs a size 1 if the addition back-pressures on the decoding.

Have you been able to play with different values to notice real differences?

flow/connectors/mongo/cdc.go
```diff
@@ -418,6 +566,40 @@ func (c *MongoConnector) PullRecords(
 				slog.Int("channelLen", req.RecordStream.ChannelLen()),
 				slog.Float64("elapsedMinutes", time.Since(pullStart).Minutes()))
 		}
+	}
+	decodeChan := make(chan recordItems, workerBufferedChanSize)
+	errChan := make(chan error, 2) // size = max number of workers
```
pfcoperez: Workers here refer to the two go routines (`decodeLoop` and `sendLoop`), right? I'd specify that in the comment, a diagonal human reader could imagine that we have more than one worker goroutine for decoding or sending.

flow/connectors/mongo/cdc.go
```diff
@@ -418,6 +566,40 @@ func (c *MongoConnector) PullRecords(
...
+	// decodeLoop starts up sendLoop.
+	wg.Add(2)
+	go c.decodeLoop(workerCtx, decodeChan, errChan, req, &wg)
+	const workerDrainTimeout = 2 * time.Minute
```
pfcoperez: nit, magic number that can be put together with the buffer size constant.

pfcoperez (commented):
Aside from the more superficial comments I left in the PR I wanted to confirm this deeper point:

The goal of the PR is to boost event processing performance through the use of parallelism in the so far serial process of:

1. Decoding BSON
2. Sending the decoded record to the record stream (`req.RecordStream.AddRecord(ctx, record)` ) were this action is, in turn, sends the record to another channel. When this final destination channel is full (back-pressure we already logged `waiting on adding record to stream` ) messages.

So I am assuming that the sought performance boost is expected to come from the added ability to decode at most 10 BSON documents when (2) back-pressure kicks in.

**Q1:** Without such back-pressure, both the serial version (prior to the PR) and the one with these changes should take the same time, right? Events are decoded, pushed to the channel and sent as they come. So, in this case, there is no huge  performance boost beyond being able to pull more records will decoding is happening, right?

In this case (no back-pressure from `AddRecord`), the bottleneck could be the decoding part. 

**Q2:** However, with this implementation as I understand it, decoding is still serial (we are not decoding two or more different events at the same time at different go-routines). **Am I understanding this correctly?**

It's true that it adds a buffer `workerBufferedChanSize` that might allow for the continued processing of events in the main loop until the decoding queue grows up to 10 (current constant value).

Then we'd allow for at most other 10 events to be sent  while decoding is working.

**Q3:** But back to point (2) above, there we already had a channel consumed by another go-routine. So, would this change actually improve things in this case? 

**Q4**: And, finally, if the bottleneck  actually happens at `AddRecord` point,  then following a similar reasoning we wouldn't expect noticeable performance boost through parallelism as decoding needs to wait to be able to send to the destination channel, the pipeline becomes serialized.

**Q5**: Would it be easy to add https://pkg.go.dev/net/http/pprof and play with some operation in the local dev env?

Jeremyyang920:
@pfcoperez Just for some more context around this issue, there was a situation that came up while you were out where it took almost 48 hours to drain 24 hours of source side lag so this is what is driving this change.

jgao54:
to answer this question first:
> Q1: Without such back-pressure, both the serial version (prior to the PR) and the one with these changes should take the same time, right? Events are decoded, pushed to the channel and sent as they come. So, in this case, there is no huge performance boost beyond being able to pull more records will decoding is happening, right?

Back-pressure kicks in only after some time (~24 hours), so even if it doesn't kick in, this optimization will reduce latency if the processing is behind but yet triggered back-pressure. The most expensive/slow part is bson deser/ser into QRecord, so by parallelizing deserialization we can saturate cpu better and get some perf gain. 

I'm a bit behind on reviewing this, taking a look now

jgao54:
After taking an initial look, here's my thoughts: 
- pipelining is going to help a bit but not give us quite the perf gain we want.  
- what i had in mind was fan out the decoding to multiple workers, so decoding itself does not have to be bottlenecked 
- we probably want to put this behind a feature flag to start, and enable it when needed (e.g. back-pressure observed); and later on assess enabling it globally by default.

itsbilal:
@jgao54 @pfcoperez thanks for the reviews! I expect that the current approach will provide some perf gains by parallelizing the event decode (in PullRecords) with the document decode (in decodeLoop). Based on how some of my in-progress stresshouse testing goes, I plan to update the code with an approach that allows for multiple parallel `decodeLoop`s that still maintain order, and if that yields a greater improvement I will go ahead with that approach.

It's also possible that the bottlenecks are elsewhere and not with full document decode. But I'm looking forward to seeing what my experimentation yields.

Also yes, I'm open to putting this behind a feature flag.

itsbilal:
I couldn't test this end-to-end using stresshouse but instead I wrote a microbenchmark that I can point at a mongoDB and a clickhouse running locally. I will polish the benchmark up in a separate PR, but in the meantime I noticed a few things:

1. Adding more workers to parallelize decode is very necessary, as that's still the most CPU-intensive part of the whole pipe. I've made this change.
2. Adding more workers isn't sufficient on its own; we also need batching because individual documents are still relatively quick to decode, so the goroutine coordination starts to dominate unless we pass larger batches of documents at once.
3. The time it takes to Next() while we wait for the next 16MB of change events from MongoDB is also significant. Some prefetching could help here; I'll try this out next.

I've updated the code to parallelize decode across multiple workers while still maintaining order of events, and with this I now see a ~40-50% reduction in the amount of time it takes to run PullRecords ( ~0.13 elapsedMinutes logged by PullRecords down to ~0.06-0.07 for my contrived benchmark of writing 2 million rows across multiple parallel writers to mongo and CDCing it over to a ClickHouse instance). Also translates to a ~40% reduction in end-to-end replication time because we were (and continue to be) bound by `PullRecords`.

This should be ready for a review again. I'll scrutinize the `checkpoint` logic a bit more as it might now be possible to write an incorrectly-forwarded checkpoint while an inflight batch errors out and breaks the pipeline. But other than that this should be good to go.

itsbilal:
> The time it takes to Next() while we wait for the next 16MB of change events from MongoDB is also significant. Some prefetching could help here; I'll try this out next.

Tried this out, didn't help significantly (< 5% improvement) so I'll keep this change as-is.

Only other TODO is to add more testing around the error / early termination cases as mentioned previously, but otherwise this is ready for a look.

jgao54:
> The time it takes to Next() while we wait for the next 16MB of change events from MongoDB is also significant. Some prefetching could help here; I'll try this out next

this would also depends on network roundtrip between peerdb node and source mongo node, and I'd expect putting them in the _same_ region should shorten this if they are tested far away from each other.

jgao54:
(i haven't yet looked at the code changes but just commenting based on what you wrote)

> I'll scrutinize the checkpoint logic a bit more as it might now be possible to write an incorrectly-forwarded checkpoint while an inflight batch errors out and breaks the pipeline.

yeah this one needs to be very careful as order matters. before we checkpoint offset after every event in-memory, and then persist it to disk at the end of a batch. if we introduce concurrent checkpointing, we have to make sure the goroutines are coordinated in a way such that we never checkpoint until all the worker are past a certain point (kind of similar to cockroachdb's quirk we had to workaround recently, only this we have to manage it ourselves)

if an _error_ is hit, it's less of a concern because we don't persist checkpoint to disk when we encounter an error, so on retry we just resume from previously checkpointed token.

edit: on second thought, maybe the worst case is we checkpoint an older offset than the latest event we send to the stream (when the events are processed out-of-order), so worst case is duplicates, not missing data... (but would still be good to avoid if possible).

itsbilal:
> if we introduce concurrent checkpointing, we have to make sure the goroutines are coordinated in a way such that we never checkpoint until all the worker are past a certain point (kind of similar to cockroachdb's quirk we had to workaround recently, only this we have to manage it ourselves)

This ended up being relatively simple to do; I just moved checkpoint responsibility over to `sendLoop` and tracked resume tokens alongside batches as they flowed through the goroutines. There's one case where we can have a resumeToken that lags behind a sent record, and that is if `RecordStream.AddRecord` itself errors out mid-batch. I can track resume tokens at record level if duplicates when resuming in that case are an issue; I had figured (possibly incorrectly) that the normalize step would flatten duplicates after an error.

If duplicates in that specific error case are a non-issue, the current approach is slightly cleaner and marginally more performant.

Also added tests for the error cases, this should be ready for a look, thanks!

flow/connectors/mongo/cdc.go
itsbilal: 10 ended up being useful once I added the dynamically increasing number of decode workers - but I can experiment with just matching the number of workers if you want?

flow/connectors/mongo/cdc.go
itsbilal: Correct, clarified!

flow/connectors/mongo/cdc.go
```diff
@@ -227,18 +230,283 @@ func decodeEvent(
...
+		items := model.NewRecordItems(2)
+
+		if len(item.documentKey) > 0 {
+			rv := item.documentKey.Lookup(DefaultDocumentKeyColumnName)
+			if rv.IsZero() || rv.Type == bson.TypeNull {
+				return nil, exceptions.NewInvalidIdValueError(item.sourceTableName)
+			}
+			qValue, err := converter.QValueStringFromId(rv, req.InternalVersion)
+			if err != nil {
+				return nil, err
```
jgao54: this was `return fmt.Errorf("failed to convert key: %w", err)` before, assuming this change is not intentional here

flow/connectors/mongo/cdc.go
```diff
@@ -227,18 +230,283 @@ func decodeEvent(
...
+// This should ideally be larger than decodeWorkerBufSize.
+const workerBufferedChanSize = 10
+
+// Number of recordItems to pass in one batch to decode/send loops. Doing per-item channel sends
+// results in too much coordination and reduces effective concurrency in practice.
+const pullRecordsItemsBatchSize = 256
+
+// maxNumDecodeWorkers is the maximum number of goroutines that decodeLoop spins up to parallelize
+// decoding of record batches.
+const maxNumDecodeWorkers = 6
```
jgao54: for parallelism, it should be a dynamic config, default to 1, and can be increased on demand. 

given there can be multiple pipes per worker, using `GOMAXPROCS` to determine pipe-level parallelism can lead to OOM given the buffering introduced in this PR

flow/connectors/mongo/cdc.go
```diff
@@ -227,18 +230,283 @@ func decodeEvent(
...
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+}
+
+func (c *MongoConnector) decodeLoop(
```
jgao54: the high-level design: (1) synchronous record pull (2) batch and parallelize for decoding (3) process each batch in order makes sense to me. i think (2) with the worker pool could be simplified with a semaphore-based approach, e.g. in `finishBatch` (and maybe rename to `dispatchBatch` since finish is a bit misleading), something like the following as a rough idea:
```
sender := make(chan chan sendBatch, decodeWorkerBufSize*maxNumDecodeWorkers)
result := make(chan sendBatch)
decodeSem := make(chan struct{}, maxNumDecodeWorkers)


// enqueue result channel before decoding goroutine starts
select {
case sender <- result:
case ... // err handling
}

// manages concurrency and backpressure
select {
case decodeSem <- struct{}{}:
case ... // err handling
}

// spun goroutine for decoding
errGroup.Go()(func() error{
	defer func() { <-decodeSem }()
	records, err := c.decodeItems(ctx, batch)
	if err != nil {
		return err
	}
	result <- sendBatch{records: records, resumeToken: batch.resumeToken}
	return nil
})

```
The semaphore makes sure there are no more than N parallel decoding goroutines; which achieves what you had originally but with less worker pool orchestration.

flow/connectors/mongo/cdc.go
```diff
@@ -227,18 +230,283 @@ func decodeEvent(
...
-func (c *MongoConnector) PullRecords(
+// Constants used by PullRecords.
+//
+// Buffered channel size for channels to pass records to decode and send loops (see below).
+// This should ideally be larger than decodeWorkerBufSize.
+const workerBufferedChanSize = 10
+
+// Number of recordItems to pass in one batch to decode/send loops. Doing per-item channel sends
+// results in too much coordination and reduces effective concurrency in practice.
+const pullRecordsItemsBatchSize = 256
```
jgao54: nit: general naming feedback, we use `batch` to refer to per-sync interval in general in PullRecord context, to avoid confusion here, maybe we can call `chunk` instead (including anywhere batch is referenced in this PR)

flow/connectors/mongo/cdc.go
```diff
@@ -369,44 +653,8 @@ func (c *MongoConnector) PullRecords(
...
-	}
-
-	addRecord := func(ctx context.Context, record model.Record[model.RecordItems]) error {
+	incrementRecordCount := func() {
 		recordCount += 1
-		if err := req.RecordStream.AddRecord(ctx, record); err != nil {
-			return err
-		}
 		if recordCount == 1 {
 			req.RecordStream.SignalAsNotEmpty()
```
jgao54: there's a subtle change here introduced which is that `req.RecordStream.SignalAsNotEmpty()` notifies downstream but then `finishBatch` fails; if this happens on the very first batch, offset may be accidentally overwritten with offset with `{ID: 0, Text: ""}`. Can be fixed by always seed PullRecord with ` req.LastOffset.Text`; or SignalAsNotEmpty only after a batch is actually send to stream

flow/connectors/mongo/cdc.go
jgao54: on the topic of framework, given the nature of this logic being fairly generic, would it be feasible to move it to to a separate shared module and import it here? it will also make the logic here easier to reason about as the `PullRecords` code is getting a bit long.

Given this PR introduces a relatively big change to mongo cdc processing, it would be ideal if we can keep the original logic for the general synchronous case, and having this logic in a separate package will make this a bit cleaner to introduce parallel-processing as a separate mode.

flow/connectors/mongo/cdc.go
itsbilal: The semaphore achieves this now, right? We'll start with 1 and go up until `min(6, GOMAXPROCS/2)`

jgao54: ah, what i meant is that runtime.GOMAXPROCS(0)/2 is not a super meaningful cap.  if there is only one pipe, we could probably get a better perf at higher value than `runtime.GOMAXPROCS(0)/2`;  if there are many pipes, runtime.GOMAXPROCS(0)/2 as ceiling is still too high since the `decodeWorkerSem` controls _per pipe_ concurrency, not service-level concurrency.  So i recommend making it a configurable value here to give more control.

flow/connectors/mongo/cdc.go
itsbilal: Implemented this approach - it significantly simplified things, thanks!

It does make sense to have a more general framework to be able to swap in parallel processing for any generic PullRecords implementation that wants to implement parallel workers while still maintaining order of records in and out. I'll do this as a follow-up, but at least with the last refactor the amount of extra code in this change has gone down significantly.

flow/connectors/mongo/cdc.go
jgao54: Or are you thinking of modifying GOMAXPROCS env var to control pipe-level concurrency.  downside  with that is it would require a deploy; vs. dynamic conf can be done on-the-fly (and it would be a service-level env-var and not something we can customize on the pipe-level)

itsbilal: Ah yes that makes sense. We don't have a per-process control or cap in a sense, which is why I went with a fairly low `maxNumDecodeWorkers` value in this case. I'll replace this cap with one defined in `flow/internal/dynamicconf.go` which is what I believe you're referring to, and get rid of the `runtime.GOMAXPROCS(0)/2` part.

itsbilal: Did this part!

flow/connectors/mongo/cdc.go
pfcoperez: >  but I can experiment with just matching the number of workers if you want?

I was mostly wondering what drove choosing that number, maybe others can drive to better results? Is 10 just good enough? 

If you experimented and observed benefits from 10 that's fine by me. I was curious as to were it's coming from :slightly_smiling_face:

pfcoperez:
> I now see a ~40-50% reduction in the amount of time it takes to run PullRecords ( ~0.13 elapsedMinutes logged by PullRecords down to ~0.06-0.07 for my contrived benchmark of writing 2 million rows across multiple parallel writers to mongo and CDCing it over to a ClickHouse instance). Also translates to a ~40% reduction in end-to-end replication time because we were (and continue to be) bound by `PullRecords`.

Wow @itsbilal, impressive! I am going through your changes again now :bow:

pfcoperez:
Adding AI generated diagram that helped me with the review:

```mermaid
flowchart TB
    CS[("MongoDB change stream")]

    subgraph main["PullRecords main goroutine (1 per batch)"]
        direction TB
        NEXT["changeStream.Next + decodeEvent<br/>(envelope only, full document kept as raw BSON)"]
        CHUNK["existingChunk []recordItems<br/>buffers up to 256 items (pullRecordsItemsChunkSize)"]
        DISPATCH["dispatchChunk()<br/>1. acquire decodeWorkerSem slot<br/>2. resultChan := make(chan sendChunk)<br/>3. sendChan ← resultChan<br/>4. workerEg.Go(decodeWorker)"]
        DRAIN["drainWorkers()<br/>dispatch partial chunk, close(sendChan), workerEg.Wait()"]
    end

    SEM{{"decodeWorkerSem<br/>chan struct{}<br/>cap = max(1, PEERDB_MONGODB_NUM_PARALLEL_DECODE_THREADS)<br/>default 1"}}
    SENDCHAN{{"sendChan<br/>chan chan sendChunk<br/>cap = 10 (workerBufferedChanSize)<br/>FIFO of per-chunk result channels"}}

    subgraph workers["decodeWorker goroutines (at most N in flight, one per chunk)"]
        direction TB
        W1["decodeWorker for chunk 1<br/>BSON → QValue, build model.Record slice"]
        W2["decodeWorker for chunk 2"]
        WK["decodeWorker for chunk k"]
    end

    RC1{{"resultChan 1<br/>chan sendChunk, unbuffered"}}
    RC2{{"resultChan 2"}}
    RCK{{"resultChan k"}}

    subgraph send["sendLoop goroutine (1 per change stream)"]
        SL["for resultChan := range sendChan<br/>chunk := ← resultChan (blocks until that chunk is decoded)<br/>RecordStream.AddRecord for each record<br/>RecordStream.UpdateLatestCheckpointText(chunk.resumeToken)"]
    end

    RS{{"req.RecordStream<br/>CDCStream records channel"}}
    SYNC["sync side of the activity<br/>(destination connector)"]

    CS -->|change events| NEXT
    NEXT -->|append| CHUNK
    CHUNK -->|"len == 256"| DISPATCH
    CHUNK -->|"batch cut: idle timeout / MaxBatchSize / stream error"| DRAIN
    DRAIN --> DISPATCH

    DISPATCH -->|"acquire: sem ← struct{}{} (blocks while N busy)"| SEM
    DISPATCH -->|"sendChan ← resultChan (blocks when 10 queued)"| SENDCHAN
    DISPATCH -.->|go| W1
    DISPATCH -.->|go| W2
    DISPATCH -.->|go| WK

    W1 -->|"resultChan ← sendChunk"| RC1
    W2 -->|"resultChan ← sendChunk"| RC2
    WK -->|"resultChan ← sendChunk"| RCK
    W1 -.->|"defer release: ← sem"| SEM
    W2 -.->|"defer release: ← sem"| SEM
    WK -.->|"defer release: ← sem"| SEM

    SENDCHAN -->|"next resultChan, in dispatch order"| SL
    RC1 -->|sendChunk| SL
    RC2 -->|sendChunk| SL
    RCK -->|sendChunk| SL
    DRAIN -.->|"close(sendChan) → sendLoop exits"| SENDCHAN

    SL -->|AddRecord| RS
    RS --> SYNC
```

flow/connectors/mongo/cdc.go
```diff
@@ -426,9 +563,67 @@ func (c *MongoConnector) PullRecords(
...
 				slog.Float64("elapsedMinutes", time.Since(pullStart).Minutes()))
 		}
+	}
+	existingChunk := make([]recordItems, 0, pullRecordsItemsChunkSize)
+	sendChan := make(chan chan sendChunk, workerBufferedChanSize)
+	numParallelDecodeWorkers, err := internal.PeerDBMongoDBNumParallelDecodeThreads(ctx, req.Env)
+	if err != nil {
+		return err
+	}
+	decodeWorkerSem := make(chan struct{}, max(1, numParallelDecodeWorkers))
```
pfcoperez, line 573: Did we choose to implement the semaphore as a bounded channel instead of https://pkg.go.dev/golang.org/x/sync/semaphore because of performance?

flow/connectors/mongo/cdc.go
```diff
@@ -426,9 +563,67 @@ func (c *MongoConnector) PullRecords(
...
+		}
+		rt := changeStream.ResumeToken()
+		var rtText string
+		if rt == nil {
+			c.logger.Warn("change stream does not currently contain a resume token")
+		} else {
+			rtText = base64.StdEncoding.EncodeToString(rt)
+		}
+		select {
+		case decodeWorkerSem <- struct{}{}:
```
pfcoperez, line 591:
```suggestion
        // Acquire worker slot in semaphore 
		case decodeWorkerSem <- struct{}{}:
```

Would it make sense to use a mini interface around `decodeWorkerSem` with methods `.AcquireOrBlock()` and `.Free()` to improve code readability?

flow/connectors/mongo/cdc.go
```diff
@@ -230,6 +231,170 @@ func decodeEvent(
...
+			case <-ctx.Done():
+				return nil
+			}
+		case <-ctx.Done():
+			return nil
+		}
+	}
+}
+
+type recordItems struct {
```
pfcoperez, line 301: I would rename this struct to something that conveys its content better. As I understand it, this struct contains the information associated with a MongoDB event.

If the event contains a document (`maybeFullDocument != nil`), then we'll generate a record which itself contains the `RecordItems` as the map of columns and their QValues

flow/connectors/mongo/cdc.go
```diff
@@ -230,6 +231,170 @@ func decodeEvent(
...
+const workerBufferedChanSize = 10
+
+// Number of recordItems to pass in one chunk to decode/send loops. Doing per-item channel sends
+// results in too much coordination and reduces effective concurrency in practice.
+const pullRecordsItemsChunkSize = 256
+
+type decodeChunk struct {
+	// base64-encoded resume token
+	resumeToken string
+	items       []recordItems
```
pfcoperez, line 247: Each element of this collection actually represents a document record which, prior to this change, was  populated with: https://github.com/PeerDB-io/peerdb/blob/37803572401232eefd6db5af8d89e38a09a21e43/flow/connectors/mongo/cdc.go#L381 function.

The world items is a bit overloaded and I think we should use it for the record items: 

```go
type RecordItems struct {
	ColToVal map[string]types.QValue
}
```

Our private `recordItems`, representing the source record envelop, clashes with `RecordItems`  representing decoded, processed and ready to be sent records.

flow/connectors/mongo/cdc.go
```diff
@@ -230,6 +231,170 @@ func decodeEvent(
...
+				SourceTableName:      item.sourceTableName,
+				DestinationTableName: item.destinationTableName,
+			}
+		}
+		return record, nil
+	}
+	modelRecords := make([]model.Record[model.RecordItems], len(chunk.items))
+	for i := range chunk.items {
+		var err error
+		if modelRecords[i], err = parseItem(chunk.items[i]); err != nil {
```
pfcoperez, line 386: This has the potential of further increasing parallelism, with N   `NewDirectBsonConverter`, given that we are decoding positional elements in the same batch, this parsing could be done in N threads, one per converter.

This makes me wonder. Could we get parallelism changes simplified by exploiting it at this level and just at this level?

Instead of having M workers processing and sending batches, we could just accumulate one batch (e.g: 256 elements) launch the decoding ops in parallel, order is given by the position. Once the whole batch is one we move to the next. No sempahores, no workers management, ....

itsbilal: This is what I had initially, as it turns out - I didn't have any batching logic and I was dispatching one item to a worker at a time. The problem there was that the goroutine coordination and channel overhead per decode item was high enough to dominate over the wins of parallelizing the decode, unless the documents themselves were very big.

Some batching was important to actually make the parallelism work out and yield a benefit in the benchmark. (The initial approach was not significantly faster than the baseline, even with ~8 decode workers).

pfcoperez: Makes sense @itsbilal , thanks for sharing!

flow/connectors/mongo/cdc.go
itsbilal: That one implements a weighted semaphore, which we don't really need here. Happy to use it still, but most Go semaphores i've seen are just buffered channels

pfcoperez: > Go semaphores i've seen are just buffered channels

Yeah, I was curious. Mostly moved by having an interface that explicitly offers "semaphore" vocabulary but I agree, buffered channels are quite popular for this. Thanks for clarifying!

flow/connectors/mongo/cdc.go
itsbilal: Renamed it to encodedMongoEvent and used the var name `event` instead of `items` here - thanks!

flow/connectors/mongo/cdc.go
itsbilal: Done

flow/connectors/mongo/cdc.go
itsbilal: I considered adding helper functions, but the fact that we use the semaphores in select{} statements makes it a bit more challenging. I ended up adding comments as you suggested instead - thanks!

itsbilal:
@jgao54 @pfcoperez I ended up adding a new generic-heavy worker pool abstraction `PullRecordsWorkerPool` in `pkg/common/workers.go` that maintains order when chunking and dispatching work to workers, along with associated ResumeTokens, then combines the results in-order and calls a Send function on the results. It neatly simplified away a lot of the concurrency code in PullRecords itself. This should be ready for another look, thanks!

pfcoperez:
@fable run a review on these changes focusing on potential data loss or event order changes regressions.

pfcoperez:
@itsbilal From the Fable review I'd address this one, specially making sure the defaults don't lead to hanging syncs:

> oncurrency is taken verbatim from the dynamic config at [cdc.go:464-469](https://github.com/PeerDB-io/peerdb/blob/7ec39fc/flow/connectors/mongo/cdc.go#L464-L477) and used as the semaphore capacity at [workers.go:54](https://github.com/PeerDB-io/peerdb/blob/7ec39fc/flow/pkg/common/workers.go#L54). If an operator sets it to 0, p.sem <- struct{}{} in Flush can never proceed (no worker ever releases a slot) and the mirror hangs in AddItem until the activity is killed; a negative value panics in make(chan struct{}, n). Since this knob exists precisely to be tuned on live mirrors, clamp it: Concurrency: max(1, int(numParallelDecodeWorkers)) (or validate in Init). [Fix this →](https://claude.ai/code?q=In%20flow%2Fconnectors%2Fmongo%2Fcdc.go%20(branch%20bilal%2FDBI-482-pipeline-decode)%2C%20the%20PullRecordsWorkerPool%20Concurrency%20is%20set%20directly%20from%20PeerDBMongoDBNumParallelDecodeThreads.%20A%20value%20of%200%20deadlocks%20Flush%20(semaphore%20never%20released)%20and%20a%20negative%20value%20panics%20in%20make(chan).%20Clamp%20with%20max(1%2C%20int(numParallelDecodeWorkers))%20and%2For%20validate%20Concurrency%20%3E%3D%201%20in%20PullRecordsWorkerPool.Init%20in%20flow%2Fpkg%2Fcommon%2Fworkers.go%2C%20and%20add%20a%20test.&repo=PeerDB-io/peerdb)

itsbilal:
@pfcoperez addressed! thanks

flow/shared/concurrency/workers.go
```diff
@@ -0,0 +1,169 @@
...
+//
+// Items of work ("events") arrive through AddItem, and are chunked into instances of WorkerFunc
+// that are spun up (up to a max of `Concurrency` in parallel). The result of WorkerFunc
+// is passed through to Send, along with the resumeToken from the last call to AddItem
+// that added an event that was part of this chunk. Order is maintained through all calls;
+// events that arrive earlier will be part of earlier chunks than events that arrived later,
+// and chunks will maintain this AddItem-implied order on their way to Send(). Within a chunk,
+// the slice of events is also ordered from earliest to latest.
+type PullRecordsWorkerPool[E, D, RT any] struct {
+	WorkerFunc func(events []E) (D, error)
```
pfcoperez, line 29: Nice!

pfcoperez (approved):
This reads like a charm now, thank you!

flow/pkg/common/workers_test.go
```diff
@@ -0,0 +1,84 @@
+package common
```
jgao54: nit: we use flow/pkg/ for anything that needs to be exported/used externally (i.e. the validation logic); given this is intended for internal use, would recommend putting it in flow/shared/

itsbilal: done, thanks!

jgao54 (approved):

flow/connectors/mongo/cdc.go
```diff
@@ -339,6 +462,21 @@ func (c *MongoConnector) PullRecords(
 			slog.Int("channelLen", req.RecordStream.ChannelLen()),
 			slog.Float64("elapsedMinutes", time.Since(pullStart).Minutes()))
 	}()
+	numParallelDecodeWorkers, err := internal.PeerDBMongoDBNumParallelDecodeThreads(ctx, req.Env)
```
jgao54, line 465: one thing i missed earlier during review that may worth a follow-up: currently this hits the db on every PullRecords, which can be as frequent as every second.  so applied to all the pipes in all services this can cause some db pressure.

given that this changes infrequently, and requires a pause/resume to update anyways, can be a connector level setting that gets read once upfront (similar to how we handle `PeerDBMongoDBExcludedOperationTypes` today).

PR #4723, opened 2026-08-22
remove errgroup legacy code

jgao54:
the normalizeLoop was wrapped in a errgroup with no other goroutine because historically it was used to group `normalizeloop` with `maintainReplConn` in https://github.com/PeerDB-io/peerdb/pull/2371; the latter was later removed in https://github.com/PeerDB-io/peerdb/pull/4324. NormalizeLoop never returns an error, so this errgroup here can be replaced with a more idiomatic "done" channel approach to avoid confusion.

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -428,7 +428,8 @@ func (a *FlowableActivity) SyncFlow(
 				normRequests, normResponses, normBufferSize, idleTimeout, &syncingBatchID, &syncState)
 		}
 
-		if syncErr != nil {
+		if syncErr != nil && ctx.Err() == nil {
+			logger.Error("failed to sync records", slog.Any("error", syncErr))
 			break
 		}
 		if syncResponse != nil {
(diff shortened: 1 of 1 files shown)


flow/activities/flowable.go
```diff
@@ -458,14 +447,15 @@ func (a *FlowableActivity) SyncFlow(
...
-		logger.Info("sync canceled", slog.Any("error", err))
-		return err
-	} else if waitErr != nil {
-		logger.Error("sync failed", slog.Any("error", waitErr))
-		return waitErr
+	<-normDone
+
+	if syncErr != nil && ctx.Err() == nil {
+		// return syncErr when it's not a top-level context cancellation
+		logger.Error("failed to sync records", slog.Any("error", syncErr))
```
ilidemi: This will be logged after normalize finishes. Would be more helpful in investigations if we still log it immediately, to 100% see what triggered what

ilidemi (approved):

PR #4726, opened 2026-08-25
fix: suppress bool->int8 false-positive column type change from TABLE_MAP metadata

shockdav:
MySQL/MariaDB binlog `TABLE_MAP` metadata cannot express the `TINYINT(1)` display width. A column that the snapshot typed as `bool` therefore always arrives on the wire as `int8`, and `shouldReportColumnTypeChange` reports it as a column type change **once per row-event, forever**.

On a busy table this is unbounded log growth. In our deployment (MariaDB 10.x → ClickHouse, `stable-v0.37.0`) a single `tinyint(1)` column on the hottest table produced ~34 GB of this one WARN line — about 83% of the flow-worker's total log volume — before it filled the host's disk and stalled the mirror:

```
{"level":"WARN","msg":"column type change detected from TABLE_MAP_EVENT, not propagating","table":"tradersync.trades","column":"symboldata","from":"bool","to":"int8"}
```

The source column was, and still is, `tinyint(1)`; nothing changed. The mirror column is `Bool` and the data is correct either way — the warning is a pure false positive.

This suppresses exactly the `bool` (schema) → `int8` (wire) pairing, mirroring the existing MariaDB UUID/INET wire-kind special case. A real change — `bool` → anything other than `int8`, or any non-`bool` schema kind — is still reported.


shockdav:
Pushed a second commit covering the same failure mode for ENUM: binlog TABLE_MAP encodes ENUM under the generic STRING column type, so a snapshot-typed enum column always arrives as string on the wire and warns once per row-event forever (we measured 4 enum columns on one table producing ~90 warn lines/min). Same shape of suppression, scoped to exactly the enum->string pairing.

ilidemi (approved):
Thanks for the contribution @shockdav! Would you be able to sign the CLA so we can merge?

PR #4727, opened 2026-08-25
feat(bigquery): implement a separate SyncFlow branch for query-based CDC

dtunikov:
# Problem

We need to support CDC for DWH sources such as BigQuery/Snowflake. These sources don't have global wal/binlog/change-stream concept. There is no channel we can listen to for events (inserts, updates, deletes).
For such sources we are going to implement query-based replication. Which, essentially, just queries the tables every SyncInterval seconds, e.g. runs similar query every N secs:
`SELECT ... FROM table WHERE col > lower AND col <= upper`

We could keep using existing `SyncFlow` PullRecords function to query the source. But it means that all mirror's tables would be coupled:
- one CDC records stream for all tables
- one table's failure blocks other tables replication
- one table can slow down replication of other tables (if the source is slow or a downstream MV is slow)

# Solution
This PR introduces a new `SyncFlow` branch for query-based CDC sources (for now just BigQuery).

It implements the following requirements:
- Back-pressure must be applied per table
- Tables must be queried in parallel (a single slow table can't affect others)
- Errors must be reported via LogFlowError, but shouldn't block other tables
- No `raw` table hop. Batches are ingested directly from s3/gcs into the final destination table

Currently, only BigQuery source connector supports new interfaces. In the future this workflow can be used for other query-based CDC connectors as well (e.g. other data warehouses or even OTLP databases in cases when a customer doesn't want to rely on wal/binlog replication).

# Tests
Added e2e tests that cover the following scenarios:
- single table's failure doesn't block other tables
- back-pressure kicks in only for the affected table, so it doesn't block sibling tables
- run cdc, pause pipe, remove table from the mirror, unpause -> replication still works

# Workflow diagram
```mermaid
flowchart TD
    SF["SyncFlow activity<br/>(flowable.go)"]
    SF -->|isIsolatedTableCDCPath?<br/>BQ source + ClickHouse dest| DEC{branch}
    DEC -->|false| OLD["pullAndSync / pullAndSyncPg<br/>+ shared CDCStream<br/>+ normalizeLoop (unchanged path)"]
    DEC -->|true| SFI["syncFlowIsolatedTables<br/>(flowable_isolated_cdc.go)"]

    SFI --> SETUP["setup: idleTimeout, channelBufferSize,<br/>pullSem(parallelism), normBufferSize,<br/>PruneTableReplicationState, batchIDCounter"]
    SETUP --> SPAWN["for each TableMapping:<br/>spawn 1 sync loop + 1 normalize loop<br/>paired via per-table LastChan(normRequests, normResponses)"]

    subgraph PERTABLE["per source table (independent goroutines)"]
        direction TB

        subgraph SYNCLOOP["isolatedTablePullSyncLoop"]
            direction TB
            S1["GetTableReplicationState"] --> S2{"synced - normalized<br/>>= normBufferSize?"}
            S2 -->|yes: backpressured| S2W["wait on normResponses<br/>(this table only)"] --> S1
            S2 -->|no| S3{"poll due?<br/>(idleTimeout)"}
            S3 -->|not yet| S3W["waitOrDone"] --> S1
            S3 -->|due| S4["acquire pullSem"]
            S4 --> S5["RecordTableReplicationAttempt"]
            S5 --> S6["errgroup:<br/>PullTableRecords -> TableCDCStream<br/>SyncTableCDC drains stream"]
            S6 --> S7["release pullSem"]
            S7 --> S8{"rows > 0?"}
            S8 -->|no| S9["RecordTableReplicationSync<br/>(cursor only)"] --> S1
            S8 -->|yes| S10["RecordTableReplicationSync<br/>(cursor + synced_batch_id++)"]
            S10 --> S11["recordIsolatedTableBatch<br/>(cdc_batches/cdc_flows monitoring)"]
            S11 --> S12["normRequests.Update(batchID)"]
            S12 --> S1
        end

        subgraph NORMLOOP["isolatedTableNormalizeLoop"]
            direction TB
            N1["resume: normResponses.Update(normalizedBatchID);<br/>if synced>normalized, normRequests.Update(synced)"] --> N2{"normRequests > normResponses?"}
            N2 -->|no| N2W["wait on normRequests"] --> N2
            N2 -->|yes| N3["NormalizeTableCDC:<br/>for batch in (lastNorm, reqBatch]:<br/>GetTableAvroStage -> INSERT...SELECT<br/>into final table -> DeleteTableAvroStage"]
            N3 --> N4{"error?"}
            N4 -->|yes| N4B["alert once + backoff"] --> N2
            N4 -->|no| N5["RecordTableReplicationNormalize"]
            N5 --> N6["normResponses.Update(batchID)<br/>(releases sync loop's backpressure wait)"]
            N6 --> N2
        end

        S6 -. "PullTableRecords" .-> BQ[("BigQuery<br/>connector")]
        S6 -. "SyncTableCDC:<br/>typed Avro -> S3/GCS<br/>+ SetTableAvroStage" .-> CH1[("ClickHouse<br/>connector")]
        N3 -. "NormalizeTableCDC" .-> CH2[("ClickHouse<br/>final table")]

        S12 -. signal .-> N2
        N6 -. signal .-> S2
    end

    SPAWN --> SYNCLOOP
    SPAWN --> NORMLOOP

    CAT1[("catalog:<br/>cdc_table_replication_state<br/>(cursor_text, last_attempt_at,<br/>synced_batch_id, normalized_batch_id)")]
    CAT2[("catalog:<br/>cdc_table_avro_stage<br/>(flow, table, batch_id -> avro_file)")]
    S1 -.-> CAT1
    S10 -.-> CAT1
    N1 -.-> CAT1
    N5 -.-> CAT1
    S6 -.-> CAT2
    N3 -.-> CAT2
```

Resolves: DBI-1045


flow/connectors/clickhouse/query_cdc_sync.go
```diff
@@ -0,0 +1,159 @@
...
+			SourceType:     protos.DBType_BIGQUERY,
+			WatermarkTable: req.SourceTableIdentifier,
+		},
+		connector: c,
+		logger:    c.logger,
+	}
+
+	chSettings := chinternal.NewInsertSettings(c.chVersion, req.Version)
+
+	for batchID := req.StartBatchID + 1; batchID <= req.EndBatchID; batchID++ {
```
dtunikov: opened a draft PR to combine multiple batches in a single insert to produce less parts on destination CH:
https://github.com/PeerDB-io/peerdb/pull/4732

flow/connectors/bigquery/cdc.go
```diff
@@ -12,8 +12,6 @@ import (
 	"time"
 
 	"cloud.google.com/go/bigquery"
-	"go.opentelemetry.io/otel/attribute"
```
dtunikov, line 16: will be easier to review `PullTableRecords` in an editor rather than looking at this diff
the diff is a bit messed up since I removed PullRecords implementation
can also bring it back for the sake for PR readability and submit PullRecords deprecation as a separate PR

flow/connectors/bigquery/cdc.go
```diff
@@ -75,137 +73,102 @@ func pollWindow(checkpoint, now time.Time, safetyLag, maxQueryWindow time.Durati
...
+		return model.PullTableRecordsResult{}, fmt.Errorf("failed to get BigQuery CDC safety lag: %w", err)
 	}
 	maxQueryWindow, err := internal.PeerDBBigQueryCDCMaxQueryWindow(ctx, req.Env)
 	if err != nil {
-		return fmt.Errorf("failed to get BigQuery CDC max query window: %w", err)
+		return model.PullTableRecordsResult{}, fmt.Errorf("failed to get BigQuery CDC max query window: %w", err)
 	}
 
-	upper, ok := pollWindow(checkpoint, now, safetyLag, maxQueryWindow)
+	upper, ok := pollWindow(start, now, safetyLag, maxQueryWindow)
```
dtunikov, line 131: This whole safety_lag + max_query_window logic might be moved to the activity lvl `syncFlowIsolatedTables`. I can imagine that other DWH connectors might need smth similar.
A reminder:
- safety_lag - keeps a poll window's upper bound this many seconds behind the source clock. (ms fabric documents that safety_lag must be at least 10 mins for the `CHANGES` mode, but it's not confirmed in official GCP docs + manual testing didn't reveal that, but still this might be relevant to configure in some cases).
- max_query_window - caps how much time a single poll can cover at once. (for BQ it can't be more than 1d I believe)

draft [PR](https://github.com/PeerDB-io/peerdb/pull/4734) to move it into the activity.

nexus/catalog/migrations/V56__query_cdc_replication_state.sql
```diff
@@ -0,0 +1,25 @@
...
+    updated_at timestamptz NOT NULL DEFAULT now(),
+    PRIMARY KEY (flow_name, source_table_identifier)
+);
+
+-- Avro files staged by a table's sync step, pending normalize (INSERT straight
+-- into the final destination table). One row per (table, synced_batch_id).
+CREATE TABLE IF NOT EXISTS cdc_table_avro_stage (
+    flow_name text NOT NULL,
+    source_table_identifier text NOT NULL,
+    batch_id bigint NOT NULL,
```
dtunikov, line 27: per table sequence - has nothing to do with global cdc_batches table
we still write batches to cdc_batches as well for stats reporting
in comparison to PullRecords path, we're going to have a batch per table

flow/connectors/clickhouse/table_function.go
```diff
@@ -64,9 +66,9 @@ func timeFieldExpressionConverter(
 		return sourceFieldIdentifier, nil
 	}
 
-	// Handle BigQuery source where TIME is exported as Parquet TIME(MICROS), which
-	// ClickHouse interprets as DateTime64(6, 'UTC'), so no manual conversion needed
-	if config.config.SourceType == protos.DBType_BIGQUERY {
+	// Parquet TIME(MICROS) (e.g. BigQuery's object export) is interpreted by
+	// ClickHouse as DateTime64(6, 'UTC') directly, so no manual conversion needed.
+	if config.parquetStaged {
```
dtunikov: the same path is used for CDC now (previously it was just snapshot path where we have parquet export from GCP)
now we need to support avro as well (cdc path)
for avro we can't skip `toTime64(toDecimal64(%s, 6) / 1000000, 6)` conversion, otherwise `Test_BigQuery_CDC_All_Types` fails

flow/connectors/core.go
```diff
@@ -136,6 +136,24 @@ type CDCPullConnector interface {
...
+// independently (no shared replication stream to isolate tables on top of),
+// letting the activity drive per-table isolation, parallelism, and
+// backpressure generically instead of each such connector reimplementing it.
+type TableCDCPullConnector interface {
+	CDCPullConnectorCore
+
+	// PullTableRecords pulls whatever is newly available for one source table
+	// and streams it into req.Stream. This method should be idempotent given
+	// the same req.Cursor.
+	PullTableRecords(
```
dtunikov, line 149: see [this](https://github.com/PeerDB-io/peerdb/pull/4727#discussion_r3862432408) for how it could be extended if we want to move more responsibilities to the activity code

flow/connectors/bigquery/bigquery.go
```diff
@@ -52,13 +53,13 @@ func NewBigQueryServiceAccount(bqConfig *protos.BigqueryConfig) (*utils.GcpServi
...
-	lastPollAt time.Time
 	// droppedExcludeColumns remembers, per source table, excluded columns that
 	// BigQuery has reported as no longer existing, so later polls stop asking BigQuery
-	// to EXCEPT them.
-	droppedExcludeColumns map[string]map[string]struct{}
-	logger                log.Logger
+	// to EXCEPT them. Guarded by droppedExcludeColumnsMu since concurrent per-table
+	// pull loops share this connector.
+	droppedExcludeColumnsMu sync.Mutex
+	droppedExcludeColumns   map[string]map[string]struct{}
```
dtunikov, line 62: dropped columns are completely removed in favor of explicit selection set in the follow-up [PR](https://github.com/PeerDB-io/peerdb/pull/4740)

flow/activities/flowable_isolated_cdc.go
```diff
@@ -0,0 +1,478 @@
...
+		}
+		normResponses.Update(reqBatchID)
+		logger.Info("[cdc] normalize done", slog.Int64("normalizedBatchID", reqBatchID))
+	}
+}
+
+// recordIsolatedTableBatch records one table's completed sync as its own tiny
+// CDC batch, reusing the same catalog bookkeeping the shared-stream pipeline
+// uses, so the mirror status UI and lag metrics keep working for this path too.
+func (a *FlowableActivity) recordIsolatedTableBatch(
```
ilidemi: Are we getting anything from cdc_batches that we couldn't get from cdc_table_replication_state, cdc_table_avro_stage and metrics?
Current uses:
* CDCMirrorStatus RowsSynced - can be summed from cdc_table_aggregate_counts
* CDC graph (rows over time) - there is an OTel metric, it's not fully propagated through ClickPipes but it could be. Don't think we need to invest in PeerDB UI if OTel is available
* E2E lag, destination lag, normalize lag - can add more timestamps to cdc_table_avro_stage and find max delta to now() (being careful with deleting them promptly and cleaning up on table removal)
* The view that shows batches for PG/MySQL/Mongo - the purpose of it for the user is to see if their pipe/mirror is healthy, and if something is delayed, what is it and by how much. For DWH pipes this could straightforwardly be the list of tables with data from cdc_table_replication_state and cdc_table_avro_stage in each - some combination of last successful pull, sync, CH ingest, watermark value. No need to expose batch ids at all

flow/activities/flowable_query_cdc.go
```diff
@@ -0,0 +1,478 @@
...
+
+	normBufferHours, err := internal.PeerDBNormalizeBufferHours(ctx, config.Env)
+	if err != nil {
+		return a.Alerter.LogFlowError(ctx, flowName, err)
+	}
+	// Same approximation pullAndSyncCore uses: normBufferHours worth of
+	// idleTimeout-cadence polls, at least 2 so a table's own sync can run
+	// ahead of its own normalize by a couple of batches under steady state.
+	normBufferSize := normBufferHours * 3600 / int64(idleTimeout.Seconds())
+	normBufferSize = max(normBufferSize, 2)
```
ilidemi, line 95: Possible that I'm missing something, but not sure we have a technical reason for persisting more than one batch in S3, and consequently needing backpressure. In Postgres/MySQL we have limited retention and consume actual disk space, in BQ we have cloud 7-day time travel for changes and data just available for watermark. Can think of a product case of really needing watermark changes at 5 minute granularity even when CH is down, but that remains to be confirmed.

If multibatch is not needed, can simplify catalog schemas, remove a little bit of logic, but would be a good time to put a block on pull when destination is unhealthy

ilidemi (commented):
High level review so far

flow/activities/flowable_query_cdc.go
ilidemi: Having pull/sync and normalize pipelined is valuable though, so 1 full batch can be unfinished while the next one is produced

ilidemi: Correction: continuously persisting into S3 is only needed for PG as reconstructing replication state with a lot of WAL can be expensive. For MySQL and Mongo 24h doesn't really do anything as we strongly suggest 2-7 days of binlog retention (but still get full fidelity stream after a pause)

ilidemi: Maybe it would make sense to buffer up to PEERDB_GROUP_NORMALIZE or +1 or 2x that (however pipelining math works out). It does provide a buffer against slow MVs that do something dumb on every insert but about the same amount of dumb. With this approach, the design wouldn't change, just the threshold would be more reasoned. Pardon for thrashing

dtunikov: Current normalize implementation doesn't respect `PEERDB_GROUP_NORMALIZE` yet. It makes an insert into CH per batch. I've put a [WIP PR](https://github.com/PeerDB-io/peerdb/pull/4732) to insert multiple batches at once into the destination table.

Regarding a need for back-pressure in general... I agree that it's less important than pg/mysql. But at the same time it keeps our source query-pattern consistent over time. I mean, we continue hitting the source every `SyncInterval` seconds, we continue getting roughly the same amount of data  on every poll interval. If we wouldn't have the back-pressure mechanism, we'd have to query a lot more data than usual from the source in short amount of time (potentially).

And regarding the buffer size. Imo it's fine to keep using `PEERDB_NORMALIZE_BUFFER_HOURS` for BQ connector. Using `PEERDB_GROUP_NORMALIZE` seems like an implicit workaround to me. It'd control two a bit different replication dimensions:
- how many batches are inserted into CH at once 
- how long do we wait before kicking in the back-pressure.
We could maybe tune the default value if you think that 24h is too much for such query-based connectors?

flow/activities/flowable_isolated_cdc.go
dtunikov: Yeah, fully agree. I'll check how it all could work together just based on the new cdc_table_replication_state and cdc_table_avro_stage tables. And yeah, agree that it doesn't have to mirror OTLP pipes exactly.

flow/activities/flowable_query_cdc.go
ilidemi: As it wouldn't influence the architecture as much as I initially thought, fine either way. Just the "why" is a bit weaker - if we didn't have Postgres we probably wouldn't bother with pulling for strictly more than optimal CH perf requires. In the scenario where CH being down starts backpressure on BQ only after a few batches, we would need some sort of artificial limit on pull size so that pipelining is still happening, so yeah still doing predictable batches helps with that.

flow/activities/flowable_isolated_cdc.go
dtunikov: So..
I added a few fields to `cdc_table_replication_state` table:
```sql
CREATE TABLE IF NOT EXISTS cdc_table_replication_state (
    flow_name text NOT NULL,
    source_table_identifier text NOT NULL,
    cursor_text text NOT NULL DEFAULT '',
    -- last time we queried the source
    last_attempt_at timestamptz,
    -- last time we synced a batch to s3/gcs
    last_synced_at timestamptz,
    -- synced_batch_id/normalized_batch_id are this table's own sync/normalize
    -- progress, independent of every other table's
    synced_batch_id bigint NOT NULL DEFAULT 0,
    normalized_batch_id bigint NOT NULL DEFAULT 0,
    last_normalized_at timestamptz,
    updated_at timestamptz NOT NULL DEFAULT now(),
    inserts_count bigint NOT NULL DEFAULT 0,
    updates_count bigint NOT NULL DEFAULT 0,
    deletes_count bigint NOT NULL DEFAULT 0,
    PRIMARY KEY (flow_name, source_table_identifier)
);
```

Also added a `GetTableReplicationState` handler to return data from this table for the mirror.
I think this gives us pretty good observability into the current state of cdc
On the UI we could display a table like this:
| Table  | Last queried at     | Last synced at      | Inserts | Updates | Deletes |
|--------|----------------------|----------------------|---------|---------|---------|
| users  | 12.01.2026 15:30:01  | 12.01.2026 15:31:04  | 10      | 2       | 3       |
| orders | 12.01.2026 15:30:05  | 12.01.2026 15:31:01  | 20      | 0       | 5       |

Here "Last synced at" means when was the last batch normalized (not synced to s3/gcs).
Can hide updates/deletes on the UI when replication mode is not `CHANGES`.
Theoretically, can also add smth about synced/normalized batch ids (like a gap or even some sort of status derived from the diff between last synced and last normalized batch ids..)

Will follow-up on other metrics (lag, rows over time) later.

dtunikov: Removed dependency on cdc_batches completely from this isolated cdc branch.
Added synced records metrics reporting from the isolatedPullAndSync func:
https://github.com/PeerDB-io/peerdb/pull/4727/commits/2e2b1a2768d48dc09bf3e449977269db9fac55c9

We'd need to propagate RecordsSyncedPerTableGauge metric through ClickPipes, then we could add smth like a bar chart with records per batch.

dtunikov: Let's handle lag metrics as a separate [PR](https://github.com/PeerDB-io/peerdb/pull/4750). Don't want to blow up this one even more.

flow/activities/flowable.go
```diff
@@ -174,27 +174,54 @@ func (a *FlowableActivity) EnsurePullability(
 	return output, nil
 }
 
+func isIsolatedTableCDCPath(
```
jgao54: high-level feedback on naming conventions in this PR, recommend a consistent convention across function calls, db, file names, etc. so it's easy to identify whenever we hit this path. If we decide to go with "Query CDC":
- isIsolatedTableCDCPath -> isQueryCdcPath
- syncFlowIsolatedTables -> syncFlowQueryCDC
- flowable_isolated_cdc.go -> flowable_query_cdc.go
- TableReplicationState/GetTableReplicationState -> QueryCDCTableReplicationState/GetQueryCDCTableReplicationState
- TableCDCPullConnector -> QueryCDCPullConnector
- PEERDB_CDC_TABLE_PARALLELISM -> PEERDB_QUERY_CDC_TABLE_PARALLELISM

nexus/catalog/migrations/V56__cdc_table_replication_state.sql
```diff
@@ -0,0 +1,25 @@
...
+    updated_at timestamptz NOT NULL DEFAULT now(),
+    PRIMARY KEY (flow_name, source_table_identifier)
+);
+
+-- Avro files staged by a table's sync step, pending normalize (INSERT straight
+-- into the final destination table). One row per (table, synced_batch_id).
+CREATE TABLE IF NOT EXISTS cdc_table_avro_stage (
+    flow_name text NOT NULL,
+    source_table_identifier text NOT NULL,
+    batch_id bigint NOT NULL,
```
jgao54: useful context, would be good to add this at the top of the sql file

flow/connectors/external_metadata/table_replication_state.go
```diff
@@ -0,0 +1,185 @@
...
+// RecordTableReplicationSync persists a table's new cursor and, if newBatchID
+// is non-zero, advances its synced_batch_id after a successful poll that
+// produced records staged for normalize. newBatchID is zero for a poll that
+// found nothing new; the cursor still advances but there's no batch to
+// normalize.
+func (p *PostgresMetadata) RecordTableReplicationSync(
+	ctx context.Context, jobName string, sourceTableIdentifier string, cursor string, syncedAt time.Time, newBatchID int64,
+) error {
+	if _, err := p.pool.Exec(ctx, `
+		INSERT INTO `+cdcTableReplicationStateTableName+
```
jgao54: is there a reason for this being an INSERT ... ON CONFLICT DO UPDATE ... statement instead of just UPDATE statement like `RecordTableReplicationNormalize`?  insert here suggest the row may not exist, but in isolatedTablePullSyncLoop it's always called after `RecordTableReplicationAttempt`

flow/connectors/external_metadata/query_cdc_replication_state.go
```diff
@@ -0,0 +1,185 @@
...
+			AND `+cdcTableReplicationStateTableName+`.synced_batch_id = 0
+			AND `+cdcTableReplicationStateTableName+`.normalized_batch_id = 0
+	`, jobName, sourceTableIdentifier, cursor); err != nil {
+		return fmt.Errorf("failed to initialize table replication state for %s: %w", sourceTableIdentifier, err)
+	}
+	return nil
+}
+
+// RecordTableReplicationAttempt records that a poll attempt for this table
+// started at attemptedAt, creating the row if this is the table's first poll.
```
jgao54, line 103: > if this is the table's first poll

this is only possible in the table-addition path right?  (could add context in comment)

flow/connectors/utils/stream.go
```diff
@@ -148,6 +148,84 @@ func recordToQRecordOrError(
...
+	var isDeleted int64
+	switch typedRecord := record.(type) {
+	case *model.InsertRecord[model.RecordItems]:
+		items = typedRecord.Items
+	case *model.UpdateRecord[model.RecordItems]:
+		items = typedRecord.NewItems
+	case *model.DeleteRecord[model.RecordItems]:
+		items = typedRecord.Items
+		isDeleted = 1
+	case *model.MessageRecord[model.RecordItems]:
```
jgao54: nit: better to let this fallback into the default path as it's only emitted by PG today; if we ever decide to use it for whatever reason this logic may need to be updated vs. silently ignored

flow/connectors/external_metadata/store.go
```diff
@@ -457,5 +457,9 @@ func SyncFlowCleanupInTx(ctx context.Context, tx pgx.Tx, jobName string) error {
 		return err
 	}
 
+	if err := deleteTableReplicationStateInTx(ctx, tx, jobName); err != nil {
```
jgao54: should also clean up cdc_table_avro_stage to avoid orphaned rows when pipe is deleted

flow/activities/flowable_isolated_cdc.go
```diff
@@ -0,0 +1,442 @@
...
+	})
+	defer shutdown()
+
+	group, groupCtx := errgroup.WithContext(ctx)
+	for _, tableMapping := range options.TableMappings {
+		normRequests := concurrency.NewLastChan()
+		normResponses := concurrency.NewLastChan()
+
+		group.Go(func() error {
+			return a.isolatedTableNormalizeLoop(groupCtx, config, tableMapping, tableNameSchemaMapping, normRequests, normResponses)
```
jgao54: looks like normalize doesn't have a way to manage concurrency, so could end up overwhelming clickhouse with too many inserts and lead to OOM. In the happy path with sync-side concurrency this is unlikely, however if normalize is behind sync for all tables (say due to clickhouse being unavailable), the catch-up can generate a lot of inserts.

so a separate tunable config/semaphore here is generally useful.

flow/activities/flowable.go
```diff
@@ -174,27 +174,54 @@ func (a *FlowableActivity) EnsurePullability(
...
+	flowConfig, err := internal.FetchConfigFromDB(ctx, a.CatalogPool, config.FlowJobName)
 	if err != nil {
-		return nil, a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
+		return nil, a.Alerter.LogFlowError(ctx, config.FlowJobName, fmt.Errorf("failed to fetch flow config: %w", err))
+	}
+
+	var rawTableIdentifier string
+	// The isolated per-table CDC path writes typed Avro straight into each
+	// destination table, so there's no _peerdb_raw_* table to create.
+	if !isIsolatedTableCDCPath(flowConfig, destinationType) {
```
jgao54: would checking `srcConn.(connectors.TableCDCPullConnector)` be more direct here (vs. flowConfig)?

flow/activities/flowable.go
```diff
@@ -392,6 +419,10 @@ func (a *FlowableActivity) SyncFlow(
 		return a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
 	}
 
+	if isIsolatedTableCDCPath(config, destinationType) {
+		return a.syncFlowIsolatedTables(ctx, config, options, srcConn.(connectors.TableCDCPullConnector))
+	}
+
```
jgao54: style nit: this path is hidden inside the middle of `SyncFlow` so isn't very obvious , doesn't need to change anything in this PR to avoid scope creep, but can add a TODO here for a small refactor later, e.g. 

```
    ...
	if if isIsolatedTableCDCPath(config, destinationType) {
		return a.syncFlowIsolatedTables(ctx, config, options, tablePull)
	}
	return a.syncFlowSharedStream(...)
}
```

flow/connectors/bigquery/cdc.go
```diff
@@ -76,141 +74,104 @@ func pollWindow(checkpoint, now time.Time, safetyLag, maxQueryWindow time.Durati
...
-			ctx, sourceTableIdentifier, req.TableNameMapping[sourceTableIdentifier], checkpoint, upper, addRecord,
-		)
-		if err != nil {
-			return err
+	// The activity waits on this signal before starting sync for this poll, so
+	// a query-based source's schema - known as soon as the first row is read
+	addRecord := func(addCtx context.Context, record model.Record[model.RecordItems]) error {
+		if !signaledNotEmpty {
+			signaledNotEmpty = true
+			req.Stream.SignalAsNotEmpty()
```
jgao54, line 159: better to signal after `AddRecord` to avoid any weird racy-ness

jgao54 (commented):
general style nit: there's a mixed terminology usage for "isolated cdc" and "query cdc" (as well as both "cdc table" and "table cdc") across variable name, method name, table definitions, etc.  would be good to introduce a single terminology/concept for the query-based cdc

flow/connectors/external_metadata/query_cdc_replication_state.go
dtunikov: yes yes
InitializeTableReplicationState only runs on initial mirror setup
so, when a new table is added, it's not going to have an entry in cdc_table_replication_state yet, so we'd need to create it

flow/connectors/external_metadata/table_replication_state.go
dtunikov: makes sense, fixed

flow/connectors/external_metadata/store.go
dtunikov: yep, fixed!

flow/activities/flowable_isolated_cdc.go
dtunikov: yeah, makes sense
https://github.com/PeerDB-io/peerdb/pull/4727/commits/f20447b6ffc03ccad5e1a129e177de4efebc0499

dtunikov: put a default to 4 for now
will adjust once we see how it works in prod

flow/activities/flowable.go
dtunikov: in the future we might introduce query-based mode for OTLP connectors as well (theoretically)
i think deriving it from the flow config is fine here

flow/activities/flowable.go
dtunikov: yeah, created a follow-up [PR](https://github.com/PeerDB-io/peerdb/pull/4768)
it also has some fixed related to context cancellation

flow/connectors/external_metadata/query_cdc_replication_state.go
dtunikov: hm
good point
i'm checking table addition consistency now
seems like there could be some gaps in the design

dtunikov: Ok, BQ actually never supported table addition. `ExportTxSnapshot` reads table mappings from a catalog flow config, which has the mappings before an update.
We update flowConfig.tableMapping only once table addition is done in `processCDCFlowConfigUpdate`.
So, `SnapshotFlowWorkflow` must send state.tableMappings to `MaintainTx` activity which runs `ExportTxSnapshot` method which exports a table to gcs.

dtunikov: I've put a separate [PR](https://github.com/PeerDB-io/peerdb/pull/4769) to fix table addition.

nexus/catalog/migrations/V56__query_cdc_replication_state.sql
```diff
@@ -0,0 +1,35 @@
...
+    -- synced_batch_id/normalized_batch_id are this table's own sync/normalize
+    -- progress, independent of every other table's; see per-table
+    -- backpressure in flow/activities/flowable_query_cdc.go.
+    synced_batch_id bigint NOT NULL DEFAULT 0,
+    normalized_batch_id bigint NOT NULL DEFAULT 0,
+    last_normalized_at timestamptz,
+    updated_at timestamptz NOT NULL DEFAULT now(),
+    inserts_count bigint NOT NULL DEFAULT 0,
+    updates_count bigint NOT NULL DEFAULT 0,
+    deletes_count bigint NOT NULL DEFAULT 0,
```
jgao54: is it intentional that we are not using `peerdb_stats.cdc_table_aggregate_counts` to store these values? assuming they are aggreagtes

flow/model/model.go
```diff
@@ -92,6 +92,77 @@ type PullRecordsRequest[T Items] struct {
...
+	Env          map[string]string
+	FlowJobName  string
+	TableMapping *protos.TableMapping
+	TableSchema  *protos.TableSchema
+	Version      uint32
+	Flags        []string
+	// StartBatchID is exclusive (the table's last normalized batch), EndBatchID
+	// is inclusive (the table's last synced batch as of this normalize call).
+	StartBatchID      int64
+	EndBatchID        int64
```
jgao54: Nit: i think in normal cdc we derive a `startBatchID` from lastNormBatchId+1, and `endBatchID` from the target normBatchId.  so this is a bit unintuitive. would it be better to name these as 'lastNormalizedBatchId' to be self-explanatory?

flow/internal/clickhouse/settings.go
```diff
@@ -74,6 +76,24 @@ func NewCHSettings(version *chproto.Version, settings ...CHSettingEntry) *CHSett
 	return chSettings
 }
 
+// NewInsertSettings builds settings common to every INSERT ... SELECT into a ClickHouse
+// destination table
+func NewInsertSettings(chVersion *chproto.Version, internalVersion uint32) *CHSettings {
```
jgao54: thanks for consolidating this, was a TODO of mine

flow/connectors/clickhouse/table_function.go
```diff
@@ -28,6 +28,8 @@ type insertFromTableFunctionConfig struct {
 	schema                    types.QRecordSchema
 	excludedColumns           []string
 	fieldExpressionConverters []fieldExpressionConverter
+	// parquetStaged is true when the table function reads Parquet
+	parquetStaged bool
```
jgao54: nit: instead of a bool, more flexible and future-proof to have a `format` field

flow/connectors/utils/monitoring/monitoring.go
```diff
@@ -64,6 +64,20 @@ func UpdateLatestLSNAtTargetForCDCFlow(ctx context.Context, pool shared.CatalogP
 	return nil
 }
 
+// GetMaxCDCBatchID returns the highest batch_id ever recorded for flowJobName
+// in cdc_batches, including batches an AddCDCBatchForFlow call started but
+// never finished (e.g. a worker crash mid-batch).
+func GetMaxCDCBatchID(ctx context.Context, pool shared.CatalogPool, flowJobName string) (int64, error) {
```
jgao54: doesn't look like this is used anywhere in this PR, can be deleted

flow/connectors/utils/stream.go
```diff
@@ -148,6 +148,82 @@ func recordToQRecordOrError(
...
+	records <-chan model.Record[model.RecordItems],
+	destinationTableName string,
+	schema types.QRecordSchema,
+	sourceColumnByDest map[string]string,
+	targetDWH protos.DBType,
+	unboundedNumericAsString bool,
+	numericTruncator model.StreamNumericTruncator,
+	rowCounts *model.RecordTypeCounts,
+) (*model.QRecordStream, error) {
+	businessFields := schema.Fields[:len(schema.Fields)-2]
```
jgao54: nit: this feels a bit brittle,  should have a uinit test at least in case the generated fields drift (or better to update implementation to not depend on magic number)

flow/connectors/clickhouse/query_cdc_sync.go
```diff
@@ -0,0 +1,192 @@
...
+// CDC path, see flow/activities/flowable_query_cdc.go. Pair with
+// NormalizeQueryCDC to insert staged batches into the final table.
+func (c *ClickHouseConnector) SyncQueryCDC(
+	ctx context.Context, req *model.SyncQueryCDCRequest,
+) (*model.RecordTypeCounts, error) {
+	if err := c.ReplayTableSchemaDeltas(
+		ctx, req.Env, req.FlowJobName, []*protos.TableMapping{req.TableMapping}, req.SchemaDeltas, req.Flags,
+	); err != nil {
+		return nil, fmt.Errorf("failed to sync schema changes: %w", err)
+	}
```
jgao54: nit: add a comment here saying this is future-proofing, or just remove if we don't have plan in near-term to support.

flow/activities/flowable.go
```diff
@@ -392,6 +419,10 @@ func (a *FlowableActivity) SyncFlow(
 		return a.Alerter.LogFlowError(ctx, config.FlowJobName, err)
 	}
 
+	if isQueryCDCPath(config, destinationType) {
+		return a.syncFlowQueryCDC(ctx, config, options, srcConn.(connectors.QueryCDCPullConnector))
```
jgao54: should check `shutdown.Load()` similar to the regular CDC path, and return nil (to avoid unnecessary error backoff on retry)

flow/model/model.go
jgao54, line 162: Nit: a bit unintuitive with cdc-based replication [here](https://github.com/PeerDB-io/peerdb/blob/04071cc2ced7a8a04401718c58bd66a3fd5af73a/flow/model/model.go#L189-L192) which represent lastNormBatchId + 1 and lastSynced batch id.

nexus/catalog/migrations/V56__query_cdc_replication_state.sql
jgao54, line 16: is it intentional that we are not using `peerdb_stats.cdc_table_aggregate_counts` to store these values?   i think the values from that table are used in the UI

flow/connectors/clickhouse/table_function.go
jgao54: nit: instead of a bool, more flexible and future-proof to pass the format field directly

jgao54 (approved):
Looks great overall.

flow/connectors/utils/monitoring/monitoring.go
dtunikov: oh, yes, it's left after some refactoring

flow/connectors/clickhouse/table_function_test.go
```diff
@@ -57,3 +57,25 @@ func TestBuildInsertFromTableFunctionQuery(t *testing.T) {
 			query)
 	}
 }
+
+func TestBuildInsertFromTableFunctionQueryArrayTime(t *testing.T) {
```
jgao54: should also add Array(Time)  in`Test_BigQuery_CDC_All_Types`

flow/activities/flowable.go
dtunikov: yeah, I've implemented it in the follow up [PR](https://github.com/PeerDB-io/peerdb/pull/4768) where we split query-cdc with shared-stream-cdc into two distinct functions for readability (also some unused shared-stream-cdc things are gone from the query-cdc branch now)

nexus/catalog/migrations/V56__query_cdc_replication_state.sql
dtunikov: so, yeah, we talked about that with @ilidemi [here](https://github.com/PeerDB-io/peerdb/pull/4727#discussion_r3879848853)
for query cdc having just this table `query_cdc_replication_state` (former `cdc_table_replication_state`) is enough to have all the data we need for the UI.
the view anyways is going to be different since we'd show per table information (see my comment in the discussion above).

flow/activities/flowable_query_cdc.go
```diff
@@ -0,0 +1,474 @@
...
+		if wait := queryCDCPollWait(state.LastAttemptAt, time.Now(), idleTimeout); wait > 0 {
+			logger.Info("[cdc] waiting before next poll", slog.Duration("wait", wait))
+			if err := waitOrDone(ctx, wait); err != nil {
+				return err
+			}
+			continue
+		}
+
+		// bounded parallelism: only pull+sync for up to parallelism tables at once
+		release, err := acquire(ctx, pullSyncSem, logger, "pull-sync")
```
jgao54: sorry for the trickling in review comments, can be done as follow up as it's not correctness issue. the release() are currently all handled correct but if additional error path get added it's often easy to miss.

good practice in general is to use defer with nested function so they don't get missed, e.g.:
```
  result, err := func() (pollResult, error) {
      release, err := acquire(ctx,, pullSyncSem, logger,, "-sync")
      if err != nil {
          return pollResult{}, err
      }
      defer release()

      ... 
  }()
```

although AI probably don't make this mistakes we human do :P

flow/connectors/clickhouse/table_function_test.go
jgao54: should also add Array(Time)  in`Test_BigQuery_CDC_All_Types` for e2e

jgao54:
the catalog migration error is expected, i forgot to mention that they just needed to be added to goose https://github.com/PeerDB-io/peerdb/blob/e9d5da54c91c0a5dca73205b120a101129908124/flow/db/README.md

i plan to remove the refinery migration scripts after the next release (or at least freeze them in some way) so going forward we'd only need to modify goose-side; but this is dependent on a release

flow/activities/flowable_query_cdc.go
```diff
@@ -0,0 +1,474 @@
...
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-ch:
+				continue
+			}
+		}
+
+		// bounded parallelism: only normalize into the destination for up to
+		// normParallelism tables at once.
+		release, err := acquire(ctx, normSem, logger, "normalize")
```
jgao54: nit-pick and not blocking for this pr: would be good practice to wrap acquire in a nested function and call defer release() for this and sync loop; so we don't end up holding locks indefinitely when we introduce new error path and forget to add release() (although more of a human error and less likely ai would make the same mistakes =P ):

e.g.
```
var normCounts *model.RecordTypeCounts
normErr, getConnectorFailed := func() (error, bool) {
	release, err := acquire(ctx, normSem, logger, "normalize")
	if err != nil {
		return err, false
	}
	defer release()
    ... 
}()
``` 
(there's probably a cleaner way to address the different errors)

dtunikov: yep, addressed it in this PR.
https://github.com/PeerDB-io/peerdb/pull/4727/commits/bf1018e1f805683d61641a9ae1497647324a1533

PR #4729, opened 2026-08-25
fix(clickhouse): accept cloud_mode_engine modes 3 and 4 in SMT validation

tpanetti:
## Problem

Creating a CDC mirror to a ClickHouse Cloud service can fail destination validation with:

> ClickHouse service is not migrated to use SharedMergeTree tables, please contact support

on services configured with `cloud_mode_engine = 3`, even though those services create SharedMergeTree tables by default.

## Cause

`CheckIfClickHouseCloudHasSharedMergeTreeEnabled` requires `cloud_mode_engine` to be exactly `2`. The setting is a mode ladder, and modes 3 and 4 are supersets of 2 that still rewrite DDLs to SharedMergeTree:

- `2` — rewrite DDLs to use SharedMergeTree
- `3` — same as 2, except when an explicitly passed remote disk is specified
- `4` — same as 3, plus use Alias instead of Distributed

ClickHouse Cloud has begun rolling out mode 3, so the exact-match check now rejects healthy SMT services. Mirrors never specify a remote disk, so their tables are created as SharedMergeTree under all three modes.

## Fix

Accept values `2`, `3` and `4`. The `changed`/`readonly` requirements are unchanged.

## Test

New test `TestCheckIfClickHouseCloudHasSharedMergeTreeEnabled` drives the check through per-user settings profiles (`CREATE SETTINGS PROFILE ... SETTINGS cloud_mode_engine = N READONLY`), which matches how Cloud pins the setting. It covers: modes 2/3/4 accepted, mode 1 rejected, non-readonly value rejected, and default (unchanged) value rejected. The mode 3/4 cases fail against the previous validation logic.

### DIFF
--- a/flow/pkg/clickhouse/validation.go
+++ b/flow/pkg/clickhouse/validation.go
@@ -57,10 +57,11 @@ func validateDatabaseEngine(ctx context.Context, logger log.Logger, conn clickho
 func CheckIfClickHouseCloudHasSharedMergeTreeEnabled(ctx context.Context, logger log.Logger,
 	conn clickhouse.Conn,
 ) error {
-	// this is to indicate ClickHouse Cloud service is now creating tables with Shared* by default
+	// cloud_mode_engine 2, 3 and 4 all create tables with Shared* engines by default (3/4 only add
+	// carve-outs for explicit remote disks / Distributed); accept any of them as SMT-enabled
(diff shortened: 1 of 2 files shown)


ilidemi (approved):
Thanks for jumping on this!

PR #4735, opened 2026-08-26
fix(flow): classify ProxySQL connection timeout as connectivity err

itsbilal:
This change updates the alerting classifier for MySQL pipes to classify error code 9001 with connection timeout related text as a connectivity error that can be retried and will not page us.. Other errors with the same codes will continue to be classified as Other and result in a page.

Resolves DBI-1040


Jeremyyang920:
If you could add the "Resolves DBI-abc"  to the comment that would be great since the ticket will auto close.

itsbilal:
@Jeremyyang920 Added! Although my experience is that having it just in the branch name is sufficient to get Linear to track the PR status and auto-close.

Jeremyyang920:
> @Jeremyyang920 Added! Although my experience is that having it just in the branch name is sufficient to get Linear to track the PR status and auto-close.

@itsbilal My bad! I didn't look at the branch name was just glancing at the description. But as long as it closes all good.

jgao54 (approved):

PR #4737, opened 2026-08-26
Fix PK and Nullable display in mirror columns UI

ilidemi:
Nullable wasn't plugged on the backend, PK wasn't reading the field correctly in the UI
Now:
Screenshot 2026-08-26 at 4 29 20 PM


jgao54 (approved):
Didn't even know we had this view 🤦‍♀️ !!

PR #4738, opened 2026-08-27
feat(bigquery): fetched bytes via custom http client

dtunikov:
Use custom http client for BigQuery to calculate transferred bytes for billing.
Previously we used json.Marshal which put a lot more load on the CPU.


flow/connectors/bigquery/bigquery.go
```diff
@@ -104,10 +104,15 @@ func NewBigQueryConnector(ctx context.Context, config *protos.BigqueryConfig) (*
 		return nil, fmt.Errorf("failed to create credentials: %v", err)
 	}
 
+	meteredHTTPClient, err := newMeteredBigQueryHTTPClient(ctx, creds)
```
itsbilal: nit: this is already inside the bigquery package, no need to put BigQuery in the name?

itsbilal (approved):

flow/connectors/bigquery/metered_http.go
```diff
@@ -0,0 +1,67 @@
...
+// BigQuery client whose RoundTripper reports consumed response body bytes via
+// withByteCounter.
+func newMeteredBigQueryHTTPClient(ctx context.Context, creds *auth.Credentials) (*http.Client, error) {
+	transport, err := htransport.NewTransport(ctx, &meteredRoundTripper{base: http.DefaultTransport},
+		option.WithAuthCredentials(creds))
+	if err != nil {
+		return nil, err
+	}
+	return &http.Client{Transport: transport}, nil
+}
```
jgao54: would be good if we stay close to google's implementation and use their `NewClient` API for creating an http client:  https://github.com/googleapis/google-api-go-client/blob/f03a404252af72da386c7cd3b80d99f7516c7008/transport/http/dial.go#L34-L60 

```suggestion
func newMeteredBigQueryHTTPClient(ctx context.Context, creds *auth.Credentials) (*http.Client, error) {
	client, _, err := htransport.NewClient(ctx, option.WithAuthCredentials(creds))
	if err != nil {
		return nil, err
	}
	client.Transport = &meteredRoundTripper{base: client.Transport}
	return client, nil
}

```

flow/connectors/bigquery/cdc.go
```diff
@@ -478,12 +468,11 @@ func (c *BigQueryConnector) pullTableChanges(
 
 	var qfields []types.QField
 	var changeCols bigQueryChangeColumns
-	var bytesForwarded int64
 	for {
 		var row []bigquery.Value
 		if err := it.Next(&row); err != nil {
 			if errors.Is(err, iterator.Done) {
-				return bytesForwarded, nil
+				return bytesTransferred.Load(), nil
```
jgao54, line 479: one thing worth checking here is that given it's HTTP response, i would imagine an idle query would still result in non-zero bytesTransferred. it may be better to conditionally return bytesTransferred.Load() when at least one row has been processed. Otherwise the metrics may be confusing as there would be bytes fetched when no data is actually replicated.

jgao54:
Two NITs, otherwise lgtm

flow/connectors/bigquery/cdc.go
dtunikov: yep, makes sense, made it conditional.

PR #4739, opened 2026-08-27
feat(bigquery): implement watermark column query based replication for bigquery sources

dtunikov:
This PR implement the third BigQuery replication mode - query based replication based on user-specified watermark/cursor column (for example `created_at`).

- Added a new field to `TableMapping` called watermark_column. It's a required field for BigQuery sources with query-based replication mode.
- Implemented new `pullTableQuery` function in biguqery/cdc.go. It fetches data from the table using a simple SQL query like: `SELECT ... FROM ... WHERE col > lower AND col <= upper`.
- Adjusted initial snapshot checkpoint calculation logic. For query-based CDC we use `SELECT max(watermark_column) FROM table` as a snapshot boundary and store it as an initial CDC checkpoint. So, in this mode we don't use `FOR SYSTEM_TIME AS OF TIMESTAMP`, instead we query: `SELECT ... FROM ... WHERE col < max(watermark_col)`.

Resolves DBI-1056.


protos/flow.proto
```diff
@@ -52,6 +52,9 @@ message TableMapping {
   // Only meaningful when the mirror's BigqueryCdcConfig.replication_mode is
   // BIGQUERY_REPLICATION_MODE_EVENTS.
   BigqueryCdcEventsFunction bigquery_cdc_events_function = 10;
+  // the column to use as a cursor for query-based CDC replication
+  // required if replication_mode is BIGQUERY_REPLICATION_MODE_QUERY
+  string watermark_column = 11;
```
dtunikov: The plan is to have an option on the UI to configure `watermark_column` per table + to be able to type in a global `watermark_column` name (since in most cases the same column name is used for all tables, like `created_at`).
I think it can be handled completely on the UI and here we just get the column for every configured table. (e.g. we won't need to store `global_watermark_column` on the flow config lvl)

dtunikov: another thing is that I decided not to add `bigquery_` prefix to its name, because it might be used for other query-based CDC connectors in the future.

flow/connectors/bigquery/qrep_object_pull.go
```diff
@@ -557,19 +638,42 @@ func (c *BigQueryConnector) SetupReplication(
...
+// watermark_column <= that same bound so CDC's first poll (watermark_column > checkpoint)
+// picks up exactly where the snapshot left off.
+func (c *BigQueryConnector) setupQueryModeReplication(
+	ctx context.Context,
+	req *protos.SetupReplicationInput,
+	cfg *protos.FlowConnectionConfigsCore,
+) (map[string]time.Time, error) {
+	checkpointByTable := make(map[string]time.Time, len(cfg.TableMappings))
+	for _, tableMapping := range cfg.TableMappings {
+		watermark, err := c.maxWatermarkValue(ctx, tableMapping.SourceTableIdentifier, tableMapping.GetWatermarkColumn())
```
dtunikov: We could also use BQ current timestamp as an initial watermark value.
I don't have a strong preference here, both ways should work fine.

protos/flow.proto
```diff
@@ -35,8 +35,9 @@ message ColumnSetting {
 // EVENTS-mode CDC polling (see BigQueryReplicationMode) for one table:
 // APPENDS() (insert-only) or CHANGES() (insert/update/delete).
 enum BigqueryCdcEventsFunction {
-  BIGQUERY_CDC_EVENTS_FUNCTION_APPENDS = 0;
-  BIGQUERY_CDC_EVENTS_FUNCTION_CHANGES = 1;
+  BIGQUERY_CDC_EVENTS_FUNCTION_UNSPECIFIED = 0;
```
dtunikov, line 38: since it's not used in production yet - it's safe to update these enums
would make validation more predictable in peerdb and clickpipes

flow/connectors/bigquery/cdc.go
```diff
@@ -525,3 +533,97 @@ func (c *BigQueryConnector) pullTableChanges(
...
+	dsTable, err := c.convertToDatasetTable(sourceTableIdentifier)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse table identifier %s: %w", sourceTableIdentifier, err)
+	}
+
+	buildQueryModePullQuery := func(dsTable string, watermarkColumn string, exclude map[string]struct{}) string {
+		col := quotedIdentifier(watermarkColumn)
+		return fmt.Sprintf("SELECT *%s FROM %s WHERE TIMESTAMP(%s) > @start AND TIMESTAMP(%s) <= @end ORDER BY %s",
+			exceptClause(exclude), dsTable, col, col, col)
+	}
```
Jeremyyang920, line 546: nit: Could we maybe extra this to a package level function, then above for runPullQuery we can use a named type instead of the closure definition. Would also ensure that if we ever change the signature, we don't need to update it in multiple places.

flow/connectors/bigquery/cdc.go
```diff
@@ -525,3 +533,97 @@ func (c *BigQueryConnector) pullTableChanges(
...
+	locateColumnIndex := func(schema bigquery.Schema, name string) int {
+		for i, field := range schema {
+			if field.Name == name {
+				return i
+			}
+		}
+		return -1
+	}
```
Jeremyyang920: This seems quite excessive for what it is used for and only used once. If I recall stdlib semantics correctly, you can just use slices.IndexFunc for this.

flow/connectors/bigquery/cdc.go
dtunikov: will do that in the follow-up PR where I replaced EXCEPT with explicit column selection set (since this code will change a bit there)

protos/flow.proto
```diff
@@ -52,6 +53,9 @@ message TableMapping {
   // Only meaningful when the mirror's BigqueryCdcConfig.replication_mode is
   // BIGQUERY_REPLICATION_MODE_EVENTS.
   BigqueryCdcEventsFunction bigquery_cdc_events_function = 10;
+  // the column to use as a cursor for query-based CDC replication
+  // required if replication_mode is BIGQUERY_REPLICATION_MODE_QUERY
+  string query_cdc_watermark_column = 11;
```
dtunikov, line 58: no bigquery prefix since it will be used by other query-cdc connectors in the future

flow/pkg/bigquery/validation.go
```diff
@@ -56,3 +77,306 @@ func TablesHaveChangeHistoryEnabled(
...
+	for _, col := range t.Columns {
+		if col.IsPK {
+			return true
+		}
+	}
+	return false
+}
+
+// GetTables returns information about the specified tables
+func GetTables(
```
dtunikov, line 103: if we encounter issues with datasets that have a lot of tables - we can consider using plain SELECT over bigquery system tables.

flow/pkg/bigquery/validation.go
```diff
@@ -56,3 +77,306 @@ func TablesHaveChangeHistoryEnabled(
 	}
 	return result, nil
 }
+
+// ColumnInfo describes a single column of a BigQuery table.
+type ColumnInfo struct {
```
dtunikov, line 82: tried to move everything related to validation here so that it'd be re-usable from the ClickPipes

flow/connectors/bigquery/source.go
```diff
@@ -4,127 +4,102 @@ import (
...
 	"cloud.google.com/go/storage"
 	"google.golang.org/api/iterator"
 
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
 	bqvalidate "github.com/PeerDB-io/peerdb/flow/pkg/bigquery"
-	"github.com/PeerDB-io/peerdb/flow/pkg/common"
 	"github.com/PeerDB-io/peerdb/flow/shared/exceptions"
 )
 
 func (c *BigQueryConnector) ValidateMirrorSource(ctx context.Context, cfg *protos.FlowConnectionConfigsCore) error {
```
dtunikov, line 16: logic was moved to pkg so that it'd be easy to share with the ClickPipes
ValidateMirrorSource just builds `bqvalidate.SourceConfig` and passed it to pkg functions

flow/connectors/bigquery/qrep_object_pull.go
```diff
@@ -451,15 +495,19 @@ func buildBigQueryExportSQL(
 	}
 
 	uri := fmt.Sprintf("%s/%s/*.parquet", snapshotStagingPath, url.PathEscape(sourceTableIdentifier))
-	snapshotLiteral := snapshotTime.UTC().Format("2006-01-02 15:04:05.999999")
+	boundLiteral := bound.UTC().Format("2006-01-02 15:04:05.999999")
+
+	from := fmt.Sprintf("%s FOR SYSTEM_TIME AS OF TIMESTAMP('%s UTC')", dsTable.stringQuoted(), boundLiteral)
+	if watermarkColumn != "" {
+		from = fmt.Sprintf("%s WHERE TIMESTAMP(%s) <= TIMESTAMP('%s UTC')",
+			dsTable.stringQuoted(), quotedIdentifier(watermarkColumn), boundLiteral)
```
itsbilal, line 503: `watermarkColumn` should never be pre-quoted right?

flow/pkg/bigquery/validation.go
itsbilal: for my own understanding: we'll also call this directly from `discovery` and that's why we moved this here?

flow/connectors/bigquery/source.go
```diff
@@ -4,127 +4,102 @@ import (
...
+		Tables:         tables,
 	}
 
-	bucket := c.storageClient.Bucket(stagingPath.Bucket())
-
-	it := bucket.Objects(ctx, &storage.Query{Prefix: stagingPath.QueryPrefix()})
-	_, err = it.Next()
-	if err != nil && !errors.Is(err, iterator.Done) {
-		return fmt.Errorf("failed to access staging bucket: %w", exceptions.NewBigQueryError(err))
+	tablesByKey, err := bqvalidate.ValidateSourceTables(ctx, sourceConfig)
```
itsbilal, line 69: nit: maybe we can have a function in bqvalidate that calls both this and `ValidateSourceCDC` in the CDC case? Feels a little awkward to thread state from one call into another (`tablesByKey`).

itsbilal (commented):
Mostly looks good to me, just some clarifying questions. I'll look at the tests next!

flow/e2e/bigquery_cdc_test.go
```diff
@@ -258,6 +272,52 @@ func (s BigQueryClickhouseSuite) Test_BigQuery_CDC_Appends_Insert_Only() {
 	RequireEnvCanceled(t, env)
 }
 
+// Test_BigQuery_CDC_Query_Mode covers BIGQUERY_REPLICATION_MODE_QUERY: a plain
+// SELECT ... WHERE watermark_column > lower AND watermark_column <= upper scan,
+// rather than APPENDS()/CHANGES(). The initial snapshot is bounded by the
+// watermark column's max value at setup time instead of FOR SYSTEM_TIME AS OF.
+func (s BigQueryClickhouseSuite) Test_BigQuery_CDC_Query_Mode() {
```
itsbilal, line 279: nit: I see this is a convention in this file but usually we don't use underscore in test function names right? It'd just be `TestBigQueryCDCQueryMode`

itsbilal (approved):

flow/e2e/bigquery_cdc_test.go
dtunikov: oh, yeah, it's a convention + it's not an actual Test function
there is an e2e tests framework that has a single Test function per source (e.g. `TestBigQueryClickhouseSuite` in this case)
it calls e2eshared.RunSuite which internally runs all methods of BigQueryClickhouseSuite that have Test prefix
So, theoretically it could have been TestAbc, but this might get confusing with actual go Test functions..

flow/connectors/bigquery/source.go
dtunikov: yeah, thought about the same, but couldn't find the right abstraction
refactored it a bit
made a single ValidateSource exposed function that has all validations inside

PR #4740, opened 2026-08-27
feat(bigquery): select only those columns that were configured via table mappings

dtunikov:
Select only the columns configured for the table instead of relying on `SELECT * EXCEPT (...)`.

## Dropped source columns

- On the first pull for a table, fetch its BigQuery metadata and cache any configured mirror columns that are already absent.
- If a column disappears after initialization, retry the failed query without that column and add it to the same in-memory cache.
- Keep the cache for the connector lifetime, avoiding repeated query failures and warning logs.
- A resume or deploy reconstructs the connector and checks metadata again, so a column re-added under the same name is detected then.

A column re-added while the same connector remains running stays excluded until connector reconstruction. Data written to that column during the gap cannot be recovered automatically.


jgao54:
> store it in in-memory connector map with TTL
This means that there could be gaps in the data for such column if it was removed and re-added later on the source (up to configured TTL)

given the TTL is going to data loss gaps anyways,  could we keep it simple by: 
- in PullTableRecords call, if the table has no key in the in-memory map,  call table.Metadata() once, and add the table to the map (and any "missing columns" detected)
- when polling fails with missing column, update the in-memory map
- during the next deploy (or a resume), connector will be reconstructed and column that is added back get detected

i think the common case for when a column is dropped is that it doesn't get added back, so it would be a bit annoying to log error every time TTL expires. 

the downside is this takes longer than the TTL approach to detect re-added column; but given how infrequent this scenario is (and given data gap is unavoidable in either case) I think we could get away with it. wdyt?

flow/shared/types/kind.go
```diff
@@ -71,6 +75,83 @@ func (kind QValueKind) IsArray() bool {
...
+		return QValuePoint{}
+	case QValueKindCIDR:
+		return QValueCIDR{}
+	case QValueKindINET:
+		return QValueINET{}
+	case QValueKindMacaddr:
+		return QValueMacaddr{}
+	default:
+		return QValueNull(kind)
+	}
```
jgao54, line 152: the method `DefaultValue` only run on non-null branch, but defaulting the value here as `QValueNull(kind)` seems contradicting, should this be erroring out instead? 

separately, should the Array types also be handled here?

jgao54: actually, `DefaultValue` may not be necessary given `input_format_null_as_default=1`  setting in clickhouse by default, i think the original `val = types.QValueNull(field.Type)` should just work ?

jgao54: ah, re: otherwise the destination will reject the row

I think less so that destination is rejecting but avro is rejecting the value

flow/connectors/bigquery/cdc.go
```diff
@@ -308,70 +341,74 @@ func (c *BigQueryConnector) runPullQuery(
...
-					slog.String("table", sourceTableIdentifier), slog.String("column", col))
-				continue
-			}
-			next[col] = struct{}{}
-		}
-		effective = next
+		c.logger.Warn("[bigquery] column no longer exists on source table, dropping from SELECT list",
+			slog.String("table", sourceTableIdentifier), slog.String("column", missingCol))
+		c.recordMissingSourceColumn(sourceTableIdentifier, missingCol)
+		effective = slices.DeleteFunc(slices.Clone(effective), func(col string) bool { return col == missingCol })
```
jgao54, line 351: we shouldn't allow ordering key to be dropped; should error if the dropped column is the watermark column

EDIT: i noticed we also don't error today for postgres/mysql either, instead we log an error on source column drop (e.g. "Detected dropped column %s in table %s, but not propagating"),  so fine to leave as is for now. and we can address it across connectors if it becomes an issue.

jgao54 (approved):

jgao54:
Slightly more efficient approach (created PR against your branch) that avoids unconditionally fetch table metadata when pipe is healthy: https://github.com/PeerDB-io/peerdb/pull/4828.

PR #4742, opened 2026-08-29
migrate migration to goose

jgao54:
Migrate db migrations from rust's refinery-based script to go's goose-based script. 

The PR:
- copies and converts all refinery migration scripts to goose, both using .sql files. 
- add mechanism that allow for bootstrap goose from any versions of PeerDB previously on refinery
- add tests that ensures the two migrations provides identical db schemas (until the cutover, we expect to maintain both migration scripts, which is trivial to maintain, and this test will fail on any divergence)
- removed migration CI check (in favor of go test that covers a superset)
- add db migration as first step of start-maintenance to allow removing the refinery-based hook, currently disabled.

Fixes: DBI-1028


flow/db/migrations.go
```diff
@@ -0,0 +1,123 @@
...
+var migrationsFS embed.FS
+
+// Goose already uses a session-level advisory lock during migrations.
+// This distinct lock protects the bootstrap-from-refinery process.
+// crc32.ChecksumIEEE([]byte("peerdb"))
+const bootstrapLockID int64 = 311574919
+
+const (
+	defaultRefineryTableName = "refinery_schema_history"
+	defaultGooseTableName    = "goose_db_version"
```
ilidemi: There's goose.DefaultTablename

flow/db/migrations.go
```diff
@@ -0,0 +1,123 @@
...
+	if err := store.CreateVersionTable(ctx, tx); err != nil {
+		return err
+	}
+	// Goose inserts version 0 when it creates its ledger.
+	if err := store.Insert(ctx, tx, database.InsertRequest{Version: 0}); err != nil {
+		return err
+	}
+	// Insert the refinery schema history into the goose ledger.
+	if _, err := tx.ExecContext(ctx,
+		fmt.Sprintf(`INSERT INTO %s (version_id, is_applied)
```
ilidemi: Should we let goose insert the versions as it sees fit via its API? Would just need to read them ourselves from refinery

flow/db/migrations_test.go
```diff
@@ -0,0 +1,327 @@
...
+	for _, migration := range files {
+		if migration.version > upToVersion {
+			break
+		}
+		tx, err := conn.Begin(ctx)
+		require.NoError(t, err)
+		_, err = tx.Exec(ctx, migration.sql)
+		require.NoError(t, err, "refinery-style apply of V%d__%s failed", migration.version, migration.name)
+		_, err = tx.Exec(ctx,
+			"INSERT INTO refinery_schema_history (version, name, applied_on, checksum) VALUES ($1, $2, $3, $4)",
```
ilidemi: Mocking refinery in a compatibility test doesn't sound great. Would this work:
* For full migration, invoke real nexus with something like this (can be opted out in local)
```
cmd := exec.CommandContext(
    ctx,
    "docker", "exec", "peerdb-server",
    "./peerdb-server",
    "--migrations-only",
    "--catalog-database", customDatabase,
)
output, err := cmd.CombinedOutput()
if err != nil {
    return fmt.Errorf("running Nexus migrations: %w\n%s", err, output)
}
```
* For intermediate version test, can check in a compressed pgdump of some real intermediate state or add a CLI hook for https://docs.rs/refinery/latest/refinery/struct.Runner.html#method.set_target to nexus

flow/db/migrations_test.go
```diff
@@ -0,0 +1,327 @@
...
+//	sequence: public.flows_id_seq
+//	rows: public.flows 3
+func schemaFingerprint(t *testing.T, ctx context.Context, connStr string) []string {
+	t.Helper()
+	conn, err := pgx.Connect(ctx, connStr)
+	require.NoError(t, err)
+	defer conn.Close(ctx)
+
+	const excludedTables = `('goose_db_version', 'refinery_schema_history', 'goose_db_version_id_seq')`
+	queries := []string{
```
ilidemi: For schema, would it be better to use pg_dump (which we already have in CI)?
```
  pg_dump
    --schema-only
    --no-owner
    --exclude-table=public.refinery_schema_history
    --exclude-table=public.goose_db_version
    --exclude-table=public.goose_db_version_id_seq
```

jgao54: good idea, simplify schema comparison a lot

flow/db/migrations_test.go
jgao54: good idea, set_target hook simplifies this a lot.

flow/db/migrations.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/PeerDB-io/peerdb/flow/internal"
+)
+
+//go:embed migrations/*.sql
+var migrationsFS embed.FS
+
+// Goose already uses a session-level advisory lock during migrations.
+// This distinct lock protects the bootstrap-from-refinery process.
+// crc32.ChecksumIEEE([]byte("peerdb"))
+const bootstrapLockID int64 = 311574919
```
ilidemi, line 25: nit: could make a var with init()

ilidemi (approved):

flow/db/migrations.go
jgao54: this is intended, so that inspecting pg_locks is easier to match (this is also what goose does). it also makes intent more clear as this is something we don't want to change.  will add a small unit test to avoid someone from accidentally attempting to modify this.

Issue #309, opened 2023-08-11
Support columns with width greater than 16MB for Snowflake QRep via Avro


heavycrystal:
There is a limitation from Snowflake's side on the maximum width of a column when copying records to Snowflake via the Avro intermediary format. The limit is 16MB and QRep mirrors will fail when they have columns with values greater than this. 

For now, we send an empty string instead of causing the mirror to fail.

PR #4743, opened 2026-08-29
Preserve large CDC values for non-Snowflake destinations

ilidemi:
## Problem

`RecordItems` carried a `TruncateThresholdBytes` field set by the source connector — 15 MiB everywhere, 16 MiB + 16 KiB for MongoDB. When a CDC record was serialized into the raw-table JSON, any `String`, `JSON`, or `HStore` value above that threshold was replaced with an empty value, for every destination.

That limit exists for Snowflake alone (#309), and it was applied everywhere. A 20 MiB text column replicated to ClickHouse landed as an empty string, with no error.

## Change

The threshold moves from the source-side record to a destination-aware serialization option, `ToJSONOptions.ClearValuesOverBytes` (0 preserves values). `rawTableJSONOptions` in `flow/connectors/utils/stream.go` sets it, and only for Snowflake.

| Destination | Values over 15 MiB (string / JSON / hstore) |
| --- | --- |
| Snowflake | cleared to `""` / `{}` — unchanged behavior |
| ClickHouse and all others | preserved |

Also in this PR:

- `shared.SnowflakeClearValueThresholdBytes` replaces the four hardcoded `15*1024*1024` literals in the Avro converter, so the raw-table path and the Avro path share one constant.
- `NewMongoRecordItems` is removed; MongoDB now uses `NewRecordItems`. Its 16 MiB + 16 KiB threshold is redundant because the MongoDB server caps documents at 16 MiB.

## Verification

New tests:

- `flow/model/record_items_test.go` — values are cleared at the configured limit and preserved when no limit is set.
- `ClickHouseSuite.Test_Large_Text_CDC` — CDC of a 100 MiB PostgreSQL `TEXT` and a 63 MiB MySQL `LONGTEXT` (just under MySQL's default 64 MiB `max_allowed_packet`), asserting the destination value length matches the source.

The e2e tests were validated to work: temporarily restoring the unconditional truncation made both ClickHouse cases fail with an empty destination value. As a stress check, PostgreSQL CDC of a 1023 MiB string also passed after temporarily raising the local ClickHouse server memory ceiling.

### DIFF
--- a/ancillary-docker-compose.yml
+++ b/ancillary-docker-compose.yml
@@ -135,6 +135,7 @@ services:
     image: ${MYSQL_GTID_IMAGE}
     command:
       - "--binlog-row-metadata=FULL"
+      - "--max-allowed-packet=128M"
     restart: unless-stopped
     volumes:
       - mysql_gtid_data:/var/lib/mysql
@@ -153,7 +154,7 @@ services:
   mysql-pos:
     container_name: peerdb-mysql-pos
     image: ${MYSQL_POS_IMAGE}
-    command: ["--log_bin=mysql-bin", "--server-id=1", "--bind-address=::"]
+    command: ["--log_bin=mysql-bin", "--server-id=1", "--bind-address=::", "--max-allowed-packet=128M"]
     restart: unless-stopped
     volumes:
       - mysql_pos_data:/var/lib/mysql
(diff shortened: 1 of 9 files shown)


dtunikov (approved):

jgao54 (approved):
lgtm, mongo-side safe to remove

PR #4744, opened 2026-08-29
Track CDC event size distribution as a histogram

ilidemi:
Add a histogram for the size of individual CDC events.

The existing `fetched_bytes` counter reports total volume, logs expose delta between timer fires but with enough events can't distinguish 10000 * 50KB from 10000 * 40KB + 100MB. This histogram gives a sense of which sizes are present.

The Postgres path uses an explicit type switch so that keepalive, relation, and transaction-boundary messages are excluded and the distribution reflects row changes only.

## Testing

Deployed to ClickPipes service `378ddc9b-b2e7-49a5-89f2-375ec23f82a6` and exercised with generated load spanning log-scale value sizes from empty to 100 MB.


dtunikov (approved):

jgao54 (commented):
nice, this pr would covers the important part of DBI-319 as well

jgao54 (approved):

PR #4747, opened 2026-08-30
fix(clickhouse): insert into local _shard on clustered replicated targets to avoid silent Distributed+quorum row loss

### LINKED ISSUE #4746: Silent row loss on clustered replicated ClickHouse target: INSERT via Distributed + insert_quorum_parallel=0 drops parts (snapshot and CDC)
## Summary

On a clustered replicated ClickHouse target (`cluster` set plus `replicated=true`), PeerDB loses rows silently. The initial snapshot and every post-snapshot CDC change can fail to show up in the destination, and nothing errors out. PeerDB just reports success and moves on.

The symptom usually looks like "snapshot loads fine, CDC never streams", but the same thing breaks the snapshot too.

## Setup

- PeerDB `stable-v0.37.5`
- Source: PostgreSQL 17, logical decoding on, `pgoutput`, pre-created publication passed via `publication_name`
- Destination: ClickHouse 26.6, `cluster` set plus `replicated=true`, so the final table is `Distributed(<t>_shard)` over a `ReplicatedReplacingMergeTree`. 1 shard, 2 replicas, GCS (S3-compatible) Avro staging
- Kubernetes, bundled Temporal, `do_initial_copy=true`

## Root cause

PeerDB runs all of its data inserts through the `Distributed` table `<t>`. On a real multi-replica cluster, an `INSERT` into a Distributed table with `insert_quorum_parallel=0` (sequential quorum) silently drops the committed part. The Distributed front returns `written_rows=1`, there's no exception, but 0 rows actually land in `<t>_shard` (0 parts, empty `system.distribution_queue`, no shard insert in `query_log`).

That's [ClickHouse #97557](https://github.com/ClickHouse/ClickHouse/issues/97557). PeerDB sets exactly this combination for replicated targets (`insert_quorum=auto`, `insert_quorum_parallel=0`, `select_sequential_consistency=1`), which #4676 added so normalize's read-back stays consistent.

The problem is there's no safe value here as long as the insert goes through the Distributed table:

- `insert_quorum_parallel=0` through Distributed drops the write (#97557).
- `insert_quorum_parallel=1` makes the write land, but read-your-writes stops being honored, so the #4676 normalize race comes back.

`insert_quorum` is really a `ReplicatedMergeTree` level thing. `Distributed` just passes it through and gives no guarantee of its own.

### How I pinned it down

- The pull side is fine. A fresh `pg_logical_slot_peek_binary_changes` on the same publication decodes the changes correctly.
- A snapshot-only mirror (no CDC, no normalize, no second writer at all) still loses the seed row. So it isn't a CDC or normalize race, and it isn't concurrency.
- Bisecting the actual insert settings on the live cluster: the row only disappears with `insert_quorum_parallel=0` and the insert going through the `Distributed` table. The same insert straight into `<t>_shard` with `insert_quorum_parallel=0` lands the row and keeps read-your-writes. Through Distributed with `insert_quorum_parallel=1` it lands but breaks read-your-writes.

I went down a couple of wrong turns first (thought it was normalize, then `replicated=true` vs non-replicated, then the #4676 quorum race). Each got closer but none was the whole story. The Distributed insert drop is the real cause, and it explains the snapshot loss too.

## Fix

Insert straight into `<t>_shard`, where `insert_quorum` actually applies, and leave the `Distributed` table for reads. That's #4747. It routes all three insert paths (snapshot QRep, raw copy, normalize) to the shard when `cluster` is set, resolving the shard name through the existing `getDistributedShardTable` so the post-`RESYNC` `<t>_shard<timestamp>` name is handled. Non-clustered targets don't change.

Verified live on the 2-replica cluster: the snapshot and a live `UPDATE`/`INSERT` both land on both replicas, and `query_log` shows every insert hitting a `_shard` table instead of the Distributed front.

agrim123:
## Problem

On a clustered replicated ClickHouse target (`cluster` set plus `replicated=true`), PeerDB creates a `Distributed` table `<t>` in front of the local `ReplicatedMergeTree` `<t>_shard` and runs all of its data inserts through the Distributed table. On a real multi-replica cluster this loses rows silently: the Distributed front returns `written_rows=1`, nothing errors, but 0 rows land in the shard (0 parts, empty `system.distribution_queue`, no shard insert in `query_log`).

It hits every insert path: the initial snapshot, the CDC raw-table write, and normalize.

### Why

It's [ClickHouse #97557](https://github.com/ClickHouse/ClickHouse/issues/97557). An `INSERT` into a Distributed table with `insert_quorum_parallel=0` (sequential quorum) drops the committed part. PeerDB sets exactly that for replicated targets (`insert_quorum=auto`, `insert_quorum_parallel=0`, `select_sequential_consistency=1`), which #4676 added to keep normalize's read-back consistent.

There's no safe setting while the insert goes through Distributed:

- `insert_quorum_parallel=0` through Distributed drops the write ([#97557](https://github.com/ClickHouse/ClickHouse/issues/97557)).
- `insert_quorum_parallel=1` makes it land, but read-your-writes stops holding, so the #4676 normalize race comes back.

`insert_quorum` is a `ReplicatedMergeTree` level contract. `Distributed` only passes it through and gives no guarantee of its own.

### Fix

Insert straight into the local `<t>_shard` `ReplicatedMergeTree`, where `insert_quorum` actually applies, and keep the `Distributed` table for reads. Direct-to-shard with `insert_quorum_parallel=0` lands the row and keeps read-your-writes.

This routes all three insert paths to the shard when `cluster` is set:

- snapshot QRep, `buildInsertFromTableFunctionQuery` (`table_function.go`)
- snapshot / raw copy, `CopyStageToDestination` (`avro_sync.go`)
- normalize, `NormalizeQueryGenerator.BuildQuery` (`normalize_query.go`), with the target resolved by the caller in `normalize.go`

The shard name comes from the existing `getDistributedShardTable` (reads `engine_full` from `system.tables`), so a post-`RESYNC` name like `<t>_shard<timestamp>` is handled instead of hardcoding `_shard`. Metadata bookkeeping (`_peerdb_destination_table_name`, last-normalized-batch tracking) still keys on the logical destination name. Only the physical `INSERT INTO` target changes, and the Distributed table stays for reads.

## Verification (live, multi-replica cluster)

ClickHouse 26.6, 1 shard, 2 replicas, `ReplicatedReplacingMergeTree`, GCS Avro staging.

- Snapshot: seed row lands on both replicas, and the insert targets `<t>_shard` rather than the Distributed table.
- CDC: a live `UPDATE` and `INSERT` on the source both reach both replicas within one sync cycle.
- `query_log` shows every insert this run (snapshot, normalize, `_peerdb_raw`) going to a `_shard` table, none to the Distributed front.

Before the change the same mirror lost the snapshot row and every post-snapshot change on the replicated target.

## Notes

- Non-clustered targets are unchanged. The `_shard`/Distributed split only exists when `cluster` is set.
- No proto or config change.

Fixes #4746. Related: #4676, ClickHouse/ClickHouse#97557.

### DIFF
--- a/flow/connectors/clickhouse/normalize_query.go
+++ b/flow/connectors/clickhouse/normalize_query.go
@@ -24,7 +24,13 @@ type NormalizeQueryGenerator struct {
 	chVersion                       *chproto.Version
 	Query                           string
 	TableName                       string
-	rawTableName                    string
+	// insertTargetTable overrides the physical INSERT INTO target. When set
+	// (clustered targets), the INSERT goes directly to the local _shard
+	// ReplicatedMergeTree table where insert_quorum is actually contracted;
+	// empty falls back to TableName (the Distributed table). Metadata/raw
+	// filtering still keys off TableName, not this.
+	insertTargetTable string
+	rawTableName      string
 	isDeletedColName                string
 	tableMappings                   []*protos.TableMapping
 	lastNormBatchID                 int64
@@ -51,6 +57,7 @@ func NewNormalizeQueryGenerator(
 	configuredSoftDeleteColName string,
 	version uint32,
 	flags []string,
+	insertTargetTable string,
 ) *NormalizeQueryGenerator {
 	isDeletedColumn := defaultIsDeletedColName
 	if configuredSoftDeleteColName != "" {
@@ -71,6 +78,7 @@ func NewNormalizeQueryGenerator(
 		isDeletedColName:                isDeletedColumn,
 		version:                         version,
 		flags:                           flags,
+		insertTargetTable:               insertTargetTable,
 	}
 }
 
@@ -366,8 +374,16 @@ func (t *NormalizeQueryGenerator) BuildQuery(ctx context.Context) (string, error
 		chSettings.Add(clickhouse.SettingJsonTypeEscapeDotsInKeys, "1")
 	}
 
+	// Physical INSERT target: on clustered replicated targets this is the
+	// resolved local _shard table (insert_quorum is contracted there);
+	// otherwise the Distributed table (== TableName). Reads still go through
+	// the Distributed table.
+	insertTarget := t.TableName
+	if t.insertTargetTable != "" {
+		insertTarget = t.insertTargetTable
+	}
 	insertIntoSelectQuery := fmt.Sprintf("INSERT INTO %s %s %s%s",
-		peerdb_clickhouse.QuoteIdentifier(t.TableName), colSelector.String(), selectQuery.String(), chSettings.String())
+		peerdb_clickhouse.QuoteIdentifier(insertTarget), colSelector.String(), selectQuery.String(), chSettings.String())
 
 	t.Query = insertIntoSelectQuery
 
--- a/flow/connectors/clickhouse/avro_sync.go
+++ b/flow/connectors/clickhouse/avro_sync.go
@@ -49,8 +49,22 @@ func (s *ClickHouseAvroSyncMethod) CopyStageToDestination(ctx context.Context, a
 		return fmt.Errorf("failed to build staging table function: %w", err)
 	}
 
+	// On clustered replicated targets, route the INSERT directly to the local
+	// _shard ReplicatedMergeTree table (where insert_quorum is actually
+	// contracted). Inserting into the Distributed table with insert_quorum
+	// silently drops the part (ClickHouse #97557). Reads still go through the
+	// Distributed table.
+	insertTarget := s.config.DestinationTableIdentifier
+	if s.Config.Cluster != "" {
+		shardTable, err := s.getDistributedShardTable(ctx, s.config.DestinationTableIdentifier)
+		if err != nil {
+			return fmt.Errorf("failed to resolve shard table for %s: %w", s.config.DestinationTableIdentifier, err)
+		}
+		insertTarget = shardTable
+	}
+
 	query := fmt.Sprintf("INSERT INTO %s SELECT * FROM %s",
-		peerdb_clickhouse.QuoteIdentifier(s.config.DestinationTableIdentifier), stagingTableFunction)
+		peerdb_clickhouse.QuoteIdentifier(insertTarget), stagingTableFunction)
 	return s.exec(ctx, query)
 }
 
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -577,6 +577,20 @@ func (c *ClickHouseConnector) NormalizeRecords(
 				continue
 			}
 
+			// On clustered replicated targets, route the physical INSERT
+			// directly to the local _shard ReplicatedMergeTree table (where
+			// insert_quorum is actually contracted). Empty => normalize writes
+			// to the Distributed table as before.
+			var insertTargetTable string
+			if c.Config.Cluster != "" {
+				insertTargetTable, err = c.getDistributedShardTable(ctx, tbl)
+				if err != nil {
+					c.logger.Error("[clickhouse] error while resolving shard table for INSERT target",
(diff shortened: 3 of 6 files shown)


agrim123:
@dtunikov @ilidemi @jgao54 
Hey guys, would it be possible to check this?

flow/connectors/clickhouse/normalize_query.go
```diff
@@ -24,7 +24,13 @@ type NormalizeQueryGenerator struct {
...
-	rawTableName                    string
+	// insertTargetTable overrides the physical INSERT INTO target. When set
+	// (clustered targets), the INSERT goes directly to the local _shard
+	// ReplicatedMergeTree table where insert_quorum is actually contracted;
+	// empty falls back to TableName (the Distributed table). Metadata/raw
+	// filtering still keys off TableName, not this.
+	insertTargetTable string
+	rawTableName      string
 	isDeletedColName                string
 	tableMappings                   []*protos.TableMapping
```
serprex, line 35: formatting

flow/connectors/clickhouse/normalize_query.go
```diff
@@ -24,7 +24,13 @@ type NormalizeQueryGenerator struct {
 	chVersion                       *chproto.Version
 	Query                           string
 	TableName                       string
-	rawTableName                    string
+	// insertTargetTable overrides the physical INSERT INTO target. When set
+	// (clustered targets), the INSERT goes directly to the local _shard
+	// ReplicatedMergeTree table where insert_quorum is actually contracted;
+	// empty falls back to TableName (the Distributed table). Metadata/raw
+	// filtering still keys off TableName, not this.
```
serprex, line 31: comments should not be sloppy

serprex:
seems like a bug which should be fixed in ClickHouse itself

serprex:
to expand on my initial take, slop evaluation:

`getDistributedShardTable` only extracts table name from engine_full; does not evaluate sharding_key or choose destination shard

That means workaround is valid for configuration tested here, one shard with multiple replicas, but not for multiple shards. It also bypasses cityHash64(_peerdb_uid) routing used by raw table. This can cause incorrect results for shard-pruned queries or operations relying on row colocation

We should fix underlying ClickHouse issue. Supporting direct multi-shard writes would require us to reproduce ClickHouse routing semantics, partition rows using arbitrary sharding_key, account for shard weights, and connect directly to each selected shard

agrim123:
Thanks for the detailed look @serprex 

You're right that `getDistributedShardTable` only pulls the local table name and doesn't reproduce the `cityHash64(_peerdb_uid)` sharding. This patch was never meant to be a general multi-shard solution; it targets the single-shard / multi-replica case where the Distributed + insert_quorum_parallel=0 path silently drops committed parts (#4746 / [ClickHouse #97557](https://github.com/ClickHouse/ClickHouse/issues/97557)).

I agree the durable fix belongs in ClickHouse (it's tracked at [ClickHouse/ClickHouse#97557](https://github.com/ClickHouse/ClickHouse/issues/97557)).

Would you accept a narrower version of this, gated behind an explicit "single shard" guard that errors out (rather than silently misrouting) the moment it detects >1 shard? That keeps stock users safe from data loss without PeerDB having to reimplement CH routing.

If neither fits the project's direction, no worries, I'll carry it as a fork. Wanted to offer the safer, narrower option first.

serprex:
I don't think it fits our direction. Distributed table support was already added as a bit of an extra feature, so I don't think there's appetite to take on workarounds for it. I've escalated CH issue on our end, hopefully it gets resolved soon

agrim123:
> I don't think it fits our direction. Distributed table support was already added as a bit of an extra feature, so I don't think there's appetite to take on workarounds for it. I've escalated CH issue on our end, hopefully it gets resolved soon

Thanks @serprex a lot.

PR #4751, opened 2026-08-31
Don't propagate schema delta for excluded columns in MySQL

ilidemi:
Data wasn't being propagated but a column add on the destination would still be unexpected


jgao54 (approved):
nice, had the same issue in CockroachDB connector that was fixed.

PR #4752, opened 2026-08-31
Symlink skills for codex and other agents

ilidemi:


pfcoperez (commented):
I think Claude can also use the .agents directory, would it make sense just to move the directory rather than using symlinks ?

ilidemi:
@pfcoperez Refreshed my knowledge, Claude is still a special snowflake here and wants the skills to be in `.claude` - so at most it could be a symlink the other way, which imo doesn't make much difference. It does support AGENTS.md as a client-side configuration change, but not really for skills.

pfcoperez (approved):

pfcoperez:
> @pfcoperez Refreshed my knowledge, Claude is still a special snowflake here and wants the skills to be in `.claude` - so at most it could be a symlink the other way, which imo doesn't make much difference. It does support AGENTS.md as a client-side configuration change, but not really for skills.

Interesting 😞 , I think it's fine to have your original proposal given the circumstances.

PR #4757, opened 2026-08-31
fix(flow): upgrade to go 1.27

itsbilal:
Bumps version of Go up to 1.27.

Only changes were some transitive
dependency bumps (pebble, swiss) and transposing the `Error()` method on
some errors that also have an Unwrap() method so that the pointer type
itself also implements the `error` interface, which govet now enforces
more strongly when an error is formatted with %w. Other lint changes
were removing excess parenthesis and always-nil return values.


itsbilal:
There's some overlap here with #4753. I'll re-trigger the bot on that branch to pick up the rest if this one merges first.

ilidemi (approved):
Nice!

PR #4759, opened 2026-08-31
fix(postgres): avoid json marshal roundtrip for json columns

itsbilal:
Previously we did a json unmarshal/marshal with a "relaxed number" extension to convert any large numbers that don't fit in a float into a string. Other than the large number case there was no reason to parse the whole document, and for large json values this was starting to greatly increase CDC time spent per row.

This change significantly reduces this processing overhead by walking through the json value token-by-token instead and just swapping out any numbers that can't be cast to float with a stringified version of them instead.


flow/connectors/postgres/cdc.go
```diff
@@ -399,16 +393,17 @@ func (p *PostgresCDCSource) decodeColumnData(
 			return nil, fmt.Errorf("failed to scan json: %w", err)
 		}
 		if text.Valid {
-			if err := p.jsonApi.UnmarshalFromString(text.String, &parsedData); err != nil {
-				p.logger.Error("[pg_cdc] failed to unmarshal json", slog.Any("error", err))
-				return nil, fmt.Errorf("failed to unmarshal json: %w", err)
+			convertedData, err := convertWithRelaxedNumbers(strings.NewReader(text.String))
```
ilidemi: As this change will start ingesting different payloads into the destination, it's better to add the new behavior under a dynamic config and do a safe rollout. We can enable it right away for the struggling service(s) but would be good to have an idea of how much impact would there be on existing mirrors, as well as a way to quickly disable if needed.

`encoding/json` does some niceties that we can live without (sorted keys, html escaping) and things that could be helping CH queries not break (sanitizing invalid utf-8), these are only the ones I know of. A good rollout plan IMO:
* Find out all the things the `encoding/json` was doing that make a difference for CH JSON functions
* Add telemetry for those, can do for all json if they're cheap (i.e. scanning through text.String here) or with sampling if they're expensive
* Add jsontext conversion under a dynconf flag, enabled as needed for heavy customers or if we want to manually gradually rollout
* If telemetry says it's not a big deal, or we can add a mitigation on our end (like pre-check and fallback), or there's a limited set of customers to reach out to, can default to enabled on the next release
* Clean up the old implementation after 1-2 months of no complaints

It would be a bit easier if we already had growthbook integration, but the bad cases should be analyzable (with AI assistance) from `encoding/json` code and service-level conf doesn't seem too bad for this one.

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -487,17 +479,34 @@ func (c *PostgresConnector) parseFieldFromPostgresOID(
...
-			return nil, fmt.Errorf("failed to parse JSON Array: %w", err)
+		// value is an array of pre-marshalled JSON documents; incrementally marshal
+		// the array instead of round-tripping the whole array through json.Marshal.
+		valArr, ok := value.([][]byte)
+		if !ok {
+			return nil, fmt.Errorf("expected pre-marshalled JSON array, got %T", value)
 		}
-		return tmp, nil
+		var builder strings.Builder
+		builder.WriteByte('[')
```
dtunikov: just to clarify
qrep/cdc use the same path for qvalue conversion?
asking because for mysql we had an issue that qrep and cdc were producing JSON values with different formatting

flow/connectors/postgres/json_roundtrip_test.go
```diff
@@ -0,0 +1,249 @@
+package connpostgres
```
dtunikov, line 1: can we add a simple benchmark showing that this new implementation is more performant that the previous one?
yeah, I know that it must be faster since we don't do unmarshal/marshal roundtrip anymore
but still would be good to be sure that `encoding/json/jsontext` doesn't have any weird hidden costs

dtunikov: we don't even have to merge this benchmark code (since we probably want to get rid of jsoniter dependency)
but would be nice to see some results

flow/connectors/postgres/cdc.go
itsbilal: Thanks for this! I ended up bringing back the old code and gating the new behaviour behind a dynconf. Also had Fable evaluate differences between jsoniter/`encoding/json` and the jsontext walkthrough. Turns out we can very easily update the jsontext tokenwise iteration to also start sanitizing invalid utf-8, so I've made that change.

The other issue that would have affected us, was duplicate keys in json objects. PG allows duplicate keys to be written to JSON fields, while for JSONB it does an internal last-duplicate-wins de-duplication. Our old path also does a last-duplicate-wins deduplication, while `jsontext` defaults to erroring out on seeing a duplicate object key token.

I've added an otel counter on the old path observing a duplicate json key. I'm hoping this is rare enough to not be an issue, but if it is, we can either pass in `jsontext.AllowDuplicateNames(true)` to the decoder (which would persist duplicate keys through to the sink, and ClickHouse would do first-duplicate-wins), or we can manually do some sort of last-key-wins logic in the token walkthrough but that would quickly get expensive especially for deep json object trees.

flow/connectors/postgres/qvalue_convert.go
itsbilal: Yes, both of them use `qvalue_convert.go`

flow/connectors/postgres/json_roundtrip_test.go
itsbilal: I wrote one, will put it up in a separate PR as it's the same overall structure as the one I used in #4722. There's a ~17% improvement in end-to-end observed sync times, but since that includes all of sync and normalize too, the actual performance gain in PullRecords is more significant:

```
                         │ ../pg_before2.bench │          ../pg_after2.bench          │
                         │       sec/op        │   sec/op     vs base                 │
PostgresClickHouseCDC-18            88.20 ± 4%   74.78 ± 14%  -15.21% (p=0.005 n=7+6)

                         │    ../pg_before2.bench    │                 ../pg_after2.bench                 │
                         │ catchup_after_insert_s/op │ catchup_after_insert_s/op  vs base                 │
PostgresClickHouseCDC-18                  72.51 ± 6%                 58.89 ± 18%  -18.79% (p=0.002 n=7+6)

                         │ ../pg_before2.bench │           ../pg_after2.bench            │
                         │   replicate_s/op    │ replicate_s/op  vs base                 │
PostgresClickHouseCDC-18            88.20 ± 4%      74.78 ± 14%  -15.22% (p=0.005 n=7+6)

                         │ ../pg_before2.bench │          ../pg_after2.bench           │
                         │       rows/s        │    rows/s     vs base                 │
PostgresClickHouseCDC-18           22.68k ± 4%   26.75k ± 12%  +17.95% (p=0.005 n=7+6)
```

dtunikov: Oh, I think you mean e2e benchmark, right?
I meant a simple unit test like benchmark that compares:
call `convertWithRelaxedNumbers` + strings.Builder for arrays vs call `json.Marshal + json.Unmarshal`
(on a raw bytes stream)

flow/internal/dynamicconf.go
```diff
@@ -566,6 +566,17 @@ var DynamicSettings = [...]*protos.DynamicSetting{
...
+	{
+		Name: "PEERDB_POSTGRES_FAST_PROCESS_JSON_COLUMNS",
+		Description: "Process JSON/JSONB columns by iterating on JSON tokens instead of a full " +
+			"unmarshal/marshal roundtrip. Faster as it avoids wasted CPU cycles on a full JSON " +
+			"marshal/unmarshal, and it still converts out-of-float64-range numbers to strings like the " +
+			"classic path.",
+		DefaultValue:     "false",
+		ValueType:        protos.DynconfValueType_BOOL,
+		ApplyMode:        protos.DynconfApplyMode_APPLY_MODE_IMMEDIATE,
+		TargetForSetting: protos.DynconfTarget_POSTGRES,
```
ilidemi: TargetForSetting would be DynconfTarget_ALL, it means destination

flow/connectors/postgres/json.go
```diff
@@ -19,24 +29,107 @@ func (d *relaxedNumberDecoder) Decode(ptr unsafe.Pointer, iter *jsoniter.Iterato
 		} else {
 			*anyPtr = numberToken.String()
 		}
+	case jsoniter.ObjectValue:
+		// decode objects ourselves to count duplicate keys; last occurrence wins.
+		obj := make(map[string]any)
+		iter.ReadMapCB(func(it *jsoniter.Iterator, field string) bool {
```
ilidemi, line 35: Would that not regress us back where we started with full deserialization and reflection? I think it's ok to be submitting duplicate keys and let CH switch from last to first. The query is showing data that customer sent anyway, we're just delivering more of it.

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -487,17 +487,45 @@ func (c *PostgresConnector) parseFieldFromPostgresOID(
 		boolVal := value.(bool)
 		return types.QValueBoolean{Val: boolVal}, nil
 	case types.QValueKindJSON, types.QValueKindJSONB:
-		tmp, err := parseJSON(value, false)
+		if preMarshalled, ok := value.([]byte); ok {
```
ilidemi: Should we introduce a new type here? `type preMarshalledJSON []byte`

flow/connectors/postgres/json_roundtrip_test.go
ilidemi: +1 to decode+encode benchmark. With just the relaxed number addition it was 5x difference for me, but curious what is the hit from unicode normalization (which we have to do, it seems) and the object decoding would show up

flow/connectors/postgres/json.go
itsbilal: This only applies in the old code path where we're already doing the full deserialization. I imagine you meant that we want to add telemetry to see how common this is in the wild, and if it's not common at all, then we don't need to worry about the subtle change in semantics when we switch to the new code path? Or if we don't really care about the semantics then we may as well go ahead and just do the `AllowDuplicateKeys(true)` option.

flow/connectors/postgres/json_roundtrip_test.go
itsbilal: Added this benchmark to this PR - see `flow/connectors/postgres/json_test.go` for the code. Here's the difference I see when running this on my laptop. Looks like deeper JSON objects benefit more than shallower ones do, but we're still not at 5x (guessing it might be the unicode normalization?):

```
bilal.akhtar@Mohammeds-MacBook-Pro flow % benchstat ./bench_oldmod.bench ./bench_new.bench
goos: darwin
goarch: arm64
pkg: github.com/PeerDB-io/peerdb/flow/connectors/postgres
cpu: Apple M5 Max
                                                │ ./bench_oldmod.bench │         ./bench_new.bench          │
                                                │        sec/op        │   sec/op     vs base               │
ConvertRelaxedNumber/numFields=4/maxDepth=32-18           40.31m ±  6%   23.31m ± 3%  -42.18% (p=0.002 n=6)
ConvertRelaxedNumber/numFields=8/maxDepth=4-18            31.73µ ±  1%   23.86µ ± 1%  -24.78% (p=0.002 n=6)
ConvertRelaxedNumber/numFields=8/maxDepth=8-18            7.507m ±  3%   4.816m ± 3%  -35.84% (p=0.002 n=6)
ConvertRelaxedNumber/numFields=8/maxDepth=16-18            9.692 ±  6%    6.314 ± 3%  -34.85% (p=0.002 n=6)
ConvertRelaxedNumber/numFields=64/maxDepth=2-18           9.303m ± 16%   8.645m ± 1%        ~ (p=0.065 n=6)
ConvertRelaxedNumber/numFields=64/maxDepth=4-18            3.351 ±  1%    3.232 ± 2%   -3.55% (p=0.002 n=6)

File: flow/connectors/postgres/json.go (hunk 5 above)
Comment (ilidemi) [reply flow/connectors/postgres/json.go:35]: I am an idiot, apologies. Let's just go with `AllowDuplicateKeys(true)`.

File: flow/connectors/postgres/json_roundtrip_test.go (hunk 3 above)
Comment (ilidemi) [reply flow/connectors/postgres/json_roundtrip_test.go:1]: Yeah, looks like it's unicode normalization (and encoding overall, which only does normalization for us), and also seems different data results in different perf.

On the same data as above but copying bytes and inserting quotes instead of calling the encoder, 1.75-2.59x become 2.97-3.85x. Data that resulted in around 5x for me (GH issues/commits/events), comes to just below 3x on this impl. The difficulty with copying bytes is that bad unicode becomes unqueryable in CH, but thanks to encoder we can do `AllowInvalidUTF8(false)` on the decoder, do the bytes approach, and if decoder fails, start over with `AllowInvalidUTF8(true)` and the encoder, at a cost of being slower than this impl and in really bad cases even compared to main (but imo it's fine and we can put in telemetry).

Full results (Legacy = main, Committed = this PR, Raw patch = just copy bytes, Raw + repair = copy bytes and restart on bad unicode):
| Payload | Size | Legacy | Committed | Raw patch | Raw + repair |
|---|---:|---:|---:|---:|---:|
| Small | 55 B | 969.0 ns | 525.5 ns · **1.84×** | 312.5 ns · **3.10×** | 313.5 ns · **3.09×** |
| Small + long number | 1.0 KB | 4.25 µs | 2.84 µs · **1.49×** | 2.32 µs · **1.83×** | 2.23 µs · **1.90×** |
| 1 MiB | 1.05 MB | 990.0 µs | 921.5 µs · **1.07×** | 598.7 µs · **1.65×** | 589.0 µs · **1.68×** |
| 1 MiB + long number | 1.05 MB | 965.9 µs | 967.2 µs · **1.00×** | 605.1 µs · **1.60×** | 572.4 µs · **1.69×** |
| Issues | 51.8 KB | 246.1 µs | 82.6 µs · **2.98×** | 44.7 µs · **5.50×** | 44.5 µs · **5.53×** |
| Issues + long number | 52.8 KB | 254.7 µs | 84.9 µs · **3.00×** | 48.1 µs · **5.29×** | 48.1 µs · **5.30×** |
| Issues + invalid Unicode | 51.8 KB | 244.7 µs | 84.5 µs · **2.90×** | 45.4 µs · **5.39×** | 109.3 µs · **2.24×** |
| Commits | 80.3 KB | 293.9 µs | 102.1 µs · **2.88×** | 52.6 µs · **5.59×** | 52.5 µs · **5.60×** |
| Commits + long number | 81.3 KB | 294.6 µs | 102.9 µs · **2.86×** | 58.1 µs · **5.07×** | 58.3 µs · **5.05×** |
| Commits + invalid Unicode | 80.3 KB | 297.0 µs | 112.2 µs · **2.65×** | 57.9 µs · **5.13×** | 141.4 µs · **2.10×** |
| Events | 411.2 KB | 1.75 ms | 654.8 µs · **2.68×** | 353.7 µs · **4.96×** | 353.2 µs · **4.96×** |
| Events + long number | 412.2 KB | 1.79 ms | 656.3 µs · **2.73×** | 370.1 µs · **4.84×** | 369.3 µs · **4.85×** |
| Events + invalid Unicode | 411.2 KB | 1.75 ms | 652.4 µs · **2.69×** | 355.7 µs · **4.92×** | 856.8 µs · **2.04×** |
| Escaped Unicode | 1.05 MB | 2.19 ms | 2.92 ms · **0.75×** | 886.1 µs · **2.47×** | 888.6 µs · **2.46×** |
| Invalid UTF-8 | 1.05 MB | 976.9 µs | 2.22 ms · **0.44×** | 696.3 µs · **1.40×** | 2.59 ms · **0.38×** |
| Lone surrogate | 1.05 MB | 3.07 ms | 2.07 ms · **1.48×** | 624.3 µs · **4.91×** | 2.30 ms · **1.34×** |
| Generated 4/32 | 13.47 MB | 44.24 ms | 21.41 ms · **2.07×** | 13.80 ms · **3.21×** | 13.81 ms · **3.20×** |
| Generated 8/4 | 10.5 KB | 35.2 µs | 20.1 µs · **1.75×** | 11.8 µs · **2.99×** | 11.6 µs · **3.04×** |
| Generated 8/8 | 2.54 MB | 8.07 ms | 4.07 ms · **1.98×** | 2.72 ms · **2.97×** | 2.71 ms · **2.98×** |
| Generated 8/16 | 3.36 GB | 15.66 s | 6.34 s · **2.47×** | 4.12 s · **3.80×** | 4.93 s · **3.18×** |
| Generated 64/2 | 2.30 MB | 9.41 ms | 3.73 ms · **2.53×** | 2.53 ms · **3.72×** | 2.54 ms · **3.70×** |
| Generated 64/4 | 930.15 MB | 3.92 s | 1.51 s · **2.59×** | 1.03 s · **3.80×** | 1.02 s · **3.85×** |

Can be reproduced by rerunning `benchmark-postgres-json.sh` on [this branch](https://github.com/PeerDB-io/peerdb/commit/89dd8871869f1e7b8ca58c505ede844abb2ea201) (takes a few minutes)

Also robot flagged that jsontext has a fast path for `*bytes.Buffer` and the branch is using strings, so a microoptimization is possible.

File: flow/connectors/postgres/json.go (hunk 5 above)
Comment (itsbilal) [reply flow/connectors/postgres/json.go]: Done!

File: flow/connectors/postgres/json_roundtrip_test.go (hunk 3 above)
Comment (ilidemi) [reply flow/connectors/postgres/json_roundtrip_test.go:1]: Do we want to remove the Encoder and get more perf in the good-unicode case?

Comment (itsbilal) [reply flow/connectors/postgres/json_roundtrip_test.go:1]: @ilidemi Done - thanks for the suggestion! We now have a fast-path when no invalid floats or strings are seen, and a slower-but-still-fast path for just invalid floats where we byte-copy the rest. The old path is now only used when we see a parse error.

Comment (itsbilal) [reply flow/connectors/postgres/json_roundtrip_test.go:1]: Forgot to drop this earlier: the same benchmark, now with this version of the code. 5x speedup is back when we hit the fast path!

```
goos: darwin
goarch: arm64
pkg: github.com/PeerDB-io/peerdb/flow/connectors/postgres
cpu: Apple M5 Max
                                                │ ./bench_oldmod.bench │          ./bench_new.bench          │
                                                │        sec/op        │    sec/op     vs base               │
ConvertRelaxedNumber/numFields=4/maxDepth=32-18           45.06m ±  4%   12.23m ± 10%  -72.85% (p=0.002 n=6)
ConvertRelaxedNumber/numFields=8/maxDepth=4-18            35.28µ ±  6%   10.25µ ±  7%  -70.95% (p=0.002 n=6)
ConvertRelaxedNumber/numFields=8/maxDepth=8-18            7.793m ±  7%   2.223m ± 10%  -71.47% (p=0.002 n=6)
ConvertRelaxedNumber/numFields=8/maxDepth=16-18           14.037 ± 25%    2.987 ± 42%  -78.72% (p=0.002 n=6)
ConvertRelaxedNumber/numFields=64/maxDepth=2-18           9.526m ± 16%   1.999m ±  4%  -79.02% (p=0.002 n=6)
ConvertRelaxedNumber/numFields=64/maxDepth=4-18          3504.0m ±  3%   798.7m ±  4%  -77.21% (p=0.002 n=6)
geomean                                                   42.39m         10.49m        -75.27%
```

File: flow/connectors/postgres/json.go [hunk 7]
```diff
@@ -40,3 +45,146 @@ func createExtendedJSONUnmarshaler() jsoniter.API {
...
+		}
+	}
+	// The encoder terminates each top-level value with a newline; callers treat
+	// the result as a single JSON value, so trim it.
+	return bytes.TrimRight(out.Bytes(), "\n"), nil
+}
+
+func convertWithRelaxedNumbers(input *bytes.Buffer, sizeHint int) (preMarshalledJson, error) {
+	// First pass: Try to copy the input byte-for-byte into an output buffer,
+	// while walking through it with a jsontext encoder that's expected to
```
ilidemi:
```suggestion
	// while walking through it with a jsontext decoder that's expected to
```

flow/connectors/postgres/cdc.go
```diff
@@ -399,13 +403,26 @@ func (p *PostgresCDCSource) decodeColumnData(
...
 		if text.Valid {
-			if err := p.jsonApi.UnmarshalFromString(text.String, &parsedData); err != nil {
-				p.logger.Error("[pg_cdc] failed to unmarshal json", slog.Any("error", err))
-				return nil, fmt.Errorf("failed to unmarshal json: %w", err)
-			}
-			if parsedData == nil {
-				// avoid confusing SQL null & JSON null by using pre-marshaled value
-				parsedData = json.RawMessage("null")
+			if p.fastProcessJsonColumns {
+				convertedData, err := convertWithRelaxedNumbers(bytes.NewBufferString(text.String), len(text.String))
```
ilidemi: To avoid extra copy, let's use `shared.UnsafeFastStringToReadOnlyBytes(text.String)` here. Also only the layer interacting with `jsontext` needs to know about the bytes.Buffer fast path, so can have it just accept `[]byte` and construct buffer on its own

flow/connectors/postgres/json.go
```diff
@@ -40,3 +45,146 @@ func createExtendedJSONUnmarshaler() jsoniter.API {
...
+	dec := jsontext.NewDecoder(
+		input,
+		jsontext.AllowInvalidUTF8(false),
+		jsontext.AllowDuplicateNames(true),
+	)
+	out := new(bytes.Buffer)
+	if sizeHint > 0 {
+		// Grow slightly past the sizeHint to account for any whitespace added by
+		// the encoder below. We want to avoid repeat allocations as much as possible.
+		out.Grow(int(float64(sizeHint) * 1.5))
```
ilidemi: There's no encoder whitespace anymore, right? And the default case would be output length equal to input length

flow/connectors/postgres/json.go
```diff
@@ -40,3 +45,146 @@ func createExtendedJSONUnmarshaler() jsoniter.API {
...
+					return nil, err
+				}
+
+				// Float is out of range. Convert tok to a string.
+				rawNumber := tok.String()
+
+				end := int(dec.InputOffset() - readSoFar)
+				start := end - len(rawNumber)
+				if out.Available() < (end + 2) {
+					out.Grow(out.Len() + end + 2)
```
ilidemi: `.Grow()` accepts extra bytes not total bytes, so this would grow it by over 2x. Anything bad about just sticking to `.Write()`/`.WriteByte()` and letting it manage memory by itself? As I understand, `.Grow()` is there for large jumps in size, and we're adding a couple of quotes at most

flow/connectors/postgres/json.go
```diff
@@ -40,3 +45,146 @@ func createExtendedJSONUnmarshaler() jsoniter.API {
...
+				}
+				if _, err := savedReader.Read(out.AvailableBuffer()[:start]); err != nil {
```
ilidemi: Wouldn't it be simpler to slice `inputBytes` into `out.Write()`? Not sure what an extra reader is adding

flow/connectors/postgres/cdc.go
```diff
@@ -399,13 +403,26 @@ func (p *PostgresCDCSource) decodeColumnData(
...
+				if err != nil {
+					p.logger.Error("[pg_cdc] failed to process json", slog.Any("error", err))
+					return nil, fmt.Errorf("failed to process json: %w", err)
+				}
+				if len(convertedData) == 0 {
+					// avoid confusing SQL null & JSON null by using pre-marshaled value
+					convertedData = jsonNullLiteral
+				}
```
ilidemi: This branch was needed for the old path as json `null` gets unmarshaled as nil pointer, in our new path we'd get `null` as four bytes already. Same for qrep_query_executor

flow/connectors/postgres/qvalue_convert.go
```diff
@@ -487,17 +487,45 @@ func (c *PostgresConnector) parseFieldFromPostgresOID(
 		boolVal := value.(bool)
 		return types.QValueBoolean{Val: boolVal}, nil
 	case types.QValueKindJSON, types.QValueKindJSONB:
-		tmp, err := parseJSON(value, false)
+		if preMarshalled, ok := value.(preMarshalledJson); ok {
+			// value is a pre-marshalled JSON document, see convertWithRelaxedNumbers.
+			if len(preMarshalled) == 0 {
+				preMarshalled = jsonNullLiteral
+			}
```
ilidemi: Not needed for the single value case

flow/connectors/postgres/json.go
itsbilal: This was all downstream from 1) taking in a bytes.Buffer in this function and not a `[]byte`, and 2) requiring that we pass a bytes.Buffer down into jsontext.NewDecoder. The decoder would have advanced the reader state of the bytes buffer, so we had to create additional readers to clone the initial state. Now that we just take a byte slice as input, we can read from it directly, so I've made that change.

flow/connectors/postgres/json.go
itsbilal: Removed. This was necessary to have the `out.AvailableBuffer` part work out in the old code.

flow/connectors/postgres/json.go
```diff
@@ -158,33 +148,31 @@ func convertWithRelaxedNumbers(input *bytes.Buffer, sizeHint int) (preMarshalled
 				// Float is out of range. Convert tok to a string.
 				rawNumber := tok.String()
 
-				end := int(dec.InputOffset() - readSoFar)
-				start := end - len(rawNumber)
-				if out.Available() < (end + 2) {
-					out.Grow(out.Len() + end + 2)
+				// It's possible this is the first time we're writing to `out`. Do the first
+				// alloc if it hasn't been done yet.
+				if out.Available() < sizeHint {
```
ilidemi, line 153:
```suggestion
				if readSoFar == 0 {
```
if the intention is to only grow the first time

flow/connectors/postgres/json.go
```diff
@@ -111,40 +111,30 @@ func reencodeWithRelaxedNumbers(input io.Reader, sizeHint int) (preMarshalledJso
...
-	}
 	for {
 		// Read a token from the input.
 		tok, err := dec.ReadToken()
 		if err != nil {
 			if err == io.EOF {
 				break
 			}
-			return reencodeWithRelaxedNumbers(savedReaderForReencode, sizeHint)
+			return reencodeWithRelaxedNumbers(bytes.NewReader(input), sizeHint)
```
ilidemi, line 137:
```suggestion
			return reencodeWithRelaxedNumbers(bytes.NewBuffer(input), sizeHint)
```
for fast path there too

ilidemi (approved):

PR #4761, opened 2026-09-01
Fix classification for missing tables

ilidemi:
Three issues came together:
* A change in snapshotting wasn't hooked up to the shared table missing type
* Temporal error wrapping didn't keep the table name in the message
* Missing table error was classified as notify connectivity when there's already a dedicated code

Fixes: DBI-1094


pfcoperez (approved):
LGTM

PR #4762, opened 2026-09-01
Validate ClickHouse table capacity before mirror creation

ilidemi:
## Summary

Validate CH table count against max_table_num_to_throw before creating the mirror.
Takes into account new CDC mirrors (+ raw table), resyncs (+N tables) and null table resyncs (+1 temp table for all)

It turned out that querying system tables needs extra permissions so if someone creates a restricted role, we're failing open and good that classification is there.

Contributes to DBI-1061


pfcoperez (approved):
LGTM.
Being a bit paranoid, I think we could concentrate all the queries of the DB into a single one thus getting the picture of the DB state in a single round-trip which should provide a more consistent state in a single point in time but I don't think it's worth of it, specially given that `AttachedTable` is not atomic in ClickHouse Cloud.

PR #4764, opened 2026-09-01
perf(ssh): 8 MiB channel receive window via x/crypto fork

ilidemi:
## What

Replaces `golang.org/x/crypto` with [PeerDB-io/crypto](https://github.com/PeerDB-io/crypto) via a `replace` directive in `flow/go.mod`. The fork carries one patch: the SSH channel receive window is raised from upstream's hardcoded 2 MiB to 8 MiB.

## Why

SSH throughput caps at window ÷ round-trip time. With the 2 MiB window, bulk CDC traffic over high-bandwidth, high-latency SSH tunnels runs below link capacity; 8 MiB covers the bandwidth-delay product of the links we see in practice.

## How the fork stays current

1. A daily workflow in the fork rebases the patch onto each new upstream release tag, tests with PeerDB's Go version, and pushes a matching fork tag (fork `vX.Y.N` = upstream `vX.Y.0` + the N extra commits in the fork - count-based patch versions give every cut a fresh tag name).
2. On failure it opens a `sync-failure` issue in the fork repo and bumps it daily until a run goes green.
3. Renovate in this repo follows the fork's tags and bumps the replace directive like any other gomod dependency, in the grouped weekly update PR.

The new Renovate rule opts fork bumps out of `minimumReleaseAge`: the fork ships Go-team x/crypto releases plus a one-line patch, so the aging gate adds no supply-chain protection and would delay CVE fixes.

Details in the [fork's README](https://github.com/PeerDB-io/crypto/blob/1f46416c52a1dc85b2fe1a3add94118def5689a3/README.md).

Closes DBI-1073


ilidemi:
Correction to the description's Renovate paragraph: fork tags do **not** derive from already-aged upstream releases — the daily sync tags within a day of an upstream release. The actual rationale for skipping `minimumRelease­Age` (amended into the commit and config comment): this is x/crypto from the Go team plus our one-line patch, not an ecosystem package where aging guards against supply-chain surprises; the gate would only delay CVE fixes.

dtunikov (approved):

PR #4765, opened 2026-09-01
fix correctness and consistency in date and datetime inference

jgao54:
We don't support type hinting in table creation for MongoDB. The default date/datetime inference settings in ClickHouse are:
- input_format_try_infer_dates = 1
- input_format_try_infer_datetimes = 1

The problem with this is depending on the first row in a insert block, a date string will be inferred as DateTime64 if it's <1970, and Date if it's >1970. The former is fine, but the latter means if subsequent rows in the insert block contains <1970 date, they get clamped to 1970 and result in incorrect data propagation that can only be fixed with updating clickhouse user settings and resync pipe.

The fix is to disable `input_format_try_infer_dates` (unfortunately there is not a setting for date32 inference), and always have both `input_format_try_infer_datetimes` and `input_format_try_infer_datetimes_only_datetime64` enabled.  This allows both date string and datetime string to be inferred as DateTime64 going forward. 

Use internal versioning for backward compatibility (although we could probably get away with applying to existing pipes without interval versioning, as the indeterminism is on the insert-block level). 

Fixes: DBI-1086

### DIFF
--- a/flow/connectors/clickhouse/object_sync.go
+++ b/flow/connectors/clickhouse/object_sync.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
 	"github.com/PeerDB-io/peerdb/flow/internal"
+	chinternal "github.com/PeerDB-io/peerdb/flow/internal/clickhouse"
 	"github.com/PeerDB-io/peerdb/flow/model"
 	peerdb_clickhouse "github.com/PeerDB-io/peerdb/flow/pkg/clickhouse"
 	"github.com/PeerDB-io/peerdb/flow/shared"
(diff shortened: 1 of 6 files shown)


flow/connectors/clickhouse/object_sync.go
```diff
@@ -322,6 +323,13 @@ func (c *ClickHouseConnector) insertFromURLBatch(
 		fieldExpressionConverters: fieldExpressionConverters,
 	}
 
+	chSettings := chinternal.NewCHSettings(c.chVersion)
```
ilidemi, line 326: If we're expecting JSON destination columns in BigQuery, should it get the full set of related settings? Then can have a helper that builds all settings QRep currently does and reuse it in three places

ilidemi (approved):

flow/connectors/clickhouse/object_sync.go
jgao54: yeah, i didn't do so in this PR as it's a bit out-of-scope, but agree it is a good follow-up.

PR #4768, opened 2026-09-02
chore(bigquery): extract shared-stream CDC path into syncFlowSharedStream

dtunikov:
Follow-up refactor on top of `bq/isolate-tables-flow`. The shared-stream path behaves the same as before. The isolated path gets two small behavior fixes.

### What changed

`SyncFlow` now does only the setup both CDC paths need: cancellable context, flow name and operation context, worker-stop handling, destination peer type, source connector plus `SetupReplConn`. It then hands off to `syncFlowIsolatedTables` or the new `syncFlowSharedStream`.

Resolves: DBI-1045

### DIFF
--- a/flow/activities/flowable.go
+++ b/flow/activities/flowable.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"slices"
 	"strconv"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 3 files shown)


flow/activities/flowable.go
```diff
@@ -354,28 +355,10 @@ func (a *FlowableActivity) SyncFlow(
...
-		}
-		return fmt.Sprintf(
-			"currentSyncFlowNum:%d, totalRecordsSynced:%d, syncingBatchID:%d (%s), normalizingBatchID:%d%s",
-			currentSyncFlowNum.Load(), totalRecordsSynced.Load(),
-			sBatchID, *syncState.Load(), nBatchID, nWaiting,
-		)
-	})
-	defer shutdown()
+	// Heartbeats connector setup below, which must not exceed the activity's heartbeat
+	// timeout. Each sync path stops it and installs its own heartbeat once it takes over.
```
jgao54: should call `stopSetupHeartbeat()` just before `if isQueryCDCPath(config, destinationType)`  to avoid having 2 heartbeat routines in this Activity and actually reflect this comment. Heartbeats are billable events in temporal, so while it doesn't hurt it's not necessary.

flow/activities/flowable_query_cdc.go
```diff
@@ -35,11 +35,19 @@ func (a *FlowableActivity) syncFlowQueryCDC(
...
 	srcConn connectors.QueryCDCPullConnector,
+	shutDown *atomic.Bool,
 ) error {
 	flowName := config.FlowJobName
 	logger := internal.LoggerFromCtx(ctx)
 	pgMetadata := connmetadata.NewPostgresMetadataFromCatalog(logger, a.CatalogPool)
 
+	var totalRecordsSynced atomic.Int64
+	shutdown := common.HeartbeatRoutine(ctx, func() string {
+		return fmt.Sprintf("isolated CDC: %d tables, totalRecordsSynced:%d",
```
jgao54: nit: "query-based CDC" before this got moved is more consistent (same as logs below)

jgao54 (approved):

PR #4769, opened 2026-09-02
fix(bigquery): give added tables a snapshot and a CDC checkpoint

dtunikov:
## What changed

Fix table additions for running query-based CDC mirrors. The child addition flow does not run `SetupReplication`, and the catalog still contains the old table mappings while it runs.

- Pass the child flow's `TableMappings` to `ExportTxSnapshot` so it snapshots the added tables.
- Initialize each added table's CDC checkpoint at `snapshotTime + 1us`, using the same helper as `SetupReplication`.
- Leave tables added with initial load skipped unchanged so CDC starts with their first poll.

## Testing

Added `Test_BigQuery_CDC_Table_Addition_Mid_CDC`, covering the added table's initial snapshot, checkpoint creation, continued CDC for both tables, and mirror status.

Resolves: DBI-1091


flow/activities/snapshot_activity.go
```diff
@@ -114,7 +121,7 @@ func (a *SnapshotActivity) MaintainTx(ctx context.Context, sessionID string, flo
 	}
 	defer connClose(ctx)
 
-	exportSnapshotOutput, tx, err := conn.ExportTxSnapshot(ctx, flowName, env)
+	exportSnapshotOutput, tx, err := conn.ExportTxSnapshot(ctx, flowName, tableMappings, env)
```
dtunikov, line 124: ExportTxSnapshot should use the mappings that are passed via state, not those that are stored in the catalog. Catalog table mappings are updated only after MaintainTx succeed.

I'll need to check if it can break existing running temporal activities since it changes the set of input parameters.

flow/connectors/bigquery/qrep_object_pull.go
```diff
@@ -333,10 +334,20 @@ func (c *BigQueryConnector) ExportTxSnapshot(
...
+	// Use the activity's mappings rather than the catalog config: during a table
+	// addition, the catalog still contains the mirror's pre-addition table list.
+	if err := c.exportTablesAsOf(ctx, flowName, tableMappings, runSnapshotStagingPath, snapshotTime); err != nil {
 		return nil, nil, err
 	}
 
+	// Table additions run as InitialSnapshotOnly child flows and do not call
+	// SetupReplication, so seed their query-CDC cursors from this snapshot.
+	if cfg.GetBigqueryCdcConfig() != nil {
+		if err := c.initializeQueryCDCCheckpoints(ctx, flowName, tableMappings, snapshotTime); err != nil {
```
jgao54, line 346: a side-effect of coupling replication requirement with snapshot logic is this will also write to `query_cdc_replication_state` which stays unused when creating a new snapshot-only pipe, not super ideal but also tolerable. 

a bigger gap is for _cdc-only_ table addition, `ExportTxSnapshot` is skipped so that path does not seed a cursor timestamp at all. 

what do you think about following postgres's path in `processTableAddition` where it runs a `AddTablesToPublication` which implements "setup replication logic" for table addition. Can rename the activity name to `AddTablesToReplication` instead to be more generic, and have each connector implement it (no-op if nothing is needed).  The side-effect is that the replication timestamp recorded for CDC will be earlier than the timestamp used for snapshot, but RMT will handle the dedupe (note this is similar to how pg behaves today for table addition, since adding table to publication implies the LSN represent the timestamp when the pipe was paused, which is not the same as the export snapshot timestamp). 

if the one-time duplication is still too expensive for BQ's case, a simpler fix would be to make sure CDC handles the "cdc-only table addition" edge case where cursor is not seeded (e.g. seed it to current timestamp).

latter is probably simpler fix; but former provides a generic framework for setting up replication for table addition (that handles both snapshot+cdc and cdc-only table addition)

PR #4771, opened 2026-09-03
MongoDB CDC: User facing log upon drop and rename collection events

pfcoperez:
Add an user facing alert when `drop` and `rename` events are detected in MongoDB CDC.

Closes: https://linear.app/clickhouse/issue/DBI-955

### DIFF
--- a/flow/connectors/mongo/cdc.go
+++ b/flow/connectors/mongo/cdc.go
@@ -17,6 +17,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
+	"github.com/PeerDB-io/peerdb/flow/alerting"
 	"github.com/PeerDB-io/peerdb/flow/generated/protos"
 	"github.com/PeerDB-io/peerdb/flow/internal"
 	"github.com/PeerDB-io/peerdb/flow/model"
@@ -34,6 +35,8 @@ const (
 	operationTypeUpdate  operationType = "update"
 	operationTypeReplace operationType = "replace"
 	operationTypeDelete  operationType = "delete"
+	operationTypeDrop    operationType = "drop"
+	operationTypeRename  operationType = "rename"
 )
 
 func parseOperationType(s string) (operationType, bool) {
@@ -234,6 +237,11 @@ func (c *MongoConnector) PullRecords(
 	req *model.PullRecordsRequest[model.RecordItems],
 ) error {
 	defer req.RecordStream.Close()
+
+	var alerter *alerting.Alerter
+	if catalogPool.Pool != nil {
+		alerter = alerting.NewAlerter(ctx, catalogPool, otelManager)
+	}
 
 	fullDocumentColumnName := DefaultFullDocumentColumnName
 	if req.InternalVersion < shared.InternalVersion_MongoDBFullDocumentColumnToDoc {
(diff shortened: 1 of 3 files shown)


flow/connectors/mongo/cdc.go
```diff
@@ -235,6 +238,11 @@ func (c *MongoConnector) PullRecords(
 ) error {
 	defer req.RecordStream.Close()
 
+	var alerter *alerting.Alerter
```
dtunikov, line 241: As of now other connectors don't use alerter from within the connector package
errors/warnings are returned back to the caller (in this case `pullAndSyncCore`) and the caller logs errs/warnings.
PullRecords currently doesn't return warnings, but a similar pattern is applied for qrep [here](https://github.com/PeerDB-io/peerdb/blob/f0be0982aa07af69fc8a963ad769fe755d7d9104/flow/activities/flowable_core.go#L550).

imo, it's better to abstract alerter logic from connectors code and just return warnings from PullRecords.
i'd try to keep connector responsibility as narrow as possible since its PullRecords logic is already quite complicated. (e.g. PullRecords responsibility is to fetch the next batch of records from the source and return them to the caller + return errs/warnings if smth went wrong).

pfcoperez: Thank you @dtunikov !

> As of now other connectors don't use alerter from within the connector package

I don't think this sentence is accurate. Postgres connector uses alerts directly. One example is:

https://github.com/PeerDB-io/peerdb/blame/f0be0982aa07af69fc8a963ad769fe755d7d9104/flow/connectors/postgres/cdc.go#L753-L759

> imo, it's better to abstract alerter logic from connectors code and just return warnings from PullRecords

It's cleaner and I prefer the decoupling when possible.

This is a CDC/PullRecords only report and we should consider if it's worth it to follow this preference now and change the contracts of this function.

> PullRecords currently doesn't return warnings

Are you suggesting that we should include in this PR a contract change on `PullRecords` and extending the signature change to all CDC pipes? Boiling it up to `pullAndSyncCore->pullAndSync` ?

If that's the case, there are two considerations:

A)  From the UX experience, the alert might be delayed or get completely lost. I think that's not a blocker but something to consider.
B) Do we want to take this refactor now for this change when we instances of other connectors using alerter directly? Maybe Alerter could be a good abstraction for alerts across the board and the approach to follow deserves some discussion.

I feel a bit uneasy about changing the contract just for this change.  I like your separation of concerns approach and I think it deserves a follow-up discussion/ticket where the abstraction is addressed for all connectors.

dtunikov: Yeah yeah, I understand the consequences. Just really don't like the fact that we put unnecessary dependencies into connector's code. (especially here this Alerter is not even an interface, it's an actual struct that depends on postgres catalog, so we can't mock it for connector unit testing).
It's not blocking comment for this PR, feel free to proceed. 

Regarding pg example, there we at least have a reason to do that - we're stuck in PullRecords due to long-running transaction and want to notify a customer exactly about that.
Here, imo, it'd be completely fine to report this warning from the pullAndSync at the end of each pull iteration.
My concern is that if we postpone things like that and never get back them, it will become a pattern that agents/humans will follow and we'll have to live with it forever :(

ilidemi: There were a few places before where the error would be logged to the user twice. It would happen when the inner func does it, then returns an error, then the outer func does it. There isn't really a way to make the outer func not do it as you'd have to audit every single place where an error can be produced and then still make sure that nesting=2 doesn't log an error from nesting=3. So the current pattern is to log to the user at the outermost point of the activity, or don't return an error (like in the case Dmitrii mentioned).

This still leaves us with a myriad alerter calls to ensure full coverage, taming that complexity is up for grabs

ilidemi: Ok now actually looked through the change and it makes sense here. Double logging doesn't apply as there's no error return, and the flow is clearer when it warns inline. One case we have with accumulation is numeric truncation, and there iirc we first count how many were truncated to report accurately, so that has to be accumulated

ilidemi: Could do a refactor with warning-only alerter, to reduce agent confusion

ilidemi (approved):

flow/connectors/mongo/cdc.go
pfcoperez: >  taming that complexity is up for grabs

I've create a detailed ticket to follow-up and to apply @dtunikov 's suggested pattern to this and other cases where it makes sense by changing the pullrecords contract: https://linear.app/clickhouse/issue/DBI-1097

>  My concern is that if we postpone things like that and never get back them, it will become a pattern that agents/humans will follow and we'll have to live with it forever

With the ticket recorded this is not going to happen :slightly_smiling_face:

PR #4773, opened 2026-09-03
feat(bigquery): classify common BigQuery errors

dtunikov:
## Summary

- classify BigQuery errors by structured API reason before HTTP status
- retry transient backend, rate-limit, stopped-job, and table-unavailable failures
- notify users about access, quota, invalid-query, resource, and other actionable failures
- add HTTP status fallbacks for errors without structured reasons
- emulate raw BigQuery REST responses through `googleapi.CheckResponse` in tests


flow/alerting/classifier.go
```diff
@@ -335,6 +340,47 @@ func (e ErrorClass) ErrorAction() ErrorAction {
...
+		"invalid",
+		"invalidQuery",
+		"notImplemented",
+		"quotaExceeded",
+		"resourceInUse",
+		"resourcesExceeded",
+		"responseTooLarge",
+		"stopped",
+		"timeout":
+		return ErrorNotifyBigQueryError, true
```
jgao54, line 377: nit: classifying errors upfront have the benefit of less noisy alert; the trade-off here is that some of these errors (e.g. invalidQuery) may actually surface implementation issues that we otherwise don't get alerted if it is something on our end that could be fixed, so could be informative for new connectors. Since you have a better grasp of errors in BQ will leave this to your decision on this trade-offs here and which errors to classify upfront vs. later.

jgao54 (approved):

flow/alerting/classifier.go
dtunikov: yeah..
i'll double check these errors and exclude those that can be added after some monitoring in prod

PR #4774, opened 2026-09-04
[DBI-91] connectors/mongodb: Typed mappings

pfcoperez:
These changes introduce the table mappings options for structured ingestion.  

Opening as draft until it addresses the actual document flattening at ingestion. 

Closes: https://linear.app/clickhouse/issue/DBI-1095
Part of: https://linear.app/clickhouse/issue/DBI-91


flow/cmd/validate_mirror.go
```diff
@@ -181,6 +177,51 @@ func (h *FlowRequestHandler) validateCDCMirrorImpl(
...
+	ctx context.Context, cfg *protos.FlowConnectionConfigsCore,
+) APIError {
+	// Check source peer type
+	peer, err := connectors.LoadPeer(ctx, h.pool, cfg.SourceName)
+	if err != nil {
+		return NewInternalApiError(fmt.Errorf("failed to load source peer %s: %w", cfg.SourceName, err))
+	}
+
+	// This condition must evolve as new connectors support structured ingestion.
+	structuredIngestionCompatible := peer.GetMongoConfig() != nil
```
dtunikov, line 191: checking if a peer/sourceConnector supports structured ingestion can be a helper, we'll definitely need the same check in other places as well

pfcoperez: This is the way it is done as I moved this into its separate PR: https://github.com/PeerDB-io/peerdb/pull/4818

PR #4775, opened 2026-09-04
mark qrep failure in db and filter out of result

jgao54:
When a QRep workflow fails (for example a snapshot clone during a CDC initial load or a table addition), its `peerdb_stats.qrep_runs` row keeps `consolidate_complete = false`. `InitialLoadSummary` reports it as in progress forever, and a retried attempt for the same table adds a second row, so the UI shows a table with more than one in-progress run.

Changes:
- Add a `failed` boolean column to `peerdb_stats.qrep_runs`
- `QRepFlowWorkflow` now runs a `MarkQRepRunFailed` activity when it returns an error
- `InitialLoadSummary` filters out failed runs, so only completed and in-progress runs reach the UI.
-  Failed rows and their partitions stay in the catalog, so the history of failed attempts is preserved for debugging.

Testing:
- add integration test
- verified peerdb UI and db locally

Fixes: DBI-1105


flow/workflows/qrep_flow.go
```diff
@@ -510,11 +510,30 @@ func updateStatus(ctx workflow.Context, logger log.Logger, state *protos.QRepFlo
...
+
 func QRepFlowWorkflow(
 	ctx workflow.Context,
 	config *protos.QRepConfig,
 	state *protos.QRepFlowState,
-) (*protos.QRepFlowState, error) {
+) (_ *protos.QRepFlowState, err error) {
+	defer func() {
+		if err != nil && !workflow.IsContinueAsNewError(err) && !temporal.IsCanceledError(err) && ctx.Err() == nil {
+			markQRepRunFailed(ctx, config, workflow.GetInfo(ctx).OriginalRunID)
```
ilidemi: This approach is generic but makes it a bit harder to match the misbehaving area with the error handling for it.
Wdyt of this approach:
- separate the first monitoring.InitializeQRepRun into its own activity right before GetQRepPartitions
- put this defer after it

jgao54: good point

jgao54: done, also separated out `monitoring.InitializeQRepRun` into `RecordQRepRun` and `RecordQRepPartitions` because the txn there for both tables was deceiving anyways as the second insert into qrep_runs is a no-op and that was always bothering me. Now it is more explicit.

ilidemi (approved):

PR #4776, opened 2026-09-05
Reduce catalog contention for RecordMetricsCritical

ilidemi:
When an instance is overloaded with an overly parallel snapshot, RecordMetricsCritical can repeatedly fail to run, not being able to get a catalog connection. This change removes a few silly ways in which catalog was hammered or its client goroutines were delayed.

Fixes:
1. The helper function for getting or constructing the catalog pool was acquiring a mutex, checking if the pool is already constructed, and doing a ping. This made all callers serialize on a network call under a mutex. Additionally, issuing a ping doesn't pin the connection to the caller or anything like that, so it just gets released back after stalling everyone. Change this to double-checked locking and keep the ping just to construction, after pool is constructed there's no outer mutex wait at all.
2. Every partition reads `PEERDB_S3_BYTES_PER_AVRO_FILE` and `PEERDB_S3_PART_SIZE` dynconf, after resolving the above they were the next biggest catalog consumers. We want them to stay hot-configurable but there's no need to reread the values this often. Added a 30 second cache on both, global so workers don't need to re-poll individually. It is making dynconf more messy, but hopefully in a contained way, as the contention problem needs to be solved but redoing dynconf is out of scope right now.
3. We do 6 catalog updates per partition for bookkeeping. 2 of them were easy to merge, so doing that and reducing to 5.

Results from running 8x8 snapshots from our mysql perf instance and analyzing 120s traces:

| Measurement | Base | +avoid poolMutex | +dynconf cache | +merged bookkeeping (avg of 2 traces) |
|---|---:|---:|---:|---:|
| Catalog poolMutex wait | 41.33m | 0 | 0 | 0 |
| Catalog query pool wait | 5.91m | 53.43m | 29.71m | 25.65m |
| Mean query pool wait | 0.47s | 5.88s | 4.08s | 3.80s |
| Dynconf pool-wait events | 181 | 145 | 14 | 13 |
| RecordMetricsCritical waiting | 29.96s | 25.59s | 18.53s | 14.16s |

As you can see from the last row, this doesn't fully solve the situation (especially given goroutines can't have priorities). Proper solution is to limit the parallelism allowed/not-discouraged per instance size, but reducing repeat calls doesn't hurt.

Closes DBI-1107

### DIFF
--- a/flow/internal/catalog.go
+++ b/flow/internal/catalog.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/jackc/pgx/v5/pgxpool"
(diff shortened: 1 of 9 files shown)


flow/internal/catalog.go
```diff
@@ -15,35 +16,39 @@ import (
...
-		}
-		config.MaxConns = 3
-		config.MaxConnIdleTime = 90 * time.Second
-		pool, err = pgxpool.NewWithConfig(ctx, config)
-		if err != nil {
-			return shared.CatalogPool{Pool: pool},
-				exceptions.NewCatalogError(fmt.Errorf("unable to establish connection with catalog: %w", err))
-		}
-	}
+	if pool.Load() == nil {
```
dtunikov, line 23: nit: maybe do early return instead of branching two times:
```go
if pool.Load() == nil {
  if pool.Load() == nil {
     // do smth
  }
}
```

dtunikov (approved):

flow/internal/catalog.go
ilidemi: This is a [Double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking) pattern to not hit the lock after the pool is initialized. If the first `if` is reversed, then `shared.CatalogPool{Pool: pool.Load()}` would have to be duplicated

PR #4781, opened 2026-09-09
feat(structured-ingestion) DBI-1096: Generic flattening schema projection for structured ingestion 

pfcoperez:
This PR adds the generic tools to implement structured ingestion (using QValue system) for any unstructured (dynamic schema) data source.

It provided two abstractions for this purpose:  

- `SchemaProjector` (db9ce4197a235160d499299a82623a804cf1f2b4):  Upon initialization, it receives the target schema (`[]*protos.ColumnSetting`) and a function to interpret these raw mappings column settings as an ordered array of `QValue`. Through `ProjectRecord` method it transforms generic dynamic schema records (abstracted behind a walk iterator so it could be JSON, BSON, or anything that a lazy walker function can take) into an ordered array of `QValue` instances matching the schema plus an extra `QValueJSON` column (`malformed_data`) reporting records not matching the schema (see next point).  In the ouput QValues record all fields are nullable as missing fields are considered `NULL` for structured logging.

One example of lazy iterator applied for MongDB document flattening using `SchemaProjector` is:

https://github.com/PeerDB-io/peerdb/blob/859e699d25c0c173b0702c640d6cbdc0a76e24ce/flow/connectors/mongo/qvalue_convert.go#L246-L273

- `MalformedData` (898e5d0dc79fbb481e38b21c177611512a3e5b12):   While processing each individual unstructured document, it is the tracker of schema violations. Its JSON marshalling method implementation generates a JSON structure that makes it possible and easy to query data that failed to fit into the schema at destination CH table:

```
{
 "<offending_field_i>": {
    "unexpected": true|false,        // True iff this field was not in the projected schema
    "type_mismatch": true|false, // True iff this field was in the schema but with an incompatible type
 }
}
````

e.g:

```json
{
    "year": {
        "unexpected": true,
        "value": 2019
      }
}
```

```json
{
    "year": {
        "type_mismatch": true,
        "value": "two thousand nineteen" 
      }
}
```

Query example at destination CH table:

```sql
SELECT name, coalesce(age, malformed_data.age.value) FROM test_people;
```

```
Carmen	seven
Sofía	4
```

Given that it provides a default CH types to QKind schema interpreter (Used for MongDB structured ingestion) and that it matches the equivalence function already present in the ClickHouse target connector implementation, this PR also factors out the conversion table so it's shared:   60a80b9bf6e222dd2ea4f78eb08cf1946ea4c630


pfcoperez:
@fable do a deep review with explanations of the changes.

flow/pkg/clickhouse/qvalue_kind.go
```diff
@@ -0,0 +1,77 @@
+package clickhouse
+
+import (
+	"fmt"
+	"strings"
+)
+
+// QValueKindForType maps a ClickHouse column type to the QValueKind expected for its values,
+// returned as the string form of the flow module's types.QValueKind.
+func QValueKindForType(columnType string) (string, error) {
```
dtunikov: hm
why can't we return `QValueKind` type from here?
because you want it to stay in pkg and pkg can't refer to flow/shared/types package?

flow/connectors/utils/structured/malformed_data.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+// isJSONRepresentable reports whether v can be encoded by encoding/json, which rejects NaN and ±Inf floats.
+func isJSONRepresentable(v any) bool {
+	switch f := v.(type) {
+	case float64:
+		return !math.IsNaN(f) && !math.IsInf(f, 0)
+	case float32:
+		return !math.IsNaN(float64(f)) && !math.IsInf(float64(f), 0)
```
dtunikov: nit: can wrap float validation into helper:
```go
func isJSONRepresentable(v any) bool {
	validateFloat := func(f float64) bool {
		return !math.IsNaN(f) && !math.IsInf(f, 0)
	}
	switch f := v.(type) {
	case float64:
		return validateFloat(f)
	case float32:
		return validateFloat(float64(f))
	default:
		return true
	}
}
```

flow/pkg/clickhouse/qvalue_kind.go
pfcoperez: Because flow/pgk doesn't have that type defined. I think that's OK since it's just a rename of `string` (from now) but it is less than ideal.

Once in the past I attempted bringing Q abstractions to flow/pkg but that's a rabbit hole for a PR like this one.

flow/connectors/utils/structured/schema.go
```diff
@@ -0,0 +1,171 @@
...
+	// Schema columns by the record field they read from.
+	columns map[string]schemaColumn
+	// Record fields in order: the schema columns as declared, then the malformed data column.
+	fields             []types.QField
+	shouldRecordValues bool
+}
+
+// NewSchemaProjector resolves the schema columns' kinds through schemaToQKind, failing on a type it does
+// not know, a column declared twice or one named as the malformed data column.
+func NewSchemaProjector(
```
dtunikov, line 35: i wouldn't add 3 `NewProjectorFromX` implementations prematurely
imo just one NewSchemaProjector is fine until we see the need for other ways to create this object

flow/connectors/utils/structured/schema.go
```diff
@@ -0,0 +1,171 @@
...
+// schema type to QValueKind conversion, i.e. for schemas declared with ClickHouse column types.
+func NewSchemaProjectorFromCHtoQValue(
+	schemaColumns []*protos.ColumnSetting,
+	shouldRecordValues bool,
+) (*SchemaProjector, error) {
+	return NewSchemaProjector(defaultCHSchemaToQKind, schemaColumns, shouldRecordValues)
+}
+
+// Columns are the structured schema (order is relevant).
+func (sc *SchemaProjector) Columns() []types.QField {
```
dtunikov: i'm not sure i understand why we'd need these two methods - Columns and QRecordSchema.
can we maybe add them once we see the need for that in the code that is going to use this SchemaProjector?
kinda the same concern as 3 constructors.

flow/pkg/clickhouse/qvalue_kind.go
dtunikov: yeah, but I'm wondering why do you want to put it in pkg at all?
afaiu we put things in pkg when we want to re-use them from the other repos.
this function refers to internal peerdb qvalue/qkind concept which is unknown to the outside world

i guess you didn't want to put it in `flow/connectors/clickhouse/clickhouse.go` because you need to reference it from utils/structured?

i'd rather keep it closer to QValueKind itself so that it would be hard to forget to update it when a new kind is added

pfcoperez: I am preparing to potentially use them in Discovery validations.

flow/connectors/utils/structured/schema.go
pfcoperez: This was used in the context of https://github.com/PeerDB-io/peerdb/pull/4774/ at a certain point but it's no longer. We can remove it indeed.

pfcoperez: @fable remove the Columns method and the unit test case exercising it. commit to the PR source branch.

pfcoperez: Done :wink:

flow/connectors/utils/structured/schema.go
pfcoperez: @dtunikov Each one of these constructors are used in https://github.com/PeerDB-io/peerdb/pull/4774 for:

- QRep replication
- CDC replication
- Schema derivation

flow/connectors/utils/structured/malformed_data.go
pfcoperez: Sure, done in 340d978deeb536e45bc30af58c4b438293d134fe

flow/pkg/clickhouse/qvalue_kind.go
dtunikov: but discovery doesn't know anything about q type system
or you want to start using it there?

pfcoperez: The validation case I have in mind is making sure this table can find a match. The only thing Discovery needs is to check if this function returns error.

I was discussing this with @jgao54 as well and there are several alternatives to achieve the same outcome without exposing this conversion logic at `flow/pkg`. 

This is something needed for the Schema inference side of the project and we can address following these alternatives or default to expose this to flow/pkg as a last resource at that stage so I am moving this function to the connectors/clickhouse package in this PR.

pfcoperez:
> 2. **`MalformedData`** (`flow/connectors/utils/structured/malformed_data.go`) — a per-record tracker of schema violations. `AddField` records a field name, a reason (`unexpected` / `type_mismatch` / `not_a_number`), and optionally the offending value. `MarshalJSON` produces the query-friendly `{"<field>": {"<reason>": true, "value": ...}}` shape, and `AsQValue` wraps it as a `QValueJSON` destined for the reserved `malformed_data` column. Non-finite floats (which `encoding/json` rejects) are caught by `isJSONRepresentable` and downgraded to a value-less `not_a_number` entry. Output is deterministic because `encoding/json` sorts map keys — and the tests assert that explicitly.

This is a valid concern and has been addressed in bf8a9a866c2a1a26115144ebfbe3e06b8543d823

flow/connectors/utils/structured/malformed_data.go
pfcoperez: Done in 340d978deeb536e45bc30af58c4b438293d134fe

pfcoperez:
> **3. `shouldRecordValues=false` still persists values of unexpected fields (medium, PII question)** — `schema.go:126` records the value for `ReasonUnexpected` unconditionally; only the `ReasonTypeMismatch` path (`schema.go:139-141`) honors the flag. The test documents this asymmetry with a NOTE, so it may be intentional — but if the flag exists so operators can avoid persisting raw source values (compliance/PII), unexpected fields are precisely the ones most likely to contain data nobody vetted, since they're by definition outside the declared schema. If intentional, a doc comment on `shouldRecordValues` stating the scope ("gates mismatched values only") would prevent misuse; if not, gate line 126 too.

This was an implementation bug, the intention was always to avoid including unexpected values if the flag sets it.

Fixed in:   1d03fde45352b42fd93458b4ff7ffc303567eceb

flow/connectors/utils/structured/schema.go
```diff
@@ -0,0 +1,172 @@
...
+
+		values[column.index] = value
+	}
+
+	if !malformedData.IsEmpty() {
+		qValue, err := malformedData.AsQValue()
+		if err != nil {
+			return nil, err
+		}
+		values[len(values)-1] = qValue
```
jgao54: nit: to avoid accidental drifts, could have `malformedDataIndex := len(fields)` and pass it into SchemaProjector as a field.

flow/connectors/utils/structured/malformed_data.go
```diff
@@ -0,0 +1,137 @@
...
+	ReasonUnexpected MalformedReason = iota
+	ReasonTypeMismatch
+	ReasonNaN
+	ReasonDuplicatedFields
+)
+
+func (r MalformedReason) String() string {
+	switch r {
+	case ReasonUnexpected:
+		return "unexpected"
```
jgao54: nit: "unexpected_field" is a bit more clear to me

flow/connectors/utils/structured/malformed_data.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+
+	"github.com/PeerDB-io/peerdb/flow/generated/protos"
+	"github.com/PeerDB-io/peerdb/flow/shared/types"
+)
+
+const MalformedDataColumn = "malformed_data"
```
jgao54: to make it consistent with other peerdb columns (and less chance of collision), maybe `_peerdb_malformed_data` or `_peerdb_errors` if we want to be more generic?

pfcoperez:
> 4. Duplicate record fields produce contradictory output (low) — BSON legally allows repeated keys and JSON tolerates them, and ProjectRecord consumes whatever the walker yields. If a field appears twice — first with a valid value, then with a mismatched one — the column keeps the first value and malformed_data reports a type_mismatch for the same field; two valid occurrences silently last-write-win into the slot. Related: AddField (malformed_data.go:55-60) updates the reason but keeps a stale value when the new value is nil, so repeated adds can pair a value with a reason it didn't belong to. Cheapest fix is documenting the iterator contract (field names must be unique, e.g. dedup in the walker); alternatively treat a repeated field as malformed itself.

This is a protection against deeply broken upstream data sources or bad connectors implementation (around the iterator).

Included in 368a9fa8c963311178566aa6597db9871bf0e7f3

flow/connectors/utils/structured/malformed_data.go
pfcoperez: @fable Apply this suggestion

flow/connectors/utils/structured/malformed_data.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		if isJSONRepresentable(jsonValue) {
+			marshaledValue, err := json.Marshal(jsonValue)
+			if err != nil {
+				field["value"] = fmt.Sprintf("%v", jsonValue)
+			} else {
+				field["value"] = json.RawMessage(marshaledValue)
+			}
+		} else {
+			fields[name] = map[string]any{ReasonNaN.String(): true}
```
jgao54, line 100: i think claude already mentioned this but here it's overriding the default reason and iiuc is only set because json.Marshal does not support it. could instead convert the NaN/-Inf/+Inf value as string and then the original reason can be used and reasonNaN can be removed (since ClickHouse supports NaN/Inf/-Inf for floats)

flow/connectors/utils/structured/malformed_data.go
pfcoperez: I'd like to keep `malformed_data`   over errors because it points to problems in the data. Error might suggest other problems like destination issues or with our own services.

That said, wouldn't a customer wonder what `_peerdb` means if they don't know anything about how the ClickPipes is implemented?  Maybe `_pipe_malformed_data` ?

jgao54: We already use _peerdb_* fields today so this won't be new and is more for consistency.

flow/pkg/clickhouse/qvalue_kind.go
pfcoperez: @fable Move `QValueKindForType` out of flow/pkg to the flow/connectors/clickhouse/clickhouse.go file and update the code in this PR using it.

flow/connectors/utils/structured/malformed_data.go
pfcoperez: Makes sense, @claude rename use `_peerdb_malformed_data` as the value for `MalformedDataColumn`

flow/connectors/utils/structured/malformed_data.go
pfcoperez: Yes Claude flagged it too. I disregarded its comment for this reason: In these cases, the data could be malformed for two different reasons: The out of domain value and the potentially being an unexpected or wrong type column.

I want to keep the malformed column simple and avoid the situation were we record or need to detect all the possible ways in which a value is broken. So I decided that it was better to flag the completely out of domain value because:

- Unexpected type is a super-category of this problem. Bad domain value implies bad value type -> Unexpected type.
- Unexpected field is less disruptive than broken data.

dtunikov (approved):

flow/connectors/utils/structured/schema.go
pfcoperez: Sure, done at ad480b667d3ffcf09e130b3f43367c352a9d2e83

flow/connectors/utils/structured/malformed_data.go
jgao54: maybe I am misunderstanding something here so want to clarify.

my understanding is that `ReasonNaN` here does not mean that the data itself cannot be replicated due to NaN (because clickhouse supports it for floats), but it's set to NaN because _peerdb_malformed_data column is JSON type and that does not support NaN if the data is deemed incompatible for some other reason (e.g. unexpected field, or type mismatch, but it shouldn't be ambiguous here since unexpected field means the field is not expected from the schema, or type mismatch which means the actual type does not match the inferred type). if my understanding is correct, then i think this code here is hiding that root cause, and overriding with NaN instead.

I can't think of a valid scenario where NaN is a valid reason on during inferencing itself: if it's a float, clickhouse can handle it; if it's not a float, then it's a typeMismatch error of some sort.

Does this match your understanding or did I miss something here

pfcoperez: @jgao54 You understood it perfectly!

There is a wrong premise in this comment of mine:

> Unexpected type is a super-category of this problem. Bad domain value implies bad value type -> Unexpected type. 

I assumed we didn't accept these float values for ingestion but you clarified that they do, thank you!

Corrected in https://github.com/PeerDB-io/peerdb/pull/4798

PR #4782, opened 2026-09-09
remove tailwind

serprex:


ilidemi (approved):
Would be nice to have a description on background/motivation. From what I understand from git history, Tailwind was added for Tremor, then Tremor got removed and Tailwind is not needed anymore, plus we have styled components in most styled places and many deps are going away with this.

serprex:
Yes. & tremor we removed because it became abandonware after acquisition, meanwhile tailwind has announced its acquisition. Tailwind has been on my list of dependencies I'd like to get rid of for awhile, but being tied up with tremor/charts was keeping it around

Net line reduction speaks for itself

PR #4783, opened 2026-09-10
fix(DBI-1096 - clickhouse destination connector): Honor nullability for destination_type overrides in normalize

pfcoperez:
With a table mapping column setting both a `destination_type` override and nullability (table- or column-level `nullable_enabled`), the DDL generator creates the destination column as Nullable(<type>), but the normalize query still extracted it as plain <type>. JSONExtract to a non-nullable type turns JSON nulls into the type's default, so NULL values silently landed as `0`, `" "`, etc. instead of NULL.

This PR makes the normalize query generator mirror the DDL: 
- Wraps the override in Nullable(...) under the same conditions.
- Guards both generators against double wrapping when the override is already spelled Nullable(...), which previously produced invalid Nullable(Nullable(<type>)) DDL.

Related to:

- https://github.com/PeerDB-io/peerdb/pull/4781
- https://github.com/PeerDB-io/peerdb/pull/4774

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -216,7 +216,8 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
 				if err != nil {
 					return nil, fmt.Errorf("error while converting column type to ClickHouse type: %w", err)
 				}
-			} else if (tableSchema.NullableEnabled || columnNullableEnabled) && column.Nullable && !colType.IsArray() {
+			} else if (tableSchema.NullableEnabled || columnNullableEnabled) && column.Nullable && !colType.IsArray() &&
+				!strings.HasPrefix(clickHouseType, "Nullable(") {
 				clickHouseType = fmt.Sprintf("Nullable(%s)", clickHouseType)
 			}
 
--- a/flow/connectors/clickhouse/normalize_query.go
+++ b/flow/connectors/clickhouse/normalize_query.go
@@ -168,6 +168,11 @@ func (t *NormalizeQueryGenerator) BuildQuery(ctx context.Context) (string, error
 			if err != nil {
 				return "", fmt.Errorf("error while converting column type to clickhouse type: %w", err)
 			}
+		} else if (schema.NullableEnabled || columnNullableEnabled) && column.Nullable && !colType.IsArray() &&
+			!strings.HasPrefix(clickHouseType, "Nullable(") {
+			// mirror the table DDL: a nullable-enabled column created as Nullable(...) must also be
+			// extracted as Nullable(...), or JSON nulls turn into the type's default value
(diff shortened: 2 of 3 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -216,7 +216,8 @@ func (c *ClickHouseConnector) generateCreateTableSQLForNormalizedTable(
 				if err != nil {
 					return nil, fmt.Errorf("error while converting column type to ClickHouse type: %w", err)
 				}
-			} else if (tableSchema.NullableEnabled || columnNullableEnabled) && column.Nullable && !colType.IsArray() {
+			} else if (tableSchema.NullableEnabled || columnNullableEnabled) && column.Nullable && !colType.IsArray() &&
```
dtunikov, line 219: afaiu tableSchema.NullableEnabled is kind of a global flag whether we want to have nullable fields at the destination at all (regardless of source column nullability)
so, I think it should be `tableSchema.NullableEnabled && columnNullableEnabled`, no?
if tableSchema.NullableEnabled is false, it means that none of the destination columns can be nullable

flow/connectors/clickhouse/normalize_query.go
```diff
@@ -168,6 +168,11 @@ func (t *NormalizeQueryGenerator) BuildQuery(ctx context.Context) (string, error
 			if err != nil {
 				return "", fmt.Errorf("error while converting column type to clickhouse type: %w", err)
 			}
+		} else if (schema.NullableEnabled || columnNullableEnabled) && column.Nullable && !colType.IsArray() &&
+			!strings.HasPrefix(clickHouseType, "Nullable(") {
+			// mirror the table DDL: a nullable-enabled column created as Nullable(...) must also be
```
dtunikov, line 173: should we move it to the helper then?
if it's exactly the same condition as in `flow/connectors/clickhouse/normalize.go`.

flow/connectors/clickhouse/normalize.go
dtunikov: ah
we have schema.NullableEnabled and column.NullableEnabled
seems like we can put false on the schema lvl and override on the column lvl..
so `OR` is correct here I guess

dtunikov (approved):

flow/connectors/clickhouse/normalize_query.go
pfcoperez: It's the same condition today but JSON extraction function and DDL are different enough IMO as to maybe make us stumble upon different conditions in future changes.

flow/connectors/clickhouse/normalize.go
pfcoperez: > seems like we can put false on the schema lvl and override on the column lvl..

Exactly, at least this is how I interpret it.

PR #4788, opened 2026-09-10
fix(mongo): report receive/process/addRecord time for mongo CDC

itsbilal:
Previously, we reported receive/process/addRecord time for mysql and postgres CDC connectors but not for mongo. This change adds this reporting for the mongodb connector as well, although this is complicated by part of the processing and addRecord time happening in parallel with the receive, so a new counter to count walltime inside parallel processing workers was added.

With this, we separately report time the main thread waits for the parallel processors to finish (in processTime), and time taken by individual processors (parallelProcessTime) which will be a multiple of concurrency.


flow/connectors/mongo/cdc.go
```diff
@@ -568,6 +585,7 @@ func (c *MongoConnector) PullRecords(
 
 	var lastEventGaugesRecordedAt time.Time
 	for recordCount < req.MaxBatchSize {
+		receiveStart := time.Now()
 		if ok := changeStream.Next(timeoutCtx); !ok {
```
ilidemi, line 589: probably a rare case but can increment receiveTime on !ok as well

ilidemi (approved):

PR #4791, opened 2026-09-11
support overriding env on the pipe-level in UI

jgao54:
Previously we already support settings override during pipe creation in the PeerDB UI:
Screenshot 2026-09-10 at 17 50 06@2x

This PR:
- adds the same option to the pipe editing UI (so we do not have to rely on API alone for this modification)
- because `updated_env` does not support env removal, add a separate `removed_env` field and let the typescript code perform a diff to determine the `removed_env` (otherwise, removing an env from `Settings Override` is a no-op)
- add validation for env override to ensure the key/value are legitimate.

edit mirror UI:
Screenshot 2026-09-10 at 17 47 01@2x

validate env key must exist:
Screenshot 2026-09-10 at 17 46 01@2x

validate env value must container valid type:
Screenshot 2026-09-10 at 17 44 40@2x

valiate invalid json:
Screenshot 2026-09-10 at 17 46 32@2x

Fixes: DBI-1118

### DIFF
--- a/flow/internal/dynamicconf.go
+++ b/flow/internal/dynamicconf.go
@@ -5,7 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"maps"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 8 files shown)


flow/internal/dynamicconf.go
```diff
@@ -602,6 +604,32 @@ var DynamicIndex = func() map[string]int {
...
+	var errs []error
+	for _, key := range slices.Sorted(maps.Keys(env)) {
+		idx, ok := DynamicIndex[key]
+		if !ok {
+			errs = append(errs, fmt.Errorf("%s is not a known setting", key))
+			continue
+		}
+		value := env[key]
+		var err error
+		switch DynamicSettings[idx].ValueType {
```
ilidemi, line 617: Exhaustive switch with default error, please

ilidemi (approved):

PR #4792, opened 2026-09-11
abstract away signalAsEmpty/signalAsNotEmpty

jgao54:
Remove the need to call `signalAsEmpty`/`signalAsNotEmpty`, which in the past had been a source of bugs when it was forgotten to be set in each of the connectors. Instead rely on cdc stream itself to manage this internal state:
- signal as not empty when the first AddRecord is called
- if stream closes and AddRecord hasn't been called at least once, than we know that the stream is empty (this is not very different from today where `SignalAsEmpty` is only called on PullRecord exit with defer anyways, and we already close the record stream on PullRecord exit with defer as well, so it can be encapsulated in handling of Close(). 

Also removed some boolean values: `needsNormalize`, `empty`, and `emptySet` can all be consolidated into into `firstRowSet`.


flow/connectors/bigquery/cdc.go
```diff
@@ -148,28 +141,14 @@ func (c *BigQueryConnector) PullTableRecords(
...
-			req.Stream.SignalAsNotEmpty()
-		}
-		return nil
-	}
-
 	var bytesProcessed int64
 	if cfg.GetBigqueryCdcConfig().GetReplicationMethod() == protos.BigQueryReplicationMethod_BIGQUERY_REPLICATION_METHOD_QUERY {
 		bytesProcessed, err = c.pullTableQuery(ctx, tm.QueryCdcWatermarkColumn, req.SourceTableIdentifier,
-			req.NameAndExclude, start, upper, addRecord)
+			req.NameAndExclude, start, upper, req.Stream.AddRecord)
```
dtunikov, line 147: i've added some logs to addRecord in one of the [PR](https://github.com/PeerDB-io/peerdb/pull/4797)s
can we keep it as is just without that signal as empty part?

jgao54: ah yes, of course

PR #4793, opened 2026-09-11
clickhouse: parallelize S3 to raw table ingestion

dtunikov:
## Summary

- add PEERDB_CLICKHOUSE_PARALLEL_RAW_INGESTION with a default of 1
- copy grouped Avro batches into the ClickHouse raw table concurrently when configured above 1
- only enable parallel ingestion when every destination table mapping uses ReplacingMergeTree or ReplicatedReplacingMergeTree
- advance the raw-table batch watermark only after the full contiguous batch range succeeds

## Failure semantics

A partial parallel failure does not advance latest_batch_id_in_raw_table. Successful out-of-order copies can therefore be retried, while the RMT version column prevents an older batch from replacing a newer value in the final table.

## Testing

- GOCACHE=/tmp/peerdb-go-build go test ./connectors/clickhouse -run Test\(AllTablesUseReplacingMergeTree\|GenerateCreateTableSQLForNormalizedTable\)$
- GOCACHE=/tmp/peerdb-go-build go test ./internal
- GOCACHE=/tmp/peerdb-go-build go vet ./connectors/clickhouse ./internal
- GOCACHE=/tmp/peerdb-go-build golangci-lint run ./connectors/clickhouse ./internal

The complete ClickHouse package suite was attempted but sandbox restrictions prevent its localhost catalog and httptest listeners from starting; CI should exercise those integration-dependent tests.

### DIFF
--- a/flow/connectors/clickhouse/normalize.go
+++ b/flow/connectors/clickhouse/normalize.go
@@ -449,7 +449,9 @@ func (c *ClickHouseConnector) NormalizeRecords(
 
 	endBatchID := min(req.SyncBatchID, lastNormBatchID+groupBatches)
 
-	if err := c.copyAvroStagesToDestination(ctx, req.FlowJobName, lastNormBatchID, endBatchID, req.Env, req.Version); err != nil {
+	if err := c.copyAvroStagesToDestination(
+		ctx, req.FlowJobName, lastNormBatchID, endBatchID, req.Env, req.Version, req.TableMappings,
+	); err != nil {
 		return model.NormalizeResponse{}, fmt.Errorf("failed to copy avro stages to destination: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


flow/connectors/clickhouse/normalize.go
```diff
@@ -708,6 +716,44 @@ func (c *ClickHouseConnector) copyAvroStagesToDestination(
...
+		c.logger.Info("[clickhouse] copying s3 batches to raw table in parallel",
+			slog.Int64("startBatchID", lastCopiedBatchID+1),
+			slog.Int64("endBatchID", endBatchID),
+			slog.Int("parallelism", parallelism))
+
+		group, groupCtx := errgroup.WithContext(ctx)
+		group.SetLimit(parallelism)
+		for batchID := lastCopiedBatchID + 1; batchID <= endBatchID; batchID++ {
+			group.Go(func() error {
+				if err := c.copyAvroStageToDestination(groupCtx, flowJobName, batchID, env, version); err != nil {
```
itsbilal, line 739: if I'm understanding this right, we parallelize the copying into the raw table part, but we then wait for all the writes to the raw table to finish (`group.Wait()` below) before we do any writes to destination tables. Seemingly it'd also be good if we could parallelize the latter? Doesn't have to be in this PR of course; just a question

dtunikov: sorry, forgot to close this PR
this was my first idea - to parallelize an ingestion into the raw table (so run insert into raw from s3 in parallel)
but later I realized that it'd be better to have a pipeline between s3->raw and raw->final ingestion
i have this draft [PR](https://github.com/PeerDB-io/peerdb/pull/4794), but it's not yet ready for review

PR #4795, opened 2026-09-11
fix(clickhouse): ignore empty allowed host domains

fallintoplace:
## What changed

- **Trim whitespace** from configured ClickHouse allowed domains.
- **Ignore empty entries** in the comma-separated list.
- Keep an empty effective list unrestricted, matching the existing behavior.
- Apply the same parsing in the UI.

## Why

With `PEERDB_CLICKHOUSE_ALLOWED_DOMAINS=clickhouse.cloud,`, the empty entry matches every host because every string ends with an empty string. An unrelated ClickHouse host could pass validation.

## Tests

- Added a regression test for a trailing empty entry.
- Added coverage for whitespace and empty entries.


dtunikov (approved):

fallintoplace:
I have signed the CLA by the way.

fallintoplace:
image

PR #4797, opened 2026-09-11
feat(bigquery): enable storage read api client for BQ

dtunikov:
I ran some performance benchmarks and without a storage read api client we can reach only 1.8mb/s throughput. 
With storage read api client we can reach ~ 18mb/s.
Which is important for BQ sources that produce a lot of data on every sync interval.
BQ go client automatically identifies whether storage read api or a "simple" client should be used.

### DIFF
--- a/flow/connectors/bigquery/bigquery.go
+++ b/flow/connectors/bigquery/bigquery.go
@@ -18,6 +18,7 @@ import (
 	"go.temporal.io/sdk/log"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
+	"google.golang.org/grpc"
 
 	metadataStore "github.com/PeerDB-io/peerdb/flow/connectors/external_metadata"
 	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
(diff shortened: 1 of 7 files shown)


flow/connectors/bigquery/bigquery.go
```diff
@@ -119,6 +119,10 @@ func NewBigQueryConnector(ctx context.Context, config *protos.BigqueryConfig) (*
 	if err != nil {
 		return nil, fmt.Errorf("failed to create BigQuery client: %v", err)
 	}
+	if err := client.EnableStorageReadClient(ctx, option.WithAuthCredentials(creds)); err != nil {
+		_ = client.Close()
+		return nil, fmt.Errorf("failed to enable BigQuery Storage Read API: %v", err)
+	}
```
jgao54, line 130: iiuc, the meteredHTTPClient won't apply in this case given it's using grpc instead of http.  will probably need to adjust that as well to the new client, but doesn't block this PR

jgao54 (approved):

PR #4799, opened 2026-09-13
improvement(clickhouse-connector): Remove the need to explicitly list type combinations with Nullable and LowCardinality

pfcoperez:
This PR aims to remove the surface of explicit types listing in CH connector `QValueKindForType`.
This expands the surface of accepted combinations with additions from the valid set.

It also improves maintainability and reduces the chances of introducing bugs by leaving combinations behind.


pfcoperez:
@fable review this PR, focus on regressions.

pfcoperez:
@fable fix linting issues reported in CI

pfcoperez:
> 5. Typos (nit). PR title: "conbinations" → "combinations"; description: "reduces the changes of" → "reduces the chances of".


@claude Fix these typos.

flow/connectors/clickhouse/clickhouse.go
```diff
@@ -394,44 +394,73 @@ func (c *ClickHouseConnector) GetFlags(ctx context.Context) ([]string, error) {
...
+
+var notNullableTypesPrefixes = []string{
+	"Array(",
+	"Map(",
+	"Tuple(", // Remove once this is supported without `enable_nullable_tuple_type = 1`
+}
+
 // QValueKindForType maps a ClickHouse column type to the QValueKind expected for its values.
 func QValueKindForType(columnType string) (types.QValueKind, error) {
+	errResolve := fmt.Errorf("failed to resolve QValueKind for %s", columnType)
```
dtunikov: Agree with claude here
"Eager allocation on the hot path (nit). errResolve is built with fmt.Errorf on every call — including every successful resolution and every recursion level — and fmt.Sprintf("Nullable(%s", ...) runs per call inside the guard loop (clickhouse.go:410, 413). Since this runs per column rather than per row it's negligible in practice; precomputing the guard strings ("Nullable(Array(", …) as package vars and constructing the error only on failure paths would be tidier."

masterashu (approved):

dtunikov (approved):

flow/connectors/clickhouse/clickhouse.go
pfcoperez: > precomputing the guard strings ("Nullable(Array(", …) as package vars and constructing the error only on failure paths would be tidier.

@claude   Apply this suggestion.

pfcoperez: I don't like this Claude proposal at all (9e5b15bbc89942a717aa30a5e2d6880adb373dff) , it makes maintainability worse. Reverted.

pfcoperez: Made error generation lazy in 4a38177fd42d59ff71fb7c13dc352a0f4c8aee79

pfcoperez:
@claude  Fix linting issues

PR #4801, opened 2026-09-14
feat(bigquery): chunk BigQuery CDC Avro staging files

dtunikov:
## Context
BigQuery CDC currently stages one Avro file per logical sync batch. A long sync interval can therefore produce an extremely large object: for example, with a 24-hour interval, a customer may run a daily ingestion job that writes around 1 TB of rows at once. Because those rows can share the same watermark timestamp, splitting the source query into smaller time windows does not reliably bound the batch size.

A single object of that size is operationally risky to upload, retry, and ingest. Reading all resulting objects in one ClickHouse insert would also leave ClickHouse processing the full terabyte as one large operation, increasing memory and resource pressure.

## Changes
- extract size-bounded Avro chunk writing into `writeToAvroFiles`
- use `PEERDB_S3_BYTES_PER_AVRO_FILE` to split one logical BigQuery CDC batch into multiple physical Avro objects
- stage the batch as an Avro file manifest
- normalize one staged file at a time, bounding each ClickHouse insert
- checkpoint the remaining manifest after each successful insert so retries normally resume from the next file

The logical CDC batch and source cursor semantics remain unchanged. There is still an existing cross-system retry gap if a ClickHouse insert succeeds but the catalog checkpoint fails; in that case at most one chunk is replayed rather than the entire batch.

Resolves: DBI-1132

### DIFF
--- a/flow/connectors/bigquery/bigquery.go
+++ b/flow/connectors/bigquery/bigquery.go
@@ -18,6 +18,7 @@ import (
 	"go.temporal.io/sdk/log"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
+	"google.golang.org/grpc"
 
 	metadataStore "github.com/PeerDB-io/peerdb/flow/connectors/external_metadata"
 	"github.com/PeerDB-io/peerdb/flow/connectors/utils"
@@ -119,7 +120,11 @@ func NewBigQueryConnector(ctx context.Context, config *protos.BigqueryConfig) (*
 	if err != nil {
 		return nil, fmt.Errorf("failed to create BigQuery client: %v", err)
 	}
-	if err := client.EnableStorageReadClient(ctx, option.WithAuthCredentials(creds)); err != nil {
+	if err := client.EnableStorageReadClient(
+		ctx,
+		option.WithAuthCredentials(creds),
+		option.WithGRPCDialOption(grpc.WithStatsHandler(&meteredGRPCStatsHandler{})),
+	); err != nil {
 		_ = client.Close()
 		return nil, fmt.Errorf("failed to enable BigQuery Storage Read API: %v", err)
 	}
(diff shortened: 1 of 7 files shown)


flow/connectors/clickhouse/avro_sync.go
```diff
@@ -156,13 +156,35 @@ func (s *ClickHouseAvroSyncMethod) pushDataToStagingForSnapshot(
 		return nil, 0, err
 	}
 
-	bytesPerAvroFile, err := internal.PeerDBS3BytesPerAvroFile(ctx, config.Env)
+	return s.writeToAvroFiles(ctx, config.Env, stream, avroSchema, partition.PartitionId,
```
dtunikov, line 159: changes in this file are purely cosmetic in order to re-use writeToAvroFiles method from cdc sync path
no changes in behavior here

flow/connectors/clickhouse/s3_stage.go
```diff
@@ -122,23 +122,23 @@ func GetQueryCDCAvroStage(
...
 				"%w for flow job %s, table %s, batch %d", ErrNoAvroStage, flowJobName, sourceTableIdentifier, batchID)
 		}
-		return utils.AvroFile{}, nil, fmt.Errorf("failed to get table avro stage: %w", err)
+		return nil, nil, fmt.Errorf("failed to get table avro stage: %w", err)
 	}
 
-	var avroFile utils.AvroFile
-	if err := json.Unmarshal(avroFileJSON, &avroFile); err != nil {
-		return utils.AvroFile{}, nil, fmt.Errorf("failed to unmarshal avro file: %w", err)
+	var avroFiles []utils.AvroFile
```
dtunikov, line 131: fine since we don't have any pipes that use this code path in production yet

itsbilal (approved):
LGTM

PR #4802, opened 2026-09-14
MongoDB: Per-type BSON to QValue converters behind BsonToQValueConverter

pfcoperez:
Expand the converter interface with QValueFromBsonValue, which dispatches on the BSON type to per-type converters.

No pre-existing behaviour change.

:warning: These conversions are bound to evolve as we align schema inference with desired destination types. 


flow/connectors/mongo/qvalue_convert.go
```diff
@@ -62,6 +93,136 @@ func (c *DirectBsonConverter) QValueStringFromId(id bson.RawValue, version uint3
...
+
+	case bsoncore.TypeSymbol: // deprecated type, kept for backwards-compatibility
+		return c.QValueStringFromString(v.Symbol()), nil
+
+	case bsoncore.TypeInt32:
+		return c.QValueInt64FromInt32(v.Int32()), nil
+
+	case bsoncore.TypeTimestamp:
+		t, i := v.Timestamp()
+		return c.QValueJSONFromTimestamp(t, i), nil
```
pfcoperez, line 148: Note that `bsoncore.TypeTimestamp` is not a direct translation to timestamp but https://www.mongodb.com/docs/manual/reference/bson-types/#timestamps

With an extra field for sequential order within the same second.

pfcoperez:
@claude fix linting issues

pfcoperez:
@claude fix linting issues

flow/connectors/mongo/qvalue_convert.go
```diff
@@ -62,6 +93,136 @@ func (c *DirectBsonConverter) QValueStringFromId(id bson.RawValue, version uint3
 	return types.QValueString{Val: string(c.stream.Buffer())}, nil
 }
 
+func (c *DirectBsonConverter) QValueFromBsonValue(rv bson.RawValue, nullKind types.QValueKind) (types.QValue, error) {
+	if rv.IsZero() {
+		return c.QValueNullFromNull(nullKind), nil
+	}
+	v := bsoncore.Value{Type: bsoncore.Type(rv.Type), Data: rv.Value}
+	switch v.Type {
+	case bsoncore.TypeDouble:
```
dtunikov, line 91: can we maybe just have for simply go types? not sure I understand how having these as separate methods helps
```
case bsoncore.TypeDouble:
    return types.QValueFloat64{Val: v.Double()}, nil

case bsoncore.TypeBoolean:
    return types.QValueBoolean{Val: v.Boolean()}, nil

case bsoncore.TypeInt64:
    return types.QValueInt64{Val: v.Int64()}, nil
....
```

dtunikov (approved):

flow/connectors/postgres/client.go
ilidemi, line 1: Are PG changes intended to be included? Don't see anything in the PR description

pfcoperez: Not at all! I recall inspecting the build failures and accidentally clicking on "Fix with co-pilot"  instead of "run failed flows" which sneaked in this change: cc1b36496a4cc64841fa0c304dac38a343efd0ad

flow/connectors/mongo/qvalue_convert.go
pfcoperez: As I was building this  I was planning to use these methods to introduce handling ranges special cases if we ever generalized this code but the truth is that there is not need to offer anything but BSON -> QValue for MongoDB and this code is specific for this connector.

@fable Remove the intermediate c.QValue[X]From[Y] methods that are not used outside this switch and use   types.QValue[X] directly instead in these cases

PR #4809, opened 2026-09-16
feat(bigquery): resume CDC Storage Read batches

dtunikov:
## Summary

- add an opt-in PeerDB-managed BigQuery Storage Read path with durable per-stream offsets
- split large CDC windows into bounded batches by rows, serialized Arrow bytes, or pull duration
- read Storage API streams in parallel and checkpoint only complete ReadRowsResponse boundaries after destination staging succeeds
- retain the existing BigQuery RowIterator path behind PEERDB_BIGQUERY_CDC_USE_EXPLICIT_STORAGE_READ for safe rollout

## Resumption semantics

cursor_text remains the last fully completed logical window. The new inflight_state stores the fixed window, result table, read session, Arrow schema, expiry, and committed offsets for every stream. A partial slice gets its own normal query CDC batch and Avro file. If a worker dies after staging but before recording offsets, retrying uses the same batch ID and safely overwrites the stage entry.

An in-progress resumable window always finishes through the explicit Storage Read path even if the dynamic config is subsequently disabled.

## Configuration

- PEERDB_BIGQUERY_CDC_USE_EXPLICIT_STORAGE_READ=false
- PEERDB_BIGQUERY_CDC_PULL_SLICE_MAX_ROWS=1000000
- PEERDB_BIGQUERY_CDC_PULL_SLICE_MAX_BYTES=1073741824
- PEERDB_BIGQUERY_CDC_PULL_SLICE_MAX_SECONDS=300

Limits are evaluated at complete response boundaries, so parallel readers can overshoot by up to roughly one response per active stream.

## Known limitation

If the BigQuery read session expires before its logical window completes, the table reports an explicit error. It does not restart the window because earlier slices may already have normalized and replaying the window could duplicate rows.

## Validation

- BigQuery connector test suite
- repository-wide Go compilation
- git diff --check


flow/connectors/bigquery/arrow_decode.go
```diff
@@ -0,0 +1,168 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// This is adapted from cloud.google.com/go/bigquery/arrow.go for Arrow v18.
+package connbigquery
```
dtunikov, line 6: copy from go bigquery library
unfortunately they don't expose this decoder from the library itself

PR #4810, opened 2026-09-16
fix(bigquery): use last_synced_at for query-cdc scheduling

dtunikov:
## What changed

- Schedule each table start-to-start from `last_attempt_at` only when that attempt completed successfully, as indicated by `last_synced_at`.
- Keep recording `last_attempt_at` when every poll starts so failed and interrupted attempts remain observable.
- Retry failed pull/sync iterations with exponential backoff starting at 5 seconds and capped at the smaller of 1 minute or the configured sync interval.
- Reset the retry backoff after a successful poll while retaining the existing once-per-lagging-episode alert behavior and per-table failure isolation.
- Rename the query-CDC-local `idleTimeout` variables to `syncInterval` to match their actual meaning.

## Why

Previously, `last_attempt_at` both described the latest attempt and controlled the next poll time. If a pull failed or the worker restarted during a batch, the table could wait an entire sync interval before querying the source again.

`last_attempt_at` and `last_synced_at` now have distinct semantics:

- `last_attempt_at`: when the latest poll began, whether or not it succeeded.
- `last_synced_at`: when the latest poll was successfully checkpointed; together the timestamps distinguish a completed attempt from a failed or interrupted one.

As a result, a 24-hour interval remains start-to-start even when a poll takes significant time. Transient failures recover on a shorter backoff, and a worker restarted during an incomplete batch can re-query immediately without losing attempt observability.

## Testing

- `GOCACHE=/tmp/peerdb-go-build go test ./activities -count=1`
- `GOCACHE=/tmp/peerdb-go-build go test ./connectors/external_metadata -run '^$' -count=1`


flow/activities/flowable_query_cdc.go
```diff
@@ -241,7 +262,7 @@ func (a *FlowableActivity) queryCDCPullSyncLoop(
 			continue
 		}
 
-		if wait := queryCDCPollWait(state.LastAttemptAt, time.Now(), idleTimeout); wait > 0 {
+		if wait := queryCDCPollWait(state.LastSyncedAt, time.Now(), syncInterval); wait > 0 {
```
dtunikov: this is the important change

flow/connectors/external_metadata/query_cdc_replication_state.go
```diff
@@ -25,8 +25,9 @@ type QueryCDCReplicationState struct {
 	// CursorText is the opaque cursor last returned by PullTableRecords for
 	// this table, empty if this table has never been synced.
 	CursorText string
-	// LastAttemptAt is when the last poll attempt for this table started,
-	// zero if never attempted.
+	// LastAttemptAt is when the latest poll attempt for this table started,
+	// whether or not it completed successfully. It is informational and does
+	// not control poll scheduling.
```
jgao54, line 30: nit: `does not control poll scheduling` comment seems a bit misleading, since `queryCDCPollWait` in happy path is determined by this and it does control the poll scheduling.

jgao54 (approved):

flow/connectors/external_metadata/query_cdc_replication_state.go
dtunikov: yeah, thanks for checking
this one has left from one of the previous iterations i've made

PR #4813, opened 2026-09-16
feat(structured-ingestion) DBI-1096: Support Nullable(JSON) columns in normalization

pfcoperez:
JSONExtractString yields `''` both for a JSON null and for a missing field.
Casting `''` or `NULL` with `::JSON` (behaviour prior to these changes) fails.

These changes route NULL for Nullable(JSON) columns in the normalize query casting to `Nullable(JSON)`.

This is required by structured ingestion, whose `_peerdb_malformed_data` column is a `Nullable(JSON)`.


itsbilal (approved):

jgao54 (commented):
Could we add e2e test for coverage (for both snapshot and cdc) ? If mongo testing is tricky, could add test to Postgres with the setting for json enabled, as nullable json column should work in general with this fix.

pfcoperez:
Thanks for the review @jgao54 !

> If mongo testing is tricky, could add test to Postgres with the setting for json enabled, as nullable json column should work in general with this fix.

MongoDB e2e tests is coming in follow-up PRs as this is a change unit from the _feature branch_. 

Yet, we can still add generic tests exercising this logic in the shape of UTs. Added in 41c91f1e2cd242020dee16f82e54276c6c8c7f76

PR #4812, opened 2026-09-16
feat(DBI-1067): Parallel test execution per source resource

pfcoperez:
Changes:

1. In tests whose entry functions are on the same file but using different resources (e.g: TestGenericCH_PG, TestGenericCH_MySQL, ...). Split the test file per resource.
2. Tag each e2e test file with buildtags for their sources resources: postgres, mysql, mongodb, cockroachdb, ... Note that them all use the catalog (PG dedicated resource) and ClickHouse as destination. The tag should depend on the source as them all use certain shared resources as these two. MySQL flavors will be tagged all mysql (no flavor distinction from now).
3. Add a parellel job to flow.yml workflow to test unit tests (those without tags).
4. Split the source resources test execution classified at (2) in a similar way to what was proposed at [PeerDB-io/peerdb#4510](https://linear.app/clickhouse/review/dbi-829-ci-specific-test-jobs-f1996945e39f) but adapting the the current form of flow.yml. In the end I want to define one job per build tag from (2). As in that obsolete pR example, I want to have one job per tested source resource and CH version combination.

This reduces CI build times from ~23' to less than 10' while also making it trivial to add new sources without having to fit them in the test table:

```yaml
          - {source: postgres, ch: lts,    version: '16'}
          - {source: postgres, ch: stable, version: '17'}
          - {source: postgres, ch: latest, version: '18'}
          - {source: mysql, ch: lts,    version: mysql-gtid, mariadb: maria-11}
          - {source: mysql, ch: stable, version: mysql-pos,  mariadb: maria-12}
          - {source: mysql, ch: latest, version: mysql-gtid, mariadb: maria-13}
          - {source: mongodb, ch: lts,    version: '6.0'}
          - {source: mongodb, ch: stable, version: '7.0'}
          - {source: mongodb, ch: latest, version: '8.0'}
          - {source: cockroachdb, ch: lts,    version: v24}
          - {source: cockroachdb, ch: stable, version: v25}
          - {source: cockroachdb, ch: latest, version: v26}
```

It also splits into different jobs the execution of UTs and e2e tests, they happen in parallel. 

The price to pay is using build tags to run e2e tests, listing explicitly the sources we want to test:

```bash
go test -tags postgres,mysql,mongodb,cockroachdb,bigquery,clickhouse ./e2e/
```

The Go changeset is big because it splits files containing functions for different function tests into separate files so the buildtag classification can be correctly applied.

Closes: https://linear.app/clickhouse/issue/DBI-1067

PR #4815, opened 2026-09-16
Optimize Tilt startup time

ilidemi:
Trying to see what further savings on top of #4812 are possible. Detailed breakdowns in the comments below.

| Changes, cumulative | Overall avg | Saved | PG avg | Saved |
|---|---:|---:|---:|---:|
| Original | 2m46s | — | 3m04s | — |
| Disable UI + migration-only Rust | 2m03s | −44s / −26.2% | 2m19s | −45s / −24.4% |
| Build Flow without waiting on dependencies | 1m23s | −39s / −32.2% | 1m35s | −44s / −31.6% |
| Parallel source provisioning (PG build) | 1m18s | −5s / −6.0% | 1m24s | −11s / −11.9% |
| **Total improvement** | | **−1m28s / −52.9%** | | **−1m40s / −54.4%** |

Added log capture to tilt startup so the data is available to do further improvements. Fixed up some dependencies and health checks to fix tilt parallelism.

Also played around with runner sizes. flow build times definitely benefit from extra cores, but the wait times can also be longer. Just leaving the data here, switch from 64 would be inconclusive.
| | 16 cores | 32 cores | 64 cores |
|---|---:|---:|---:|
| Avg runner wait | 40s (n=108) | 3s (n=16) | 1m23s (n=120) |
| Flow API image build | 2m16s | 1m35s | 1m09s |


ilidemi:
Initial breakdown

image

ilidemi:
After disabling peerdb-ui in CI and trimming Rust build to just migrations:

image

ilidemi:
After separating flow build so it doesn't wait for temporal&co

image

ilidemi:
After allowing parallel provisioning (removes the wait on PG build)

image

PR #4817, opened 2026-09-17
feat: table level resync

dtunikov:
Add ResyncTables to UpdateCDCFlow request
It allows users to resync individual tables as well as modify destination table settings (such as ordering key or partition by key) 

Resolves: DBI-1148


flow/workflows/cdc_flow.go
```diff
@@ -180,13 +199,104 @@ func processCDCFlowConfigUpdate(
...
+			}
+		}
 	}
 
 	telemetry.LogActivityUpdateFlowConfig(context.Background(), cfg.FlowJobName, oldValues, flowConfigUpdate)
 	syncStateToConfigProtoInCatalog(ctx, cfg, state)
 	return nextRunNone, nil
 }
 
+func resolveResyncTableMappings(
```
dtunikov, line 220: just makes sure that we don't try to resync tables that aren't in the mirror

flow/workflows/cdc_flow.go
```diff
@@ -180,13 +199,104 @@ func processCDCFlowConfigUpdate(
...
+		if requested.DestinationTableIdentifier != current.DestinationTableIdentifier {
+			return nil, fmt.Errorf("destination for table %s does not match the mirror configuration",
+				requested.SourceTableIdentifier)
+		}
+		resolved = append(resolved, proto.CloneOf(requested))
+	}
+	return resolved, nil
+}
+
+func validateResyncTableUpdate(
```
dtunikov, line 253: just checks that user is not trying to resync and remove/add the same table at once

PR #4818, opened 2026-09-17
feat(structured-ingestion-mongodb) DBI-1096: Structured ingestion settings through table mappings with validation

pfcoperez:
Add `structured_ingestion` and `drop_unexpected_values` to TableMapping and QRepConfig, and thread them from the mirror configuration down to the connectors: through NameAndExclude for CDC pulls, and through the QRep and snapshot workflows for the initial load and standalone QRep mirrors.

Mirror validation gains the structured ingestion rules, factored into the `structured` package so CDC and QRep creation share them.

Nothing acts on the settings yet: the connectors keep landing documents as JSON. The e2e tests cover typed mappings passing through and the validation errors.

On top of the contract, PeerDB API and UI have been expanded be compatible with the structured ingestion mappings. The UI changes are not visible uless the `PEERDB_UI_STRUCTURED_INGESTION_ENABLED` dynamic setting is set to `true`:

Screenshot From 2026-09-18 11-42-55

The dynamic setting will be set to `true` default once the flattening logic is merged.

UI changes are isolated in the second commit of this PR.
  

protos/flow.proto
```diff
@@ -56,12 +56,16 @@ message TableMapping {
   // the column to use as a cursor for query-based CDC replication
   // required if replication_method is BIGQUERY_REPLICATION_METHOD_QUERY
   string query_cdc_watermark_column = 11;
+  StructuredIngestionConfig structured_ingestion_config = 12;
+}
+
+message StructuredIngestionConfig {
```
jgao54: Just follow up to what @dtunikov recommended, and in a similar spirit as https://github.com/PeerDB-io/peerdb/pull/4819#discussion_r4052769847, 
how about:
```
message MongoTableConfig {
   bool structured_ingestion = 1;
   bool drop_unexpected_values = 2;
}

message BigQueryTableConfig {
    EventsFunction events_function = 1;
    string query_cdc_watermark_column = 2; 
}

message TableMapping {
  ...
  oneof source_table_config {
    BigqueryTableConfig bigquery_table_config = 29;
    MongoTableConfig mongo_table_config = 30
  }
}
```

protos/flow.proto
```diff
@@ -489,6 +499,13 @@ message QRepConfig {
   repeated string flags = 31; // internal
   // if true, then a separate null partition will be created for rows with null values in the watermark column
   bool add_null_partition = 32; // internal
+  // If true, source documents are projected onto the columns declared in `columns` using their
+  // `destination_type`, instead of landing whole in a single JSON column.
+  // Only for structured ingestion enabled sources.
+  bool structured_ingestion = 33;
+  // Mirrors TableMapping.drop_unexpected_values for the table being replicated: if true, reports of
+  // unfitting values will not include the values themselves, just their field and why they did not fit.
+  bool drop_unexpected_values = 34;
```
jgao54: same as above, can be:
```
message QRepConfig {
  ...
  oneof source_table_config {
    MongoTableConfig mongo_table_config = 33;
  }
}
```

protos/flow.proto
```diff
@@ -56,6 +56,16 @@ message TableMapping {
...
+  StructuredIngestionConfig structured_ingestion_config = 12;
+}
+
+message StructuredIngestionConfig {
+  // If true, the mirror will use structured ingestion on sources without schemas
+  // using the column types given by the `columns` (`destination_type`) field in the TableMapping.
+  bool structured_ingestion = 1;
+  // If true, when structured ingestion is enabled, reports of unfitting values will not include
+  // the values themselves, just their field and why they did not fit.
+  bool drop_unexpected_values = 2;
```
jgao54, line 70: ooc, what's a use case for drop_unexpected_values?

protos/flow.proto
jgao54: or if we'd like to keep the structured ingestion config generic:
```
message StructuredIngestionTableConfig {
  bool enabled = 1;
  bool drop_unexpected_values = 2
}

message MongoTableConfig {
  StructuredIngestionConfig structured_ingestion = 1;
}

message QueryCdcTableConfig {
    watermark_column = 1;
}

message BigqueryTableConfig {
  EventsFunction events_function = 1;
  QueryCdcTableConfig  query_cdc = 2;
}

message TableMapping {
  ...

  oneof source_table_config {
    MongoTableConfig mongo_table_config = 12;
    BigqueryTableConfig bigquery_table_config = 13;
  }
}
```

The bq changes are out-of-scope for this PR, but just showcasing here for completion.

flow/internal/dynamicconf.go
```diff
@@ -847,6 +856,12 @@ func PeerDBQueueParallelism(ctx context.Context, env map[string]string) (int64,
 	return dynamicConfSigned[int64](ctx, env, "PEERDB_QUEUE_PARALLELISM")
 }
 
+// PeerDBUIStructuredIngestionEnabled reports whether the UI offers configuring structured ingestion
+// on new mirrors. It only gates the setting being shown, not mirrors already using it or the API.
```
jgao54: nit: not worth duplicating comment here given we already have `Description` field

flow/model/model.go
```diff
@@ -14,8 +14,12 @@ import (
 )
 
 type NameAndExclude struct {
```
jgao54: instead of extending `nameAndExclude`, and if we go with the above proto change, can rename to:
```
type SourceTableMapping struct {
      Name string
      Exclude   map[string]struct{}
      SourceTableConfig SourceTableConfig
}
```

protos/flow.proto
pfcoperez: We discussed that some users might like to default to not reporting unexpected actual values for privacy reasons. For example, if there are columns known to be PII/Secret upon schema discovery users can exclude them but what if a document brings a new value and it has a secret? If you think your collection might have occurrences of these cases   you can opt in to stop reporting the value. Of course, the fact that the value is unexpected and the name of the field is still reported. 

Context: https://app.notion.com/p/MongoDB-Schema-Inference-and-structured-ingestion-3c82bd19f29080569788d2e28024cda1?d=3ca2bd19f29080c4a036001ccc6b668c&source=copy_link#3ca2bd19f2908014b4dfda8e5d4b40a8

protos/flow.proto
pfcoperez: Indeed, addressed in be59b4a3acfe82b96b5cb651e73279db66fd00c6 . I thought about wrapping `source_table_config` into a message re-usable in both places but discarded the idea as it adds too many levels of indirection  to be comfortable.

protos/flow.proto
pfcoperez: I kike the third  option you proposed. Applied in be59b4a3acfe82b96b5cb651e73279db66fd00c6 . I thought about wrapping `source_table_config` into a message re-usable in both places but discarded the idea as it adds too many levels of indirection  to be comfortable.

flow/internal/dynamicconf.go
pfcoperez: good point, done in b556cb2d1471602b25c9b9158bc48bbd505dfe4c

flow/model/model.go
pfcoperez: I held myself back from changing `NameAndExclude` to limit the change set but let's bite the bullet and have a proper name for it. 

Rename done in 9cfae8a00174c4d98fca1a432f3053e2d45c615f

`SourceTableConfig`  however is not a type from pbuff as the union type is implemented through the private `isTableMapping_SourceTableConfig` so I opted for using `maybeStructuredIngestion`  with a pointer to `StructuredIngestionConfiguration` to be filled when it makes sense. I think that's a good approach at this level of table specification. To ease the review, I've done it in a separate commit: bbe19905a7d7c79519b6662b4bfc16c15d555e85

protos/flow.proto
pfcoperez: As I consolidated these changes into the PR closing the functionality loop I found a problem with this approach. 

`TableMapping` is serialized to JSON as part of SyncFlowOptions to be used by the temporal workflow. Then it's deserialzied.

The union type is not easily deserializable there without custom marshalling/unmarshalling methods introducing JSON disambiguation tags. 

Given that this is the only MongoDB special configuration, I think the cleaner appeoach here is to just have StructuredIngestionTableConfig encapsulated in an optional field.

Connectors might or might not use this field depending on whether they implement structured ingestion. 

I've reverted to that in:    b5138a5d77559a98faa5b24fd505dd4455d22a45

The end result is: We just have a StructuredIngestionTableConfig optional field in the QRepConfig and TableMapping protocol buffers object.

protos/flow.proto
jgao54: Thanks for explaining, reading @dtunikov's comment:

> It just seems to me like for some users this “structured ingestion” feature could be a way to make sure that not ALL mongo columns go into ClickHouse. So, it’d be nice to have it configurable (the values, column names are fine of course, it’s not sensitive info). 

My read of this is that @dtunikov is referring to the the column exclusion feature, which is a frequently-asked feature; with this feature, any column that user wants to exclude would not replicate to malformed_data already (and any column that user wants to include as part of structured schema should already be non-sensitive)

Personally, i don't know if having column names here without their value for malformed data (which wouldn't help with backfill but would cause storage bloat) is worth the added complexity, vs. a simpler toggle like "ignore malformed data" where anything that doesn't fit the structured schema gets ignored.

Just my 2c! but given this is more of a product decision will leave to you and @morsapaes !

jgao54:
nits on UI, would be cleaner to:
- have the toggle for structured ingestion with the other table-level options at the bottom for consistency
- move the description for 'structured column' into a tooltip for consistency

and a question on the `Add Column` button, iirc i recall the columns will be inferred but not mutable, is that still the case  or do we plan to allow for user to specify destination type

flow/cmd/handler.go
```diff
@@ -264,6 +264,10 @@ func (h *FlowRequestHandler) CreateQRepFlow(
 		cfg.Flags = flags
 	}
 
+	if apiErr := h.checkQRepTableConfig(ctx, cfg); apiErr != nil {
```
jgao54, line 267: nit: we've deprecated qrep flow and don't plan to support new features on it. so not needed here

protos/route.proto
```diff
@@ -207,7 +207,10 @@ message SchemaTablesRequest {
   bool cdc_enabled = 3;
 }
 
-message SchemaTablesResponse { repeated TableResponse tables = 1; }
+message SchemaTablesResponse {
+  repeated TableResponse tables = 1;
+  bool structured_ingestion_supported = 2;
```
jgao54, line 212: ~iiuc, given the config for StructuredIngestionTableConfig.enabled is on the table level, should this be a field in TableResponse vs. a global config here?~ thinking more about it, given this is an option that user-specify,  not something we derive from server-side, TableResponse is the wrong direction so nvm that!

or if the goal is just to surface whether `PEERDB_UI_STRUCTURED_INGESTION_ENABLED` is enabled, 
we have a precedence today with `PEERDB_UI_MAINTENANCE_TAB_ENABLED` where the UI directly fetches call the GetDynamicSettings RPC (so you check for `isMongo && structuredIngestionEnabled` without having to wire it through here).

jgao54 (approved):
added a few UI-related comments and nits. also i think it may be better to work on the UI in a feature branch until things are properly wired up, just in case anyone who discovered/enabled the FF in OSS get confused by it and think it is ready. will leave you to decide on next steps.

docs/deep-dive-design-document.md
```diff
@@ -76,7 +76,7 @@ sequenceDiagram
 type PostgresCDCSource struct {
     *PostgresConnector
     srcTableIDNameMapping     map[uint32]string          // OID → table name
-    tableNameMapping          map[string]model.NameAndExclude
+    tableNameMapping          map[string]model.SourceTableMapping
```
jgao54, line 79: sorry, was not thinking straight when I typed sourceTableMapping; this is not a mapping but maybe `SourceTableInfo` or `SourceTableConfig` 🤦‍♀️

pfcoperez: It made sense because it stills represent a subset of the mapping.   And actually the field value is called `tableNameMapping` I think we can keep it.

protos/flow.proto
pfcoperez: I am still debating to myself if this option is going to be used in production as part of ClickPipes. That said, I think it's good to have the option wired in and take a final decision at ClickPipes UI level, that's why I chose a toggle whose default is including the values.

pfcoperez:
> and a question on the Add Column button, iirc i recall the columns will be inferred but not mutable

That's right @jgao54 

> is that still the case or do we plan to allow for user to specify destination type

Inference is part of the ClickPipes product, still PeerDB UI allows its users to use structured ingestion by manually providing the types at creation time.

pfcoperez:
Thanks @jgao54 !

> have the toggle for structured ingestion with the other table-level options at the bottom for consistency

Yes, it looks better there:

Applied in:  04294c0e40de17d0157fc6fc69176227b964dd17

> move the description for 'structured column' into a tooltip for consistency

Good point, done in 55678e02d459b4ee9e6b105b058f149d272e3233

protos/route.proto
pfcoperez: > iiuc, given the config for StructuredIngestionTableConfig.enabled is on the table level, should this be a field in TableResponse vs. a global config here?

This field is just to indicate the connector capability to use structured ingestion. 

So it actually is an AND between `PEERDB_UI_STRUCTURED_INGESTION_ENABLED` and the particular connector:

https://github.com/PeerDB-io/peerdb/blob/55678e02d459b4ee9e6b105b058f149d272e3233/flow/cmd/peer_data.go#L182-L184  

> isMongo && structuredIngestionEnabled without having to wire it through here.

As we evolve and add more connectors we'll have to update these conditions in the UI but keeping an unique source of truth removes this need. Actually, that method in the backend is the only decision point both at UI and backend.

pfcoperez:
@fable Review for potential regressions introduced here taking into account that the functionality loop will be closed in   https://github.com/PeerDB-io/peerdb/blob/DBI-91/connectors/mongodb/structured-ingestion/4-structured-cdc-qrep

flow/cmd/handler.go
pfcoperez: Good point, I think we can keep it anyway since its there already. But it's a good thing to use for the fhe future.

PR #4826, opened 2026-09-19
switch Tilt and CI to goose

jgao54:
Prior to this PR, db migration ran automatically as part of Nexus on startup. 

With this PR, we create a flow-migrate container (and add a flow-api/flow-worker/flow-snapshot-worker dependency on it) to run db migration and then exits, and disable Nexus running db migration on startup. 

Tested locally and validated the goose_db_versions table is created.

Fixes: DBI-1167


jgao54:
cherry-picked into https://github.com/PeerDB-io/peerdb/pull/4827 due to dependency

pfcoperez (approved):
LGTM

PR #4827, opened 2026-09-19
Refinery db migration cleanup

jgao54:
This PR:
- switch TILT/CI to use goose (commit is a copy of https://github.com/PeerDB-io/peerdb/pull/4826), added here as well so PR does not fail CI
- remove refinery scripts from codebase
- remove refinery dependency in Nexus, delete `--migration_only` arg (no longer used), and make `--migrations_disabled` a no-op (helm chart compatibility)
- update migration test. since we no longer have refinery migration scripts or Nexus dependency, I created pg_dump of these migrations and they are added as test fixtures 

split into commits for each of the task above.

Fixes: DBI-1147, DBI-1167

Note: while it's safe to merge this cleanup PR, will wait for one more release to let the cutover happen to reduce the amount of moving pieces in a release.


ilidemi (approved):

ilidemi:
@claude could you re-review, does the comment above still apply

PR #4829, opened 2026-09-20
feat(bigquery): use max watermark column value as checkpoint

dtunikov:
## What changed

- Use the maximum watermark value observed in a query as the next checkpoint instead of the query interval upper bound.
- When a bounded query window is empty, scan subsequent non-overlapping windows up to the safety-lag frontier. Each BigQuery job remains capped by `PEERDB_BIGQUERY_CDC_MAX_QUERY_WINDOW_SECONDS`.

## Why

This matters for long-running ingestions where the watermark column is `DEFAULT CURRENT_TIMESTAMP()`. BigQuery assigns that value at the beginning of the transaction, while the rows only become visible after the transaction finishes. Advancing the checkpoint to the query upper bound could therefore skip those rows.

There is also a sparse/idle-table edge case. If the checkpoint is `t0`, the maximum query window is 24 hours, and `(t0, t0+24h]` contains no rows, a max-seen checkpoint remains at `t0`. Without searching beyond that empty interval, every subsequent poll remains capped at `t0+24h`, so a later row with a watermark after that point is never discovered. The new bounded loop walks forward window by window until it finds a watermark or reaches `now-safetyLag`.

If all windows are empty, the original checkpoint is retained so a later backfill into the scanned range can still be discovered.

Resolves: DBI-1173


flow/connectors/bigquery/cdc.go
```diff
@@ -682,17 +683,17 @@ func (c *BigQueryConnector) pullTableQuery(
 		// column isn't present.
 		commitTimeNano := start.UnixNano()
 		if watermarkColIdx >= 0 {
-			switch v := row[watermarkColIdx].(type) {
-			case time.Time:
+			if v, ok := row[watermarkColIdx].(time.Time); ok {
 				commitTimeNano = v.UnixNano()
-			case civil.Date:
```
dtunikov, line 688: we don't allow DATE watermark column anyways
cleaned it up

pfcoperez (approved):

